PR #1, opened 2019-03-26
Health and ping endpoints

doctorpod:


Gemfile.lock
```diff
@@ -156,4 +156,4 @@ RUBY VERSION
    ruby 2.6.2p47
 
 BUNDLED WITH
-   2.0.1
+   1.17.2
```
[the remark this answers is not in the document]
doctorpod: <captain-mainwaring-voice>I was wondering who'd be first to notice this</captain-mainwaring-voice>
Yep. Will fix. Cheers @matthewrudy

app/controllers/status_controller.rb
```diff
@@ -0,0 +1,31 @@
+class StatusController < ApplicationController
+  def ping
+    json = {
+      'build_date' => ENV['APP_BUILD_DATE'] || 'Not Available',
```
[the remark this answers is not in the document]
doctorpod: They are made available as build arguments in the circle CI config file: https://github.com/ministryofjustice/pecs-move-platform-backend/blob/master/.circleci/config.yml#L82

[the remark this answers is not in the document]
matthewrudy: cool.
thanks

app/controllers/status_controller.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+class StatusController < ApplicationController
+  def ping
+    json = {
+      'build_date' => ENV['APP_BUILD_DATE'] || 'Not Available',
+      'commit_id' => ENV['APP_GIT_COMMIT'] || 'Not Available',
+      'build_tag' => ENV['APP_BUILD_TAG'] || 'Not Available'
```
stevehook: Is it worth making the `'Not Available'` literals a constant here to avoid duplication?

spec/requests/status_spec.rb
```diff
@@ -0,0 +1,86 @@
+# frozen_string_literal: true
+
+# rubocop:disable Metrics/BlockLength
```
stevehook: Can't think of a way to avoid these - long blocks are pretty common in specs so it's going to be recurring annoyance. Just wondering if there's a way to do this with a global rule for `*_spec.rb` files. Could we have a `.rubocop.yml` in the spec directory?

doctorpod: Yep - good shout

app/controllers/status_controller.rb
doctorpod: Yaas!

spec/requests/status_spec.rb
cesidio: Add this to rubocop.yml

```
Metrics/BlockLength:
  Enabled: false
```

stevehook: This works in the root `.rubocop.yml` I think ...

```
Metrics/BlockLength:
  Exclude:
    - 'spec/**/*'
```

PR #7, opened 2019-04-01
Add rspec gem

cesidio:


.gitignore
```diff
@@ -28,4 +28,5 @@
 
 # Ignore user's individual .env files
 .env
+.envrc
```
doctorpod: Don't we normally keep this in the repo? It just tells direnv to use the .env file

PR #9, opened 2019-04-01
Add rubocop-rspec

cesidio:


matthewrudy:
Happy to try this.
I've not used it before because it threw too many warnings that I didn't agree with.

PR #14, opened 2019-04-09
[P4-71] Initial database schema

stevehook:
This PR tracks the very start of work on the domain model and underlying database schema.

I've sketched out a first attempt at the domain model in https://drive.google.com/drive/folders/19yos6fCHXD0f3ROEJVmbU2QOQnznYRcP

Everything here could change as we learn more, this PR is meant to start the conversation about some of the choices we are going to have to make and to get something implemented that we can start building on.

In this PR we have sketched an outline of the first 4 database tables / models:

- `Move` - a move is for one person on a particular date with a from and a to location. It will consist of one or more Journeys (not covered here)
- `Person` - is any individual being moved
- `Profile` - represents the details for a person at a given point in time and possibly within the context of a specific move. One of the requirements here is that the profile details for a given person are versionable or auditable and so I'm modelling by allowing each person to have one or more timestamped profiles.
- `Location` - a prison, police station, court or hospital etc.


app/models/person.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+class Person < ApplicationRecord
+  has_many :profiles
```
cesidio: Should have has_many :moves too I think

config/routes.rb
```diff
@@ -1,5 +1,6 @@
 Rails.application.routes.draw do
   get '/ping', to: 'status#ping', format: :json
   get '/health', to: 'status#health', format: :json
-  # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+
+  resources :moves, only: :index
```
cesidio: let's leave it out for now

PR #15, opened 2019-04-10
Move to live-1 cluster

cesidio:


Dockerfile
```diff
@@ -13,4 +13,13 @@ ENV RAILS_ENV production
 ENV PUMA_PORT 3000
 EXPOSE $PUMA_PORT
 
+# Run the application as user `moj` (created in the base image)
+# uid=1000(moj) gid=1000(moj) groups=1000(moj)
+# Some directories/files need to be chowned otherwise we get Errno::EACCES
+#
+# RUN chown $APPUSER:$APPUSER ./db/schema.rb
```
stevehook: Why is this commented out?

cesidio: We do not have a schema.rb file, but in the future, when we activate the command `rails db:migrate` that is commented out too at the moment, I guess we will require to uncomment this line

PR #19, opened 2019-04-26
[P4-125] Implement pagination with pager_api gem

cesidio:


config/initializers/kaminari_config.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+Kaminari.configure do |config|
+  config.default_per_page = 20
```
doctorpod: Is it worth making these env vars?

spec/requests/api/v1/moves_controller_spec.rb
```diff
@@ -41,6 +53,18 @@
         get '/api/v1/moves', headers: valid_headers
         expect(JSON.parse(response.body)).to include_json(data: [{ id: move_id }])
       end
+
+      it 'paginates 20 results per page' do
+        get '/api/v1/moves', headers: valid_headers
+
+        expect(JSON.parse(response.body)['data'].size).to eq 20
+      end
```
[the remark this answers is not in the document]
cesidio: If you specify a page size of 101,  it still returns 100 results

config/initializers/kaminari_config.rb
cesidio: Not sure, in the document Matt was talking about having 20 results as a default and it is a value that can be overridden. What is the advantage of having this as an env variable?

doctorpod: Yes you're right, there's no advantage as these are overridable defaults.

PR #94, opened 2019-08-05
[P4-537] Extended input for trans gender

stevehook:
- [x] Add a new text attribute to Profile called say additional_gender_information to store the extra input data.
~- [x] Add a new boolean attribute to Gender called say prompt_for_additional_information that indicates to the front-end whether we need to display and prompt for the additional info.~
~- [x] Change the gender importer so that it sets the prompt_for_additional_information attribute to true for the  built-in Trans gender.~

### DIFF
--- a/app/models/gender.rb
+++ b/app/models/gender.rb
@@ -3,4 +3,13 @@
 class Gender < ApplicationRecord
   validates :title, presence: true
   validates :key, presence: true
+  validates :prompt_for_additional_information, inclusion: [true, false]
+
+  before_validation :set_prompt_for_additional_information
+
+  private
+
+  def set_prompt_for_additional_information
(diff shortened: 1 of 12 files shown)


stevehook:
I've renamed `additional_gender_information` to `gender_additional_information`.

@teneightfive I guess it feels a bit odd to me to have the options for these kinds of questions (the list of genders for this one) data driven and have hard-coded rules in the frontend (and backend) that depend on specific values. Having said that simple should always be preferred, it depends on whether this pattern is going to be needed for other values for other questions?

Just to clarify does the validation rule make `gender_additional_information` mandatory if and only if `trans` is selected? And enforce that it's blank otherwise?

teneightfive:
I spoke to @andy-Millar again and I think we can simplify it even more by dropping the validation. 

I’m happy for the client to be responsible for how it handles display of the questions too.

PR #95, opened 2019-08-06
Allow filtering of location by NOMIS agency ID

slorek:
HMPPS SSO provides the frontend application with a list of locations for the user in the form of their NOMIS agency ID(s). The frontend application therefore needs a way to look those up and resolve them to the IDs used in the API.

This provides an additional filter to find locations by their NOMIS agency ID. It is theoretically possible for there to be more than one with the same ID, but I haven't tried to resolve this here. For now the frontend will just use the first one that isn't disabled.

### DIFF
--- a/app/controllers/api/v1/reference/locations_controller.rb
+++ b/app/controllers/api/v1/reference/locations_controller.rb
@@ -24,7 +24,7 @@ module Api
           render json: location, status: status
         end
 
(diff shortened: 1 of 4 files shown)


slorek:
@teneightfive Yes, that is how it will work :)

teneightfive:
> @teneightfive Yes, that is how it will work :)

My bad, misread the original comment as this change will do that.

PR #98, opened 2019-08-07
Change Prometheus collection condition

teneightfive:
Prometheus will currently attempt to collect stats for all environments
apart from `test`. This causes regular warnings when running the app
locally without Prometheus correctly configured.

This change makes the check look for `production` before setting up
Prometheus.

### DIFF
--- a/config/initializers/prometheus_exporter.rb
+++ b/config/initializers/prometheus_exporter.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-unless Rails.env == 'test'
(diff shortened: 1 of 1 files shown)


config/initializers/prometheus_exporter.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-unless Rails.env == 'test'
+if Rails.env == 'production'
```
stevehook: Can also use `if Rails.env.production?`

PR #100, opened 2019-08-13
[P4-564] NOMIS API data fixtures

stevehook:
The purpose of this PR is to setup some realistic but anonymised fixture data containing moves from the NOMIS API.

The fixture data covers court moves for upto 2 working days in the past to 2 working days in the future for the following three locations: 

- LEI - HMP Leeds
- PVI - HMP Pentonville
- MRI - HMP Manchester

The data fixtures are initially generated by running the following rake task:

```
rails nomis_fixtures:import_moves
```

This task depends on environment variables containing credentials for the NOMIS API - we use `preprod` to get realistic data. (Use something like `foreman` to make environment variables in `.env` available to the rake task at runtime).


slorek:
Thanks Steve. I previously worked with Andy on https://github.com/ministryofjustice/elite2-api-mock, which I think might be a less obtrusive and more controlled way of limited API requests to fixture data. This would ensure that all requests were served with something we control, and without overriding behaviour. What do you think?

stevehook:
@slorek I've had a go at making the stand-alone elite2-api-mock server work with our. It will need a bit of work but could probably be made to work to replace _test mode_ albeit with the fixture generation remaining in the Rails app (as a fairly decoupled rake task).

There are a few things that have tripped me up but I've got the rails apps nomis client library talking to elite2-api-mock. The auth stuff was mounted on the wrong path it seems and there is no route for `/movements/transfers` so that would need to be added.

I can see the attraction of decoupling the test server, it keeps the rails app cleaner. The downsides I can foresee are:

1.  Maintaining the separate elite2-api-mock may be a bit of extra work.
2.  Running it in a development environment is very easy but if we want to use it on a test server on cloud platform there is a bit more work to do to set it up as a separate environment.
3. It doesn't have any tests. Maybe fine but makes me a bit less confident about extending it.
4. It would make it much harder (maybe just not practical) to use the data fixtures as part of our acceptance specs because it would introduce a dependency on a separate process.

I guess the last point is me not fully understanding how we'll be using the fixture data - is it going to be for test automation, just developer testing or manual testing in a `staging` or review app environment? I think having a better understanding of the use case would help me.

stevehook:
I've removed the _test mode_ stuff, at least for now. This PR now just contains the tools needed to generate data fixtures - `rails nomis_fixtures:import_moves`

PR #103, opened 2019-08-19
Fix `Location#nomis_agency_id` for Guildford Police

stevehook:
Was causing issues on staging now that we have capability for multiple locations because the original agency id did not match what was (later) added to NOMIS itself. Should be `SRY016` rather than `GLFD`.


cesidio:
I am not sure about this approach. This is relevant only for the staging environment, it is run once and forgotten about. I don't see the point of a migration that needs to be deployed on staging to run. Personally I would just SSH into staging pod and fix that record.

stevehook:
@cesidio yes I'm happy to do that too. I don't know how this data was set up in the first place. The only reason for doing it this way was so that is some traceability.

stevehook:
I've corrected the data manually on the `staging` DB so no need for this PR any longer.

PR #104, opened 2019-08-19
Add NomisClient People and Moves libraries

cesidio:


.rubocop.yml
```diff
@@ -19,6 +19,9 @@ Layout/AlignParameters:
 Metrics/LineLength:
   Max: 120
 
+Metrics/MethodLength:
+  Max: 15
```
stevehook: Depends how often we are hitting this limit but if it's infrequently we should probably stick with the default and opt-out on a per-method basis. I guess you hit it here where you are returning a multiline hash even though it's logically just one line of code.

app/lib/nomis_client/people.rb
```diff
@@ -0,0 +1,28 @@
...
+        attributes_for(
+          NomisClient.get("/prisoners/#{prison_number}").parsed.first
+        )
+      end
+
+      def attributes_for(person)
+        {
+          nomis_prison_number: person['offenderNo'],
+          last_name: person['lastName'],
+          first_names: [person['firstName'], person['middleNames']].compact.join(' '),
```
stevehook: Makes sense to map the keys from camel to snake case but I don't think this kind of mapping stuff belongs in the client library. It's a separate concern.

.rubocop.yml
cesidio: correct. I can turn it off just for that instance.

app/lib/nomis_client/people.rb
cesidio: I agree with that, will change it

PR #106, opened 2019-08-21
Change moves time_due to be a datetime

cesidio:
All the moves in production have a time_due that is null, so dropping and recreating the column in the database is safe

### DIFF
--- a/db/migrate/20190821095338_change_time_due_to_datetime.rb
+++ b/db/migrate/20190821095338_change_time_due_to_datetime.rb
@@ -1,0 +1,6 @@
+class ChangeTimeDueToDatetime < ActiveRecord::Migration[5.2]
+  def change
+    remove_column :moves, :time_due
+    add_column :moves, :time_due, :datetime
+  end
+end
--- a/spec/lib/nomis_client/moves_spec.rb
+++ b/spec/lib/nomis_client/moves_spec.rb
@@ -14,7 +14,7 @@ RSpec.describe NomisClient::Moves, with_nomis_client_authentication: true do
           from_location_nomis_agency_id: 'BXI',
           to_location_nomis_agency_id: 'BXI',
           date: '2019-08-19',
-          time_due: '17:00:00',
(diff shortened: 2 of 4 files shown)


db/migrate/20190821095338_change_time_due_to_datetime.rb
```diff
@@ -0,0 +1,6 @@
+class ChangeTimeDueToDatetime < ActiveRecord::Migration[5.2]
+  def change
+    remove_column :moves, :time_due
```
stevehook: I don't think we can get away with this now that we are in production. We have to rename the old column, create the new one, copy the data from old to new and then drop the old column.

spec/lib/nomis_client/moves_spec.rb
```diff
@@ -14,15 +14,15 @@
...
+          time_due: '2019-08-19T17:00:00',
           nomis_event_id: 468_536_961
         },
         {
           person_nomis_prison_number: 'G7157AB',
           from_location_nomis_agency_id: 'BXI',
           to_location_nomis_agency_id: 'WDGRCC',
           date: '2019-08-19',
-          time_due: '09:00:00',
+          time_due: '2019-08-19T09:00:00',
```
stevehook: We need to ensure that the `time_due` date value is the same as the `date`. Otherwise we will hit nasty daylight savings bugs (off by one hour errors).

It seems to me that the best way to avoid these is to store just one value that contains the date-time value for the move. However I suspect this won't always be practical - it's possible that the `time_due` could be on the next day?

So not sure what to suggest here. A starting point would be to check what data we are getting input via the API.

db/migrate/20190821095338_change_time_due_to_datetime.rb
cesidio: All the moves in production have a time_due that is set to null, so dropping and recreating the column in the database is safe

stevehook: fair enough

PR #107, opened 2019-08-21
Set default sort order to ID on moves

teneightfive:
We were previously sorting moves by location title to help with
the frontend. This currently causing issues with pagination where
results may be returned across multiple pages as location is not
always guaranteed.

This change amends the sort to be by ID so the order returned during
pagination will always be the same.

### DIFF
--- a/app/services/moves/finder.rb
+++ b/app/services/moves/finder.rb
@@ -9,7 +9,7 @@ module Moves
     end
 
     def call
-      apply_filters(Move).order('locations.title')
+      apply_filters(Move).order('moves.id')
     end
 
     private


app/services/moves/finder.rb
```diff
@@ -9,7 +9,7 @@ def initialize(filter_params)
     end
 
     def call
-      apply_filters(Move).order('locations.title')
+      apply_filters(Move).order('moves.id')
```
teneightfive: Is this the default sort? Wondering if this would do the same?

```ruby
def call
  apply_filters(Move)
end
```

stevehook: I don't get an `ORDER BY` clause in a quick test. (`apply_filters(Move).to_sql` gave me something like `SELECT moves.* FROM moves WHERE moves.from_location_id = 'a11f6867-bd35-409e-a034-7f64a8b91c13'`)

PR #108, opened 2019-08-21
Add moves and people importer

cesidio:


spec/services/moves/importer_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+    end
+
+    it 'sets the date of the move' do
+      importer.call
+      expect(move.date).to eq Date.parse('2019-08-19')
+    end
+
+    it 'sets the time_due of the move' do
+      importer.call
+      expect(move.time_due).to eq DateTime.parse('2019-08-19T17:00:00')
```
stevehook: For me I get...
```
DateTime.parse('2019-08-19T17:00:00')
=> Mon, 19 Aug 2019 17:00:00 +0000
```
This seems wrong to me - the `time_due` string surely represents a local time so should it be ` Mon, 19 Aug 2019 17:00:00 +0100`?

Maybe we need...
```
Time.zone.parse('2019-08-19T17:00:00')
=> Mon, 19 Aug 2019 17:00:00 BST +01:00
```

Either way I think we need some more checks here, to validate any assumptions made.

spec/services/people/importer_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+      aliases: nil,
+      pnc_number: '82/18053V',
+      cro_number: '018053/82G',
+      gender: 'F',
+      ethnicity: 'White: Eng./Welsh/Scot./N.Irish/British',
+      nationalities: 'British'
+    }
+  end
+
+  let!(:ethnicity) { create(:ethnicity, title: 'White: Eng./Welsh/Scot./N.Irish/British') }
```
stevehook: Can we assume that there is a one-one mapping between ethnicity names coming out of NOMIS with those in our system? What happens if we get an unknown ethnicity?

spec/services/people/importer_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+  let!(:gender) { create(:gender, nomis_code: 'F') }
```
stevehook: Can we assume that there is a one-one mapping between gender codes coming out of NOMIS with those in our system?

spec/services/moves/importer_spec.rb
stevehook: It may also be a problem that we don't specify a timezone in `application.rb`

spec/services/people/importer_spec.rb
cesidio: At the moment we have to assume that the mapping is in place, in fact it should be managed in a different table from where this field is populated. I think what we have to do is to improve the logging system for when an event like this happens.

spec/services/people/importer_spec.rb
cesidio: Same as above

spec/services/moves/importer_spec.rb
cesidio: Added
`config.time_zone = 'London'
to `application.rb`

spec/services/moves/importer_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+  context 'with one existing record' do
+    let!(:move) { create(:move, nomis_event_id: 468_536_961) }
+
+    it 'creates 1 move' do
+      expect { importer.call }.to change(Move, :count).by(1)
+    end
+  end
+
+  context 'with one existing record with different attributes' do
+    let(:time_due) { DateTime.parse('2019-08-19T09:00:00') }
```
stevehook: Should this be `Time.zone.parse`?

PR #111, opened 2019-08-28
Added cancellation reason and comment, configurable DB_USERNAME

ngw:
FIX https://dsdmoj.atlassian.net/browse/P4-631

- Adds properties `cancellation_reason` and `cancellation_reason_comment`
- `database.yml` uses `DB_USERNAME` environment variable

### DIFF
--- a/app/models/move.rb
+++ b/app/models/move.rb
@@ -12,6 +12,12 @@ class Move < ApplicationRecord
     prison_transfer: 'prison_transfer'
   }
 
+  enum cancellation_reason: {
+    created_by_mistake: 'created_by_mistake',
+    supplier_cannot_move: 'supplier_cannot_move',
+    other: 'other'
+  }
+
   belongs_to :from_location, class_name: 'Location'
   belongs_to :to_location, class_name: 'Location', optional: true
   belongs_to :person
--- a/app/models/move.rb
+++ b/app/models/move.rb
@@ -13,8 +13,8 @@ class Move < ApplicationRecord
   }
 
   enum cancellation_reason: {
-    created_by_mistake: 'created_by_mistake',
-    supplier_cannot_move: 'supplier_cannot_move',
+    made_in_error: 'made_in_error',
+    supplier_declined_to_move: 'supplier_declined_to_move',
     other: 'other'
   }
 
--- a/spec/factories/move.rb
+++ b/spec/factories/move.rb
@@ -10,7 +10,5 @@ FactoryBot.define do
     status { 'requested' }
     additional_information { 'some more info about the move that the supplier might need to know' }
     move_type { 'court_appearance' }
-    cancellation_reason { 'other' }
-    cancellation_reason_comment { 'other reason' }
   end
 end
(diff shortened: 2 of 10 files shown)


slorek:
Thank you. Please could you also add these properties to the [Swagger schema](https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/swagger/v1/move.json)

swagger/v1/move.json
```diff
@@ -63,6 +63,15 @@
...
             ],
             "description": "Additional information about the move that the supplier should be made aware of"
+          },
+          "cancellation_reason": {
+            "type": "string",
+            "enum": ["error", "null", "other", "supplier_cannot_move"],
+            "description": "Reason the move has been cancelled"
+          },
+          "cancellation_reason_comment": {
+            "type": "string",
```
stevehook: I think this needs to be nullable

spec/factories/move.rb
```diff
@@ -10,5 +10,7 @@
     status { 'requested' }
     additional_information { 'some more info about the move that the supplier might need to know' }
     move_type { 'court_appearance' }
+    cancellation_reason { 'other' }
+    cancellation_reason_comment { 'other reason' }
```
stevehook: I don't think this is changing the tests. The new attributes should be declared in the `MovesSerializer` and in param whitelists - https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/controllers/api/v1/moves_controller.rb#L47-L53

app/models/move.rb
```diff
@@ -12,6 +12,12 @@ class Move < ApplicationRecord
     prison_transfer: 'prison_transfer'
   }
 
+  enum cancellation_reason: {
+    error: 'error',
```
teneightfive: Could we change this to something like `created_in_error` or `created_by_mistake` please? Feels like it's a little more descriptive as to what the "error" relates to.

At the moment this option is to capture cases like a user creating a move that is no longer needed or if they made a mistake.

For context, the content Leigh has come up with for the form option is:

>**Made in error**
>For example, there's a mistake or the detainee no longer needs to move

swagger/v1/move.json
```diff
@@ -63,6 +63,16 @@
               { "type": "null" }
             ],
             "description": "Additional information about the move that the supplier should be made aware of"
+          },
+          "cancellation_reason": {
+            "type": "string",
+            "enum": ["error", "null", "other", "supplier_cannot_move"],
```
teneightfive: Do we need to include `null` as an option in the enum? Or can we state that the attribute can also be null?

ngw: To be honest I'm not sure, I will change it to be consistent with the other property

app/models/move.rb
```diff
@@ -12,6 +12,12 @@ class Move < ApplicationRecord
     prison_transfer: 'prison_transfer'
   }
 
+  enum cancellation_reason: {
+    created_by_mistake: 'created_by_mistake',
```
teneightfive: Sorry to be a pain. Have had a chat with @lmoney-moj about the option names and I think we should match them to the language we use in the interface to avoid any confusion when looking at the data from the API.

Could we use these reasons as the enum options please:
- `made_in_error`
- `supplier_declined_to_move`
- `other`

PR #112, opened 2019-08-29
Import moves from NOMIS on GET moves

cesidio:


app/controllers/api/v1/moves_controller.rb
```diff
@@ -73,6 +74,25 @@ def find_move
...
+        return unless from_location_nomis_agency_id && date
+
+        moves = NomisClient::Moves.get(from_location_nomis_agency_id, date)
+        Moves::Importer.new(moves).call
+      end
+
+      def from_location_nomis_agency_id
+        return unless filter_params[:from_location_id]
+
+        Location.find(filter_params[:from_location_id]).nomis_agency_id
```
slorek: You could use `find_by(from_location_id: filter_params[:from_location_id])&.nomis_agency_id` here in case the supplied `from_location_id` doesn't resolve. Depends if you want to fail gracefully or throw an exception.

app/controllers/api/v1/moves_controller.rb
```diff
@@ -73,6 +74,25 @@ def find_move
           .includes(:from_location, :to_location, person: { profiles: %i[gender ethnicity] })
           .find(params[:id])
       end
+
+      def import_moves_from_nomis
```
slorek: It's worth considering wrapping this in a `begin...rescue` block. If Elite2 is unavailable this would block the whole request I think. We probably want to show the cached moves still, in this case.

ngw: This can be private right?
My suggestion (to take with a truckload of salt as I'm the new kid on the block here and I'm most probably missing lots of context) would be to extract this behavior in a service object with a well defined interface and defer to this `nomis` service to interact with it.

app/controllers/api/v1/moves_controller.rb
```diff
@@ -73,6 +74,25 @@ def find_move
...
+        return unless from_location_nomis_agency_id && date
```
ngw: Surely effective, could be a good idea to have this in a private method called in a `before_action`.

app/controllers/api/v1/moves_controller.rb
ngw: In case: this is an article I found helpful in the past https://www.toptal.com/ruby-on-rails/rails-service-objects-tutorial

stevehook: It makes sense to return cached data if it's available rather than failing the request. I just wonder whether we need some way to warn the user that the data they are seeing is out of date. This might be best tackled when implementing a cache expiry function (later).

app/controllers/api/v1/moves_controller.rb
stevehook: I reckon that if a request specifies a `from_location_id` that doesn't exist we should return some kind of error. Maybe a `422` rather than a `500` though.

app/controllers/api/v1/moves_controller.rb
cesidio: > This can be private right?
> My suggestion (to take with a truckload of salt as I'm the new kid on the block here and I'm most probably missing lots of context) would be to extract this behavior in a service object with a well defined interface and defer to this `nomis` service to interact with it.

It is already private

app/controllers/api/v1/moves_controller.rb
cesidio: in the `ApiController` what we do is

```
rescue_from ActiveRecord::RecordNotFound, with: :render_resource_not_found_error
```

I think that returning a 404 error is the most appropriate thing here, which is in line with what we do elsewhere, but happy to discuss about it.

stevehook: makes sense to me

cesidio:
> All good, but I think it would be worth testing explicitly for the Nomis client failing with an exception and this being captured in the controller.

Added a spec for that too

spec/requests/api/v1/moves_controller_index_spec.rb
```diff
@@ -85,6 +85,87 @@
...
+        let(:filters) do
+          {
+            from_location_id: from_location.id,
+            date_from: date.to_s
+          }
+        end
+        let(:params) { { filter: filters } }
+
+        before do
+          allow(NomisClient::Moves).to receive(:get).and_raise(StandardError)
```
slorek: Is this `before` block needed if we specify `skip_before: true` in the only test?

spec/requests/api/v1/moves_controller_index_spec.rb
```diff
@@ -85,6 +85,87 @@
...
+
+          moves_finder = instance_double('Moves::Finder', call: Move.all)
+          allow(Moves::Finder).to receive(:new).and_return(moves_finder)
+
+          get '/api/v1/moves', headers: headers, params: params
+        end
+
+        it 'invokes the Moves::Finder service', skip_before: true do
```
slorek: Is this test working if we are skipping the `before` block?

spec/requests/api/v1/moves_controller_index_spec.rb
cesidio: I think this `skip_before` is a bit confusing, @stevehook didn't like it from the beginning and he might have been right. So `skip_before` is referred to the first `before` in the file, the one on top of everything where it checks for the `metadata[:skip_before]` and if it's present it doesn't make the request. I use it to have a default `before` where we make the request, and in the few exceptions where I set doubles and method call expectations, I have to call it manually.

slorek: Got it thanks. I've seen other usage of `skip_before` in the past which is a bit confusing.

PR #113, opened 2019-08-30
[P4-606] NOMIS alerts 'importer'

stevehook:
To ensure that we have all the information we need to map NOMIS alerts to our own model without dropping any information that doesn't map to our own model we need a catalogue of NOMIS alert types and codes.

This PR provides a new rake task that creates or updates a local record of the alerts and codes from NOMIS: 

```
bundle exec rail reference_data:create_nomis_alerts
```

This information is stored in a new model called `NomisAlert` that lists the type and alert codes. In future we will use this model to also link `NomisAlert`s to our `AssessmentQuestion`s so that we can map alerts for people to our own model based on the NOMIS alert code returned by their API.


stevehook:
I think we need to do some renaming - we have two kinds of thing called _importers_ now. There are the reference data importers that are run periodically in the background to import reference data into our system from the NOMIS API. There are also importers that import move data from the API and are triggered by `GET /moves` API requests. This could get confusing so I am wondering whether we should rename the reference data importers?

slorek:
> I think we need to do some renaming - we have two kinds of thing called _importers_ now. There are the reference data importers that are run periodically in the background to import reference data into our system from the NOMIS API. There are also importers that import move data from the API and are triggered by `GET /moves` API requests. This could get confusing so I am wondering whether we should rename the reference data importers?

I'm not sure they are two different types of things. They are doing the same thing, albeit triggered at different times. I think if the behaviour of the thing is similar a consistent name is appropriate.

stevehook:
thanks @slorek I don't think it's a problem either having pondered on it for a bit. And we haven't got into a situation where the same model needs two different 'kinds of importer'. Will leave as is.

app/services/nomis_alerts/importer.rb
```diff
@@ -0,0 +1,165 @@
...
+    KNOWN_ALERTS = {
+      'Self harm' => :self_harm,
+      'Must be segregated' => :hold_separately,
+      'Violent' => :violent,
+      'Escape' => :escape,
+      'Not to be released' => :not_for_release,
+      'Health and medical' => :health_issue
+    }.freeze
+
+    def initialize(alert_types:, alert_codes:)
```
slorek: Not a biggie, but to me the interface to this class is a little non-obvious. I think it would be cleaner to accept just `alert_codes`, with the `alert_types` retrieved on-demand (we're already pulling in `NomisAlert` later).

stevehook: ok thanks, I am happy to change that...

stevehook: that should be done now

app/services/nomis_alerts/importer.rb
```diff
@@ -126,8 +126,9 @@ class Importer
       'Health and medical' => :health_issue
     }.freeze
 
-    def initialize(alert_types:, alert_codes:)
-      self.alert_types = alert_types.map { |alert_type| [alert_type[:code], alert_type] }.to_h.with_indifferent_access
+    def initialize(alert_codes:)
+      self.alert_types =
```
slorek: Sorry to be nitpicky. This doesn't need to go in the initialiser, it could be lazily evaluated in a `alert_types` method. For example:

```ruby
def alert_types
  @alert_types ||= NomisClient::AlertTypes.get.map { |alert_type| [alert_type[:code], alert_type] }.to_h.with_indifferent_access
end
```
Not a blocker however!

db/schema.rb
```diff
@@ -74,8 +74,8 @@
     t.string "reference", null: false
     t.string "move_type"
     t.string "additional_information"
-    t.integer "nomis_event_id"
     t.datetime "time_due"
+    t.integer "nomis_event_id"
```
cesidio: Probably you want to revert just this line, as it seems to be wrong if you do a `rails db:drop db:create db:migrate` command

app/services/nomis_alerts/importer.rb
slorek: Even better, you could also consider some convenience method on `NomisClient::AlertTypes` to return it in the format you want.

PR #125, opened 2019-09-16
Import status from NOMIS

cesidio:


app/lib/nomis_client/moves.rb
```diff
@@ -28,10 +29,12 @@ def attributes_for(nomis_data, event_type)
             to_location_nomis_agency_id: item['toAgency'],
             date: item['eventDate'],
             time_due: item['startTime'],
+            status: Move::NOMIS_STATUS_TYPES[item['eventStatus']],
```
stevehook: Is there are 'fallback' or default status? In other words what happens if `item['eventStatus']` is a value that we didn't know about?

cesidio: There is no fallback at the moment, not sure what that could be. Maybe the initial status of `requested`?

PR #131, opened 2019-09-24
NomisClient now anonymises data when in test_mod

ngw:
Solves ticket https://dsdmoj.atlassian.net/browse/P4-652

### DIFF
--- a/app/services/moves/anonymiser.rb
+++ b/app/services/moves/anonymiser.rb
@@ -2,25 +2,22 @@
 
 module Moves
   class Anonymiser
-    attr_accessor :nomis_offender_number, :day_offset, :move_response
+    attr_accessor :move
 
-    def initialize(nomis_offender_number:, day_offset:, move_response:)
-      self.nomis_offender_number = nomis_offender_number
-      self.day_offset = day_offset
-      self.move_response = move_response
+    def initialize(move:)
+      self.move = move
     end
 
     def call
-      start_date_time = move_response['startTime'].present? ? Time.parse(move_response['startTime']) : nil
+      nomis_offender_number =
+        People::Anonymiser.encrypt_offender_number(offender_number: move['offenderNo'])
+      start_date_time = move['startTime'].present? ? Time.parse(move['startTime']) : nil
       start_time = start_date_time&.strftime('%H:%M:%S') || '00:00:00'
 
-      move_response.merge(
-        offenderNo: nomis_offender_number,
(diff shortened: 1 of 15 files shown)


spikeheap:
This looks good to me as a tactical solution to remove identifiable data from the API responses. I'll pick up improving the T3 dataset with the DPS team, which should mean we can roll back these changes in `n` months...

app/services/moves/anonymiser.rb
```diff
@@ -2,26 +2,31 @@
...
-        commentText: nil,
-        createDateTime: "<%= date.toISOString().split('.')[0] %>",
-        eventDate: "<%= date.toISOString().split('T')[0] %>",
-        startTime: "<%= date.toISOString().split('T')[0] + 'T' + '#{start_time}' %>"
+      move.merge(
+        person_nomis_prison_number: nomis_offender_number,
+        from_location_nomis_agency_id: from_location,
+        to_location_nomis_agency_id: to_location,
+        date: "<%= date.toISOString().split('T')[0] %>",
+        time_due: "<%= date.toISOString().split('T')[0] + 'T' + '#{start_time}' %>"
```
stevehook: I think we should still set the `judgeName` and `commentText` to `nil` even if we don't use them.

app/services/people/anonymiser.rb
```diff
@@ -6,8 +6,18 @@ module People
...
+          Rails.application.secrets.secret_key_base[0..31]
+        )
+        cipher.encrypt_and_sign(offender_number)
+      end
+    end
+
     def initialize(nomis_offender_number:)
-      self.nomis_offender_number = nomis_offender_number
+      self.nomis_offender_number =
+        People::Anonymiser.encrypt_offender_number(offender_number: nomis_offender_number)
```
stevehook: Can we need to use the offender number from the move here than repeat the encryption?

stevehook: I guess this makes the interface a little cleaner so leave as is I think.

app/services/moves/anonymiser.rb
ngw: AFAIK that wouldn't make a lot of sense, before we were anonymising the data that nomis was returning, I changed that to anonymise the data the client is returning for consistency on how for example the people anonymiser was working.
I wouldn't even be sure what parameter to add because that data is not what NomisClient#Moves is returning.

PR #135, opened 2019-09-27
Validator for move params

ngw:
Fixes https://dsdmoj.atlassian.net/browse/P4-701

### DIFF
--- a/app/controllers/api/v1/moves_controller.rb
+++ b/app/controllers/api/v1/moves_controller.rb
@@ -10,7 +10,7 @@ module Api
           moves = Moves::Finder.new(filter_params).call
           paginate moves, include: MoveSerializer::INCLUDED_ATTRIBUTES
         else
-          render json: { error: moves_params.errors }
+          render json: { error: moves_params.errors }, status: :bad_request
         end
       end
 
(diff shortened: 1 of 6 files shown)


ngw:
Force pushed indentation and spacing to make rubocop happy

app/controllers/api/v1/moves_controller.rb
```diff
@@ -4,9 +4,14 @@ module Api
...
-        import_moves_from_nomis
-        moves = Moves::Finder.new(filter_params).call
-        paginate moves, include: MoveSerializer::INCLUDED_ATTRIBUTES
+        moves_params = Moves::ParamsValidator.new(params[:filter])
+        if moves_params.valid?
+          import_moves_from_nomis
+          moves = Moves::Finder.new(filter_params).call
+          paginate moves, include: MoveSerializer::INCLUDED_ATTRIBUTES
+        else
+          render json: { error: moves_params.errors }
```
stevehook: I think there should be a non-200 status code in here somewhere

stevehook: Also can you adapt this to fit with the JSONAPI spec - https://jsonapi.org/format/#errors ?

I think there are a couple of examples elsewhere in the code. Maybe you can reuse https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/controllers/api_controller.rb#L40 ?

stevehook: You might need to change the Swagger docs depending on whether this case is covered by an existing scenario.

PR #136, opened 2019-09-27
[P4-707] Support for multiple location downloads

stevehook:
We can now use the environment variable `NOMIS_AGENCY_IDS` to specify a comma-separated list of NOMIS ids for the prisons that form part of the pilot program. This value will be used for requests to `GET /moves` that do not specify a `from_location_id`, e.g.

```
NOMIS_AGENCY_IDS=HLI,PVI
```

### DIFF
--- a/app/controllers/api/v1/moves_controller.rb
+++ b/app/controllers/api/v1/moves_controller.rb
@@ -94,6 +94,8 @@ module Api
         @from_locations ||=
           if filter_params[:from_location_id]
             [Location.find(filter_params[:from_location_id])]
+          elsif ENV['NOMIS_AGENCY_IDS']
+            Location.where('nomis_agency_id IN (?)', ENV['NOMIS_AGENCY_IDS'].split(',').map(&:strip))
           else
             []
           end
--- a/app/controllers/api/v1/moves_controller.rb
+++ b/app/controllers/api/v1/moves_controller.rb
@@ -82,18 +82,21 @@ module Api
 
       def import_moves_from_nomis(move: nil)
         if move.present?
-          Moves::NomisSynchroniser.new(location: move.from_location, date: move.date).call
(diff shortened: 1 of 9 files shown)


app/controllers/api/v1/moves_controller.rb
```diff
@@ -82,18 +82,23 @@ def find_move
...
       end
 
-      def from_location
-        return unless filter_params[:from_location_id]
-
-        @from_location ||= Location.find(filter_params[:from_location_id])
+      def from_locations
+        @from_locations ||=
+          if filter_params[:from_location_id]
+            [Location.find(filter_params[:from_location_id])]
```
spikeheap: Hi @stevehook, if `NOMIS_AGENCY_IDS` specifies agencies that are part of the pilot, should this restrict `from_locations` to _only_ IDs in that set?

stevehook: @spikeheap I guess that makes sense. My only concern would be that we may grant access to people to view moves for other prisons, possibly as we roll out to other establishments. If that were the case you might log in be able to select a location which is a prison outside the pilot programme and then not be able to see any moves. If we did do this check then I think we would need to make sure that there was a clear error message for any attempt to look at prisons outside the list.

spikeheap: That sounds reasonable @stevehook. Digital Prisons use this approach to feature-toggle prisons for new NOMIS. I'm easy either way – if we keep the existing behaviour should we change `NOMIS_AGENCY_IDS` to `DEFAULT_NOMIS_AGENCY_IDS` or something that conveys the meaning of the env var?

stevehook: I've changed the env var name - makes sense. I've logged a separate ticket for filtering based on the env var - https://dsdmoj.atlassian.net/browse/P4-713 (mainly to keep this PR small(ish))

PR #138, opened 2019-10-03
Attempt to fix to_location_id and fetch multiple from_location_id

ngw:
The new `from_location_id` and `to_location_id` (which wasn't working previously) are now accepting arrays.
The goal is to follow json api recommendations (https://jsonapi.org/recommendations/#filtering) without breaking HTTP query strings (which require array parameters to be like "from_location_id[]=foo&from_location_id[]=bar")

### DIFF
--- a/app/controllers/api/v1/moves_controller.rb
+++ b/app/controllers/api/v1/moves_controller.rb
@@ -41,7 +41,8 @@ module Api
 
       private
 
-      PERMITTED_FILTER_PARAMS = %i[date_from date_to from_location_id location_type status].freeze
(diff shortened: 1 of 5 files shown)


app/controllers/api/v1/moves_controller.rb
```diff
@@ -98,7 +99,7 @@ def import_moves_from_nomis(move: nil)
       def from_locations
         @from_locations ||=
           if filter_params[:from_location_id]
-            [Location.find(filter_params[:from_location_id])]
+            filter_params[:from_location_id].collect { |id| Location.find(id) }
```
spikeheap: Could this be `Location.where(id: filter_params[:from_location_id])` to reduce the number of database queries?

PR #139, opened 2019-10-04
reference/supplier has been added

ngw:
Solves https://dsdmoj.atlassian.net/browse/P4-730

### DIFF
--- a/app/controllers/api/v1/reference/suppliers_controller.rb
+++ b/app/controllers/api/v1/reference/suppliers_controller.rb
@@ -19,11 +19,7 @@ module Api
         private
 
         def find_supplier
-          if ::UUID.validate(params[:id])
-            Supplier.find(params[:id])
-          else
-            Supplier.find_by(key: params[:id])
-          end
+          Supplier.find_by(key: params[:id])
         end
       end
     end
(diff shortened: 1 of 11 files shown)


app/controllers/api/v1/reference/suppliers_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+
+        private
+
+        def find_supplier
+          if ::UUID.validate(params[:id])
+            Supplier.find(params[:id])
+          else
+            Supplier.find_by(key: params[:id])
+          end
+        end
```
spikeheap: Hi @ngw, are we moving away from the service objects approach (e.g. https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/services/locations/finder.rb)?

spikeheap: Also, are we sure we need to look this up abiguously? I've asked on https://dsdmoj.atlassian.net/browse/P4-730.

ngw: About service objects no, we're not moving away, I just didn't think about it at that moment because we mostly don't use finders and query directly the model (see for example `find_location`). I guess I'll wait to see what happens with the find by key or id which is actually the only logic that could go in a service object and eventually refactor. Good enough?

spikeheap: Sounds good to me :+1:

ngw:
Will wait a bit to merge this to see how we will filter by supplier.

PR #140, opened 2019-10-07
Add habtm locations-suppliers

ngw:
Soves https://dsdmoj.atlassian.net/browse/P4-725 by adding a many to many relationship between locations and suppliers

### DIFF
--- a/app/controllers/api/v1/moves_controller.rb
+++ b/app/controllers/api/v1/moves_controller.rb
@@ -41,7 +41,8 @@ module Api
 
       private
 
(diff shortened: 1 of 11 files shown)


ngw:
I can do that @spikeheap, I normally don't on habtm because that kind of relationship tends to avoid querying the join table directly. But if you think it's a good idea sure, why not.

spikeheap:
Hi @ngw, I've been thankful for the timestamps a couple of times when I've been debugging something awkward (or trying to replicate the state of the system at a point in time). :+1:

ngw:
When rebases make github cry...

PR #141, opened 2019-10-08
locations accept filter supplier_ids

ngw:
Solves https://dsdmoj.atlassian.net/browse/P4-731 by adding a filter on `reference/locations` that accepts `supplier_ids`.

### DIFF
--- a/app/controllers/api/v1/reference/locations_controller.rb
+++ b/app/controllers/api/v1/reference/locations_controller.rb
@@ -24,7 +24,7 @@ module Api
           render json: location, status: status
         end
 
-        PERMITTED_FILTER_PARAMS = %i[location_type nomis_agency_id].freeze
+        PERMITTED_FILTER_PARAMS = [:location_type, :nomis_agency_id, supplier_ids: []].freeze
 
         def filter_params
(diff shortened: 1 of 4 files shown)


app/controllers/api/v1/reference/locations_controller.rb
```diff
@@ -24,7 +24,7 @@ def render_location(location, status)
           render json: location, status: status
         end
 
-        PERMITTED_FILTER_PARAMS = %i[location_type nomis_agency_id].freeze
+        PERMITTED_FILTER_PARAMS = [:location_type, :nomis_agency_id, supplier_ids: []].freeze
```
teneightfive: Maybe this should be singular: `supplier_id`. I don't think we'll ever have a case where we need to filter on multiple suppliers.

A user should never work for both suppliers, in theory.

ngw: `supplier_id` is something that doesn't exist on that model, as the relationship has been set as a many to many. That would require me intercepting the filter and rewrite the parameter in `supplier_ids`. Can do it, but extremely sloppy - eventually this is not a many to many.

PR #147, opened 2019-10-17
task for removing duplicates + unique contraint on nomis_event_id

ngw:


app/controllers/api/v1/moves_controller.rb
```diff
@@ -16,7 +16,7 @@ def index
 
       def show
         move = find_move
-        import_moves_from_nomis(move: move)
+        # import_moves_from_nomis(move: move)
```
feedmypixel: best not to commit commented out code?

lib/tasks/data_maintenance.rake
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
```
feedmypixel: does this need to be in here?

spec/services/moves/sweeper_spec.rb
```diff
@@ -149,52 +149,5 @@
         end
       end
     end
-
```
feedmypixel: bit scary that there are no tests around this? I always find not only are they good for testing but they are great for describing what the code is expected to do. Considering this has been such a "work it out and see" it would be good to put some tests in to explain to your future self or others what is intended for this code to do

lib/tasks/data_maintenance.rake
ngw: yep: https://stackoverflow.com/questions/37799296/what-does-the-comment-frozen-strinliteralgliteral-literal-true-do

spec/services/moves/sweeper_spec.rb
ngw: The code has been removed... The sweeper doesn't remove duplicates anymore...

app/controllers/api/v1/moves_controller.rb
ngw: removed

PR #149, opened 2019-10-22
police national computer filter

ngw:
Fixes https://dsdmoj.atlassian.net/browse/P4-740 by adding a `/people?police_national_computer=$PNC` endpoint

### DIFF
--- a/app/controllers/api/v1/people_controller.rb
+++ b/app/controllers/api/v1/people_controller.rb
@@ -3,6 +3,11 @@
 module Api
   module V1
     class PeopleController < ApiController
+      def index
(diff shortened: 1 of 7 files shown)


feedmypixel:
To confirm this work works with the associated frontend work in https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/pull/288

spikeheap:
The syntax looks good to me :+1:. @ngw would it be onerous to rename `police_national_computer to `police_national_number` so the name matches what's held?

ngw:
I think PNC really stands for Police National Computer?

PR #150, opened 2019-10-23
Logstash formatting

spikeheap:
This PR introduces the `lograge` gem to output JSON-formatted logs to `stdout`, which are then consumed by fluentd and pushed into the Cloud Platform Kibana instance.

Switching from plain-text to JSON logs should allow us to graph (for example) request/response times, filter and query by status, action, controller, etc.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -43,3 +43,7 @@ group :development do
   gem 'spring'
   gem 'spring-watcher-listen', '~> 2.0.0'
 end
+
+gem "lograge", "~> 0.11.2"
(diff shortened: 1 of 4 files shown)


Gemfile
```diff
@@ -43,3 +43,7 @@ group :development do
   gem 'spring'
   gem 'spring-watcher-listen', '~> 2.0.0'
 end
+
+gem 'lograge', '~> 0.11.2'
```
ngw: maybe you can add the gems before the groups like all others?

PR #156, opened 2019-11-06
Hotfix duplicate moves

ngw:
The old fix dealing with duplicate moves has been ported to deal with the new `nomis_event_ids` field

### DIFF
--- a/app/services/moves/sweeper.rb
+++ b/app/services/moves/sweeper.rb
@@ -5,7 +5,7 @@ module Moves
     attr_accessor :items, :date, :locations
 
     def initialize(locations, date, items)
-      self.locations = locations
+      self.locations = locations.select(&:prison?)
       self.date = date
       self.items = items
     end
--- a/app/services/moves/sweeper.rb
+++ b/app/services/moves/sweeper.rb
@@ -36,6 +36,8 @@ module Moves
         date: date,
         from_location_id: locations.map(&:id)
       )
+
+      update_nomis_event_ids_when_duplicate(scope)
       update_nomis_event_ids_containing_currents(scope)
       update_nomis_event_ids_not_containing_currents(scope)
     end
@@ -45,6 +47,15 @@ module Moves
         'ARRAY[?] && nomis_event_ids', current_nomis_event_ids
       ).each do |move|
         move.update(nomis_event_ids: move.nomis_event_ids & current_nomis_event_ids)
+      end
+    end
+
+    def update_nomis_event_ids_when_duplicate(scope)
+      items.map do |item|
+        move = scope.find_by(
+          person: Person.where(nomis_prison_number: item[:person_nomis_prison_number])
+        )
+        move.update(nomis_event_ids: move.nomis_event_ids << item[:nomis_event_id])
       end
     end
 
--- a/app/services/moves/sweeper.rb
+++ b/app/services/moves/sweeper.rb
@@ -18,7 +18,11 @@ module Moves
 
     def cancel_outdated_moves!
       update_nomis_event_ids!
-      Move.where(nomis_event_ids: []).update(status: Move::MOVE_STATUS_CANCELLED)
+      Move.where(
+        date: date,
+        from_location_id: locations.map(&:id),
+        nomis_event_ids: []
+      ).update(status: Move::MOVE_STATUS_CANCELLED)
     end
 
     def current_nomis_event_ids
--- a/app/services/moves/sweeper.rb
+++ b/app/services/moves/sweeper.rb
@@ -42,20 +42,21 @@ module Moves
       update_nomis_event_ids_not_containing_currents(scope)
     end
 
+    def update_nomis_event_ids_when_duplicate(scope)
+      items.map do |item|
+        move = scope.find_by(
+          person: Person.where(nomis_prison_number: item[:person_nomis_prison_number]),
+          to_location: Location.where(nomis_agency_id: item[:to_location_nomis_agency_id])
+        )
+        move.update(nomis_event_ids: move.nomis_event_ids << item[:nomis_event_id])
+      end
+    end
+
     def update_nomis_event_ids_containing_currents(scope)
       scope.where(
         'ARRAY[?] && nomis_event_ids', current_nomis_event_ids
       ).each do |move|
         move.update(nomis_event_ids: move.nomis_event_ids & current_nomis_event_ids)
-      end
-    end
-
-    def update_nomis_event_ids_when_duplicate(scope)
-      items.map do |item|
-        move = scope.find_by(
-          person: Person.where(nomis_prison_number: item[:person_nomis_prison_number])
-        )
-        move.update(nomis_event_ids: move.nomis_event_ids << item[:nomis_event_id])
       end
     end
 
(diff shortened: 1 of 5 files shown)


app/services/moves/sweeper.rb
```diff
@@ -36,6 +36,8 @@ def update_nomis_event_ids!
         date: date,
         from_location_id: locations.map(&:id)
       )
+
+      update_nomis_event_ids_when_duplicate(scope)
```
spikeheap: Are we using the same logic to detect potential duplicates as the unique index? Should this (L35-38 above) also have to_location and time_due?

spec/services/moves/sweeper_spec.rb
```diff
@@ -54,117 +56,105 @@
...
-        it 'removes the outdated nomis_event_id' do
-          sweeper.call
-          expect(multiple_nomis_ids_move.reload.nomis_event_ids).to eq([attributes[:nomis_event_id]])
-        end
+      it 'cancels the outdated move' do
+        sweeper.call
+        expect(outdated_move.reload.status).to eq('cancelled')
       end
     end
   end
```
spikeheap: Do we need an additional test case where there are no NOMIS ids? I may have missed this one...

app/services/moves/sweeper.rb
```diff
@@ -48,6 +50,15 @@ def update_nomis_event_ids_containing_currents(scope)
       end
     end
 
+    def update_nomis_event_ids_when_duplicate(scope)
+      items.map do |item|
+        move = scope.find_by(
+          person: Person.where(nomis_prison_number: item[:person_nomis_prison_number])
+        )
+        move.update(nomis_event_ids: move.nomis_event_ids << item[:nomis_event_id])
```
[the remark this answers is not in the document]
ngw: Yes, if you look into the model it will call uniq before validation

app/services/moves/sweeper.rb
ngw: The whole sweeper has been built around the concept that from_location + date + person = same move, this is the logic that was there before and the spec. Eventually using a timestamp could be a very bad idea, it normally gives a lot of problems.
If you guys think the to_location is relevant so be it

spec/services/moves/sweeper_spec.rb
ngw: I thought it was implicit that that case was covered, because the one you're commenting works (id gets removed from the array, then state changes to cancelled because the array is empty). Makes sense?

spikeheap: Sorry, I meant for moves that never belonged to NOMIS (i.e. the data we lost before)

app/services/moves/sweeper.rb
spikeheap: Hi @ngw, from the spec doc:

> A Person could have multiple Moves on the same date, given they were for a different purpose (eg. a Move to court in the morning, and a Hospital Move in the evening).

It's important the uniqueness validation matches this logic, so I'll add a note on this PR next to the schema

db/migrate/20191106121438_unique_index_on_moves.rb
```diff
@@ -0,0 +1,5 @@
+class UniqueIndexOnMoves < ActiveRecord::Migration[5.2]
+  def change
+    add_index :moves, [:from_location_id, :to_location_id, :person_id, :date, :time_due], name: 'index_on_move_uniqueness', unique: true
```
spikeheap: Given https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/156#discussion_r343683943, this index should be `:from_location_id, :to_location_id, :person_id, :date`.

spec/services/moves/sweeper_spec.rb
ngw: There's no clear way to identify that, they all have empty nomis_event_ids. We have same move from nomis and outside of it now?

lib/tasks/data_maintenance.rake
```diff
@@ -1,6 +1,17 @@
...
+  task remove_duplicate_moves: :environment do
+    duplicates = Move.select(:nomis_event_ids).group(:nomis_event_ids).having('COUNT(*) > 1').size
+    duplicates.each_pair do |k, _|
+      next if k.nil?
+
+      moves = Move.order(:created_at).where(nomis_event_ids: k).offset(1)
+      moves.destroy_all
+    end
+  end
+
```
spikeheap: Hi @ngw, could we add a test for this please? I'm probably reading this wrong, but it looks like `duplicates` is set to be a number (the result of `.size`).

Could we also add some comments, and name variables more explicitly. It's not clear to me what `k` represents, and why it's pulled out with `each_pair`.

spec/services/moves/sweeper_spec.rb
jason-sims: There is a really clear way to identify locations that should be synced with NOMIS. If the NOMIS client is hitting the **movements** api requesting moves from.. lets say HMP Pentonville (PVI) then this site and all the moves originating from it, should be synced. The movements api is key as this tells us users are not entering moves in our service for that site.(they also have the pecs_prison role & caseload - our workflow doesn't allow us create moves in our service) What sites a user seeing is due to HMPPS Authentication (If it is an Auth user (outside of NOMIS) its a police station, If its non Auth (inside NOMIS - we have to do a secondary lookup to get caseload) then its a Prison). Not all requests from the frontend invoke the NOMIS lookup - all that logic is in the codebase.

spikeheap: :+1: @ngw this might be what you're after: https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/models/location.rb#L21-L23

lib/tasks/data_maintenance.rake
spikeheap: Morning :Wave:. I've dug into this code a bit more and I think the behaviour is correct. I'd recommend refactoring it to make it more readable, for example:

```
    duplicate_event_ids = Move
                            .group(:nomis_event_ids)
                            .having('COUNT(*) > 1')
                            .pluck(:nomis_event_ids)
                            .compact
    
    duplicate_event_ids.each  do |event_ids|
      Move
        .where(nomis_event_ids: event_ids)
        .order(:created_at)
        .offset(1)
        .destroy_all
    end
```

I also think this needs a test, as it destroys data.

spec/services/moves/sweeper_spec.rb
ngw: Ok, but then the logic we currently have is 100% correct and covering all the cases, right? The sweeper does not act on "all the moves", but limited to moves that are passed as arguments. Once the cancellation of the moves is scoped on the locations provided (which are provided directly by nomis) we should be perfectly safe.

spikeheap: Hi @ngw, I think we should just merge it at this point, but your comment above makes me feel quite uneasy. We should be coding defensively, particularly on this sort of thing where important data is being modified. The assumption that it's "100% correct" relies on the logic in another class not changing, which means we have a dependency on the thing that invokes this... it's easy to foresee this being called somewhere else in future, and without that defensive check we _could_ alter data in a way we might not expect. 

It's good to be kind to our future selves and protect ourselves from that sort of error.

As an example, you could write a test case here that invokes the sweeper for a location which is not a prison and has no `nomis_event_ids`. Even if we don't add that test before merging this (and I believe @jason-sims is keen to manually test this in staging today), I'd like to see this test in a subsequent PR.

PR #163, opened 2019-11-12
Batching calls for /alerts/

ngw:
Closes https://dsdmoj.atlassian.net/browse/P4-786

Batches calls to import alerts

### DIFF
--- a/app/lib/nomis_client/alerts.rb
+++ b/app/lib/nomis_client/alerts.rb
@@ -3,14 +3,21 @@
 module NomisClient
   class Alerts < NomisClient::Base
     class << self
-      def get(prison_number)
-        response(prison_number).map do |alert|
+      def get(prison_numbers)
+        get_response(nomis_offender_numbers: prison_numbers).map do |alert|
           attributes_for(alert)
         end
       end
 
-      def get_response(nomis_offender_number:)
-        NomisClient::Base.get("/bookings/offenderNo/#{nomis_offender_number}/alerts")
+      def get_response(nomis_offender_numbers:)
+        NomisClient::Base.post(
+          '/bookings/offenderNo/alerts',
+          body: nomis_offender_numbers.to_json,
+          headers: {
+            'Accept': 'application/json',
+            'Content-Type': 'application/json'
+          }
+        ).parsed
       end
 
       # rubocop:disable Metrics/MethodLength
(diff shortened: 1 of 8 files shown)


app/lib/nomis_client/alerts.rb
```diff
@@ -3,14 +3,21 @@
...
-      def get_response(nomis_offender_number:)
-        NomisClient::Base.get("/bookings/offenderNo/#{nomis_offender_number}/alerts")
+      def get_response(nomis_offender_numbers:)
+        NomisClient::Base.post(
+          '/bookings/offenderNo/alerts',
+          body: nomis_offender_numbers.to_json,
+          headers: {
+            'Accept': 'application/json',
+            'Content-Type': 'application/json'
+          }
```
spikeheap: Hi @ngw, could the headers be set in `NomisClient::Base`? It feels like this should be consistent for all requests to NOMIS.

app/lib/nomis_client/base.rb
```diff
@@ -11,6 +11,7 @@ def get(path, params = {})
       end
 
       def post(path, params = {})
+        params[:headers] = update_json_headers(params[:headers]) if params[:headers]
```
spikeheap: Hi @ngw, should this add/update the headers regardless of their presence when this method is called? I think I'd  expect to be able to call this with no `params[:headers]` specified and still see the headers added to the call.

app/lib/nomis_client/base.rb
```diff
@@ -54,6 +55,15 @@ def benchmark_request(path)
...
       end
+
+      def update_json_headers(headers)
+        return unless headers.key?(:Accept) && headers.key?(:'Content-Type')
+
+        {
+          'Accept': 'application/json',
+          'Content-Type': 'application/json'
+        }
+      end
```
spikeheap: Would this remove existing headers passed into the `post` method? The logic I think I'd expect from this and L14 is:

- Set the `Accept` header unless it has been specified in `params[:headers]`.
- Set the `Content-Type` header unless it has been specified in `params[:headers]`.

That way we'd preserver any explicit headers added by the calling method, but add them otherwise. If that matches your understanding we might want to rename this `add_json_headers`, and merge those with the existing headers rather than overwriting it.

app/lib/nomis_client/base.rb
ngw: why would you add an instance variable for this? what's the point?

spikeheap: Sorry, I don't think I'm suggesting that. My understanding is that the code currently will *only* call update_json_headers` when `params[:headers]` is set. The behaviour I think I'd expect is to always call that method... so if `post` is called with a `params` hash that doesn't contain `:headers` the json headers will still be added.

ngw: Oh gotcha. Sorry misunderstood. See if you're happy now.

ngw: ah wait, ok I see the point now

ngw: we should be there now

PR #170, opened 2019-11-28
P4 807 support ability to delete documents via api

ngw:
Solves https://dsdmoj.atlassian.net/browse/P4-807

It also removes description and document_type from the Document model


app/controllers/api/v1/documents_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+
+      def document_attributes
+        document_params[:attributes].merge(
+          move: Move.find(params.dig(:move_id))
+        )
+      end
+
+      def render_document(document, status)
+        render json: document, status: status
+      end
```
spikeheap: Hi @ngw, could we un-refactor this? It appears to only swap named parameters for positional ones, and IMHO the `render` call is clearer in-line.

feedmypixel:
I can confirm this branch works with the frontend for:
• calls to `[POST]: /move/<moveId>/documents` to upload a `document` to this `move`
• calls to `[GET]: /moves/<moveId>` to obtain `documents` uploaded to this `move`
• calls to `[DELETE]: /move/<moveId>/documents/<documentId>` to delete a specified `document` that has been uploaded to this `move`

I am getting back everything I need from this calls.
I am yet to check the error response but if this is the same as all API calls then I am _functionally_ happy with this work for the frontend.

PR #174, opened 2019-12-03
Hotfix limit multipart to document creation

ngw:
Limit multipart Content-Type to document creation and leave everything else to JSON API

The reason for doing this is that we're returning JSON with content-type `multipart/form-data` which doesn't get parsed by the HTTP library the frontend is using.
This limits the `multipart/form-data` to the document creation, and returns JSON API for everything else.

### DIFF
--- a/app/controllers/api/v1/documents_controller.rb
+++ b/app/controllers/api/v1/documents_controller.rb
@@ -3,17 +3,19 @@
 module Api
   module V1
     class DocumentsController < ApiController
+      prepend_before_action :set_force_content_type
+
       CONTENT_TYPE = 'multipart/form-data'
 
       def create
         document = Document.create!(document_attributes)
-        render json: document, status: 201
+        render json: document, status: 201, content_type: ApiController::CONTENT_TYPE
       end
 
       def destroy
         document = Document.find(params[:id])
         document.destroy!
-        render json: document, status: 200
+        render json: document, status: 200, content_type: ApiController::CONTENT_TYPE
       end
 
       private
@@ -30,6 +32,10 @@ module Api
         document_params[:attributes].merge(
           move: Move.find(params.dig(:move_id))
         )
+      end
+
+      def set_force_content_type
+        @force_content_type = true
       end
     end
   end
--- a/app/controllers/api/v1/documents_controller.rb
+++ b/app/controllers/api/v1/documents_controller.rb
@@ -3,11 +3,11 @@
 module Api
   module V1
     class DocumentsController < ApiController
-      prepend_before_action :set_restricted_content_type, only: :create
+      prepend_before_action :set_restricted_request_content_type, only: :create
 
       def create
         document = Document.create!(document_attributes)
-        render json: document, status: 201, content_type: ApiController::CONTENT_TYPE
+        render json: document, status: 201
       rescue ActiveSupport::MessageVerifier::InvalidSignature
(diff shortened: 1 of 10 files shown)


app/controllers/api_controller.rb
```diff
@@ -2,8 +2,8 @@
 
 class ApiController < ApplicationController
   before_action :doorkeeper_authorize!
-  before_action :restrict_content_type
-  after_action :set_content_type
+  before_action :restrict_content_type, unless: -> { @force_content_type }
+  after_action :set_content_type, unless: -> { @force_content_type }
```
spikeheap: Hi @ngw, just taking the ability to set response content-type explicitly, the flag used here feels a bit convoluted, and it's easy to see someone copy/pasting the `content_type` and then wondering why it's not being honoured. 

Could we just make this a `before_action` (without the `unless`? That way it's easy to override it explicitly with `content_type: ApiController::CONTENT_TYPE` (as you have) without needing to also say you don't want to set it for anything in that controller.

spikeheap: Given the update to the description, I think the change I've suggested here would resolve your issue entirely (without changing `restrict_content_type`).

`restrict_content_type` prevents requests from arriving at the API with an incorrect `Content-Type`. If I'm understanding correctly, that doesn't sound like an issue here as you're expecting the correct `Content-Type`, but returning a different one.

spec/requests/api/v1/documents_controller_destroy_spec.rb
```diff
@@ -53,11 +63,5 @@
 
       it_behaves_like 'an endpoint that responds with error 404'
     end
-
-    context 'with an invalid CONTENT_TYPE header' do
-      let(:content_type) { 'application/xml' }
-
-      it_behaves_like 'an endpoint that responds with error 415'
-    end
```
spikeheap: I think we still want to validate that we reject invalid content type headers

app/controllers/api/v1/documents_controller.rb
```diff
@@ -31,6 +33,12 @@ def document_attributes
           move: Move.find(params.dig(:move_id))
         )
       end
+
+      protected
+
+      def set_restricted_request_content_type
+        @restricted_request_content_type = 'multipart/form-data'
+      end
```
spikeheap: :+1: this approach lets us have different endpoints with custom content types on each :+1:

spec/requests/api/v1/documents_controller_destroy_spec.rb
spikeheap: Hi @ngw wouldn't this test still be valid? I think it'll still pass with the changes you've proposed.

app/controllers/api/v1/documents_controller.rb
```diff
@@ -3,11 +3,13 @@
...
   module V1
     class DocumentsController < ApiController
-      CONTENT_TYPE = 'multipart/form-data'
+      prepend_before_action :set_restricted_request_content_type, only: :create
 
       def create
         document = Document.create!(document_attributes)
         render json: document, status: 201
+      rescue ActiveSupport::MessageVerifier::InvalidSignature
+        Document.create!(file: nil)
```
spikeheap: Sorry, I've just noticed this. How come we create an empty file if the signature's invalid?

PR #175, opened 2019-12-03
Swagger ui fixes

spikeheap:
This PR:

- Adds auth to Swagger, so you can complete the OAuth cycle and test the endpoints directly.
- Fixes `filter[location_from]` and `filter[location_to]` so the UI renders and allows an array of UUIDs to be specified.

While testing this I came across a few problems with the `Content-Type` header (`GET`, `DELETE` and `HEAD`) requests don't have a body, so don't have `Content-Type` set, and setting it in the Swagger config renders a redundant example body field. The API requires the `Content-Type` header to be set correctly for all requests.

To resolve this, this PR updates the API controller to validate the `Content-Type` header as follows:
- If the `Content-Type` header is set, it must be correct.
- If the request is body-less (`GET`, `DELETE` and `HEAD`), `Content-Type` may be omitted.
- If the request has a body (`POST`, `PUT` and `PATCH`), `Content-Type` must be specified and correct.

The direct effect is that this PR enables Swagger docs to have `GET` requests without a body field, which is neater and clearer for suppliers to consume.


spikeheap:
I've tried to add tests around this, but I'm being thwarted. Setting `CONTENT_TYPE` or `Content-Type` to `nil` (or removing it from the headers hash) doesn't actually remove the `Content-Type` from the header in the request... it just drops in `application/x-www-form-urlencoded`. I've burned a bit of time trying to explicitly set things, but to no avail.

Given that, the existing test suite validates that correct and incorrect headers are handled correctly, we just don't have tests for accepting no Content-Type on bodiless requests, and rejecting for others.

@ngw do you have any ideas for working around this? If not I think I'd accept the risk and merge this...

PR #180, opened 2019-12-04
Fixes validation when file is missing + attachment url

ngw:
This fixes a weird bug happening when POSTing to `documents/` with an empty `file` attribute, for some reasons the server returns a `ActiveSupport::MessageVerifier::InvalidSignature` instead of generating a validation error (https://groups.google.com/forum/#!topic/rubyonrails-core/xI5HPnxIAIE).

It also adds the url attribute to the Document's JSON serializer.

### DIFF
--- a/app/serializers/document_serializer.rb
+++ b/app/serializers/document_serializer.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class DocumentSerializer < ActiveModel::Serializer
+  include Rails.application.routes.url_helpers 
+  
   attributes :id, :url, :filename, :content_type
 
   def url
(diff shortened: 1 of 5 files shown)


app/serializers/document_serializer.rb
```diff
@@ -1,7 +1,11 @@
 # frozen_string_literal: true
 
 class DocumentSerializer < ActiveModel::Serializer
-  attributes :id, :filename, :content_type
+  attributes :id, :url, :filename, :content_type
+
+  def url
+    rails_blob_path(object.file)
+  end
```
spikeheap: :+1: these links auto-expire, which is 👍

PR #185, opened 2019-12-12
Change db table declaration to be nice with UUIDS

ngw:
This changes the DB schema to use UUIDs in the ActiveStorage attachments table.

This is a framework bug: https://github.com/rails/rails/issues/35746

1) Pull
2) `cd $YOURLOCATION/hmpps-book-secure-move-api`
3) `bundle exec rake db:migrate VERSION=20191106121438` # this will wipe the broken table from your local db
4) `bundle exec rake db:migrate`

### DIFF
--- a/db/migrate/20191114122923_create_active_storage_tables.active_storage.rb
+++ b/db/migrate/20191114122923_create_active_storage_tables.active_storage.rb
@@ -1,7 +1,7 @@
 # This migration comes from active_storage (originally 20170806125915)
 class CreateActiveStorageTables < ActiveRecord::Migration[5.2]
   def change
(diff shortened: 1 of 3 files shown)


feedmypixel:
This is working locally for me but I feel there needs to be a tests around:

`When I create a document`
`And add it to a move`
`Then the move contains the document`

`When I create multiple documents`
`And add them to a move`
`Then the move contains these documents`

feedmypixel:
Approved on the fact that this works for me locally. Would be good to get @spikeheap eyes on it retrospectively

spikeheap:
:+1: LGTM. I've applied the rollback/apply on all our cloud platform environments

PR #188, opened 2020-01-03
Integrate Swagger into move tests

spikeheap:
This PR gets RSwag working, and provides an example of its use for testing requests.

The bulk of the interesting changes are in https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/188/files#diff-600b31dafc21b1314c7d06fed4fca13d, which is completely overhauled to define the requests/responses using the Rswag DSL. This DSL does a bunch of things:

1. It validates the path, response code, and schema of the response.
2. It outputs OpenAPI 3.0.1 definitions for the tested endpoint to a swagger file.

A few things aren't validated by default by Rswag:

1. The `Content-Type` header of the response
2. The content is validated against the schema, but not compared to an expected payload

The `200` response includes these additional tests to demonstrate additional assertions within a `run_test!` call. If we find we prefer the `it(...)` syntax for consistency we can achieve that (see [the docs](https://github.com/rswag/rswag), but I found it to be less readable for this example.

I needed to work around a [bug in rswag](https://github.com/rswag/rswag/issues/268) to create the model in swagger UI automatically.

The biggest challenge in these changes was maintaining the JSON relative file loading, which was pretty painful. We use the relative load paths when validating the JSON schemas in existing tests, and swagger UI loads the paths successfully, however in RSpec tests these relative files would fail to load, because the validator would attempt to load the file from the Rails root directory rather than the swagger JSON directory. I created https://github.com/rswag/rswag/issues/270 to track this.

In the meantime I copied the JSON files into `spec/swagger/definition`, and loaded the files manually into the `definitions` block of our API spec in `swagger_helper.rb`. I then updated these files to use internal references (within `#/definitions`), so schemas validate and the end Swagger document is generated in-line.

Once we have moved our tests to rswag, we can simply remove all the JSON files in `/swagger/v1/`.

To save us from a big-bang migration, I copied every path _except_ `/moves/{moveId}` into https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/188/files#diff-2468a08fddb20e8e07409c15408dab19, an load these in `swagger_helper.rb`. The resulting swagger file and UI are complete, being a mix of hand-coded JSON definitions and the new, automatically generated `/moves/{moveId}` path.

The examples are currently generated from the schemas, however we can provide higher-fidelity examples by automatically saving the responses returned in tests as examples. This is commented out to start with so the generated `swagger.yaml` isn't unique each time, but for reference: https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/188/files#diff-688c6203765aa2db488a0f69b2db33d4R109.

Because of https://github.com/rswag/rswag/issues/268 if we run `rails rswag:specs:swaggerize` to generate the swagger doc it doesn't include the models. I've added a switch to RSpec to generate the swagger doc every run (in `.rspec`), so `bundle exec rspec ./spec/requests` will build a new `swagger/v1/swagger.yaml` as needed.

Here's an example of the generated output:
Screenshot 2020-01-03 16 03 42

and the model section taking `schema`:

Screenshot 2020-01-03 12 49 30

#### Testing locally
To test it out, go to http://localhost:3000/api-docs/index.html. Here you can click the padlock and use your frontend client credentials to authenticate. From this point click "Try it out" and run some sample queries. This has only been tested for `/moves/{moveId}`, but should work for all the others (if they worked previously...)


spec/requests/api/v1/moves_controller_show_spec.rb
```diff
@@ -1,62 +1,86 @@
...
-      get "/api/v1/moves/#{move_id}", headers: headers
-    end
-
-    context 'when successful' do
-      it_behaves_like 'an endpoint that responds with success 200'
-
-      it 'returns the correct data' do
-        expect(response_json).to eq resource_to_json
+  let(:move) { create :move }
+  let(:moveId) { move.id }
```
ngw: Why the capitalization change? In Ruby symbols, methods and variables are always snake_case - https://github.com/rubocop-hq/ruby-style-guide#snake-case-symbols-methods-vars

spec/support/with_client_authentication.rb
```diff
@@ -19,6 +19,9 @@
   end
 
   let(:auth_headers) { { 'Authorization': "Bearer #{@access_token}" } }
```
ngw: Can't you reuse the :Authorization symbol you defined under here?

spikeheap: Good call. `:Authorization` causes all sorts of problems when you try to reference it (because it's capitalised but not a constant), so I've just created another variable which we use in both instead :+1:

spec/requests/api/v1/moves_controller_show_spec.rb
spikeheap: This isn't a capitalisation change – it's porting the existing swagger document and that uses camel case throughout.

That said, I'm happy to change this one, and we can keep consistency for all new tests. At some point we'll need to back-fill the tests for the existing endpoints so we can fix it up then.

spikeheap:
@ngw I've addressed both those comments in the code.

@starswan this is how we're proposing to write functional tests to generate the swagger documentation automatically.

PR #189, opened 2020-01-08
Basic RBAC implementation

ngw:
Solves https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=302&projectKey=P4&modal=detail&selectedIssue=P4-794

Implements a basic RBAC model that permits the association of a Supplier to an OAuth application.


config/environments/development.rb
```diff
@@ -53,4 +53,7 @@
 
   # Default host for url generation
   routes.default_url_options[:host] = "localhost:#{ENV['PORT'] || 3000}"
+
+  # Enable curl for testing oauth flow
+  config.action_controller.allow_forgery_protection = false
```
spikeheap: What error does this resolve? I'd expect our endpoints to work with cURL (or any non-browser system) in production.

config/initializers/doorkeeper.rb
```diff
@@ -155,15 +155,15 @@
...
+  enable_application_owner confirmation: false
 
   # Define access token scopes for your provider
   # For more information go to
   # https://github.com/doorkeeper-gem/doorkeeper/wiki/Using-Scopes
-  #
+  
   # default_scopes  :public
-  # optional_scopes :write, :update
+  # optional_scopes :read, :write
```
spikeheap: This line's been changed, but it's still a comment. Is that intentional?

lib/tasks/create_application.rake
```diff
@@ -1,22 +1,39 @@
...
+    name = prompt.ask("What's the application's name?")
+
+    scopes = %w[read write]
+    scope = prompt.multi_select("What's the scope of the #{name} application?", scopes)
+
+    application = Doorkeeper::Application.new(name: name, scopes: scope)
+
+    suppliers = Supplier.all.each_with_object({}) do |s, hsh|
+      hsh[s.name] = s.id
+    end.merge('No owner' => nil)
```
spikeheap: `hsh` and `s` aren't meaningful names. Could this be refactored to (e.g.) `memo` and `supplier`?

lib/tasks/create_application.rake
```diff
@@ -1,22 +1,39 @@
...
+    owner_id = prompt.select('Is this application owned by a supplier?', suppliers)
+    if owner_id != 'No owner'
```
spikeheap: Should `'No owner'` be a constant so we can guarantee consistency with L21?

spec/requests/api/v1/moves_controller_destroy_spec.rb
```diff
@@ -3,6 +3,7 @@
 require 'rails_helper'
 
 RSpec.describe Api::V1::MovesController, with_client_authentication: true do
+  let!(:application) { Doorkeeper::Application.create(name: 'test') }
```
spikeheap: This line looks like it's repeated quite a bit. Would it be a good candidate to refactor into (or in the same way as) `with_client_authentication`

spec/requests/api/v1/moves_controller_index_spec.rb
```diff
@@ -32,14 +33,19 @@
...
+        end
 
         it 'delegates the query execution to Moves::Finder with the correct filters', skip_before: true do
           moves_finder = instance_double('Moves::Finder', call: Move.all)
           allow(Moves::Finder).to receive(:new).and_return(moves_finder)
 
           get '/api/v1/moves', headers: headers, params: params
 
-          expect(Moves::Finder).to have_received(:new).with(from_location_id: from_location_id)
+          expect(Moves::Finder).to have_received(:new).with({ from_location_id: from_location_id }, ability)
```
spikeheap: I know you haven't introduced this mocking (🤷‍♂) , but would this be a good opportunity to remove it? I'd expect our functional tests to build test data and assert the behaviour, rather than checking a particular method has been called somewhere. This doesn't block this one, but maybe a candidate for another PR?

spec/requests/api/v1/moves_controller_role_access_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+  describe 'GET /moves' do
+    let!(:application) { Doorkeeper::Application.create(name: 'test', owner: supplier) }
+
+    let(:schema) { load_json_schema('get_moves_responses.json') }
+
+    let(:supplier) { create :supplier }
+    let(:pentonville) { create :location, suppliers: [supplier] }
+    let(:birmingham) { create :location, key: 'hmp_birmingham', title: 'HMP Birmingham', nomis_agency_id: 'BMI' }
+    let!(:pentonville_moves) { create_list :move, 10, from_location: pentonville }
+    let!(:birmingham_moves) { create_list :move, 10, from_location: birmingham }
```
spikeheap: Could this be a `with_moves` trait on the `location` factory? It feels like we do/should need suppliers with moves for a bunch of our tests.

spec/requests/api/v1/moves_controller_show_spec.rb
```diff
@@ -44,20 +40,12 @@
...
     end
 
     context 'with an invalid CONTENT_TYPE header' do
       let(:content_type) { 'application/xml' }
 
       it_behaves_like 'an endpoint that responds with error 415'
-
-      it 'doesn\'t sync data' do
-        expect(Moves::NomisSynchroniser).not_to have_received(:new)
-      end
```
spikeheap: Is this test not needed anymore?

spec/requests/api/v1/moves_controller_role_access_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+      let(:detail_404) { "Couldn't find Move with 'id'=#{birmingham_move.id} [WHERE (from_location_id IN ('#{pentonville.id}'))]" }
+
+      before do
+        get "/api/v1/moves/#{birmingham_move.id}", headers: headers
+      end
+
+      it_behaves_like 'an endpoint that responds with error 404'
+    end
+  end
+end
```
spikeheap: In preparation for https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/188, would you mind refactoring this into the Rswag DSL, so we don't need to revisit it later?

app/models/ability.rb
```diff
@@ -0,0 +1,13 @@
...
+  include CanCan::Ability
+
+  def initialize(user = nil)
+    if user
+      can :manage, Move, Move.served_by(user)
+    else
+      can :manage, Move
+    end
+  end
+end
```
spikeheap: This needs tests, so we can build on them as it becomes more complex. Something like https://github.com/CanCanCommunity/cancancan/wiki/Testing-Abilities#rspec would be great.

spikeheap: If we're introducing CanCanCan as an authorisation layer, we should plan to abstract the Doorkeeper authorisation checks into here. With that in mind, could we have a default of "can do nothing"? As an example, in the above we could receive the Doorkeeper application and only give the `:manage` ability to authenticated users, scoping it down if the supplier is present.

ngw: Correct me if I'm wrong: the doorkeeper authentication is in the `api_controller`, if someone is unauthenticated it won't be able to do anything and will never get to `cancancan`, which will evaluate `current_user` only AFTER authentication, right?

config/initializers/doorkeeper.rb
ngw: Yes, I've started the task using scopes and then moved to `cancancan`

app/models/ability.rb
spikeheap: That is correct, but it makes me a bit nervous to split our authorisation scheme across two mechanisms as it's easy to see a flaw being introduced in future, e.g. if we introduced unauthenticated endpoints. If we build this ability assuming nothing it should be easier to test, and describes our authorisation rules to a new member of the team without needing to dig through other parts of the system

spec/requests/api/v1/moves_controller_show_spec.rb
ngw: Well what's the point of expecting "not to" do something?

spikeheap: If, for example, we wanted to assert that the NomisSynchroniser wasn't called inappropriately. What was the rationale when this test was added. It'd be good to understand that in order to be comfortable removing this.

spec/requests/api/v1/moves_controller_role_access_spec.rb
ngw: I don't have experience with that, do you have any reference?

spec/requests/api/v1/moves_controller_show_spec.rb
ngw: If I remember correctly, we used to sync with Nomis under certain conditions in the /show action, but that has been removed when we discovered it was messing with duplicates. The tests don't fail because they weren't very relevant at the time and they're even less relevant now. There's no reference to the `NomisSynchroniser` right now in the whole action.

config/environments/development.rb
ngw: Leftover from one of my tests, sorry

spec/requests/api/v1/moves_controller_destroy_spec.rb
ngw: Solved.

spec/requests/api/v1/moves_controller_index_spec.rb
ngw: I can do it, but I think that being a rather big change in direction we must think about duplication.
Finders are actually "in charge" of retrieving the data, and are tested. I have nothing against removing the mocks but we'll most probably end up with a lot of overlapping tests on the same functionalities.
IMVHO it makes sense to have request specs mock the finders and test how a request to the controller turns into a call to the service object and then test the service object in isolation.

spec/requests/api/v1/moves_controller_role_access_spec.rb
ngw: Done in the supplier filter branch, will "integrate" properly once the other PR reaches master.

spec/requests/api/v1/moves_controller_index_spec.rb
spikeheap: Happy to leave this for this PR. I'd assumed we use the request specs as a functional end-to-end test, but if that's not the case we'll want to think about how we introduce the latter. Maybe something to discuss with @starswan and Martyn once he's settled in next week?

spec/requests/api/v1/moves_controller_role_access_spec.rb
spikeheap: :+1:

spec/requests/api/v1/moves_controller_role_access_spec.rb
spikeheap: #188 should be a good reference (the `/moves/:id` route is refactored into Rswag), alongside the https://github.com/rswag/rswag docs in the repo. Happy to jump on a call/pair/etc. 

Also happy to merge this PR and see this refactor in a separate PR to get the bulk of these changes merged...

spec/requests/api/v1/moves_controller_show_spec.rb
spikeheap: > There's no reference to the NomisSynchroniser right now in the whole action.

Sounds good to remove it then :+1:

spec/requests/api/v1/moves_controller_role_access_spec.rb
```diff
@@ -0,0 +1,267 @@
...
+      end
+    end
+  end
+
+  path '/moves/{move_id}' do
+    let!(:pentonville_move) { create :move, from_location: pentonville }
+    let!(:birmingham_move) { create :move, from_location: birmingham }
+
+    get 'Returns the details of a move' do
+      tags 'RBAC'
```
spikeheap: `tags` is used to group calls in the UI, so I think this wants to be "Moves"

PR #193, opened 2020-01-10
added basic test coverage support

starswan:


spikeheap:
Great to see us starting to track coverage and prevent regressions 🙌

PR #199, opened 2020-01-14
Supplier/Location rake task replaces locations for each supplier

spikeheap:
Prior to this commit the rake task would attempt to add any locations listed in the hash to each supplier, but would not remove any already in place.

This commit alters that behaviour to make the list in the Rake task authoritative. This enables (for example) moving a location from one supplier to another with a single change/command.

### DIFF
--- a/lib/tasks/reference_data.rake
+++ b/lib/tasks/reference_data.rake
@@ -161,14 +161,12 @@ namespace :reference_data do
       ]
     }
 
+    # Set the supplier locations. This is an authoritative change, making the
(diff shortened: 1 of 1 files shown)


spikeheap:
I'd like to add some tests around this as there are none at present.

ngw:
Maybe you can build some kind of "importer" and call it from the rake mimicking the same interface?

spikeheap:
I've tested this locally, but want to write an automated test. I'll submit that as a separate PR so this branch doesn't end up long-lived

PR #201, opened 2020-01-14
P4 881 log nomis api calls

starswan:
Refactored tests to support new structure. Added logging to moves importer and sweeper


app/services/people/importer.rb
```diff
@@ -15,9 +15,10 @@ def initialize(nomis_attributes)
     end
 
     def call
-      person = Person.find_or_create_by!(nomis_prison_number: nomis_attributes[:prison_number])
+      person = Person.find_or_initialize_by(nomis_prison_number: nomis_attributes[:prison_number])
```
spikeheap: What's the intention behind initialising instead of creating? If I'm reading this correctly, an effect of this change is that the returned `person` may or may not be persisted, which may not be obvious to consumers.

app/services/people/importer.rb
```diff
@@ -15,9 +15,10 @@ def initialize(nomis_attributes)
...
       profile = person.latest_profile || person.profiles.build
-      profile.update(attributes)
+      profile.update_attributes(attributes)
```
spikeheap: It looks like `update_attributes` is just an alias of `update` (https://apidock.com/rails/ActiveRecord/Persistence/update_attributes). Is there a reason we're switching?

app/services/personal_care_needs/importer.rb
```diff
@@ -17,7 +17,6 @@ def call
         personal_care_needs.map { |personal_care_need| build(personal_care_need) },
         ASSESSMENT_ANSWER_CATEGORY
       )
-      profile.save!
```
spikeheap: Would it be worth adding a comment here to point out that `merge_assessment_answers` triggers a save? This `save!` statement does appear unnecessary, but without it we have an implicit reliance on `merge_assessment_answers` doing a save. Any ideas on how we could be defensive here?

config/environments/production.rb
```diff
@@ -39,9 +39,9 @@
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
   # config.force_ssl = true
 
-  # Use the lowest log level to ensure availability of diagnostic information
-  # when problems arise.
-  config.log_level = :warn
+  # set to :info so that logging for synchroniser shows up in staging/preprod/production
+  #
+  config.log_level = :info
```
spikeheap: :+1:

spec/services/alerts/importer_spec.rb
```diff
@@ -4,17 +4,15 @@
 
 RSpec.describe Alerts::Importer do
   subject(:importer) do
-    described_class.new(
-      profile: profile,
-      alerts: alerts
-    )
+    Moves::Importer.new(moves)
```
spikeheap: This looks like it changes the importer being tested from `Alerts::Importer` to `Moves::Importer`. Is that intentional, and if so should the test be moved/renamed to reflect it?

spec/services/alerts/importer_spec.rb
```diff
@@ -4,17 +4,15 @@
...
   end
 
-  let(:person) { create :person }
+  let(:person) { create :person, nomis_prison_number: 'G3239GV' }
```
spikeheap: `nomis_prison_number` doesn't look like it needs to be explicit here. Could it be added to the factory so we have an `:prisoner` person type which automatically gets a faked NOMIS prison number?

spec/services/moves/importer_spec.rb
```diff
@@ -37,34 +37,32 @@
   let!(:profile_one) { create(:profile, person: prisoner_one) }
   let!(:profile_two) { create(:profile, person: prisoner_two) }
 
-  let(:people_importer) { instance_double('People::Importer', call: true) }
+  # let(:people_importer) { instance_double('People::Importer', call: true) }
```
spikeheap: Looks like this commented-out code snuck into a commit

spec/services/moves/importer_spec.rb
```diff
@@ -37,34 +37,32 @@
...
     # create fallback questions for PersonalCareNeeds importer and Alerts importer
     create(:assessment_question, :care_needs_fallback)
     create(:assessment_question, :alerts_fallback)
   end
 
-  it 'calls the People::Importer service twice' do
-    importer.call
-    expect(people_importer).to have_received(:call).twice
-  end
-
```
spikeheap: Do we not want to assert the service being called anymore?

spec/services/personal_care_needs/importer_spec.rb
```diff
@@ -4,17 +4,15 @@
...
   subject(:importer) do
-    described_class.new(
-      profile: profile,
-      personal_care_needs: personal_care_needs
-    )
+    Moves::Importer.new(moves)
   end
 
-  let(:person) { create :person }
+  let(:person) { create :person, nomis_prison_number: 'G3239GV' }
```
spikeheap: (also in previous comment), looks like a good candidate for a factory trait/type

app/services/moves/importer.rb
```diff
@@ -10,63 +10,72 @@ def initialize(items)
...
       end
     end
-
-    def update_move_with_same_nomis_event_id(move)
-      same_nomis_event_id_move = Move.find_by_nomis_event_ids([move[:nomis_event_id]])
-      same_nomis_event_id_move&.update(move_params(move))
-    end
+    # rubocop:enable Style/GuardClause
+    # rubocop:enable Metrics/MethodLength
+    # rubocop:enable Metrics/AbcSize
```
spikeheap: This refactor seems to introduce two complex, long methods which violate our Rubocop conventions. Could this be submitted in a separate PR to the logging and test changes, so we can demonstrate the tests are successful against the original code?

app/services/people/importer.rb
starswan: There is only 1 consumer. These 'services' are less than ideal as the contract between them is less than obvious. Yes I've changed the contract as I wanted to do all the persistence in the same place (and know when we had created a new perosn record)

app/services/people/importer.rb
starswan: Yes this should be assign_attributes...

app/services/personal_care_needs/importer.rb
starswan: merge_assessment_answers! doesn't trigger a save.

spec/services/alerts/importer_spec.rb
starswan: Yes its intentional. No the test shouldn't be renamed, as it's goal is still to test the Alerts part of the importer. The original test was poor because it tested the Alerts importer in isolation and so failed to find several defects. Configuring the Moves::Importer with a focus on the Alerts::Importer rather than creating a 'naked' Alerts::Importer results in a much more effective test

spec/services/alerts/importer_spec.rb
starswan: +1

starswan: +1

spec/services/moves/importer_spec.rb
starswan: Nope - see comments above. The test now uses more concrete classes - the fact that the people_importer was called twice (note not even a check for arguments) wasn't a very strong test anyway.

app/services/moves/importer.rb
starswan: The changes are needed to support the logging. IMHO this new method isn't particularily long or complex (our Rubocop method length is set to the default of 10 which is extremely  short). Open to suggestions.

starswan: These exceptions all vanish now we have the GDS style guides in place

PR #206, opened 2020-01-16
P4 885 associate move with profile

starswan:


app/serializers/move_serializer.rb
```diff
@@ -4,13 +4,13 @@ class MoveSerializer < ActiveModel::Serializer
...
 
-  has_one :person, serializer: PersonSerializer
+  has_one :profile, serializer: ProfileSerializer
   has_one :from_location, serializer: LocationSerializer
   has_one :to_location, serializer: LocationSerializer
   has_many :documents, serializer: DocumentSerializer
 
   INCLUDED_ATTRIBUTES = {
-    person: %i[first_names last_name date_of_birth assessment_answers indentifiers ethnicity gender],
+    profile: %i[first_names last_name date_of_birth assessment_answers indentifiers ethnicity gender],
```
spikeheap: This looks like it removes an attribute the API response, which we're not able to do. The API is closed for modification, but open for extension. As galling as it is to introduce inconsistency, we may need to keep the `person` name here.

starswan: pretty sure it doesn't do that. There are no test changes in this PR. The new 'ProfileSerializer' class has a type 'people' line to keep the API consistent

db/migrate/20200116083122_associate_move_with_profile.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+class AssociateMoveWithProfile < ActiveRecord::Migration[5.2]
```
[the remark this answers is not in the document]
starswan: Yes maybe? I used a hand-rolled migratoion system years ago, and we always did the data along with the schema so we didn't forget (apart from anything else). But yes, don't really know the best way to do this (or if there is a best way)

PR #220, opened 2020-01-28
P4 885 part 1 migration

starswan:


db/migrate/20200127150100_associate_move_with_profile_part1.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class AssociateMoveWithProfilePart1 < ActiveRecord::Migration[5.2]
+  def up
+    change_table :moves do |t|
+      t.uuid :profile_id
```
martyn-w: @starswan could you make this a foreign key?

starswan: it comes later I think

PR #221, opened 2020-01-28
P4 885 part 2 data fixup

starswan:


martyn-w:
@starswan could you pull in the latest master into your branch ?

app/serializers/profile_serializer.rb
```diff
@@ -0,0 +1,27 @@
...
+  )
+
+  # although we are serialising a 'profile', to the outside world this is a person.
+  type 'people'
+
+  has_one :ethnicity, serializer: EthnicitySerializer
+  has_one :gender, serializer: GenderSerializer
+
+  def id
+    object.person.id
```
martyn-w: @starswan should this definitely be the `person.id` and not the `profile.id` ? There is a danger that different profiles for the same person could be returned - each with the same apparent id. That might confuse api consumers?

starswan: Yes - there's no profile endpoint. The profile returned by the person endpoint is always the 'latest_profile' (yes I'm not sure whether this works either)

martyn-w:
@starswan can you pull in the latest master branch?

PR #228, opened 2020-01-31
added special vehicle assessment question

ngw:
Added special vehicle assessment question

https://dsdmoj.atlassian.net/browse/P4-926


app/services/assessment_questions/importer.rb
```diff
@@ -18,6 +18,7 @@ class Importer
       { key: :solicitor, category: :court, title: 'Solicitor or other legal representation' },
       { key: :interpreter, category: :court, title: 'Sign or other language interpreter' },
       { key: :other_court, category: :court, title: 'Any other information' },
+      { key: :special_vehicle, category: :health, title: 'Special vehicle' },
```
teneightfive: Maybe move to line 18 so the categories are still grouped?

ngw: Amended!

PR #230, opened 2020-02-03
P4 922 sync completed moves

starswan:
sync completed moved by setting movements=true in NOMIS API calls

### DIFF
--- a/app/lib/nomis_client/moves.rb
+++ b/app/lib/nomis_client/moves.rb
@@ -3,6 +3,8 @@
 module NomisClient
   class Moves
     class << self
+      MAX_POSSIBLE_MOVES_IN_ONE_DAY = 500
+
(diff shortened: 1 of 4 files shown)


app/lib/nomis_client/moves.rb
```diff
@@ -3,6 +3,8 @@
 module NomisClient
   class Moves
     class << self
+      MAX_POSSIBLE_MOVES_IN_ONE_DAY = 500
```
martyn-w: @starswan what is the purpose of this constant and is 500 a reasonable value? I thought there are about ~600k moves a year which is roughly ~2k moves per day?

starswan: It's per-institution. I introduced it as a named constant - it used to be a hard-coded value

PR #233, opened 2020-02-03
Remove swagger yaml

starswan:
Take 2 at removing the swagger.yaml file from source control. 
Needed to save it in a circle:ci workspace


martyn-w:
To note: `swagger.yml` is being removed as it is frequently regenerated in a random order

PR #234, opened 2020-02-05
P4 951 webhook models

martyn-w:
new models for webhooks


Gemfile.lock
```diff
@@ -359,10 +362,11 @@ DEPENDENCIES
...
   tty-prompt
+  validate_url (~> 1.0.8)
   webmock
 
 RUBY VERSION
    ruby 2.6.2p47
 
 BUNDLED WITH
-   2.0.2
+   2.1.4
```
starswan: changing the bundler version doesn't seem part of this PR?

spec/services/encryptor_spec.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
```
starswan: please don't resurrect this file

swagger/v1/swagger.yaml
```diff
@@ -0,0 +1,1153 @@
+---
```
starswan: please don't resurrect this file

db/migrate/20200205151906_make_subscription_secret_encrypted.rb
```diff
@@ -0,0 +1,5 @@
+class MakeSubscriptionSecretEncrypted < ActiveRecord::Migration[5.2]
+  def change
+    rename_column :subscriptions, :secret, :encrypted_secret
```
starswan: It would have been nicer to have this in the original migration. Also I'm not sure whether 'string' is the correct datatype for encrypted fields? I've often has to use binary-type DB fields in tha past

app/serializers/subscription_serializer.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
```
starswan: I don't understand what the purpose of this class is? It isn't linked into the code in any way (it only has unit tests)

app/models/move.rb
```diff
@@ -33,6 +33,7 @@ class Move < ApplicationRecord
   belongs_to :to_location, class_name: 'Location', optional: true
   belongs_to :person
   has_many :documents, dependent: :destroy
+  has_many :notifications, as: :topic, dependent: :destroy # NB: polymorphic association
```
starswan: I'm not sure why this is polymorphic? I don't see the need for it in the current codebase?

app/services/encryptor.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+class Encryptor
+  KEY = ActiveSupport::KeyGenerator.new(
+    ENV.fetch('SECRET_KEY_BASE'),
```
starswan: Would it be more idiomatic to use Rails.application.secret_key_base here?

spec/factories/subscriptions.rb
```diff
@@ -0,0 +1,10 @@
+FactoryBot.define do
+  factory :subscription do
+    association :supplier
+    callback_url { 'http://foo.bar/?bla=bla' }
+    topic { '*' }
```
starswan: slightly confused by this, as subscription doesn't appear to have a 'topic' property

db/schema.rb
```diff
@@ -208,6 +229,20 @@
     t.integer "latest_nomis_booking_id"
   end
 
+  create_table "subscriptions", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.uuid "supplier_id", null: false
+    t.string "topic", default: "*", null: false
```
starswan: Not sure what the purpose of this 'topic' field is. I'm finding it difficult to tell the difference between this 'topic' and the 'topic' in notifications/subscription relationship.

spec/models/subscription_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+  end
+
+  describe 'encrypted_secret' do
+    subject(:subscription) { create(:subscription) }
+
+    context 'when writing' do
+      it { expect(subscription.encrypted_secret).not_to eq('Secret') }
+    end
+  end
```
starswan: I think you need a test to prove that you can retrieve a secret back from the DB.

spec/serializers/subscription_serializer_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  it 'contains an id property' do
+    expect(result[:data][:id]).to eql subscription.id
+  end
+
+  it 'contains a first_names attribute' do
+    expect(result[:data][:attributes][:callback_url]).to eql 'http://foo.bar/?bla=bla'
+  end
+
+  it 'contains a last_name attribute' do
```
starswan: This typo looks like a copy-n-paste test. IMHO it would be better to simply assert the contents of result[:data]

spec/services/encryptor_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+require 'rails_helper'
+
+RSpec.describe Encryptor do
+  subject { described_class }
+
+  let(:secret_key_base) { SecureRandom.hex(64) }
+  let(:encryptor_salt) { SecureRandom.random_bytes(ActiveSupport::MessageEncryptor.key_len) }
+
+  before do
```
starswan: is this required?

app/models/subscription.rb
```diff
@@ -0,0 +1,15 @@
+class Subscription < ApplicationRecord
+  has_many :notifications, dependent: :restrict_with_error
```
starswan: This implies that once we have a set of notifications (which hang around forever?) then we can't remove a subscription. But supplier has_many subscriptions with cascade destroy - so we can't remove a supplier either. IMHO this decision needs a least a supporting comment

spec/services/encryptor_spec.rb
starswan: Because it covers the decryption path I think this test hides the 'missing decryption test' (see other comments).

app/services/encryptor.rb
martyn-w: @starswan good point 👍

spec/services/encryptor_spec.rb
martyn-w: now removed 👍

db/migrate/20200205151906_make_subscription_secret_encrypted.rb
martyn-w: @starswan agreed 1 migration would be simpler 👍 
Regarding the datatype, the output from ActiveSupport::MessageEncryptor is a (base64?) string, and so the string datatype is perfectly suited:
```
Encryptor.encrypt('foo')
=> "0AK1bcbDunemQ5akCQ==--huW00cOaMYxsWWuO--vnr6Px1ghnOkjUpdQbV+eA=="
```

starswan: Ok that wasn't obvious to me. Maybe a comment is required?

martyn-w: @starswan a comment will be added 👍

app/models/move.rb
martyn-w: @starswan it is polymorphic because other classes will have notifications, e.g.  people/profiles

swagger/v1/swagger.yaml
martyn-w: oops I will de-zombify that

Gemfile.lock
martyn-w: sorry, my bad

db/schema.rb
martyn-w: @starswan the topic field here was intended as a filter, so that you could subscribe to a particular move or person, rather than all of them. However it is not in the requirements so I'll remove it for now

app/models/subscription.rb
martyn-w: The notification records are like move records and shouldn't be hard-deleted...
...I've just see there is another PR to soft-delete moves, so will follow that pattern here

app/serializers/subscription_serializer.rb
martyn-w: @starswan it was left over from when there was an API for administering subscriptions: but that has been descoped so this can be removed 👍

spec/serializers/subscription_serializer_spec.rb
martyn-w: this file has been deleted

spec/models/subscription_spec.rb
martyn-w: @starswan surely the test above does that (line 31)?:
```
  context 'when writing' do
      it { expect(subscription.secret).to eq('Secret') }
    end
```

How else should this be tested? do you mean doing a `.reload` on the `subscription` first?

spec/services/encryptor_spec.rb
martyn-w: @starswan I'm afraid I disagree: this is testing encryption and decryption, as opposed to reading an encrypted field in the db and automatically decrypting it.  I.e. it seems perfectly legitimate to test encryption outside of the database.

 Could you clarify what the issue is? Perhaps a quick call would help?

starswan: I'm not sure that it tests enough. If you disagree that's fine, I'm probably wrong

starswan: sorry my mistake - combination of mis-reading several things.

PR #237, opened 2020-02-06
P4-822 Papertrail for auditing writes

starswan:
Added Papertrail to record versioning (and who did the changes) to 3 API models.


app/controllers/api/v1/moves_controller.rb
```diff
@@ -90,7 +90,9 @@ def find_move
       end
 
       def import_moves_from_nomis
-        Moves::NomisSynchroniser.new(locations: from_locations, date: date).call
+        PaperTrail.request(whodunnit: nil) do
+          Moves::NomisSynchroniser.new(locations: from_locations, date: date).call
+        end
```
spikeheap: :+1: to removing the current user from synchroniser changes. Very neat. Would it be worth adding an intentional comment saying why we decided not to associate those changes with a user?

app/controllers/api_controller.rb
```diff
@@ -17,6 +18,10 @@ def current_user
     doorkeeper_token&.application
   end
 
+  def user_for_paper_trail
+    current_user.owner_id
+  end
+
```
spikeheap: :+1: it looks like this will extend nicely once we have the user information from the frontend :100:

app/models/person.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class Person < ApplicationRecord
+  has_paper_trail
+
```
spikeheap: Should we consider adding this to the `ApplicationRecord` so models have versioning by default?

app/services/moves/importer.rb
```diff
@@ -28,7 +28,6 @@ def import_people
       people.each do |offender_no, person_data|
         profile = People::Importer.new(person_data).call
         new_person_count += 1 if profile.new_record?
-        profile.save!
```
spikeheap: Where does the save occur if we remove this line? I'm assuming there's a test which would fail if it didn't, but `profile.changed?` on L34/35 returns `false` if the record was never persisted.

app/services/moves/importer.rb
```diff
@@ -56,7 +55,10 @@ def import_moves
           end
         else
           new_count += 1
-          new_move.save!
+          new_move.save
+          unless new_move.persisted?
+            raise new_move.errors
+          end
```
spikeheap: What's the intention behind this? I'm probably just missing something obvious, but wouldn't `new_move.save!` raise consistently with L54/55 for an existing move?

db/migrate/20200203145050_create_versions.rb
```diff
@@ -0,0 +1,23 @@
...
+  # The largest text column available in all supported RDBMS is
+  # 1024^3 - 1 bytes, roughly one gibibyte.  We specify a size
+  # so that MySQL will use `longtext` instead of `text`.  Otherwise,
+  # when serializing very large objects, `text` might not be big enough.
+  TEXT_BYTES = 1_073_741_823
+
+  def change
+    create_table :versions do |t|
+      t.string   :item_type, {:null=>false}
+      t.integer  :item_id,   null: false, limit: 8
```
spikeheap: Should ` item_id` be a UUID rather than an integer? I think all our models use UUIDs

starswan: Yes - now I'm curious how any of the tests pass

app/services/moves/importer.rb
starswan: That was an unintentional commit that I thought I had removed

app/models/person.rb
starswan: do we want to audit every model including locations etc? Otherwise I agree it would make sense to have a common base class for auditable things

app/controllers/api/v1/moves_controller.rb
starswan: Yes agreed. Big fan of intention comments, not always perfect at writing them

app/services/moves/importer.rb
starswan: See other branch. profile.changed? actually returns true in this scenario, hence why the test doesn't fail

db/migrate/20200203145050_create_versions.rb
starswan: good spot - it just happens to work by accident.

spikeheap:
Hi @starswan, can this PR be trimmed down to _just_ introduce papertrail? There's a bunch of changes which don't seem related to that change (e.g. spec/support/with_client_authentication.rb, spec/requests/api/v1/moves_controller_create_spec.rb) which should be in separate PRs. I'm :+1: with the papertrail changes, so don't want questions around the other bits to stop that from being merged...

app/models/profile.rb
```diff
@@ -20,11 +20,42 @@ class Profile < ApplicationRecord
...
+    end
+    changed = profile_identifiers.select do |old|
+      new_id = new_identifiers.detect { |pi| pi[:identifier_type] == old.identifier_type }
+      new_id && new_id[:value] != old.value
+    end
+
+    unless deleted.empty? && inserted.empty? && changed.empty?
+      # self[:profile_identifiers] = new_identifiers
+      super
+    end
```
spikeheap: Hi @starswan, as we discussed these blocks could do with some comments explaining their purpose. Once they're in this is good to merge

PR #238, opened 2020-02-06
remove webmock (for now) as it doesn't work in this project

starswan:
Something to do with our usage of OAuth2::Client means that webmock doesn't work in this project. As we have a consistent API for accessing NOMIS, the lack of webmock can be effectively worked-around by mocking NomisClient::Moves#get_response (et al)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -50,7 +50,6 @@ group :development, :test do
   gem 'rspec-rails'
   gem 'shoulda-matchers'
   gem 'timecop'
-  gem 'webmock'
 end
 
 group :development do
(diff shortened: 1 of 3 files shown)


spikeheap:
Presumably we're not using WebMock in any of our tests if it can be removed with just those 3 changes? In principle I'm happy to remove a dependency if we're not using it :D.

My only concern around this is that it _looks_ like WebMock disables external network connections as part of its setup. Are we sure we want to remove that? 

Has CircleCI been configured with (for example) T3 credentials? If so, we might inadvertently end up with tests that make network requests.

starswan:
Webmock was in the project but was not being used. There appears to be a conflict between it and our JSON/token handling in our API layer that stops it working in weird slightly silent ways (.parsed method returns nil rather than a JSON object). I attempted to use it and got burnt spending 2 hours down a rabbit-hole - so thought it was worth taking out to prevent further confusion

PR #239, opened 2020-02-06
remove duplicated profile save during move imports

starswan:
Remove duplicate save! call on profile when importing people

### DIFF
--- a/app/services/moves/importer.rb
+++ b/app/services/moves/importer.rb
@@ -28,7 +28,6 @@ module Moves
       people.each do |offender_no, person_data|
         profile = People::Importer.new(person_data).call
         new_person_count += 1 if profile.new_record?
-        profile.save!
         Alerts::Importer.new(profile: profile, alerts: alerts.fetch(offender_no, [])).call
         PersonalCareNeeds::Importer.new(profile: profile,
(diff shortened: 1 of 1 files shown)


app/services/moves/importer.rb
```diff
@@ -28,7 +28,6 @@ def import_people
       people.each do |offender_no, person_data|
         profile = People::Importer.new(person_data).call
         new_person_count += 1 if profile.new_record?
-        profile.save!
```
spikeheap: Copying comment over from https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/237#discussion_r376307736:
> Where does the save occur if we remove this line? I'm assuming there's a test which would fail if it didn't, but profile.changed? on L34/35 returns false if the record was never persisted.

starswan: Your assertion is false. Profile.new(first_name: 'fred').changed? returns true

spikeheap: Apologies, when I tested it in the console I just updated the `:id` attribute, which then returned `false`. 🤦‍♂

starswan: Yes it's a bit violation of principle of least surprise I agree.

PR #240, opened 2020-02-06
Add ADR to keep track of the Architech Decisions

alexdesi:
This PR add a template for ADR (see [Michael Nygard's template](https://github.com/joelparkerhenderson/architecture_decision_record/blob/master/adr_template_by_michael_nygard.md) )

It also adds a simple ADR about API params naming.

### DIFF
--- a/adr/0002-API-params-naming.md
+++ b/adr/0002-API-params-naming.md
@@ -1,0 +1,26 @@
+# 2. Name API parameters to match DPS APIs
+Date: 2020-02-06
+
+## Status
+
+Accepted
+
+## Context
+
+The naming of the params used in the APIs needs be consistent.
+
+For example:
+nomis_offender_id, prison_number and nomis_prison_number refer to the same data.
+
+## Decision
+Use param names similar to those used on [DPS APIs](https://gateway.prod.nomis-api.service.hmpps.dsd.io/elite2api/swagger-ui.html#//offenders/getAddressesByOffenderNoUsingGET)
+
+Use snake_case in place of camelCase.
+Consider to prefix the param with `nomis_` to clarify the source.
+
+For example:
+`offenderNo` -> `nomis_offender_no`
+
+
+## Consequences
+We will be able achieve consistency on naming API params.
--- a/adr/0001-record-architecture-decisions.md
+++ b/adr/0001-record-architecture-decisions.md
@@ -1,0 +1,19 @@
+# 1. Record architecture decisions
+
+Date: 2020-02-06
+
+## Status
+
+Accepted
+
+## Context
+
+We need to record the architectural decisions made on this project (Pecs).
+
+## Decision
+
+We will use Architecture Decision Records, as [described by Michael Nygard](https://github.com/joelparkerhenderson/architecture_decision_record/blob/master/adr_template_by_michael_nygard.md).
+
+## Consequences
+
+We can captures an important architectural decision made along with its context and consequences.


adr/0002-API-params-naming.md
```diff
@@ -0,0 +1,27 @@
+# 1. Record architecture decisions
```
spikeheap: Should this be `2. Name API parameters to match DPS APIs` or similar?

alexdesi: good point.

adr/0002-API-params-naming.md
```diff
@@ -0,0 +1,27 @@
...
+## Context
+
+The naming of the params used in the API needs be consistent.
+
+For example:
+nomis_offender_id, prison_number and nomis_prison_number refer to the same data.
+
+## Decision
+Use param names similar to those used on [DPS APIs](https://gateway.prod.nomis-api.service.hmpps.dsd.io/elite2api/swagger-ui.html#//offenders/getAddressesByOffenderNoUsingGET)
+
```
starswan: sadly Elite2 is uncredibly inconsistent, so following Elite2 naming conventions isn't possible as they use different names all the time

adr/0002-API-params-naming.md
```diff
@@ -0,0 +1,27 @@
...
+## Context
+
+The naming of the params used in the API needs be consistent.
+
+For example:
+nomis_offender_id, prison_number and nomis_prison_number refer to the same data.
+
+## Decision
+Use param names similar to those used on [DPS APIs](https://gateway.prod.nomis-api.service.hmpps.dsd.io/elite2api/swagger-ui.html#//offenders/getAddressesByOffenderNoUsingGET)
```
starswan: The 'DPS API' is usually referred to as 'Elite2' in the MoJ

adr/0002-API-params-naming.md
```diff
@@ -0,0 +1,27 @@
...
+
+Use snake_case in place of camelCase.
+Consider to prefix the param with `nomis_` to clarify the source.
+
+For example:
+`offenderNo` -> `nomis_offender_no`
```
starswan: Unfortunately I believe this is the opposite of the decision that we have actually taken. I think we have decided to use prison_number for this field throughout.

starswan: and actually the word 'consider' implies an optional behaviour rather than a decision?

adr/0002-API-params-naming.md
spikeheap: Hi @starswan do you have an example of this? `offenderId` and `bookingId` appear to be used consistently in https://gateway.prod.nomis-api.service.hmpps.dsd.io/elite2api/swagger-ui.html

adr/0002-API-params-naming.md
spikeheap: DPS have a number of APIs, of which elite2 is one. To aid comprehension, maybe this could be "Use param names similar to those used on Digital Prison Studio APIs, such as [elite2]"?

adr/0002-API-params-naming.md
spikeheap: Hi @starswan this is a new decision to introduce consistency with DPS and remove an ambiguous identifier. As a result of this decision, we will change all references to `prison_number` (except if/where it's published in our Swagger docs).

alexdesi: @starswan I'll remove consider, it is  ambiguous

alexdesi: I am going to merge this, @starswan in case we'll find cases that for some reasons cannon use this convention, we can amend the ADR.

PR #242, opened 2020-02-07
re-work serializers - remove PersonSerializer

starswan:
Reworked serializers after own personal review


app/controllers/api/v1/people_controller.rb
```diff
@@ -5,7 +5,7 @@ module V1
     class PeopleController < ApiController
       def index
         people = People::Finder.new(filter_params).call
-        paginate people, include: PersonSerializer::INCLUDED_DETAIL
+        paginate Kaminari.paginate_array(people.map(&:latest_profile)), include: ProfileSerializer::INCLUDED_DETAIL
```
spikeheap: Am I right in thinking the use of `Kaminari.paginate_array` will make this portion less efficient (as it will load the entire set of people and profiles in the controller, but the other refactor/removal of person_serializer removes an n+1 set of queries where the serializer for each person loads their profile?

starswan: I think I might have created a massive N+1 problem if this is ever called without a filter. Really PeopleController is a ProfileController

alexdesi: I noticed 
`latest_profile` is a method in the model`Person`
You might want to use that instead of `people.map(&:latest_profile`

alexdesi: It might be clearer to use scopes in the Person Model:

```
scope :latest, -> { order(:created_at).last }
```
so you can perform  `people.profile.latest`

alexdesi: What happens if `filter_params` is empty? apparently it'll try to load the entire people's latest profile. 😨

starswan: The scope you suggested is probably wrong - there's no guarantee that latest profile is the one that was created last. Having said that of course the current code looks wrong too.

PR #243, opened 2020-02-07
POST /documents to upload a document without a move

ngw:
It is now possible to POST to /documents and create a Document without a move

### DIFF
--- a/app/controllers/api/v1/documents_controller.rb
+++ b/app/controllers/api/v1/documents_controller.rb
@@ -34,7 +34,7 @@ module Api
 
       def document_attributes
         document_params[:attributes].merge(
-          move: Move.find(params.dig(:move_id)),
+          move: Move.find_by(id: params.dig(:move_id)),
         )
       end
 
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_01_27_150100) do
+ActiveRecord::Schema.define(version: 2020_02_07_111034) do
 
   # These are extensions that must be enabled in order to support this database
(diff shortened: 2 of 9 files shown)


app/controllers/api/v1/documents_controller.rb
```diff
@@ -34,7 +34,7 @@ def document_params
 
       def document_attributes
         document_params[:attributes].merge(
-          move: Move.find(params.dig(:move_id)),
+          move: Move.find_by(id: params.dig(:move_id)),
```
starswan: did you mean this change? It causes a very slight behavioural change

db/schema.rb
```diff
@@ -139,6 +139,27 @@
     t.datetime "updated_at", null: false
   end
 
+  create_table "notifications", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
```
starswan: oops this has escaped from your other branch

db/schema.rb
```diff
@@ -208,6 +229,20 @@
     t.integer "latest_nomis_booking_id"
   end
 
+  create_table "subscriptions", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
```
starswan: ditto

app/controllers/api/v1/documents_controller.rb
```diff
@@ -34,7 +34,7 @@ def document_params
 
       def document_attributes
```
starswan: as we are re-using this controller for 2 different (but very similar) endpoints, I think an intention comment would be useful.

ngw: well, the intention is for the code to do exactly what it does... there's a ticket about this, the reasoning behind it is not entirely clear to me (and I don't think it's relevant)

PR #246, opened 2020-02-10
Add support for nomis_offender_no to the /api/v1/people - Part1

alexdesi:
This is part of the ticket: BE - Add support for nomis_offender_no to the /api/v1/people

We'll need to add the support to search for `nomis_offender_no`, this first PR simplyadd the search by nomis_offender_no to the `People::Finder`

Ticket: https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=303&projectKey=P4&modal=detail&selectedIssue=P4-977

### DIFF
--- a/app/services/people/finder.rb
+++ b/app/services/people/finder.rb
@@ -19,14 +19,25 @@ module People
     end
 
     def apply_police_national_computer_filters(scope)
-      return unless filter_params.key?(:police_national_computer)
-
       scope = scope.joins(:profiles)
-      scope.where('profiles.profile_identifiers @> ?', police_national_computer)
+
+      if filter_params.key?(:police_national_computer)
+        scope.where('profiles.profile_identifiers @> ?', police_national_computer.to_json)
+      end
+
+      if filter_params.key?(:nomis_offender_no)
+        scope.where('profiles.nomis_offender_no @> ?', nomis_offender_no.to_json)
+      end
+
+      scope
     end
 
     def police_national_computer
-      [{ identifier_type: 'police_national_computer', value: filter_params[:police_national_computer] }].to_json
+      [{ identifier_type: 'police_national_computer', value: filter_params[:police_national_computer] }]
+    end
+
+    def nomis_offender_no
+      [{ identifier_type: 'prison_number', value: filter_params[:nomis_offender_no] }]
     end
   end
 end
--- a/spec/services/people/finder_spec.rb
+++ b/spec/services/people/finder_spec.rb
@@ -16,5 +16,13 @@ RSpec.describe People::Finder do
         expect(people_finder.call.pluck(:id)).to eq [person.id]
       end
     end
+
+    context 'when matching nomis_offender_no filter' do
+      let(:filter_params) { { nomis_offender_no: 'ABCDEFG' } }
+
+      it 'returns people matching the police_national_computer' do
+        expect(people_finder.call.pluck(:id)).to eq [person.id]
+      end
+    end
   end
 end
--- a/spec/factories/profile.rb
+++ b/spec/factories/profile.rb
@@ -5,7 +5,10 @@ FactoryBot.define do
     first_names { 'Bob' }
     last_name { 'Roberts' }
     date_of_birth { Date.new(1980, 10, 20) }
-    profile_identifiers { [{ identifier_type: 'police_national_computer', value: 'AB/1234567' }] }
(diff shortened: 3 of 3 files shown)


app/services/people/finder.rb
```diff
@@ -19,14 +19,27 @@ def apply_filters(scope)
     end
 
     def apply_police_national_computer_filters(scope)
-      return unless filter_params.key?(:police_national_computer)
+      return unless filter_params.key?(:police_national_computer) || filter_params.key?(:nomis_offender_no)
```
starswan: This return is now superfluous as the body checks both values in if statements

app/services/people/finder.rb
```diff
@@ -19,14 +19,27 @@ def apply_filters(scope)
...
 
       scope = scope.joins(:profiles)
-      scope.where('profiles.profile_identifiers @> ?', police_national_computer)
+
+      if filter_params.key?(:police_national_computer)
```
starswan: It would have been nice to DRY up this class, especially as it now deals with 2 types of identifiers.

spec/services/people/finder_spec.rb
```diff
@@ -16,5 +16,13 @@
         expect(people_finder.call.pluck(:id)).to eq [person.id]
       end
     end
+
+    context 'when matching nomis_offender_no filter' do
```
starswan: I'm not a fan of this pattern (relying on data set in the factory in the test) as it's not very self-describing (Yes I know that's what the other example did).

spec/services/people/finder_spec.rb
```diff
@@ -16,5 +16,13 @@
...
+      let(:filter_params) { { nomis_offender_no: 'ABCDEFG' } }
+
+      it 'returns people matching the police_national_computer' do
```
starswan: This appears to be a cut-n-paste error - it shoud metnion nomis_offender_id.

spec/services/people/finder_spec.rb
```diff
@@ -16,5 +16,13 @@
         expect(people_finder.call.pluck(:id)).to eq [person.id]
       end
     end
+
+    context 'when matching nomis_offender_no filter' do
+      let(:filter_params) { { nomis_offender_no: 'ABCDEFG' } }
```
starswan: according to the ticket, the name of this parameter should be prison_number not nomis_offender_no. We have deliberately chosen to use prison_number I believe rather than the NOMIS terminology

app/services/people/finder.rb
alexdesi: If I remove this, the behavior would be: return all the profile if there are not filters.   
That actually it makes sense.
I just need to check if there are not any calls that relies on the current behavior: return nil if there are not filters.

alexdesi: removed! thans

app/services/people/finder.rb
alexdesi: yes, we could using `select`, `maps`, ... to DRY it,
but it would be less readable. This method is still quite small, I'd keep it as it is if you are happy.

spec/services/people/finder_spec.rb
alexdesi: @starswan you are right, sorry the ticket was not updated. I had a chat with @spikeheap and we decided to name it `nomis_offender_no` .

See https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/adr/0002-API-params-naming.md

alexdesi:
@starswan I see your point, this PR adds a functionality important to complete the ticket.
The reason I did it is to release incremental value without creating a giant PR that would wait for long time.
Not sure what is the practice in this team, being able to merge often has many advantages and would help the team to a more agile style.

Maybe @spikeheap and @jason-sims can clarify this point.

spec/services/people/finder_spec.rb
starswan: ok - you need to make sure that the front-end knows about the change too, otherwise we will have integration problems

spikeheap:
👍 to smaller PRs that incrementally add features. It carries the responsibility that devs will revert changes that end up in dead-ends, of course.

In this case, I'm happy to merge the smaller change that adds offender ID to the API of an internal component, because there's a boundary around the NOMIS client classes. I'd also be happy to see both this and the controller changes in one PR 🤷‍♂, whatever works for the author.

PR #247, opened 2020-02-10
P4-960 - provisionally add assessment questions for prison to court j…

electrified:
…ourney

### DIFF
--- a/app/models/assessment_question.rb
+++ b/app/models/assessment_question.rb
@@ -8,6 +8,8 @@ class AssessmentQuestion < ApplicationRecord
     health: 'health',
     risk: 'risk',
     court: 'court',
+    hearing: 'hearing',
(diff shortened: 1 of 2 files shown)


ngw:
Yo mate, this makes some tests fail but don't despair, it's a very easy fix! 
I suppose you're adding assessment questions, right? The number of questions is fixed, and we (someone :p) wrote tests on it:
https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/p4-960-add-assessment-questions/spec/services/assessment_questions/importer_spec.rb#L10
https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/p4-960-add-assessment-questions/spec/services/assessment_questions/importer_spec.rb#L37

I think if you just increment those 2 numbers you're good to go...

PR #251, opened 2020-02-11
P4-999 placeholder documentation pages

starswan:
This introduces a /docs endpoint that serves a set of static HTML pages. Currently all the content (including the MoJ styling and branding) is done as placeholders, however the MoJ design toolkit has been integrated so that the look and feel of the placeholders is not the HTML defaults.

Once this has been done, you need to run yarn:install locally to import the MOJ design system


Gemfile
```diff
@@ -57,7 +60,7 @@ group :development do
   gem 'rubocop-govuk'
   gem 'rubocop-rspec'
   gem 'spring'
-  gem 'spring-watcher-listen', '~> 2.0.0'
+  # gem 'spring-watcher-listen', '~> 2.0.0'
```
spikeheap: Did this slip in from local?

spikeheap:
Hi @starswan, when I test this locally I see:
Screenshot 2020-02-11 16 11 06

This is a great start, but could we set up the header/footer/nav to get it a little closer to something like this:
Screenshot 2020-02-11 16 21 32

Also, could this be updated to use markdown to author the content. I'm keen to make authoring and maintaining this as easy as possible, so we don't really want to be editing HTML/ERB directly. https://github.com/thoughtbot/high_voltage might be an easy way to achieve this, though obviously open to suggestions too :).

martyn-w:
Switching to markdown is a good idea 👍 - its definitely more readable without a browser

starswan:
Added some basic GOV.UK HTML styling

PR #252, opened 2020-02-11
P4 950 async notification job

ngw:
Implements webhooks.

Follows specs in https://docs.google.com/document/d/10ESi3PwgKzAVX23mFMIs1wD15KLLiCCs1y-XfrgdmUY/edit


spec/jobs/notify_job_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+        expect(client).to have_received(:post).with(notification.subscription.callback_url, data, headers)
+      end
+
+      it 'updates delivered_at' do
+        expect(notification.reload.delivered_at).not_to be_nil
+      end
+    end
+
+    context 'when notification is not a success' do
+      it 'updates delivery_attempts' do
```
alexdesi: it is not clear from the context (probably it is implicit) why this is not a success, and what does actually mean "is not a success"

spec/jobs/notify_job_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  context 'when notification is discarded' do
+    before { notification.discard! }
+
+    it 'raises a Record Not Found error' do
+      expect { perform! }.to raise_error(ActiveRecord::RecordNotFound)
+    end
+  end
+
+  context 'when subscription is discarded' do
```
alexdesi: probably a duplicate 😄

spec/serializers/notification_serializer_spec.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe NotificationSerializer do
+  subject(:serializer) { described_class.new(notification) }
+
+  let(:notification) { create(:notification) }
+
+  describe 'move?' do
```
alexdesi: should we use the convention `#` to describe we are testing a method?
`describe '#move?' do`

spec/serializers/notification_serializer_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+    it { expect(serializer.move?).to be true }
+  end
+
+  describe 'move_url' do
```
alexdesi: maybe
`describe '#move_url' do`

spec/serializers/notification_serializer_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+    it { expect(serializer.move_url).to eql "http://localhost:4000/api/v1/moves/#{notification.topic.id}" }
+  end
+
+
```
alexdesi: nitpick: 2 empty rows

spec/serializers/notification_serializer_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+    it 'contains a move relationship data' do
+      expect(result[:data][:relationships][:move][:data]).to eql(id: notification.topic.id, type: 'moves')
+    end
+
+    it 'contains a move relationship links' do
+      expect(result[:data][:relationships][:move][:links]).to eql(self: "http://localhost:4000/api/v1/moves/#{notification.topic.id}")
+    end
+
+    it 'renders the correct json' do
+      expect(result).to eql(expected_result)
```
alexdesi: should this be the only test necessary? 🤔

app/jobs/notify_job.rb
```diff
@@ -0,0 +1,26 @@
...
+  def perform(notification_id:)
+    notification = Notification.kept.find(notification_id)
+    raise 'Subscription disabled' unless notification.subscription.enabled?
+
+    data = ActiveModelSerializers::Adapter.create(NotificationSerializer.new(notification)).to_json
+    hmac = Encryptor.hmac(notification.subscription.secret, data)
+
+    response = client.post(notification.subscription.callback_url, data, 'PECS-SIGNATURE': hmac, 'PECS-NOTIFICATION-ID': notification_id)
+
+    notification.update(delivered_at: DateTime.now) if response.success?
```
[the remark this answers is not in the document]
ngw: Yes we need it: delivered_at is basically saying the notification has been delivered successfully, under this line it updates the timestamps ALSO in case the notification wasn't a success.

app/jobs/prepare_move_notifications_job.rb
```diff
@@ -0,0 +1,19 @@
...
+  def perform(topic_id:, action_name:)
+    move = Move.find(topic_id)
+
+    move.suppliers.each do |supplier|
+      supplier.subscriptions.kept.each do |subscription|
+        next unless subscription.enabled?
+
+        notification = subscription.notifications.create!(topic: move,
+                                                          event_type: "#{action_name}_move")
+        NotifyJob.perform_later(notification_id: notification.id)
```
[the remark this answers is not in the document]
ngw: We run that job only if we need to run it. This way both jobs (which are reentrant) can run async in separate threads. We could have only one job and that would be slightly less complex, but we'll need to group a bunch of writes in the same job for no valid reason.

spec/serializers/notification_serializer_spec.rb
ngw: I wonder how rubocop didn't get this...

app/controllers/api/v1/moves_controller.rb
```diff
@@ -31,12 +32,14 @@ def update
...
         move.update!(patch_move_attributes)
+        Notifier.prepare_notifications(topic: move, action_name: 'update')
         render_move(move, 200)
       end
 
       def destroy
         move = find_move
-        move.destroy!
+        move.destroy! # TODO: we probably should not be destroying moves
+        Notifier.prepare_notifications(topic: move, action_name: 'destroy')
```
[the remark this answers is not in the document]
ngw: how to infer the `action_name`?

app/jobs/notify_job.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class NotifyJob < ApplicationJob
+  queue_as :webhooks
+
+  def perform(notification_id:)
+    notification = Notification.kept.find(notification_id)
+    raise 'Subscription disabled' unless notification.subscription.enabled?
```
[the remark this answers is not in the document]
ngw: Well deliver notifications to disabled subscriptions doesn't look like a good idea. That's a duplicated control though, will remove.

app/services/notifier.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class Notifier
+  def self.prepare_notifications(topic:, action_name:)
+    case topic
+    when Move
+      ::PrepareMoveNotificationsJob.perform_later(topic_id: topic.id, action_name: action_name)
+    end
+    # when Person
```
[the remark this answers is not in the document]
ngw: It's not a TODO, but something they might want in the future

spec/jobs/notify_job_spec.rb
ngw: it responds with `false` to the `success?` method

ngw: hopefully it's clearer now

spec/serializers/notification_serializer_spec.rb
ngw: Can't speak for @martyn-w but I believe it's checking that the actual result is what he's expecting (and he built "manually" inside the `let`)

app/jobs/notify_job.rb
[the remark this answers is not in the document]
martyn-w: Hi @ngw @alexdesi keep in mind that a notification could be scheduled for delivery when the subscription is enabled; and then the delivery attempted later when then subscription has been disabled: in that case, the delivery should _not_ be attempted - it is necessary to check if the subscription is _still enabled_ when the delivery is attempted.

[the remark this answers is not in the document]
ngw: I will then remove the one in the prepare, the check in the notify job makes the first one superfluous and just a slight quite confusing optimization

spec/jobs/notify_job_spec.rb
martyn-w: Hi @alexdesi @ngw  it is not a duplicate - a subscription could be discarded, or a notification could be discarded (or both). Hence testing for both scenarios.

spec/serializers/notification_serializer_spec.rb
martyn-w: hi @ngw @alexdesi ... true, there is a little duplication here. I had intended the first set of tests to check the values, and this test to check the shape of the json (in particular, getting the <links> object inside of the <relationships> object. But whether all tests are strictly necessary I'll leave to your better judgement :-)

spec/serializers/notification_serializer_spec.rb
martyn-w: 👍

app/serializers/notification_serializer.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+# NB: at the moment this serializer only supports Move topics. In the future we should also support Person topics.
+class NotificationSerializer < ActiveModel::Serializer
+  include Rails.application.routes.url_helpers
+
+  attributes :id, :event_type
+  attribute :created_at, key: :timestamp
+
+  # this is a little ugly but it is the only way (?!) to get the serializer to behave exactly as required
```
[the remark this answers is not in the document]
martyn-w: @ngw @alexdesi yep just remove this comment 👍

app/jobs/notify_job.rb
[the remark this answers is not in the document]
martyn-w: @ngw I believe that both checks are necessary: in the prepare and in the notify. 
1. Notifications should not be queued if the subscription is disabled
2. Queued Notifications should not be sent if the subscription is disabled
these are two different cases (which can happen at different times / system states)

app/controllers/api/v1/moves_controller.rb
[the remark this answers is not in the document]
alexdesi: it's actually called `action_name`  (it returns the name of the controller)

But actually, keep it as it is, I like the explicit way you wrote it.

app/jobs/notify_job.rb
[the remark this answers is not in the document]
alexdesi: ah ok, I got it.
Could this work, so you can use just an update? 
```
notification.delivered_at = DateTime.now if response.success?
assign_attributes(delivery_attempted_at: DateTime.now, delivery_attempts: notification.delivery_attempts.succ)
notification.save
```

spec/jobs/notify_job_spec.rb
alexdesi: ops sorry, I did not see the `success?: false` 
that's perfect!

spec/serializers/notification_serializer_spec.rb
alexdesi: 👍  for me, I'm not strongly opinionated on this.

app/jobs/notify_job.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    data = ActiveModelSerializers::Adapter.create(NotificationSerializer.new(notification)).to_json
+    hmac = Encryptor.hmac(notification.subscription.secret, data)
+
+    response = client.post(notification.subscription.callback_url, data, 'PECS-SIGNATURE': hmac, 'PECS-NOTIFICATION-ID': notification_id)
+
+    notification.update(delivered_at: DateTime.now) if response.success?
+
+    notification.update(delivery_attempts: notification.delivery_attempts.succ,
+                        delivery_attempted_at: DateTime.now)
```
spikeheap: I'd not seen `.succ` before. Neat.

app/jobs/notify_job.rb
```diff
@@ -0,0 +1,25 @@
...
+class NotifyJob < ApplicationJob
+  queue_as :webhooks
+
+  def perform(notification_id:)
+    notification = Notification.kept.find(notification_id)
+
+    data = ActiveModelSerializers::Adapter.create(NotificationSerializer.new(notification)).to_json
+    hmac = Encryptor.hmac(notification.subscription.secret, data)
+
+    response = client.post(notification.subscription.callback_url, data, 'PECS-SIGNATURE': hmac, 'PECS-NOTIFICATION-ID': notification_id)
```
spikeheap: I feel like this may have been discussed already, so apologies. In case the subscription is disabled between the notification being queued and this job running, should we guard this post on `subscription.enabled?`

config/environments/preprod.rb
```diff
@@ -59,7 +59,7 @@
   # Set this to true and configure the email server for immediate delivery to raise delivery errors.
   # config.action_mailer.raise_delivery_errors = false
 
-  # config.action_mailer.default_url_options[:host] = ENV["SERVER_FQDN"]
+  config.action_mailer.default_url_options[:host] = ENV["SERVER_FQDN"]
```
spikeheap: Does this change want to be applied to staging and production as well?

app/jobs/notify_job.rb
martyn-w: @spikeheap very pertinent question - @ngw and I wanted to get a view from the team on this. To be clear, at the guard is in place here thanks to: `notification = Notification.kept.find(notification_id)` - the `.kept` filters out notifications which have a disabled subscription

config/environments/preprod.rb
ngw: wow nice catch, recommenting

app/jobs/notify_job.rb
ngw: we should yes, commit incoming.

martyn-w: todo: add subscription check

ngw: that felt extremely weird actually :D

PR #258, opened 2020-02-13
Add support for nomis offender id to v1 people endpoint

alexdesi:
This PR add support for the params `nomis_offender_no` to the `People` controller.

Also, whenever `nomis_offender_no` is used, it update the data regarding the Person from Nomis.
(Note that the way we sync this could change in the future)

Tech Details:
- A new service `Moves::ImportPeople` was added, to sync just the person
- `Moves::Importer` was refactored to use `Moves::ImportPeople`

### DIFF
--- a/app/controllers/api/v1/people_controller.rb
+++ b/app/controllers/api/v1/people_controller.rb
@@ -4,7 +4,12 @@ module Api
   module V1
     class PeopleController < ApiController
       def index
+        person_nomis_prison_number = filter_params[:nomis_offender_no]
+
+        Moves::ImportPeople.new([person_nomis_prison_number: person_nomis_prison_number]).call if person_nomis_prison_number
+
         people = People::Finder.new(filter_params).call
+
         paginate people, include: PersonSerializer::INCLUDED_DETAIL
       end
 
@@ -20,7 +25,7 @@ module Api
 
     private
 
-      PERMITTED_FILTER_PARAMS = [:police_national_computer].freeze
+      PERMITTED_FILTER_PARAMS = %i[police_national_computer nomis_offender_no].freeze
       PERSON_ATTRIBUTES = [
         :first_names,
         :last_name,
--- a/app/services/moves/import_people.rb
+++ b/app/services/moves/import_people.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+module Moves
+  class ImportPeople
+    attr_accessor :items
+
+    def initialize(items)
+      self.items = items
+    end
+
+    def call
+      import_people
+    end
+
+  private
+
+    def import_people
+      people_nomis_prison_numbers = items.map { |item| item.fetch(:person_nomis_prison_number) }
+      people = NomisClient::People.get(people_nomis_prison_numbers).map { |p| [p.fetch(:prison_number), p] }.to_h
+      alerts = NomisClient::Alerts.get(people_nomis_prison_numbers).group_by { |p| p.fetch(:offender_no) }
+      personal_care_needs = NomisClient::PersonalCareNeeds
+                            .get(people_nomis_prison_numbers)
+                            .group_by { |p| p.fetch(:offender_no) }
+
+      new_person_count = 0
+      changed_profile_count = 0
+      people.each do |offender_no, person_data|
+        profile = People::Importer.new(person_data).call
+        new_person_count += 1 if profile.new_record?
+        Alerts::Importer.new(profile: profile, alerts: alerts.fetch(offender_no, [])).call
+        PersonalCareNeeds::Importer.new(profile: profile,
+                                        personal_care_needs: personal_care_needs.fetch(offender_no, [])).call
+        if profile.changed?
+          changed_profile_count += 1
+          profile.save!
+        end
+      end
+      if new_person_count.positive? || changed_profile_count.positive?
+        Rails.logger.info("[Moves::Importer] people new[#{new_person_count}] updated[#{changed_profile_count}]")
+      end
+    end
+  end
+end
--- a/spec/requests/api/v1/people_controller_index_spec.rb
+++ b/spec/requests/api/v1/people_controller_index_spec.rb
@@ -7,12 +7,12 @@ RSpec.describe Api::V1::PeopleController, with_client_authentication: true do
   let(:content_type) { ApiController::CONTENT_TYPE }
   let(:response_json) { JSON.parse(response.body) }
 
+  let(:schema) { load_json_schema('get_people_responses.json') }
+
   describe 'GET /people' do
-    let(:schema) { load_json_schema('get_people_responses.json') }
-
-    let!(:people) { create_list :person, 5 }
-
     describe 'filtering the results' do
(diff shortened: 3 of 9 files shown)


spec/requests/api/v1/people_controller_index_spec.rb
```diff
@@ -32,5 +32,21 @@
...
     end
+
+    context 'when Nomis offender number no is provided' do
+      let(:nomis_offender_no) { 'G5033UT' }
+      let(:params) { { filter: { nomis_offender_no: 'G5033UT' } } }
+      let(:people_finder) { instance_double('People::Finder', call: Person.all) }
+
+      it 'requests data from NOMIS', with_json_schema: true do
+        allow(People::Finder).to receive(:new).and_return(people_finder)
+        allow(Moves::ImportPeople).to receive(:new).with([person_nomis_prison_number: nomis_offender_no])
```
ngw: You're preparing the test here, you can extract this in a before block like in the other tests you're committing. The before filter and the context look like a very good match too. My 0.2.

app/services/moves/import_people.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  private
+
+    def import_people
+      people_nomis_prison_numbers = items.map { |item| item.fetch(:person_nomis_prison_number) }
+      people = NomisClient::People.get(people_nomis_prison_numbers).map { |p| [p.fetch(:prison_number), p] }.to_h
+      alerts = NomisClient::Alerts.get(people_nomis_prison_numbers).group_by { |p| p.fetch(:offender_no) }
+      personal_care_needs = NomisClient::PersonalCareNeeds
+                            .get(people_nomis_prison_numbers)
+                            .group_by { |p| p.fetch(:offender_no) }
```
ngw: Most of these look like they can be extracted in other private methods you can call as you do with the respective variables. This would make the method much less dense and defer fetching data to well named methods, IMHO adding clarity.

app/services/moves/import_people.rb
```diff
@@ -0,0 +1,44 @@
...
+        profile.save!
+        Alerts::Importer.new(profile: profile, alerts: alerts.fetch(offender_no, [])).call
+        PersonalCareNeeds::Importer.new(profile: profile,
+                                        personal_care_needs: personal_care_needs.fetch(offender_no, [])).call
+        if profile.changed?
+          changed_profile_count += 1
+          profile.save!
+        end
+      end
+      if new_person_count.positive? || changed_profile_count.positive?
```
ngw: I'd remove the logging from here and put it in the call, mostly because this method does a lot and is rather big.

app/services/moves/import_people.rb
```diff
@@ -0,0 +1,44 @@
...
+      people_nomis_prison_numbers = items.map { |item| item.fetch(:person_nomis_prison_number) }
+      people = NomisClient::People.get(people_nomis_prison_numbers).map { |p| [p.fetch(:prison_number), p] }.to_h
+      alerts = NomisClient::Alerts.get(people_nomis_prison_numbers).group_by { |p| p.fetch(:offender_no) }
+      personal_care_needs = NomisClient::PersonalCareNeeds
+                            .get(people_nomis_prison_numbers)
+                            .group_by { |p| p.fetch(:offender_no) }
+
+      new_person_count = 0
+      changed_profile_count = 0
+      people.each do |offender_no, person_data|
```
ngw: About this rather convoluted and complex loop, my *******PERSONAL TASTE******** (:D) is to extract in external methods everything I can, using the writes to the db as some kind of guidance to have each write in a separate method. That normally also guide me through naming the method properly by greatly limiting its scope.

app/services/moves/import_people.rb
```diff
@@ -0,0 +1,44 @@
...
+        profile = People::Importer.new(person_data).call
+        new_person_count += 1 if profile.new_record?
+        profile.save!
+        Alerts::Importer.new(profile: profile, alerts: alerts.fetch(offender_no, [])).call
+        PersonalCareNeeds::Importer.new(profile: profile,
+                                        personal_care_needs: personal_care_needs.fetch(offender_no, [])).call
+        if profile.changed?
+          changed_profile_count += 1
+          profile.save!
```
ngw: You save! profile twice, why?

app/controllers/api/v1/people_controller.rb
```diff
@@ -4,7 +4,14 @@ module Api
   module V1
     class PeopleController < ApiController
       def index
+        person_nomis_prison_number = filter_params[:nomis_offender_no]
+
+        if person_nomis_prison_number
+          Moves::ImportPeople.new([person_nomis_prison_number: person_nomis_prison_number]).call
+        end
```
ngw: This looks like a one liner: ``` do_this if person_nomis_prison_number

app/controllers/api/v1/people_controller.rb
```diff
@@ -4,7 +4,14 @@ module Api
   module V1
     class PeopleController < ApiController
       def index
+        person_nomis_prison_number = filter_params[:nomis_offender_no]
```
ngw: Is there a reason for this? I mean other than liking/disliking the naming. Maybe we can fix this at the source in the filter name? the variable is used only once apparently in the next line.

app/services/moves/import_people.rb
alexdesi: I agree 💯 , this code is not nice and not readable.
I've just copy/paste it from a `Moves::Importer`,
no time for refactoring atm.

app/services/moves/import_people.rb
alexdesi: yes, this smells a lot, we need to refactor it (not at the moment! 😄 )

app/services/moves/import_people.rb
starswan: Let's not copy and paste - it leads to lots of problems. 
The existing code has an import_people method that could probably just be made public. 
If we don't have time do it it right now, we will never have time later when we're really under pressure.

app/services/moves/import_people.rb
starswan: This is a bug in the original copied code - that has now been fixed. Another reason to not just copy/paste at this early stage.

alexdesi: right, I'll remove it.

app/services/moves/import_people.rb
alexdesi: I agree, but as I said, I didn't write this, and am not going to refactor this it for now.

spec/requests/api/v1/people_controller_index_spec.rb
alexdesi: moved to before block

app/controllers/api/v1/people_controller.rb
alexdesi: it is used twice, the assignment make things more readable imo

app/controllers/api/v1/people_controller.rb
alexdesi: ok, not a big issue

PR #262, opened 2020-02-14
P4 950 async notification job

ngw:
PR reopened after revert


db/schema.rb
```diff
@@ -234,19 +232,6 @@
     t.integer "latest_nomis_booking_id"
   end
 
-  create_table "request_audits", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
```
starswan: Please don't remove tables from schema.rb

ngw: I obviously need to merge it again.

PR #265, opened 2020-02-16
Only run production build/deploy jobs on Git tags

teneightfive:
This updates the CircleCI workflow to only run production jobs and
the approval process on Git tags.

This will prevent the master status showing as "On hold" on merges.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,6 +11,22 @@ aliases:
   - &notify_slack
     slack/notify-on-failure:
       only_for_branches: master
+  - &all_tags
+    filters:
+      tags:
+        only: /.*/
+  - &only_master_and_tags
+    filters:
+      tags:
+        only: /.*/
+      branches:
+        only: master
+  - &only_deploy_tags
+    filters:
+      tags:
+        only: /^v.*/
+      branches:
+        ignore: /.*/
 
 references:
   app_containers: &app_containers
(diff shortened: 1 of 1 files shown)


spikeheap:
:+1: to making both frontend & API deploy consistently

Could the [docs](https://dsdmoj.atlassian.net/wiki/spaces/PECS4/pages/1518993502/Deployment+and+Rollback) be updated to reflect the new workflow?

I _believe_ this can all be done through the UI with Github releases, and it'd be good to capture that path in the docs so (e.g.) Jason could continue to trigger a prod deploy without needing a local copy.

teneightfive:
> 👍 to making both frontend & API deploy consistently
> 
> Could the [docs](https://dsdmoj.atlassian.net/wiki/spaces/PECS4/pages/1518993502/Deployment+and+Rollback) be updated to reflect the new workflow?
> 
> I _believe_ this can all be done through the UI with Github releases, and it'd be good to capture that path in the docs so (e.g.) Jason could continue to trigger a prod deploy without needing a local copy.

Ah. This may have broken the ability to deploy without having a local copy to increase the version number.

Might be worth catching up about this tomorrow as to our ideal workflow. But for me, having every workflow that was run on master sat as "On hold" didn't feel as useful.

teneightfive:
> > 👍 to making both frontend & API deploy consistently
> > Could the [docs](https://dsdmoj.atlassian.net/wiki/spaces/PECS4/pages/1518993502/Deployment+and+Rollback) be updated to reflect the new workflow?
> > I _believe_ this can all be done through the UI with Github releases, and it'd be good to capture that path in the docs so (e.g.) Jason could continue to trigger a prod deploy without needing a local copy.
> 
> Ah. This may have broken the ability to deploy without having a local copy to increase the version number.
> 
> Might be worth catching up about this tomorrow as to our ideal workflow. But for me, having every workflow that was run on master sat as "On hold" didn't feel as useful.

I don't think a GitHub release will currently trigger the jobs around production build/deployment as the filter is based on Git tags. Which currently, we create manually on the FE. I quite like the workflow of creating a manual release in this way, but maybe we can automate how a release is created, including creating a new Git tag.

teneightfive:
@starswan / @spikeheap  I've requested another review as I had to update it. Basically it wasn't working as expected on the frontend config which I'd copied it from but since updating and getting that working I've transferred the changes here.

Basically you have to ensure all dependent jobs are also run on all tags, not just branches, for the production jobs to be run.

PR #266, opened 2020-02-17
DRY up rswag tests

starswan:
Remove a lot of duplication from the RSwag tests


spikeheap:
Hi @starswan, why have you moved the RSwag tests out of `requests` and reinstated the previous tests?

I'd be interested to see a PR to deduplicate the contents of spec/requests/api/v1/moves_controller_role_access_spec.rb, perhaps by adding a shared example, or rolling it into the controller_show spec.

starswan:
The approach I took here appears to be contraversial - the team has decided to look at alternatives to RSwag which might be less painful to implement

PR #267, opened 2020-02-17
P4 1030 add nomis offender no to api docs

martyn-w:
Adds `filter[nomis_offender_no]` param to /people swagger document

https://dsdmoj.atlassian.net/browse/P4-1030


starswan:
Interesting. This change appears to convert a mandatory parameter into a non-mandatory one (according to the docs anyway). Does this matter?

martyn-w:
hi @starswan so, the filter[X] param is still required - but it can be either a `filter[police_national_computer]` or a `ilter[nomis_offender_no]` (or both). 
OpenAPI 3.0 does not support parameter dependencies and mutually exclusive parameters. See the Parameter Dependencies at the bottom of this page: https://swagger.io/docs/specification/describing-parameters/

martyn-w:
Hi @starswan the underlying API endpoint change was done for ticket https://dsdmoj.atlassian.net/browse/P4-977 - I think the requirements for that have already discussed. In any case, if further changes to the API are required, then I think a new ticket should be opened ?

PR #269, opened 2020-02-17
P4 983 documentation anomalies

starswan:
Fixup a number of documentation anomalies, and fixed a couple of minor bugs 
e.g. when to_location not present in a move, don't serialize the to_location field

### DIFF
--- a/app/serializers/document_serializer.rb
+++ b/app/serializers/document_serializer.rb
@@ -15,7 +15,7 @@ class DocumentSerializer < ActiveModel::Serializer
   end
 
   def filesize
-    number_to_human_size(object.file.byte_size, precision: 2)
+    object.file.byte_size
   end
 
   def content_type
--- a/spec/requests/api/v1/moves_controller_destroy_spec.rb
+++ b/spec/requests/api/v1/moves_controller_destroy_spec.rb
@@ -13,7 +13,7 @@ RSpec.describe Api::V1::MovesController, with_client_authentication: true do
 
   let(:detail_404) { "Couldn't find Move with 'id'=UUID-not-found" }
 
-  describe 'DELETE /moves/{moveId}' do
+  describe 'DELETE /moves/{move_id}' do
     let(:schema) { load_json_schema('delete_move_responses.json') }
 
     let!(:move) { create :move }
--- a/spec/requests/api/v1/moves_controller_role_access_spec.rb
+++ b/spec/requests/api/v1/moves_controller_role_access_spec.rb
@@ -226,7 +226,7 @@ RSpec.describe Api::V1::MovesController, with_client_authentication: true do
     end
   end
 
-  describe 'DELETE /moves/{moveId}' do
+  describe 'DELETE /moves/{move_id}' do
     let(:schema) { load_json_schema('delete_move_responses.json') }
 
     let!(:pentonville_move) { create :move, from_location: pentonville }
--- a/spec/serializers/document_serializer_spec.rb
+++ b/spec/serializers/document_serializer_spec.rb
@@ -21,7 +21,7 @@ RSpec.describe DocumentSerializer do
   end
 
   it 'contains a `filesize` attribute' do
-    expect(result[:data][:attributes][:filesize]).to eql '98 KB'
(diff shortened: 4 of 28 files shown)


spec/requests/api/v1/moves_controller_destroy_spec.rb
```diff
@@ -13,7 +13,7 @@
 
   let(:detail_404) { "Couldn't find Move with 'id'=UUID-not-found" }
 
-  describe 'DELETE /moves/{moveId}' do
+  describe 'DELETE /moves/{move_id}' do
```
spikeheap: :+1: to making these consistent

spec/requests/api/v1/moves_controller_role_access_spec.rb
```diff
@@ -46,62 +46,32 @@
...
+        expect(JSON.parse(response.body)).to eq resource_to_json
       end
     end
+
+    context 'when the wrong supplier' do
+      let(:move_id) { birmingham_move.id }
+      let(:detail_404) { "Couldn't find Move with 'id'=#{move_id} [WHERE (from_location_id IN ('#{pentonville.id}'))]" }
+
+      it_behaves_like 'an endpoint that responds with error 404'
+    end
```
spikeheap: these changes look like they've bled in from another branch

spec/requests/api/v1/moves_controller_show_spec.rb
```diff
@@ -1,87 +1,63 @@
...
+      let(:content_type) { 'application/xml' }
+
+      it_behaves_like 'an endpoint that responds with error 415'
 
-      response '415', 'invalid content type' do
-        let(:"Content-Type") { 'application/xml' }
-        it_behaves_like 'a swagger 415 error'
-        it_behaves_like 'it does not trigger NomisSynchroniser'
+      it 'doesn\'t sync data' do
+        expect(Moves::NomisSynchroniser).not_to have_received(:new)
```
spikeheap: these changes look like they've bled in from another branch

app/serializers/document_serializer.rb
```diff
@@ -15,7 +15,7 @@ def filename
   end
 
   def filesize
-    number_to_human_size(object.file.byte_size, precision: 2)
+    object.file.byte_size
```
spikeheap: Have you checked with @teneightfive that this change from `96 KB` to a byte count doesn't have an impact on the frontend? If so, :+1:

spec/serializers/document_serializer_spec.rb
```diff
@@ -7,24 +7,26 @@
 
   let(:document) { create :document }
   let(:result) { JSON.parse(ActiveModelSerializers::Adapter.create(serializer).to_json).deep_symbolize_keys }
+  let(:result_data) { result[:data] }
+  let(:attributes) { result_data[:attributes] }
```
spikeheap: The use of nested intermediate variables here smells a bit to me because it obscures the shape of the result hash in the `expect` statements. I'm not tied to this point, but I'm not sure there's a meaningful benefit to deduplicating those 3 calls to attributes (and 5 to data) that's worth the cost of the added complexity.

spec/swagger/definitions/hand_coded_paths.json
```diff
@@ -926,7 +926,7 @@
           "schema": {
             "type": "string",
             "enum": [
-              "risks",
+              "risk",
```
spikeheap: 🙌 nice catch on these small typos

spec/requests/api/v1/moves_controller_role_access_spec.rb
starswan: this is built on top of the DRY swagger tests branch

app/serializers/document_serializer.rb
starswan: he suggested it

martyn-w: could this transform be handled in the front end instead? that seems a more obvious place to do it ?

martyn-w: sorry ignore my comment, haven't had ☕️

PR #275, opened 2020-02-18
Use CircleCI to deploy to sidekiq pods

martyn-w:
This should *hopefully* deploy the api to the sidekiq pods, as well to the api pods.
https://dsdmoj.atlassian.net/browse/P4-1057


spikeheap:
We can watch the post-merge build to validate the changes apply successfully

martyn-w:
@spikeheap  it looks like both sidekiq and the api were correctly deployed 👍

PR #280, opened 2020-02-20
Move asset pipeline from docker file to run.sh

starswan:


run.sh
```diff
@@ -2,4 +2,5 @@
 cd /usr/src/app
 
 bundle exec rake db:create db:migrate
+bundle exec rails assets:precompile
```
spikeheap: 😬 we can't carry this out here, as it'll be run on every pod and slow down the startup cycle (which could lead Kubernetes to think the startup has failed). We'll actually need to move the DB migration out of this soon as well, in line with https://user-guide.cloud-platform.service.justice.gov.uk/documentation/deploying-an-app/running-rails-migrations.html#how-do-i-run-rails-database-migrations.

PR #281, opened 2020-02-20
converted metrics to only run if env var is actively set

starswan:


spikeheap:
https://github.com/ministryofjustice/hmpps-book-secure-move-api-deploy/pull/38/files

config/initializers/prometheus_exporter.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-if Rails.env.production?
+if ENV['PROMETHEUS_METRICS'].present?
```
spikeheap: One minor thing after looking at https://github.com/ministryofjustice/hmpps-book-secure-move-api-deploy/pull/38#pullrequestreview-361761369. We're checking for presence, but using `on`. Would it be worth checking for `"on"`, `"true"`, or `true` here?

config/initializers/prometheus_exporter.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
-if Rails.env.production?
+# check for environment variable so that we can run production rake task in Dockerfile
+# otherwise we try to load the metrics outside of the actual app itself
+if ENV['PROMETHEUS_METRICS'].in? ['on', 'true']
```
alexdesi: 👍

PR #284, opened 2020-02-20
remove inline rails migration

starswan:
Fixes issue with inline rails migration

## Proposed Changes

Remove rails db:migrate from run.sh

## To test/demo change

Rollback environment by 1 database migration. Deploy and check that migration runs forward correctly

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -1,5 +1,4 @@
 #!/bin/bash
 cd /usr/src/app
 
-bundle exec rake db:create db:migrate
 bundle exec puma -p $PUMA_PORT


run.sh
```diff
@@ -1,5 +1,4 @@
 #!/bin/bash
 cd /usr/src/app
 
-bundle exec rake db:create db:migrate
```
martyn-w: @starswan if this is removed at what step will migrations be done? What is the reason for _not_ doing it in `run.sh`

martyn-w:
@starswan @spikeheap Temporarily closing this PR for now - a new database migration process will be required.

PR #283, opened 2020-02-20
add Node, Yarn, and run asset precompile with dummy secret. WIP

spikeheap:
## Proposed Changes

- Adds Node & Yarn depedencies to build assets

## To test/demo change

- `docker build .`

This is a WIP. Running this locally gives the following error in the asset pipeline:

```
Step 15/16 : RUN SECRET_KEY_BASE=doesntmatter rails assets:precompile
 ---> Running in 9db7dd5eb007
I, [2020-02-20T12:33:20.919971 #7]  INFO -- : Writing /usr/src/app/public/assets/app-logotype-crest@2x-32c082fbb9218f7886ae62395dbe9bb63c90bd7144b6a51b31f0cd5a0fdde3fc.png
I, [2020-02-20T12:33:20.931610 #7]  INFO -- : Writing /usr/src/app/public/assets/application-4c392bc960339ef7f65ba4d6124b304a9997169d944bd886f2197ea6189a2a69.js
I, [2020-02-20T12:33:20.931880 #7]  INFO -- : Writing /usr/src/app/public/assets/application-4c392bc960339ef7f65ba4d6124b304a9997169d944bd886f2197ea6189a2a69.js.gz
rails aborted!
SassC::SyntaxError: Error: "calc(0px)" is not a number for `max'
        on line 1159:20 of stdin, in function `max`
        from line 1159:20 of stdin
>> @supports (margin: max(calc(0px))) {

   -------------------^
stdin:1159
/usr/local/bundle/gems/sassc-2.2.1/lib/sassc/engine.rb:49:in `render'
/usr/local/bundle/gems/sassc-rails-2.1.2/lib/sassc/rails/compressor.rb:29:in `call'
/usr/local/bundle/gems/sprockets-3.7.2/lib/sprockets/sass_compressor.rb:28:in `call'
/usr/local/bundle/gems/sprockets-3.7.2/lib/sprockets/processor_utils.rb:75:in `call_processor'
/usr/local/bundle/gems/sprockets-3.7.2/lib/sprockets/processor_utils.rb:57:in `block in call_processors'
/usr/local/bundle/gems/sprockets-3.7.2/lib/sprockets/processor_utils.rb:56:in `reverse_each'
/usr/local/bundle/gems/sprockets-3.7.2/lib/sprockets/processor_utils.rb:56:in `call_processors'
/usr/local/bundle/gems/sprockets-3.7.2/lib/sprockets/loader.rb:134:in `load_from_unloaded'
[...]

PR #286, opened 2020-02-21
P4 999 static docs

starswan:
Fixes P4-999

## Proposed Changes

Adds a set of templated (empty) docs at /docs

## To test/demo change

Go to URL and check existence of docs. Click links to check navigation


Dockerfile
```diff
@@ -19,7 +19,21 @@ EXPOSE $PUMA_PORT
...
+RUN apt-get update && apt-get install yarn
+
+RUN yarn install
+# Have to set SECRET_KEY_BASE here to arbitrary string, otherwise task doesn't run
+RUN SECRET_KEY_BASE=valuenotactuallyused rails assets:precompile
+
 ENV APPUID 1000
+
+RUN chown -R $APPUID:$APPUID /usr/src/app/tmp/cache
+
```
spikeheap: Hi @starswan, could you refactor this to match the approach in https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/283/files, in particular

```
RUN curl -sL https://deb.nodesource.com/setup_13.x | bash - \
   && curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - \
   && echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list \
   && apt-get update \
   && apt-get install -y nodejs yarn \
   && apt-get clean
```

Using one `RUN` for the install (with a clean at the end) reduces image size because we don't end up with cache layers containing APT cache entries.

It looks like the other PR failed because it was pulling 10.x rather than 13.x.

PR #288, opened 2020-02-21
P4 1070 new attributes on move

ngw:
Fixes P4-1070

New attributes on move

It is possible to add a `reason` when creating a move.
Reasons can be fetched on the `reference/reasons` endpoint

## Things to check & link
- [ ] Reason table shouldn't be empty, but actually it is
- [ ] I avoided the moves scope in the reference data as we don't have such namespace and I don't think `moves` belong to reference data. It can very easily be factored in


martyn-w:
@barbaracassani @starswan this is blocked until P4-885 is unreverted

app/controllers/api/v1/reference/reasons_controller.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    module Reference
+      class ReasonsController < ApiController
```
alexdesi: Without context, the word `reason` is not easy to understand.
It'd be nice name it `MoveReasons`

app/controllers/api/v1/reference/reasons_controller.rb
```diff
@@ -0,0 +1,14 @@
...
+        def index
+          reasons = Reason.all
```
alexdesi: Same apply here `Reason` is very generic.

db/fixtures/reasons.yml
```diff
@@ -0,0 +1,63 @@
...
+  key: security
+  title: Security
+
+yjb:
+  key: yjb
+  title: YJB
+
+other:
+  key: another_reason
+  title: Another reason
```
alexdesi: could we rename the filenae to `move_reasons.yml`?

spec/requests/api/v1/moves_controller_create_spec.rb
```diff
@@ -87,6 +89,10 @@
         expect(move.documents).to eq([document])
       end
 
+      it 'associates a reason with the newly created move' do
+        expect(move.reason).to eq(reason)
+      end
+
```
alexdesi: sure we do need this test?
should this be implicit in other tests?

spec/requests/api/v1/moves_controller_show_spec.rb
```diff
@@ -5,7 +5,7 @@
 RSpec.describe Api::V1::MovesController, :with_client_authentication, :rswag do
   let(:headers) { { 'CONTENT_TYPE': content_type }.merge(auth_headers) }
   let(:content_type) { ApiController::CONTENT_TYPE }
-  let(:move) { create :move }
+  let(:move) { create :move, reason: build(:reason) }
```
alexdesi: 👍 for using build instead of create!

just double check if we really need it, we do not need specify other associations (but this)?

spec/requests/api/v1/reference/reasons_controller_spec.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
+
+require 'swagger_helper'
+
+RSpec.describe Api::V1::Reference::ReasonsController, :rswag, :with_client_authentication do
+  let(:headers) { { 'CONTENT_TYPE': content_type }.merge(auth_headers) }
+  let(:content_type) { ApiController::CONTENT_TYPE }
+  let(:response_json) { JSON.parse(response.body) }
+  let!(:reason) { create :reason }
+  let!(:reasons) { [reason] }
```
alexdesi: maybe we are overusing let here  :)

spec/serializers/reason_serializer_spec.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+RSpec.describe ReasonSerializer do
+  subject(:serializer) { described_class.new(reason) }
+
+  let(:reason) { create :reason }
+  let(:result) { JSON.parse(ActiveModelSerializers::Adapter.create(serializer).to_json).deep_symbolize_keys }
+
+  it 'contains a type property' do
```
alexdesi: for consistency you might want to put "type" between back ticks  ```

spec/swagger/definitions/get_reasons_responses.json
```diff
@@ -0,0 +1,44 @@
...
+  "200": {
+    "type": "object",
+    "required": [
+      "data"
+    ],
+    "properties": {
+      "data": {
+        "type": "array",
+        "items": {
+          "$ref": "#/definitions/reason/Reason"
```
alexdesi: there is no hint that reason is about  `move` here, neither in the path, and in the filename.

spec/requests/api/v1/moves_controller_create_spec.rb
starswan: Yes! I was worried that we didn't associate a reason with a move when I first read the code :-)

lib/tasks/reference_data.rake
```diff
@@ -39,6 +39,13 @@ namespace :reference_data do
     ActiveRecord::FixtureSet.create_fixtures(Rails.root.join('db/fixtures'), 'suppliers')
   end
 
+  desc 'create prison transfer reasons'
+  task create_prison_transfer_reasons: :environment do
+    require 'active_record/fixtures'
+
+    ActiveRecord::FixtureSet.create_fixtures(Rails.root.join('db/fixtures'), 'prison_transfer_reasons')
```
alexdesi: What happens if we run this with already existent transfer reasons?
Is the id going to change?
what happens if we add a new reason?

alexdesi: in `AssessmentQuestions::Importer` we used a different approach, have a look.

PR #289, opened 2020-02-24
P4-885 step 2 = add task to back-fill profile data

starswan:
Part of P4-885

## Proposed Changes

- Backfill task so that the migration can be deployed without having to do all the back-filling

## To test/demo change

- rake move_profile:backfill

- task need to be run on production a few times so that actual migration is quick

### DIFF
--- a/lib/tasks/fake_data.rake
+++ b/lib/tasks/fake_data.rake
@@ -140,18 +140,23 @@ namespace :fake_data do
     people = Person.all
     prisons = Location.where(location_type: 'prison').all
     courts = Location.where(location_type: 'court').all
-    500.times do
+    20000.times do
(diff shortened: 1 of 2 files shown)


lib/tasks/fake_data.rake
```diff
@@ -140,18 +140,23 @@ namespace :fake_data do
...
-        status: %w[requested completed cancelled].sample,
-      )
+      person = people.sample
+      from_location = prisons.sample
+      to_location = courts.sample
+      unless Move.find_by(date: date, person: person, from_location: from_location, to_location: to_location)
+        Move.create!(
+          date: date,
+          time_due: time,
+          person: person,
```
martyn-w: @starswan should this definitely be `person` and not `profile` ?

PR #290, opened 2020-02-24
Add field move_agreed and move_agreed_by with relevant validations.

alexdesi:
## Proposed Changes
This PR add the fields
- move_agreed
- move_agreed_by 

and the relevant validations.

## Jira card
https://dsdmoj.atlassian.net/browse/P4-1078

## TODO
Next step is to add the support for those params in the API endpoints.


db/migrate/20200221120126_add_moved_agreed_to_moves.rb
```diff
@@ -0,0 +1,6 @@
+class AddMovedAgreedToMoves < ActiveRecord::Migration[5.2]
+  def change
+    add_column :moves, :move_agreed, :boolean
+    add_column :moves, :move_agreed_by, :string
```
starswan: null: false, default: false is always a good idea for booleans, so that find_by(move_agreed: false) works as expected.

app/models/move.rb
```diff
@@ -46,6 +46,18 @@ class Move < VersionedModel
   validates :move_type, inclusion: { in: move_types }
   validates :person, presence: true
   validates :reference, presence: true
+  validates(
+    :move_agreed,
+    inclusion: { in: [true, false] },
+    if: ->(move) { move.move_type == 'prison_transfer' },
```
martyn-w: no biggee but could 'prison_transfer' be a constant ?

db/migrate/20200221120126_add_moved_agreed_to_moves.rb
alexdesi: @starswan thanks for the review!
`default: false` almost always a good idea, 
in this case `move_agreed` is  relevant only if `move_type` is a prison_transfer.
It does not apply in other cases.

app/models/move.rb
alexdesi: I'll change this in the following PR (that expose that field in the API) if it works for you

martyn-w: @alexdesi sure :+1:

alexdesi:
@starswan in a meeting with @teneightfive we decided the field names and logic of the validation

- move_agreed is required only if move_type  is a "prison_transfer"
- move_agreed_by is required only if move_type is prison_transfer" and the move is agreed

PR #293, opened 2020-02-24
better handling of failed webhook status

martyn-w:
P4-1074

Fixes P4-1074

## Proposed Changes

- this updates the `app/jobs/notify_job.rb` job to handle the case where the endpoint returns a failure status and forces an error to be raised, triggering Sidekiq to retry the job later.

## To test/demo change

- tests updated in `spec/jobs/notify_job_spec.rb`

### DIFF
--- a/app/jobs/notify_job.rb
+++ b/app/jobs/notify_job.rb
@@ -5,18 +5,29 @@ class NotifyJob < ApplicationJob
 
   def perform(notification_id:)
     notification = Notification.kept.includes(:subscription).find(notification_id)
-
     return unless notification.subscription.enabled?
 
     data = ActiveModelSerializers::Adapter.create(NotificationSerializer.new(notification)).to_json
     hmac = Encryptor.hmac(notification.subscription.secret, data)
+    success = false # assume failure
 
-    response = client.post(notification.subscription.callback_url, data, 'PECS-SIGNATURE': hmac, 'PECS-NOTIFICATION-ID': notification_id)
-
-    notification.delivered_at = DateTime.now if response.success?
+    begin
+      response = client.post(notification.subscription.callback_url, data, 'PECS-SIGNATURE': hmac, 'PECS-NOTIFICATION-ID': notification_id)
+      if response.success?
+        notification.delivered_at = DateTime.now
+        success = true
+      else
+        puts "Error: non-success status received from #{notification.subscription.callback_url} (#{response.status})"
+      end
+    rescue Faraday::ClientError => e
+      puts "Error notifying #{notification.subscription.callback_url}: #{e.inspect}"
+    end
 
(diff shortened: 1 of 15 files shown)


app/jobs/notify_job.rb
```diff
@@ -5,18 +5,29 @@ class NotifyJob < ApplicationJob
...
-
     return unless notification.subscription.enabled?
 
     data = ActiveModelSerializers::Adapter.create(NotificationSerializer.new(notification)).to_json
     hmac = Encryptor.hmac(notification.subscription.secret, data)
-
-    response = client.post(notification.subscription.callback_url, data, 'PECS-SIGNATURE': hmac, 'PECS-NOTIFICATION-ID': notification_id)
-
-    notification.delivered_at = DateTime.now if response.success?
+    success = false # assume failure
```
alexdesi: do we need this `success` var?
It sounds a bit redundant (I might be wrong) since `response.success?` could be used for it,
`raise 'Notification failed' unless success` could be put inside the `else`
and maybe rescue should raise a different kind of exception.

app/jobs/notify_job.rb
```diff
@@ -5,18 +5,29 @@ class NotifyJob < ApplicationJob
...
+    begin
+      response = client.post(notification.subscription.callback_url, data, 'PECS-SIGNATURE': hmac, 'PECS-NOTIFICATION-ID': notification_id)
+      if response.success?
+        notification.delivered_at = DateTime.now
+        success = true
+      else
+        puts "Error: non-success status received from #{notification.subscription.callback_url} (#{response.status})"
+      end
+    rescue Faraday::ClientError => e
+      puts "Error notifying #{notification.subscription.callback_url}: #{e.inspect}"
```
alexdesi: do we need these puts?
should we not Log in some way instead of just puts it? (maybe sentry?)

app/jobs/notify_job.rb
martyn-w: @alexdesi thanks for looking - the tricky thing is that I want to do raise an exception _after_ I have updated the notification record, so the status needs to be stored somewhere and the response object might not always exist. I'll have another go at making this cleaner...

app/jobs/notify_job.rb
martyn-w: good point - I'll use the rails.logger instead

app/jobs/notify_job.rb
martyn-w: i've removed the `success` variable; however I don't think the shape of the logic can be much improved ?

alexdesi: I agree, that code does a few things.
Nitpick: if you wish you can remove the `response.present?` check
and use the safe navigation  (&.):
`raise 'Notification failed' unless response&.success?`

I noticed it's a common practice in our project.
Up to you.

PR #294, opened 2020-02-25
P4 885 move profile rework

starswan:
Fixes P4-885

## Proposed Changes

- This brings back the move->profile->person re-structuring. 

## To test/demo change

- Nothing should have changed anywhere from an interface perspective

- rake move_profile:backfill_data needs to be run before deploying this to ensure that the data migration  is trivial.

### DIFF
--- a/db/migrate/20200210150200_associate_move_with_profile_part2.rb
+++ b/db/migrate/20200210150200_associate_move_with_profile_part2.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+class AssociateMoveWithProfilePart2 < ActiveRecord::Migration[5.2]
+  def up
+    Move.find_each do |move|
+      move.profile_id = Person.find_by!(id: move.person_id).latest_profile.id
+      move.save!
+    end
+
+    change_column_null :moves, :profile_id, false
+    add_foreign_key :moves, :profiles
+
+    change_table :moves do |t|
+      t.index [:from_location_id, :to_location_id, :profile_id, :date], name: "index_move_loc_profile_date", unique: true
+    end
+
+    # person_id is unused after this migration, so it has to be nullable
+    change_column_null(:moves, :person_id, true)
+    remove_foreign_key :moves, :people
+  end
+
+  def down
+    remove_foreign_key :moves, :profiles
+    change_column_null :moves, :profile_id, true
+
+    Move.find_each do |move|
+      move.person_id = move.profile.person.id
+      move.save!
+    end
+
+    remove_index :moves, name: "index_move_loc_profile_date"
+
+    change_column_null(:moves, :person_id, false)
+    add_foreign_key :moves, :people
+  end
+end
--- a/db/migrate/20200210150200_associate_move_with_profile_part2.rb
+++ b/db/migrate/20200210150200_associate_move_with_profile_part2.rb
@@ -10,13 +10,11 @@ class AssociateMoveWithProfilePart2 < ActiveRecord::Migration[5.2]
     change_column_null :moves, :profile_id, false
     add_foreign_key :moves, :profiles
 
-    change_table :moves do |t|
-      t.index [:from_location_id, :to_location_id, :profile_id, :date], name: "index_move_loc_profile_date", unique: true
-    end
-
     # person_id is unused after this migration, so it has to be nullable
     change_column_null(:moves, :person_id, true)
     remove_foreign_key :moves, :people
+    # old unique index has to be removed now person_id is unused
+    remove_index :moves, name: "index_on_move_uniqueness"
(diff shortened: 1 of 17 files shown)


db/migrate/20200210150200_associate_move_with_profile_part2.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+class AssociateMoveWithProfilePart2 < ActiveRecord::Migration[5.2]
+  def up
+    Move.find_each.reject { |m| m.profile_id.present? }.each do |move|
```
martyn-w: This will instantiate every move record in the the system _and then_ check if it has a `profile_id`. Instead use the database to filter the records first, e.g.
```
Move.where(profile_id: nil).each do |move|
```

db/migrate/20200210150200_associate_move_with_profile_part2.rb
```diff
@@ -0,0 +1,36 @@
...
+    # person_id is unused after this migration, so it has to be nullable
+    change_column_null(:moves, :person_id, true)
+    remove_foreign_key :moves, :people
+  end
+
+  def down
+    remove_foreign_key :moves, :profiles
+    change_column_null :moves, :profile_id, true
+
+    Move.find_each do |move|
```
martyn-w: Should probably just be setting the person_id for those records which don't have one, i.e.
```
Move.where(person_id: nil).each do |move|
```

spec/serializers/move_serializer_spec.rb
```diff
@@ -65,7 +65,7 @@
     let(:expected_json) do
       [
         {
-          id: move.person_id,
+          id: move.profile.person.id,
```
martyn-w: is this definitely the `person_id` and not the `profile_id` ?

db/migrate/20200210150200_associate_move_with_profile_part2.rb
starswan: This is idiomatic for migrations. Your proposal will instantiate every move record (on first run). ActiveRecord::Base#find_each maps to ActiveRecord::Base#find_in_batches (which discards the records after they have been iterated)

db/migrate/20200210150200_associate_move_with_profile_part2.rb
starswan: again see above. Yes this could be marginally more efficient, but the code you suggest will instantiate every Move object in the DB.

spec/serializers/move_serializer_spec.rb
starswan: Yes - the API exposes the Person object. Profile is currently hidden by the API.

db/migrate/20200210150200_associate_move_with_profile_part2.rb
martyn-w: hmmm, as a compromise how about:

    Move.where(profile_id: nil).find_each do |move|

That will use the efficiency of the database and won't keep all records in memory.

db/migrate/20200210150200_associate_move_with_profile_part2.rb
martyn-w: see above as a compromise: 

    Move.where(person_id: nil).find_each do |move|

db/migrate/20200210150200_associate_move_with_profile_part2.rb
starswan: I'm not sure that works? find_each goes through the whole table in ID order

martyn-w: hi @starswan yes I think it works:

 ```
pry(main)> Move.where(profile_id: nil).find_each { |m| puts m }
  Move Load (3.7ms)  SELECT  "moves".* FROM "moves" WHERE "moves"."profile_id" IS NULL ORDER BY "moves"."id" ASC LIMIT $1  [["LIMIT", 1000]]
#<Move:0x00007faaf50116f0>
#<Move:0x00007faaf5010390>
#<Move:0x00007faaf5010200>
#<Move:0x00007faaf50100c0>
#<Move:0x00007faaf501bf38>
...
```

starswan: I mean I don't think it is memory efficient

martyn-w:
@starswan and I have had a chat about this and we suggest this is paused until there is some real production-like data to test it on; otherwise there is concern that it will fail with memory problems

starswan:
We should hang back on merging this until we have the preprod/production database sync done so we can test it properly before release

cwrw:
closing as this work has been ongoing in a different stream

PR #297, opened 2020-02-25
Removed validation for move_agreed and move_agreed_by

alexdesi:
## Proposed Changes
remove validation for move_agreed and move_agreed_by

### DIFF
--- a/app/models/move.rb
+++ b/app/models/move.rb
@@ -46,17 +46,6 @@ class Move < VersionedModel
   validates :move_type, inclusion: { in: move_types }
   validates :person, presence: true
   validates :reference, presence: true
-  validates(
-    :move_agreed,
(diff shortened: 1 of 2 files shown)


app/models/move.rb
```diff
@@ -46,17 +46,6 @@ class Move < VersionedModel
...
-    :move_agreed,
-    inclusion: { in: [true, false] },
-    if: ->(move) { move.move_type == 'prison_transfer' },
-  )
-
-  validates(
-    :move_agreed_by,
-    presence: true,
-    if: ->(move) { move.move_type == 'prison_transfer' && move.move_agreed? },
-  )
```
starswan: I don't think you need to remove this - if move_agreed is true, then this is ok

PR #298, opened 2020-02-26
port specs to doorkeeper test auth

starswan:
Speeds up tests by around 100% (4 minutes -> 2mins 20)

## Proposed Changes

- convert quite a few tests from using request-level actual auth to doorkeeper test auth

## To test/demo change

- run all tests

### DIFF
--- a/spec/requests/api/v1/documents_controller_create_spec.rb
+++ b/spec/requests/api/v1/documents_controller_create_spec.rb
@@ -2,75 +2,83 @@
 
 require 'rails_helper'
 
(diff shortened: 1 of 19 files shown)


starswan:
> This looks good (and it's great to speed up the tests 🙌 ). There are a few places where `with_client_authentication` has been applied to a context (rather than the whole spec). Could these be refactored to use the faked access token, or was there some limitation which means we need to use the older approach?

Yes there are some tests (particularly the RSwag ones) where it looked too difficult to use the new approach, and passing a parameter (since worked around by using a header) did cause pagination test issues. More could certainly be done in this area

PR #299, opened 2020-02-26
P4-1069 add proposed status to move

starswan:
Fixes P4-1069

## Proposed Changes

- Adds the status of 'proposed' to move

## To test/demo change

- able to update/create a move with a status of 'proposed'

## Things to check & link
- [X ] API docs has been updated if necessary


swagger/v1/move.json
```diff
@@ -35,7 +35,7 @@
           },
           "status": {
             "type": "string",
-            "enum": ["requested", "cancelled", "completed"],
+            "enum": ["proposed", "requested", "cancelled", "completed"],
```
alexdesi: I think there is another file to update (the definitions/move.json)
They are basically duplicated and not sure why we do need both.

starswan: I've deleted the duplicate and updated the one place that it was loaded

alexdesi: 🤞

PR #300, opened 2020-02-26
P4-1091 added created_at to move with filters

starswan:
Fixes P4-1091

## Proposed Changes

- adds created_at to move model, and created_at_from and created_at_to as filter parameters

## To test/demo change

- execute query with new parameters, show that created_at date range is respected

## Things to check & link

### DIFF
--- a/spec/requests/api/v1/moves_controller_filter_spec.rb
+++ b/spec/requests/api/v1/moves_controller_filter_spec.rb
@@ -14,17 +14,42 @@ RSpec.describe Api::V1::MovesController, with_client_authentication: true do
       let!(:filtered_out_moves) { create_list :move, 10 }
       let(:filter_params) { { filter: { supplier_id: supplier.id } } }
 
-      before do
-        get '/api/v1/moves', headers: headers, params: filter_params
-      end
-
       describe 'by supplier_id' do
+        before do
+          get '/api/v1/moves', headers: headers, params: filter_params
+        end
+
         it 'returns the right amount of moves' do
           expect(response_json['data'].size).to eq(10)
         end
 
         it 'returns the right moves' do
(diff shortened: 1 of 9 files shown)


spec/requests/api/v1/moves_controller_filter_spec.rb
```diff
@@ -27,6 +27,30 @@
...
+          create(:move, created_at: middle_date)
+          create(:move, created_at: last_date)
+
+          get '/api/v1/moves', headers: headers, params: filter_params
+        end
+
+        it 'returns the right amount of moves' do
+          expect(response_json['data'].map { |x| x['attributes']['created_at'] }).
+            to match_array [first_date, middle_date, last_date].map(&:to_datetime).map(&:iso8601)
+        end
```
alexdesi: that's very dense,  can you think something less "smart" :) 
it would make more readable

spec/requests/api/v1/moves_controller_filter_spec.rb
```diff
@@ -27,6 +27,30 @@
...
+        let(:target_move2) { Move.find_by(created_at:  middle_date) }
+        let(:target_move3) { Move.find_by(created_at:  last_date) }
+        let(:filter_params) { { filter: { created_at_from: first_date.to_s, created_at_to: last_date.to_s } } }
+
+        before do
+          create(:move, created_at: first_date)
+          create(:move, created_at: middle_date)
+          create(:move, created_at: last_date)
+
+          get '/api/v1/moves', headers: headers, params: filter_params
```
alexdesi: what about putting the action inside the test?

spec/requests/api/v1/moves_controller_filter_spec.rb
starswan: less 'smart'? I'm asserting that we get back the data we expect, not sure how to improve it?

PR #310, opened 2020-03-02
p4-1114 add disabled_at to prison_transfer_reasons entity

starswan:
Fixes P4-1114

## Proposed Changes

- Adds disabled_at to prison transfer reason

## To test/demo change

- call prison_transfer_reasons endpoint and see disabled_at property for a disabled entity

## Things to check & link

### DIFF
--- a/swagger/v1/prison_transfer_reason.json
+++ b/swagger/v1/prison_transfer_reason.json
@@ -35,6 +35,19 @@
             "type": "string",
             "example": "Others",
             "description": "The label of the reason"
+          },
+          "disabled_at": {
+            "oneOf": [
+              {
+                "type": "string",
+                "format": "date-time"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "example": "2017-07-21T17:32:28Z",
+            "description": "The date-time at which a value was disabled, in ISO-8601 format, or null if it is enabled"
           }
         }
       }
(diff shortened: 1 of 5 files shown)


swagger/v1/prison_transfer_reason.json
```diff
@@ -35,6 +35,19 @@
...
+              {
+                "type": "string",
+                "format": "date-time"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "example": "2017-07-21T17:32:28Z",
+            "description": "The date-time at which a value was disabled, in ISO-8601 format, or null if it is enabled"
```
alexdesi: I do not get the meaning in the description, it sounds quite generic "which a value was disabled",
the value of what?

martyn-w:
how about using the [Discard](https://github.com/jhawthorn/discard) gem which we have used on other models (see e.g. Document, Subscription, Notification)

swagger/v1/prison_transfer_reason.json
starswan: This is copied from the other fields with a same name

starswan:
> how about using the [Discard](https://github.com/jhawthorn/discard) gem which we have used on other models (see e.g. Document, Subscription, Notification)

The semantics of this are completely different. This is a field exposed to the client for it to work with - it is on all our reference data.

martyn-w:
> > how about using the [Discard](https://github.com/jhawthorn/discard) gem which we have used on other models (see e.g. Document, Subscription, Notification)
> 
> The semantics of this are completely different. This is a field exposed to the client for it to work with - it is on all our reference data.

Sorry, I misunderstood the use-case 👍

PR #312, opened 2020-03-03
P4-1117 remove superfluous index as it causes crashes

starswan:


martyn-w:
This should hopefully fix https://dsdmoj.atlassian.net/browse/P4-1117

spikeheap:
Do we want to be able to create multiple moves with the same from/to/person/date combination, or put another way: is the error in P4-117 caused by us creating valid moves that fail because of the index, or invalid moves which this index protects against?

spikeheap:
From P4-117:
> Its believed the problem is caused by a unique index which doesn’t handle cancellations and then re-creating moves

So this looks good :+1:

PR #315, opened 2020-03-04
add faker data to locations and profiles

starswan:
Add Faker data to locations and profiles to help testing

### DIFF
--- a/spec/services/moves/reference_generator_spec.rb
+++ b/spec/services/moves/reference_generator_spec.rb
@@ -10,8 +10,10 @@ RSpec.describe Moves::ReferenceGenerator do
   let(:existing_reference) { '12345678' }
   let!(:move) { create :move, reference: existing_reference }
 
+  # Faker uses random numbers, so this test will need to be changed if
+  # more Faker data is introduced to move (or any of its associations)
   it 'generates a new reference' do
-    expect(generator.call).to eql 'JUW9825J'
+    expect(generator.call).to eql 'KRE5873T'
   end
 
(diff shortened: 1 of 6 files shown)


spec/services/moves/reference_generator_spec.rb
```diff
@@ -10,8 +10,10 @@
   let(:existing_reference) { '12345678' }
   let!(:move) { create :move, reference: existing_reference }
 
+  # Faker uses random numbers, so this test will need to be changed if
+  # more Faker data is introduced to move (or any of its associations)
   it 'generates a new reference' do
```
martyn-w: This doesn't seem very reliable as presumably if a random number is used elsewhere then a different ID will be returned... how about instead redefining `PERMISSIBLE_CHARACTERS = ['A']` and `PERMISSIBLE_NUMBERS=[1]` (with a spy?) so you can guarantee the generated reference ?

martyn-w: however, not a bigly issue so approving as-is

spec/services/moves/reference_generator_spec.rb
```diff
@@ -10,8 +10,10 @@
   let(:existing_reference) { '12345678' }
   let!(:move) { create :move, reference: existing_reference }
 
+  # Faker uses random numbers, so this test will need to be changed if
```
alexdesi: Faker can generate any custom of custom string, can't it?
If it can't we can just create our own method, it's just some random chars + some random number

PR #319, opened 2020-03-04
P4-1123 adding sort_by to moves#index

starswan:
Fixes P4-1123

## Proposed Changes

- Adds sort -> by and direction to moves index action

## To test/demo change

- test out sort parameters in postman or swagger

## Things to check & link

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -29,7 +29,10 @@ Metrics/BlockLength:
     - "spec/**/*"
 
 RSpec/NestedGroups:
-  Max: 5
+  Max: 6
+
+RSpec/MultipleExpectations:
+  Max: 2
 
 RSpec/LetSetup:
   Enabled: false
--- a/app/controllers/api/v1/moves_controller.rb
+++ b/app/controllers/api/v1/moves_controller.rb
@@ -4,10 +4,10 @@ module Api
   module V1
     class MovesController < ApiController
       def index
-        moves_params = Moves::ParamsValidator.new(params[:filter])
+        moves_params = Moves::ParamsValidator.new(filter_params, params[:sort] || {})
         if moves_params.valid?
           import_moves_from_nomis
-          moves = Moves::Finder.new(filter_params, current_ability).call
+          moves = Moves::Finder.new(filter_params, current_ability, params[:sort] || {}).call
           paginate moves, include: MoveSerializer::INCLUDED_ATTRIBUTES
         else
           render json: { error: moves_params.errors }, status: :bad_request
--- a/app/services/moves/finder.rb
+++ b/app/services/moves/finder.rb
@@ -18,6 +18,12 @@ module Moves
 
     def call
       scope = apply_filters(Move)
+      apply_ordering(scope)
+    end
+
+  private
+
+    def apply_ordering(scope)
       case @order_by
       when :name
         scope.joins(person: :profiles).merge(Profile.ordered_by_name(@order_direction))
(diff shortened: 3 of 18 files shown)


.rubocop.yml
```diff
@@ -29,7 +29,10 @@ Metrics/BlockLength:
     - "spec/**/*"
 
 RSpec/NestedGroups:
-  Max: 5
+  Max: 6
+
+RSpec/MultipleExpectations:
+  Max: 2
```
alexdesi: 👍 I like it

app/controllers/api/v1/moves_controller.rb
```diff
@@ -4,10 +4,10 @@ module Api
   module V1
     class MovesController < ApiController
       def index
-        moves_params = Moves::ParamsValidator.new(params[:filter])
+        moves_params = Moves::ParamsValidator.new(filter_params, params[:sort] || {})
```
alexdesi: Filters and params used for the model are different entities,
so I think the validation for the `filter` should be apart from the move validation (the model).

The advantage is separation of concern and the chance to reuse the filter validator (high chance)

app/services/moves/finder.rb
```diff
@@ -2,15 +2,40 @@
 
 module Moves
   class Finder
-    attr_accessor :filter_params, :ability
+    attr_reader :filter_params, :ability
 
-    def initialize(filter_params, ability = nil)
-      self.filter_params = filter_params
-      self.ability = ability
+    def initialize(filter_params, ability, order_params)
```
alexdesi: we should remove the word `params` from the params :)

`def initialize(filter, ability, order)`

app/services/moves/finder.rb
```diff
@@ -2,15 +2,40 @@
...
+        scope.left_outer_joins(:prison_transfer_reason).merge(PrisonTransferReason.ordered_by_title(@order_direction))
+      when :created_at
+        scope.order(created_at: @order_direction)
+      when :date_from
+        scope.order(date_from: @order_direction)
+      when :date
+        scope.order(date: @order_direction)
+      else
+        scope
+      end
```
alexdesi: this method could be simplified a bit, 
maybe creating more private class with meaningful names,

app/services/moves/finder.rb
```diff
@@ -2,15 +2,40 @@
...
+        scope.joins(:to_location).merge(Location.ordered_by_title(@order_direction))
+      when :prison_transfer_reason
+        scope.left_outer_joins(:prison_transfer_reason).merge(PrisonTransferReason.ordered_by_title(@order_direction))
+      when :created_at
+        scope.order(created_at: @order_direction)
+      when :date_from
+        scope.order(date_from: @order_direction)
+      when :date
+        scope.order(date: @order_direction)
+      else
```
alexdesi: this seems to be the same as:
```
when [:created_at, :date_from, :date]
        scope.order(@order_by => @order_direction)
```

app/services/moves/params_validator.rb
```diff
@@ -4,19 +4,26 @@ module Moves
   class ParamsValidator
     include ActiveModel::Validations
 
-    attr_accessor :date_from, :date_to
+    attr_reader :date_from, :date_to, :sort_by, :sort_direction
```
alexdesi: `sort_by, :sort_direction` should not be here imo

app/controllers/api/v1/moves_controller.rb
starswan: YAGNI

PR #325, opened 2020-03-06
add variable 'date' field in move factory

starswan:
Factory + test tweak to allow 'date' field on move to no longer have a fixed value

### DIFF
--- a/spec/services/moves/finder_spec.rb
+++ b/spec/services/moves/finder_spec.rb
@@ -76,10 +76,16 @@ RSpec.describe Moves::Finder do
     end
 
     context 'with matching date range' do
-      let(:filter_params) { { date_from: Date.today.to_s, date_to: 5.days.from_now.to_date.to_s } }
+      before do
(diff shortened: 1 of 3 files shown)


spec/services/moves/finder_spec.rb
```diff
@@ -76,10 +76,16 @@
     end
 
     context 'with matching date range' do
-      let(:filter_params) { { date_from: Date.today.to_s, date_to: 5.days.from_now.to_date.to_s } }
+      before do
+        create(:move, date: move.date + 6.days)
+        create(:move, date: move.date - 1.day)
+      end
+
+      let!(:next_week_move) { create(:move, date: move.date + 5.days) }
```
martyn-w: the naming here is a little unclear: `next_week_move` is 5 days after `move` - but they could be in the same week?
maybe a comment might make it clearer that the test is checking that the filter excludes move-1 and move+6 ?

PR #326, opened 2020-03-06
P4-1027 Non-unique move return error rather than crashing

starswan:
Fixes P4-1027

## Proposed Changes

- Add validation to move to prevent duplicate creation, but allow multiple cancelled moves

## To test/demo change

- create a move, then create another move with the same person/to/from/date - nice error message returned

## Things to check & link

Front-end will not understand this error particularily well - it probably needs some specific work to allow it to cope nicely

### DIFF
--- a/app/controllers/api/v1/moves_controller.rb
+++ b/app/controllers/api/v1/moves_controller.rb
@@ -21,13 +21,9 @@ module Api
       def create
         move = Move.new(move_attributes)
         authorize!(:create, move)
-        if move.save
-          Notifier.prepare_notifications(topic: move, action_name: 'create')
-          render_move(move, 201)
-        else
-          render(json: { errors: move_validation_errors(move) },
-                 status: :unprocessable_entity)
-        end
+        move.save!
+        Notifier.prepare_notifications(topic: move, action_name: 'create')
+        render_move(move, 201)
       end
 
       def update
@@ -102,18 +98,6 @@ module Api
           .accessible_by(current_ability)
           .includes(:from_location, :to_location, person: { profiles: %i[gender ethnicity] })
           .find(params[:id])
-      end
-
-      def move_validation_errors move
-        if move.errors.details[:date].map { |x| x[:error] }.include?(:taken)
-          errors = validation_errors(move.errors)
-          errors.select { |e| e[:code] == :taken }.each do |error|
-            error.merge!(meta: { move_id: move.existing.id })
-          end
-          errors
-        else
-          validation_errors(move.errors)
-        end
       end
     end
   end
--- a/app/controllers/api/v1/moves_controller.rb
+++ b/app/controllers/api/v1/moves_controller.rb
@@ -21,10 +21,13 @@ module Api
       def create
         move = Move.new(move_attributes)
         authorize!(:create, move)
-        move.save!
-        move.documents.each { |doc| doc.update(move: move) }
-        Notifier.prepare_notifications(topic: move, action_name: 'create')
-        render_move(move, 201)
+        if move.save
+          Notifier.prepare_notifications(topic: move, action_name: 'create')
+          render_move(move, 201)
+        else
+          render(json: { errors: move_validation_errors(move) },
+                 status: :unprocessable_entity)
+        end
       end
 
       def update
@@ -99,6 +102,18 @@ module Api
           .accessible_by(current_ability)
           .includes(:from_location, :to_location, person: { profiles: %i[gender ethnicity] })
           .find(params[:id])
+      end
+
+      def move_validation_errors move
+        if move.errors.details[:date].map { |x| x[:error] }.include?(:taken)
+          errors = validation_errors(move.errors)
+          errors.select { |e| e[:code] == :taken }.each do |error|
+            error.merge!(meta: { move_id: move.existing.id })
+          end
+          errors
+        else
+          validation_errors(move.errors)
+        end
       end
     end
   end
--- a/app/models/move.rb
+++ b/app/models/move.rb
@@ -53,6 +53,10 @@ class Move < VersionedModel
   validates :reference, presence: true
 
   validates :status, inclusion: { in: statuses }
+  # we need to avoid creating/updating a move with the same person/date/from/to if there is already one in the same state
+  # except that we need to allow multiple cancelled moves
+  validates :date, uniqueness: { scope: %i[status person_id from_location_id to_location_id] },
+            unless: -> { status == MOVE_STATUS_CANCELLED }
 
   validate :date_to_after_date_from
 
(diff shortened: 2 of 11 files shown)


spec/requests/api/v1/moves_controller_create_spec.rb
```diff
@@ -201,5 +201,43 @@
...
+      context 'when duplicate is active' do
+        let!(:old_move) { create(:move, person: person, from_location: from_location, to_location: to_location) }
+        let(:errors_422) do
+          [
+            {
+              title: 'Unprocessable entity',
+              detail: 'Date has already been taken',
+              source: { 'pointer' => '/data/attributes/date' },
+              code: 'taken',
+              meta: { 'move_id' => old_move.id },
```
teneightfive: Does this only support one active duplicate?

spec/requests/api/v1/moves_controller_create_spec.rb
```diff
@@ -201,5 +201,43 @@
...
+        it_behaves_like 'an endpoint that responds with success 201'
+      end
+
+      context 'when duplicate is active' do
+        let!(:old_move) { create(:move, person: person, from_location: from_location, to_location: to_location) }
+        let(:errors_422) do
+          [
+            {
+              title: 'Unprocessable entity',
+              detail: 'Date has already been taken',
```
teneightfive: Is Date our only duplicate error?

spec/requests/api/v1/moves_controller_create_spec.rb
starswan: given we allow duplicates, logically there can only be one duplicate.

spec/requests/api/v1/moves_controller_create_spec.rb
starswan: Yes - you're right it is a little fragile from that point of view. Can't think off-hand how to deal with that.

martyn-w:
I've had a quick look over this and would like a bit more time to double check the behaviour if there are multiple cancellations - in particular the `.existing` method on `Move` doesn't seem to expect multiples?

starswan:
> I've had a quick look over this and would like a bit more time to double check the behaviour if there are multiple cancellations - in particular the `.existing` method on `Move` doesn't seem to expect multiples?

Yes you might be right - existing might find something non-existent...

app/models/move.rb
```diff
@@ -51,6 +51,8 @@ class Move < VersionedModel
   validates :reference, presence: true
 
   validates :status, inclusion: { in: statuses }
+  # we need to avoid creating a move with the same person/date/from/to if there is already an 'active' one.
+  validates :date, uniqueness: { scope: %i[status person_id from_location_id to_location_id] }, on: :create
```
martyn-w: this index will prevent multiple cancellations: I think they should be permitted

app/controllers/api/v1/moves_controller.rb
```diff
@@ -22,10 +22,21 @@ def show
...
-        Notifier.prepare_notifications(topic: move, action_name: 'create')
-        render_move(move, 201)
+        if move.save
+          move.documents.each { |doc| doc.update(move: move) }
+          Notifier.prepare_notifications(topic: move, action_name: 'create')
+          render_move(move, 201)
+        elsif move.errors.details[:date].map { |x| x[:error] }.include?(:taken)
+          errors = validation_errors(move.errors)
+          errors.select { |e| e[:code] == :taken }.each do |error|
+            error.merge!(meta: { move_id: move.existing.id })
```
alexdesi: This logic should not be here in the controller, 
at first sight it could go in the move model win a meaningful name

app/controllers/api/v1/moves_controller.rb
```diff
@@ -22,10 +22,21 @@ def show
       def create
         move = Move.new(move_attributes)
         authorize!(:create, move)
-        move.save!
-        move.documents.each { |doc| doc.update(move: move) }
-        Notifier.prepare_notifications(topic: move, action_name: 'create')
-        render_move(move, 201)
+        if move.save
+          move.documents.each { |doc| doc.update(move: move) }
```
alexdesi: What about moving this to `move`,
with a name like `move.update_documents`:

Advantages:
- you have less noisy in the controller
- it is testable (in case you need a test)

alexdesi:
Sorry mi comment is a  bit late, I really think we should keep controller cleaner, and it looks easy to do in this case (see comments).

alexdesi:
By the way, I am not advocating *Fat Models and skinny Controllers*, but fat controller it might be even worst.

alexdesi:
This does not look to be ready IMO:
For example: the method
`def move_validation_errors`
has been just put in the move controller, I do not think that's a good place (let' keep controller thin).

It's natural place would be the move model (or even different validator class).

If we have some time it'd be nice to clean this a bit,
what do you think Jonny? 
cc @jcavell

spec/requests/api/v1/moves_controller_create_spec.rb
smoothcontract: @teneightfive Amended implementation to be more generic - this can now support any kind of uniqueness check, the only difference being in the `meta` key below which now uses `existing_id` rather than `move_id` so that the same approach can be extended to other models in future.

For now, only moves will return this error and they only have a uniqueness check on the date attribute.

app/models/move.rb
```diff
@@ -73,7 +77,11 @@ def from_nomis?
   end
 
   def existing
-    Move.find_by(date: date, person_id: person_id, from_location_id: from_location_id, to_location_id: to_location_id)
+    self.class.find_by(date: date, person_id: person_id, from_location_id: from_location_id, to_location_id: to_location_id)
```
smoothcontract, line 80: `find_by` only ever returns one record (although the order is not guaranteed). See https://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-find_by

Hence I don't think it's particularly valuable to add a test for the scenario of multiple existing moves

martyn-w: should `status` also be a factor in the `find_by` call? otherwise you may find legitimate cancelled moves, rather than the dodgy duplicated move?

spec/requests/api_controller_spec.rb
```diff
@@ -67,7 +84,7 @@ class Model
     before { model.valid? }
 
     it 'returns the correct errors' do
-      expect(api_controller.send(:validation_errors, model.errors)).to eq errors
+      expect(api_controller.send(:validation_errors, model)).to eq errors
```
martyn-w, line 87: not sure if `validation_errors` will always return these errors in the prescribed order - maybe use `match_array` rather than `eq`?

smoothcontract: In the general case, you're right - it's model dependant, but here the errors will be in the order specified in the `Model` class defined within the spec. I think from a documentation point of view it's better to change as suggested to indicate the order is not really defined.

app/models/move.rb
smoothcontract: GPWM!

spec/requests/api_controller_spec.rb
martyn-w: 👍

PR #327, opened 2020-03-06
P4-1142 created_at_to search needs to be inclusive

starswan:
Fixes P4-1142
## Proposed Changes

- Make the created_at_to filter inclusive rather than exclusive (as its a date/time not a date)

## To test/demo change

- check the created_at_to filter on moves to make sure that it is inclusive

### DIFF
--- a/app/services/moves/finder.rb
+++ b/app/services/moves/finder.rb
@@ -63,8 +63,9 @@ module Moves
     def apply_date_range_filters(scope)
       scope = scope.where('date >= ?', filter_params[:date_from]) if filter_params.key?(:date_from)
       scope = scope.where('date <= ?', filter_params[:date_to]) if filter_params.key?(:date_to)
+      # created_at is a date/time, so inclusive filtering has to be subtlely different
       scope = scope.where('moves.created_at >= ?', filter_params[:created_at_from]) if filter_params.key?(:created_at_from)
-      scope = scope.where('moves.created_at <= ?', filter_params[:created_at_to]) if filter_params.key?(:created_at_to)
+      scope = scope.where('moves.created_at < ?', Date.parse(filter_params[:created_at_to]) + 1) if filter_params.key?(:created_at_to)
       scope
     end
 
--- a/spec/requests/api/v1/moves_controller_filter_spec.rb
+++ b/spec/requests/api/v1/moves_controller_filter_spec.rb
@@ -41,30 +41,47 @@ RSpec.describe Api::V1::MovesController do
       end
 
       describe 'by created_at' do
-        let(:first_date) { Date.new(2019, 12, 25) }
-        let(:last_date) { Date.new(2019, 12, 27) }
-        let(:middle_date) { Date.new(2019, 12, 26) }
+        let(:first_date) { DateTime.new(2019, 12, 25, 12) }
+        let(:last_date) { DateTime.new(2019, 12, 27) }
+        let(:middle_date) { DateTime.new(2019, 12, 26, 12) }
 
         let(:target_move) { Move.find_by(created_at: first_date) }
         let(:target_move2) { Move.find_by(created_at:  middle_date) }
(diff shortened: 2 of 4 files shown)


app/services/moves/finder.rb
```diff
@@ -38,8 +38,9 @@ def apply_filter(scope, param_name)
     def apply_date_range_filters(scope)
       scope = scope.where('date >= ?', filter_params[:date_from]) if filter_params.key?(:date_from)
       scope = scope.where('date <= ?', filter_params[:date_to]) if filter_params.key?(:date_to)
+      # created_at is a date/time, so inclusive filtering has to be subtlely different
       scope = scope.where('created_at >= ?', filter_params[:created_at_from]) if filter_params.key?(:created_at_from)
-      scope = scope.where('created_at <= ?', filter_params[:created_at_to]) if filter_params.key?(:created_at_to)
+      scope = scope.where('created_at <= ?', Date.parse(filter_params[:created_at_to]) + 1) if filter_params.key?(:created_at_to)
```
martyn-w: should this be `'created_at < ?'` not `'created_at <= ?'` ?

spec/requests/api/v1/moves_controller_filter_spec.rb
```diff
@@ -29,27 +29,46 @@
...
           create(:move, created_at: first_date)
           create(:move, created_at: middle_date)
           create(:move, created_at: last_date)
-        end
 
-        it 'returns the right amount of moves' do
           get '/api/v1/moves', headers: headers, params: filter_params
+        end
+
+        context 'when happy case' do
```
martyn-w: would be nice to add a test that checks the created_at_from/to filter actually works

app/services/moves/finder.rb
starswan: nope

spec/requests/api/v1/moves_controller_filter_spec.rb
starswan: there is one - this is a very small big fix to that PR

app/services/moves/finder.rb
martyn-w: Hi @starswan this test fails because `<=` will include 00:00:00 in the following day: 

```
RSpec.describe Move do

  let!(:move1) { create(:move, created_at: '2000/01/01 00:00:00') }
  let!(:move2) { create(:move, created_at: '2000/01/02 00:00:00') }
  let!(:move3) { create(:move, created_at: '2000/01/03 00:00:00') }
  let!(:move4) { create(:move, created_at: '2000/01/04 00:00:00') }

  let(:created_at_from) {'2000/01/02'}
  let(:created_at_to) {'2000/01/03'}

  subject { Move
                .where('created_at >= ?', created_at_from)
                .where('created_at <= ?', Date.parse(created_at_to) + 1) }


  it { is_expected.to match_array([move2, move3]) }
end
```

Would you mind adding a test like this and fixing the scope ?

spec/requests/api/v1/moves_controller_filter_spec.rb
martyn-w: I'm sorry - I don't see the filter test you are referring too? can you point me in the right direction

spec/requests/api/v1/moves_controller_filter_spec.rb
```diff
@@ -41,30 +41,46 @@
...
+        end
+
+        context 'with a valid range' do
+          let(:filter_params) { { filter: { created_at_from: first_date.to_date.to_s, created_at_to: middle_date.to_date.to_s } } }
+
+          it 'returns the right amount of moves' do
+            expect(response_json['data'].map { |x| x['attributes']['created_at'] }).
+              to match_array [first_date, middle_date].map(&:to_datetime).map(&:iso8601)
+          end
+        end
```
alexdesi: a bit hard to read, but not blocking

PR #328, opened 2020-03-10
P4 1047 1122 email notifications

martyn-w:
Fixes P4-1122

## Proposed Changes

- adds a new Email Notify Job
- adds a new mailer using the Gov.UK Notify service
- refactors existing Webhook Notifier job

## To test/demo change

- see tests

## Things to check & link
- [ ] New environment variables have been added to staging configuration (awaiting GovUK Notify go-live)

- get new api key from gov uk service and setup for Staging
- get new api key from gov uk service and setup for Pre-production
- get new api key from gov uk service and setup for Production


app/jobs/notify_email_job.rb
```diff
@@ -0,0 +1,42 @@
...
+# It runs as a retryable job to handle the case where the Gov.UK service is temporarily offline.
+class NotifyEmailJob < ApplicationJob
+  queue_as :notifications
+
+  def perform(notification_id:)
+    notification = Notification.emails.kept.includes(:subscription).find(notification_id)
+    return unless notification.subscription.enabled?
+
+    if ENV.fetch('GOVUK_NOTIFY_API_KEY', nil).blank?
+      Rails.logger.error('[NotifyEmailJob] please set the GOVUK_NOTIFY_API_KEY env variable')
```
starswan: we won't get any notification of this happening - maybe we should be calling Sentry instead?

config/initializers/gov_uk_notify.rb
```diff
@@ -0,0 +1,9 @@
+if ENV['GOVUK_NOTIFY_API_KEY'].blank?
```
starswan: we seem to be violating DRY with the NOTIFY_KEY (twice) and the TEMPLATE_ID (3 times). Do we need this initialiser at all, as we are checking in the code anyway?

app/models/subscription.rb
```diff
@@ -21,4 +23,11 @@ def secret=(sekret)
   def secret
     Encryptor.decrypt(encrypted_secret)
   end
+
+private
+
+  def email_address_or_callback_url_required
+    errors.add(:callback_url, 'is required if email_address is blank') if callback_url.blank? && email_address.blank?
```
starswan: I think both being present should also be an error?

app/jobs/notify_email_job.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+    if ENV.fetch('GOVUK_NOTIFY_TEMPLATE_ID', nil).blank?
+      Rails.logger.error('[NotifyEmailJob] please set the GOVUK_NOTIFY_TEMPLATE_ID env variable')
+      return # no point retrying later
+    end
+
+    response = nil
+    begin
+      response = MoveMailer.notify(notification).deliver!
```
starswan: why this over deliver_now? It's not clear what the response for this is - can it ever return the error checked below?
https://github.com/ministryofjustice/govuk_notify_rails doesn't mention the deliver! method

app/mailers/move_mailer.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class MoveMailer < GovukNotifyRails::Mailer
+  TIME_FORMAT = '%d/%m/%Y %T'
+
+  def notify(notification)
+    set_template(ENV.fetch('GOVUK_NOTIFY_TEMPLATE_ID'))
+    set_reference(notification.id)
+    notification.topic.tap do |move|
```
starswan: 'tap' call seems superfluous here

martyn-w:
Following some feedback from @starswan I will:
* add some request-based tests
* clarify the request _state_ (requested/cancelled) based on the action and the request _status_ 
* improve job exception handling so that it raises the caught exception rather than throws a new exception
* improve checks of the ENV VARs - probably only do this once and log to Sentry/Raven if there is something which is not set.

app/jobs/notify_email_job.rb
alexdesi: it seems that `.deliver!` is deprecated
https://coderwall.com/p/8yjt0a/actionmailer-method-deliver-is-already-deprecated-use-this-instead

martyn-w: thanks, I'll switch to `deliver_now!`

spec/jobs/notify_email_job_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+
+  context 'when GOVUK_NOTIFY_TEMPLATE_ID env var is not set' do
+    let(:govuk_notify_template_id) { nil }
+
+    it 'raises an ArgumentError' do
+      expect { perform! }.to raise_error(ArgumentError, /Missing template ID/)
+    end
+  end
+
+  context 'when notification and subscription are active' do
```
alexdesi: this context is empty, usually `rubocop` complain about it, 🤔

spec/jobs/notify_email_job_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+      it 'updates delivery_attempts' do
+        expect { perform! }.to change { notification.reload.delivery_attempts }.from(0).to(1)
+      end
+
+      it 'updates response_id' do
+        expect { perform! }.to change { notification.reload.response_id }.from(nil).to(response_id)
+      end
+    end
+
+    context 'when Gov.uk Notify does not respond as expected' do
```
alexdesi: maybe we can be more explicit about  `...does not respond as expected`

spec/jobs/notify_email_job_spec.rb
```diff
@@ -0,0 +1,172 @@
+# frozen_string_literal: true
+
+RSpec.describe NotifyEmailJob, type: :job do
+  subject(:perform!) { described_class.new.perform(notification_id: notification.id) }
+
+  let(:perform_ignore_errors!) { perform! rescue nil }
```
alexdesi: it looks like `ignore_errors` is the action here,
maybe
`perform_ignoring_errors!`
or 
`perform_and_ignore_errors!`
 are better name.

spec/models/subscription_spec.rb
```diff
@@ -6,7 +6,32 @@
   it { is_expected.to have_many(:notifications) }
   it { is_expected.to belong_to(:supplier) }
   it { is_expected.to validate_presence_of :supplier }
-  it { is_expected.to validate_presence_of :callback_url }
+
+  describe 'email_address or callback_url (or both) are required' do
```
alexdesi: With  "(both)"  you mean that there is a case where email_address AND callback_url are required,
this might be true, but not clear from the test.

app/jobs/prepare_move_notifications_job.rb
```diff
@@ -10,10 +11,40 @@ def perform(topic_id:, action_name:)
...
+            event_type: event_type(action_name, topic),
+          ).id }
+  end
+
+  def should_email?(move, action_name)
+    # NB: only notify by email on move requested and move cancellation
+    [Move::MOVE_STATUS_REQUESTED, Move::MOVE_STATUS_CANCELLED].include?(move.status) && %w(create update_status).include?(action_name)
+  end
+
+  def event_type(action_name, topic)
```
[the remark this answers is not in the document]
martyn-w: yep, that is a bit clearer 👍

spec/jobs/notify_email_job_spec.rb
martyn-w: the context isn't empty? it contains multiple sub-contexts. It is intended to mark the section when notifications are enabled vs the section where they are not.

spec/jobs/notify_email_job_spec.rb
martyn-w: ok, have renamed it to: `when Gov.uk Notify does not respond as with a govuk_notify_response object`

spec/jobs/notify_email_job_spec.rb
martyn-w: ok, have renamed to `perform_and_ignore_errors!`

spec/jobs/notify_email_job_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+    allow(ENV).to receive(:fetch).with('GOVUK_NOTIFY_API_KEY', nil).and_return(govuk_notify_api_key)
+    allow(ENV).to receive(:fetch).with('GOVUK_NOTIFY_TEMPLATE_ID', nil).and_return(govuk_notify_template_id)
+    ActionMailer::Base.add_delivery_method :govuk_notify, GovukNotifyRails::Delivery, api_key: govuk_notify_api_key
+  end
+
+  context 'when GOVUK_NOTIFY_API_KEY env var is not set' do
+    let(:govuk_notify_api_key) { nil }
+
+    it 'raises an NoMethodError' do
+      expect { perform! }.to raise_error(NoMethodError, /undefined method `length' for nil/)
```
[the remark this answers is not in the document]
martyn-w: it used to do that but following a previous review I moved those checks to the `config/initializers/gov_uk_notify.rb` initialiser

spec/models/subscription_spec.rb
martyn-w: yes, line 29 is the test for both: `when email_address and callback_url provided`

PR #329, opened 2020-03-11
P4 1090 send azure events

starswan:
Fixes P4-1090

## Proposed Changes

- use a new base image for the app, as we were using an outdated MoJ image. Activate AppInsights on staging, preprod and production

## To test/demo change

- Browse environment, and check matching Azure App Insights data

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,14 +21,6 @@ aliases:
         only: /.*/
       branches:
         only: master
-  - &only_master_tags_and_p4_1090
-    filters:
-      tags:
-        only: /.*/
-      branches:
-        only:
-        - master
-        - p4-1090-send-azure-events
   - &only_deploy_tags
     filters:
       tags:
@@ -147,10 +139,6 @@ jobs:
             sudo gem update --system
             gem install bundler:2.1.4
 
-      - run:
-          name: Which bundler?
-          command: bundle -v
-
       # Download and cache dependencies
       - restore_cache:
           keys:
@@ -229,9 +217,9 @@ workflows:
       - test:
           <<: *all_tags
       - build_staging:
-          <<: *only_master_tags_and_p4_1090
+          <<: *only_master_and_tags
       - deploy_staging:
-          <<: *only_master_tags_and_p4_1090
+          <<: *only_master_and_tags
           requires:
           - test
           - build_staging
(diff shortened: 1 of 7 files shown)


alexdesi:
Broken tests here.
@starswan how we can test this?
since it changes Circle CI it could be cause of potential issues on today's deploy (12th march),
it might be wise to merge it later.

alexdesi:
By the way, I am not advocating *Fat Models and skinny Controllers*, but fat controller it might be even worst.

starswan:
> Broken tests here.
> @starswan how we can test this?
> since it changes Circle CI it could be cause of potential issues on today's deploy (12th march),
> it might be wise to merge it later.

Yes not planning to release this today. The only test is to run it on staging - but it doesn't change much on a fundamental level

Dockerfile
```diff
@@ -27,12 +30,23 @@ RUN curl -sL https://deb.nodesource.com/setup_13.x | bash - \
    && apt-get clean
 
 RUN yarn install
+
 # Have to set SECRET_KEY_BASE here to arbitrary string, otherwise task doesn't run
-RUN SECRET_KEY_BASE=valuenotactuallyused rails assets:precompile
+RUN SECRET_KEY_BASE=valuenotactuallyused bundle exec rails assets:precompile
+
+# Run the application as user `moj` (created in the base image)
```
martyn-w: is this comment still relevant given that the base image has changed from `ministryofjustice/ruby:2.6.2-webapp-onbuild` to `ruby:2.6.2-stretch`
...and on a related note, what is the reason for changing the base image?

.circleci/config.yml
```diff
@@ -21,6 +21,14 @@ aliases:
...
       branches:
         only: master
+  - &only_master_tags_and_p4_1090
+    filters:
+      tags:
+        only: /.*/
+      branches:
+        only:
+        - master
+        - p4-1090-send-azure-events
```
martyn-w: this seems a bit clunky: a comment might help explain why only  this branch and master should be deployed?

Dockerfile
starswan: Yes the comment is wrong. The base image didn't work, and turned out to be deprecated and undocumented

.circleci/config.yml
starswan: Its because I wanted to test deployments.

martyn-w:
ok, looks good - just need to get tests to pass 👍

PR #330, opened 2020-03-11
P4-1147 remove NOMIS sync from moves#index

starswan:
Fixes P4-1147

## Proposed Changes

- Removes NOMIS sync of moves and associated code 

## To test/demo change

- NOMIS is no longer synched by moves index endpoint


spec/rails_helper.rb
```diff
@@ -15,7 +15,7 @@
 
     # The intention of this value is that it should never go down after a PR
     # It is a (very) naive attempt to prevent untested code entering the codebase
-    minimum_coverage 98.82
+    minimum_coverage 98.75
```
martyn-w: @starswan is it absolutely necessary to reduce test coverage? the natural preference is to maintain/improve coverage

starswan: Yes - the massive reduction in code lines results in the coverage percentage dropping e.g. 2000/2500 is 80%, but delete 1000 covered lines from that and you have 1000/1500 66%

app/controllers/api/v1/people_controller.rb
```diff
@@ -36,6 +36,15 @@ def image
 
     private
 
+      def import_person_from_nomis prison_number
+        # This prevents us from blaming the current user/application for the NOMIS sync
+        PaperTrail.request(whodunnit: nil) do
+          Moves::ImportPeople.new([person_nomis_prison_number: prison_number]).call
+        end
+      rescue StandardError => e
```
martyn-w: maybe catch a more specific error, rather than everything ?

starswan: Its deliberate. It's a pattern that was in the other synchroniser - whatever goes wrong during sync, we don't want the application to fail, but we do want it reported to sentry

PR #332, opened 2020-03-12
Save picture in S3 and return S3 URL

alexdesi:
Jira: https://dsdmoj.atlassian.net/browse/P4-1143

## Proposed Changes

When the Image is retrieved from Nomis, it store it on S3 and return the URL

## To test/demo change

call the API get `/people/person_id/image`

## Things to check & link
- [ ] New environment variables have been added to staging configuration

### DIFF
--- a/app/controllers/api/v1/people_controller.rb
+++ b/app/controllers/api/v1/people_controller.rb
@@ -26,12 +26,14 @@ module Api
       def image
         person = Person.find(params[:person_id])
 
-        if !person.picture.attached?
+        if !person.image.attached?
           image_blob = NomisClient::Image::get(person)
-          person.attach_picture(image_blob)
+          person.attach_image(image_blob)
         end
 
-        render json: { url: url_for(person.picture) }
+        image = Image.new(person.id, url_for(person.image))
+
+        render json: image
       end
 
     private
--- a/app/controllers/api/v1/people_controller.rb
+++ b/app/controllers/api/v1/people_controller.rb
@@ -26,14 +26,13 @@ module Api
       def image
         person = Person.find(params[:person_id])
 
-        if !person.image.attached?
-          image_blob = NomisClient::Image::get(person)
-          person.attach_image(image_blob)
+        success = People::RetrieveImage.call(person)
+
+        if success
+          render json: Image.new(person.id, url_for(person.image))
+        else
+          render_resource_not_found_error(Exception.new('Image not found'))
         end
-
-        image = Image.new(person.id, url_for(person.image))
-
-        render json: image
       end
 
     private
--- a/app/controllers/api/v1/people_controller.rb
+++ b/app/controllers/api/v1/people_controller.rb
@@ -25,13 +25,13 @@ module Api
 
       def image
         person = Person.find(params[:person_id])
-        image_data = NomisClient::Image::get(person.latest_profile.latest_nomis_booking_id)
 
-        if image_data
-          send_data image_data, type: 'image/jpg', disposition: 'inline'
-        else
-          render status: :not_found
+        if !person.picture.attached?
+          image_blob = NomisClient::Image::get(person)
+          person.attach_picture(image_blob)
         end
+
+        render json: { url: url_for(person.picture) }
       end
 
     private
--- a/app/lib/nomis_client/image.rb
+++ b/app/lib/nomis_client/image.rb
@@ -5,7 +5,9 @@ module NomisClient
   # :nocov:
   class Image < NomisClient::Base
     class << self
-      def get booking_id
+      def get(person)
+        booking_id = person.latest_profile.latest_nomis_booking_id
+
         image_route = "/bookings/#{booking_id}/image/data"
         begin
           NomisClient::Base.get(image_route).body
--- a/app/lib/nomis_client/image.rb
+++ b/app/lib/nomis_client/image.rb
@@ -5,9 +5,7 @@ module NomisClient
   # :nocov:
   class Image < NomisClient::Base
     class << self
-      def get(person)
-        booking_id = person.latest_profile.latest_nomis_booking_id
-
+      def get(booking_id)
         image_route = "/bookings/#{booking_id}/image/data"
         begin
           NomisClient::Base.get(image_route).body
--- a/app/models/image.rb
+++ b/app/models/image.rb
@@ -1,0 +1,10 @@
+class Image
+  include ActiveModel::Serialization
+
+  attr_accessor :id, :url
+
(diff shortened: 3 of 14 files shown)


app/controllers/api/v1/people_controller.rb
```diff
@@ -25,13 +25,13 @@ def update
...
-        if image_data
-          send_data image_data, type: 'image/jpg', disposition: 'inline'
-        else
-          render status: :not_found
+        if !person.picture.attached?
+          image_blob = NomisClient::Image::get(person)
+          person.attach_picture(image_blob)
         end
+
+        render json: { url: url_for(person.picture) }
```
starswan: don't think this is right. I think JSON API spec requires a data: element with an array in it

app/lib/nomis_client/image.rb
```diff
@@ -5,7 +5,9 @@ module NomisClient
   # :nocov:
   class Image < NomisClient::Base
     class << self
-      def get booking_id
+      def get(person)
+        booking_id = person.latest_profile.latest_nomis_booking_id
```
starswan: not happy with this change - it ties the Nomis::Image client to our person model (additional coupling) - all our other Nomis::<x> modules are independent, so this is inconsistent as well

app/models/person.rb
```diff
@@ -4,7 +4,22 @@ class Person < VersionedModel
   has_many :profiles, dependent: :destroy
   has_many :moves, dependent: :destroy
 
+  has_one_attached :picture
+
   def latest_profile
     profiles.last
   end
+
+  def attach_picture(image_blob)
```
starswan: was it not possible to use a StringIO object here? If not, I think an intention comment would be useful

spec/requests/api/v1/people_controller_image_spec.rb
```diff
@@ -13,28 +13,46 @@
         allow(NomisClient::Image).to receive(:get).and_return(nil)
       end
 
-      it 'returns 404' do
-        get "/api/v1/people/#{person.id}/image", params: { access_token: token.token }
+      it 'returns resource not found 404' do
```
starswan: This was a different test - to check that if the person didn't have an image then we return 404. The non-existant-id case should IMHO be a new test

app/models/image.rb
```diff
@@ -0,0 +1,10 @@
+class Image
+  include ActiveModel::Serialization
+
+  attr_accessor :id, :url
```
starswan: these should be attr_readers

spec/requests/api/v1/people_controller_image_spec.rb
```diff
@@ -13,28 +13,47 @@
...
         expect(response).to be_successful
       end
 
-      it 'returns the image' do
-        expect(response.body).to eq(image_data)
+      it 'contains the url of the image' do
+        get_image
+
+
+        expect(JSON.parse(response.body)['data']['attributes']['url']).to include person.id + '.jpg'
```
starswan: IMHO we can just check for equality of the JSON reply - it's not too big or onerous

app/models/image.rb
```diff
@@ -0,0 +1,10 @@
+class Image
+  include ActiveModel::Serialization
+
```
starswan: I'm a little unsure about this behaviour. You're claiming that this is an 'image' object, but its ID field is that of a 'person' object. Doesn't feel quite right?

app/controllers/api/v1/people_controller.rb
alexdesi: Fixed, this is now using a serializer.

app/models/person.rb
alexdesi: I think I got an error about the format, I think it's clear from the fact we handle an image here that it is a blob, Binary Large Object.

app/models/image.rb
```diff
@@ -0,0 +1,10 @@
+class Image
+  include ActiveModel::Serialization
```
starswan: As we have an ImageSerializer, I don't think this line is required. This class is just a property holder

app/serializers/person_serializer.rb
```diff
@@ -13,8 +13,6 @@ class PersonSerializer < ActiveModel::Serializer
     :gender_additional_information,
   )
```
starswan: This deletion introduces new behaviour. It's now not possible to tell if a person should have an image property or not. I think we should therefore have a test to show that if we don't have a prison_number on the person, then /people/<id>/images behaves correctly (which I think means 404)

app/lib/nomis_client/image.rb
alexdesi: I got your point, it was an attempt to clean a bit the controller,
since that's a query, I'll move it in the model.

starswan: This isnt a model - it's a service to get images from NOMIS

spec/requests/api/v1/people_controller_image_spec.rb
alexdesi: fixed, I've created a different context for that.

spec/requests/api/v1/people_controller_image_spec.rb
alexdesi: that would imply to generate in full the URL, I think this test works and it is explicit

spec/requests/api/v1/people_controller_image_spec.rb
starswan: we still seem to have 2 deletions according to github?

app/models/image.rb
alexdesi: It is required, if you do not put it you get

```
 NoMethodError: undefined method `read_attribute_for_serialization' 
```

app/models/image.rb
alexdesi: I see you point, but the initialize helps to understand what actually the id is:

```
def initialize(person_id, image_url)
    @id = person_id
    @url = image_url
  end
```

Image is an entity per se, so it's fine to have it as a model imo,
the models folders do not necessary need to contain only record model, there are other classes not activerecord.

app/serializers/person_serializer.rb
alexdesi: This was discussed with @teneightfive, it complicates the FE logic, we preferred to simplify the logic removing it,
We might want to put it back in the future, so FE would need to retrieve the images only for the people that do not have yet URL,

app/lib/nomis_client/image.rb
alexdesi: I've created a services for that.

PR #334, opened 2020-03-18
Improve tasks recreate_all fake data.

alexdesi:
##Problem
the tasks `recreate_all` takes to long to run and does not have any output.

## Proposed Changes
Improving the tasks `recreate_all`
- puts comment on the various tasks
- speed up the tasks reducting the number of

### DIFF
--- a/lib/tasks/fake_data.rake
+++ b/lib/tasks/fake_data.rake
@@ -3,6 +3,7 @@
 namespace :fake_data do
   desc 'create fake people'
   task create_people: :environment do
+    puts 'create_people...'
     ethnicities = Ethnicity.all.to_a
     genders = Gender.all.to_a
     50.times do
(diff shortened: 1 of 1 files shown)


lib/tasks/fake_data.rake
```diff
@@ -109,38 +112,43 @@ namespace :fake_data do
...
   end
 
   desc 'create fake moves'
   task create_moves: :environment do
+    puts 'create_moves...'
     people = Person.all
     prisons = Location.where(location_type: 'prison').all
     courts = Location.where(location_type: 'court').all
-    20000.times do
+    1000.times do
```
starswan: rather than fighting over the value of this, could it not be a parameter?

alexdesi: I see your point, but the effort is not worth,
it is not trivial, since this task is called from another tasks
we need to change the 2 tasks, making things more complicated for something that probably nobody will use.
I'll be happy to do the change if there is a real need.

PR #337, opened 2020-03-19
Add link Pentonville (PVI) -> Serco supplier

alexdesi:
Trello: https://dsdmoj.atlassian.net/browse/P4-1187

## Proposed Changes

- Add link Pentonville (PVI) - Serco
- Extract locations in exernal Yaml file (instead to be hardcoded)

Run the following 
```
 bundle exec rake reference_data:link_suppliers
```
on staging and production


alexdesi:
@starswan I've splitted this in 2 different commits, so it is easier to see the changes (one line).
Regarding the huge list of code hardcoded in the rake task, that is not a risky operation, there is not code involved, just loading a file Yaml and in any case, I'll test it in staging.

PR #340, opened 2020-03-24
Remove random uuid gen for ActiveStorage

alexdesi:
Fixes Document/Images Bug

## Proposed Changes

The UUID for the :record_id of ActiveStorage table should not be random generated, since rails will take care of it.


starswan:
Lots of tests fail with this change

PR #341, opened 2020-03-24
support new 'dev' environment

starswan:
Relates to P4-1200

## Proposed Changes

- push to new 'dev' environment in circle

## To test/demo change

- check that new 'dev' env gets pushed when this branch is built.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,6 +21,14 @@ aliases:
         only: /.*/
       branches:
         only: master
+  - &only_master_and_create_dev
+    filters:
+      tags:
+        only: /.*/
+      branches:
+        only:
+        - master
+        - p4-1200-create-dev-env
   - &only_deploy_tags
     filters:
       tags:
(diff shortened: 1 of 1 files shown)


alexdesi:
@starswan if you can write a comment that explain why you need these changes would be great

.circleci/config.yml
```diff
@@ -21,6 +21,14 @@ aliases:
...
       branches:
         only: master
+  - &only_master_and_create_dev
+    filters:
+      tags:
+        only: /.*/
+      branches:
+        only:
+        - master
+        - p4-1200-create-dev-env
```
alexdesi: I am not CI expert, we are putting this here so we do not need to merge it on master?

.circleci/config.yml
```diff
@@ -80,7 +88,7 @@ commands:
               -t app .
 
       - run:
-          name: push docker image
+          name: push app image
```
alexdesi: what was wrong with `docker`?

PR #344, opened 2020-03-25
Add GOVUK_NOTIFY_ENABLED flag

martyn-w:
Fixes P4-1211

## Proposed Changes

Adds a GOVUK_NOTIFY_ENABLED flag which should only be set on Sidekiq containers

## To test/demo change

- TODO

## Things to check & link

- [ ] New environment variables have been added to staging configuration


martyn-w:
Hi @smoothcontract yes, good question - I don't think we have any tests for initializers; but generally they don't do much more than set a parameter to a value.

PR #346, opened 2020-03-25
Fix for empty latest_nomis_booking_id when retrieving image from nomis

alexdesi:
##Bigfix 
Jira: P4 https://dsdmoj.atlassian.net/browse/P4-1201
Sentry: https://sentry.service.dsd.io/mojds/bookasecuremove-production/issues/45695

## Proposed Changes
If the `latest_nomis_booking_id` is nil it does not get the image from Nomis

## To test/demo change
Run the enpoint /people/<ID>/images
passing a person with no latest_nomis_booking_id

### DIFF
--- a/app/services/people/retrieve_image.rb
+++ b/app/services/people/retrieve_image.rb
@@ -4,6 +4,7 @@ module People
   class RetrieveImage
     def self.call(person)
       return true if person.image.attached?
+      return false unless person.latest_nomis_booking_id
 
       image_blob = NomisClient::Image::get(person.latest_nomis_booking_id)
 
--- a/spec/services/people/retrieve_image_spec.rb
+++ b/spec/services/people/retrieve_image_spec.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe People::RetrieveImage do
+  subject(:retrieve_image) { described_class.call(person) }
+
+  let(:person) { build(:person) }
+
+  it 'returns false if latest_nomis_booking_id is empty' do
+    person.latest_profile.update(latest_nomis_booking_id: nil)
+
+    expect(retrieve_image).to eq(false)
+  end
+end


app/services/people/retrieve_image.rb
```diff
@@ -3,6 +3,7 @@
 module People
   class RetrieveImage
     def self.call(person)
+      return false unless person.latest_nomis_booking_id
```
smoothcontract: I am very new to this, so may be talking nonsense, but should this line not be placed after line 7 so that if there's an existing image attachment that gets returned irrespective of the missing `latest_nomis_booking_id` ... not sure if that's a valid scenario even?

spec/services/people/retrieve_image_spec.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe People::RetrieveImage do
+  it 'returns false if latest_nomis_booking_id is empty' do
+    person = build(:person)
```
martyn-w: how about a `let(:person) { build(:person) } ` outside of the test, and the test can just check if something is true or not

app/services/people/retrieve_image.rb
alexdesi: That's a good point, I'll make the change.

(It does not change in case you got only one profile, since `latest_nomis_booking_id` is gonna be the same for the first and subsequent calls.)

spec/services/people/retrieve_image_spec.rb
alexdesi: The test I wrote has a different "style", it follows the simple AAA Assign, Act, Assert that maximize readability and avoid all the rspec DSL, 
but I agree with you that since we are using a certain style, mainly based on http://www.betterspecs.org/ (DRY and fragmented style), 
I should adapt the code.

martyn-w: 👍 yep, you are right, it is purely a style issue - it would be good if to be consistent with other tests

alexdesi: I tried to mix up the best of the 2 approaches, I dried it without making it too fragmented

PR #349, opened 2020-03-26
Reintroduce dependency for staging build on test job

smoothcontract:
Fixes P4-1189

## Proposed Changes

This reverts the previous parallel running of build and test jobs for staging environment (done in an effort to improve the build time) but unfortunately this introduces a race condition where the swagger.yaml file is not available when performing the docker build; hence the docker image does not include the swagger files so online API docs are not available.

Also took the opportunity to clean up some inconsistent indentation across the workflow definitions.

I've explored a few other options to resolve this but none really are viable sadly..

1. Running `bundle exec rake rswag:specs:swaggerize` as part of the build rather than test phase.
This works but only when the `development` and `test` gems are included in the docker image. It would be bad practice to add those as it bloats the size of the deploy image and increases surface area for potential security issues.

2. Add some synchronisation to the test and build jobs.
Something like `dockerize -wait file:///swagger/v1/swagger.yaml -timeout 30s` as an extra run step in the build job would work to ensure that the test job has generated the swagger file before continuing. Unfortunately this needs to be right at the start of the build job so the amount of parallel work remaining would be tiny meaning the added complexity really isn't worthwhile. Ideally some of the docker steps such as fetching the OS base image could be performed in parallel while waiting for tests but this too would mean extensive refactoring and added complexity to achieve.

On balance, I think the simple course of action is to remove the parallelism for now. In future we could look at improving test execution time perhaps by optimising the slowest tests in the suite or looking at running individual tests in parallel within the test job.

## To test/demo change

- Requires merge and deployment to staging

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,6 +2,8 @@
 #
 # Check https://circleci.com/docs/2.0/language-ruby/ for more details
 #
+#
+#
 version: 2.1
 
 orbs:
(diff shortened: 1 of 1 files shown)


starswan:
There is a way of doing this - we generate the swagger docs at the moment as part of the 'test' phase - so we have swagger.yaml in a circle 'workspace', there is no need to try and buld the docs with dev/test during the image build step - runnigng a seperate 'swagger' step and depending on that should be possible.

smoothcontract:
> There is a way of doing this - we generate the swagger docs at the moment as part of the 'test' phase - so we have swagger.yaml in a circle 'workspace', there is no need to try and buld the docs with dev/test during the image build step - runnigng a seperate 'swagger' step and depending on that should be possible.

Yes, I suppose it could be done by replicating most of the `test` job and getting that just to generate the swagger docs. Happy to have another look at this if we think it's worthwhile - it would mean quite a lot of extra stuff in the circle config.. but maybe I can split out common parts 🤔

smoothcontract:
Closing in favour of #352

PR #351, opened 2020-03-26
P4-1216 rename prison transfer reasons

martyn-w:
Fixes P4-1216

## Proposed Changes

Migration to renames acronyms CPA, ISP, YJB and MAPPA to full names

Tested locally.


martyn-w:
Thanks for the feedback @starswan and @smoothcontract .  I've tested the rake task locally and for some reason which I can't fathom, it the id (uuid) of each row is the same on my local machine as it is on staging. Moreover, if I delete a row locally or change a row locally and then run the rake task, it does somehow reset the table to have exactly the same keys, titles *and ids* as it does on staging. My guess is that the table is wiped and then re-inserted and possibly the `id` is being seeded by the `key` field ??
I don't know.... but I am wary of this kind of magic: if production / pre-prod were somehow seeded differently then the ids might change and we would loose referential integrity. So my inclination is to keep with the migration based approach as that is very clear in exactly what it is changing.
Can anyone offer insight into how the UUIDs can be the same between local and staging ?

starswan:
@smoothcontract's comments are incorrect - I tested this originally, and it works exactly as you would hope it would - replacing data with new versions as appropriate

martyn-w:
@starswan could you explain how IDs are preserved? e.g., look at the existing record for `YJB` and note the `id` of the record. Delete that row, run the rake task, and then the row is re-inserted with the same id. How does the system get the same ID as it used previously? understanding that mechanism will re-assure me that the rake task is sufficient.

starswan:
> @starswan could you explain how IDs are preserved? e.g., look at the existing record for `YJB` and note the `id` of the record. Delete that row, run the rake task, and then the row is re-inserted with the same id. How does the system get the same ID as it used previously? understanding that mechanism will re-assure me that the rake task is sufficient.

I'm not sure I understand it - I just know it works. Its loaded as a fixture. I worked on the original story, and this was one of my goals that the data should be changable if required - as Barbara said the descriptions should change

smoothcontract:
I was a bit puzzled by the mystery around this fixture loading stuff, so have been doing some digging. When testing locally I can see that updating any fixture attribute and then rerunning the fixture import appears to only update the modified attributes, and leaves id's unchanged.

The story is a little more nuanced than that though. From digging into the ActiveRecord code, the method `ActiveRecord::FixtureSet.create_fixtures` works by first deleting all rows and then reinserting the fixture rows. That occurs within a transaction and temporarily suspends referential integrity, so should be safe to use on a live system - providing there are no other data changes that have been applied outside of the fixture code.

Here's the relevant code snippet:

```ruby
def insert_fixtures_set(fixture_set, tables_to_delete = [])
  fixture_inserts = build_fixture_statements(fixture_set)
  table_deletes = tables_to_delete.map { |table| "DELETE FROM #{quote_table_name(table)}" }
  statements = table_deletes + fixture_inserts

  with_multi_statements do
    disable_referential_integrity do
      transaction(requires_new: true) do
        execute_batch(statements, "Fixtures Load")
      end
    end
  end
end
```

So, if the data is being destroyed and then recreated, how are the ID's preserved? There's a clue in the Fixture documentation here: https://github.com/rails/rails/blob/5-2-stable/activerecord/lib/active_record/fixtures.rb#L227

```
# Hashing each fixture's label yields a consistent ID:
#
#   george: # generated id: 503576764
#     name: George the Monkey
#
#   reginald: # generated id: 324201669
#     name: Reginald the Pirate
#
# Active Record looks at the fixture's model class, discovers the correct
# primary key, and generates it right before inserting the fixture
# into the database.
```

Here's the corresponding code for that: https://github.com/rails/rails/blob/5-2-stable/activerecord/lib/active_record/fixtures.rb#L579

```ruby
# Returns a consistent, platform-independent identifier for +label+.
# Integer identifiers are values less than 2^30. UUIDs are RFC 4122 version 5 SHA-1 hashes.
def self.identify(label, column_type = :integer)
  if column_type == :uuid
    Digest::UUID.uuid_v5(Digest::UUID::OID_NAMESPACE, label.to_s)
  else
    Zlib.crc32(label.to_s) % MAX_ID
  end
end
```

So, this works with a few provisos - firstly, it's important to never change the `label` for ` given fixture entry. Doing so will generate a new UUID and so break RI - I've proved that locally. Secondly, this fixture technique should only be used on tables where there is no chance of any external source of modification.

TL:DR; The migration turns out not to be necessary, so long as fixtures are used with caution. Here be dragons and all that... 🐉

martyn-w:
@smoothcontract thanks very much for getting to the bottom of this, as this magical behaviour was quite perplexing. I think we can stick with the rake task+fixture for now (as that is the established process), keeping in mind the short-comings of the fixed IDs. I'll update the PR to remove the migration.

PR #354, opened 2020-04-01
Get court cases from Nomis

alexdesi:
Ticket: https://dsdmoj.atlassian.net/browse/P4-1164

## Proposed Changes
This retrieves the court cases from Nomis,
transform them using a PORO serializer,
and provide an endpoint for the FE.
  
## To test/demo change

Call the new /people/ID/court_cases endpoint

## Things to check & link


app/models/court_case.rb
```diff
@@ -0,0 +1,35 @@
+class CourtCase
+  include ActiveModel::Serialization
+
+  attr_reader :id, :case_seq, :begin_date, :case_type, :case_info_number, :case_status, :agency_id
+
+  def build_from_nomis(court_case)
+
+    @case_info_number = court_case['caseInfoNumber']
+    @id = @case_info_number
```
smoothcontract, line 9: Not sure why this has a duplicate identifier? I'd be inclined to drop `case_info_number` completely from this model and rely solely on `id`. It will likely cause confusion and inconsistency having two attributes with the same purpose.

app/models/court_case.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    @case_seq = court_case['caseSeq']
+    @begin_date = court_case['beginDate']
+    @case_type = court_case['caseType']
+    @case_status = court_case['caseStatus']
+
+    @agency_id = court_case['agency']['agencyId']
```
smoothcontract, line 16: Maybe use `dig` here just in case the `agency` key is missing in the JSON payload

app/models/court_case.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    self
+  end
+
+  def location_id
+    location = Location.find_by nomis_agency_id: @agency_id
+    if location
+      location.id
+    end
+  end
```
smoothcontract, line 26: Could simplify this a little bit to something like

```ruby
Location.find_by(nomis_agency_id: @agency_id)&.id
```

app/models/court_case.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def location
+    location = Location.find_by nomis_agency_id: @agency_id
```
smoothcontract, line 29: Looks like an opportunity to DRY up the `Location` lookup - perhaps by amending this method to return the `Location` model rather than the `LocationSerializer`? I would have thought that it's not the models responsibility to perform JSON serialisation 🤔 

So this would then simply be

```ruby
def location
  Location.find_by nomis_agency_id: @agency_id
end
```

or, even better this gives an opportunity to cache the location lookup:

```ruby
def location
  @location ||= Location.find_by nomis_agency_id: @agency_id
end
```

This then means the method above can be amended to

```ruby
def location_id
  location&.id
end
```

(or potentially this is simple enough that you can just use `location&.id` in the calling code).

app/serializers/court_case_serializer.rb
```diff
@@ -0,0 +1,41 @@
...
+  #   end
+  # end
+
+  # def location
+  #   location = Location.find_by nomis_agency_id: object.agency_id
+  #
+  #   if location
+  #     LocationSerializer.new(location)
+  #   end
+  # end
```
smoothcontract, line 40: Remove commented code?

app/serializers/court_case_serializer.rb
```diff
@@ -0,0 +1,41 @@
+class CourtCaseSerializer < ActiveModel::Serializer
+  attributes :nomis_case_id, :nomis_case_status, :nomis_case_start_date,
+             :nomis_case_type, :nomis_case_number
+
+  belongs_to :location, serializer: LocationSerializer
+
+  def nomis_case_id
+    object.id
```
smoothcontract, line 8: Same thing here with the inconsistent use of `id` - amending this to just `id` means it's more compatible with JSONAPI spec, right?

swagger/v1/court_case.json
```diff
@@ -0,0 +1,76 @@
...
+        ],
+        "properties": {
+          "nomis_case_id": {
+            "type": "string",
+            "example": "T20167984",
+            "description": "The case id that uniquely identifies the court case"
+          },
+          "nomis_case_status": {
+            "type": "string",
+            "oneOf": ["foo", "bar"],
```
willfish, line 39: Need to enumerate these

swagger/v1/court_case.json
```diff
@@ -0,0 +1,76 @@
...
+            "description": "The state of the nomis court case"
+          },
+          "nomis_case_start_date": {
+            "type": "string",
+            "example": "2020-01-01",
+            "description": "ISO8601-compatible date"
+          },
+          "nomis_case_type": {
+            "type": "string",
+            "oneOf": ["foo", "bar"],
```
willfish, line 50: And these

swagger/v1/court_case.json
```diff
@@ -0,0 +1,76 @@
...
+            "example": "",
+            "description": "The type of the nomis court case"
+          },
+          "nomis_case_number": {
+            "type": "string",
+            "example": "T20167984",
```
willfish, line 56: Is it right that we have a case number and a case id. Would be good to get the nomis definitions of these, too

app/models/court_case.rb
alexdesi: Good point, In general ID is a UUID in our models/tables, but in this case we do not have a ID, because the court case is not persistent. That's why I'm setting it to the `case_info_number`. 

Not sure about removing the `case_info_number`, that would be unexpected by the FE (I think it's already using it)

app/models/court_case.rb
alexdesi: The agency is an important field here (it identify the Court), so we expect it to be there. I think it's better to raise a exception here so we can investigate it's missing.

app/models/court_case.rb
alexdesi: yes I like the `&.id` operator, I'm gonna use 2 lines just for readability:

```
location = Location.find_by nomis_agency_id: @agency_id
location&.id
```

app/models/court_case.rb
alexdesi: I agree, this should be simplified, potentially we do not need the `location_id` or the explicit use of the serializer

app/serializers/court_case_serializer.rb
alexdesi: I am using the `object.case_id` now.

swagger/v1/court_case.json
```diff
@@ -0,0 +1,76 @@
...
+          "nomis_case_id",
+          "nomis_case_status",
+          "nomis_case_start_date",
+          "nomis_case_type",
+          "nomis_case_number"
+        ],
+        "properties": {
+          "nomis_case_id": {
+            "type": "string",
+            "example": "T20167984",
```
alexdesi, line 34: the example is actually a case_number, it should be a case_id.

swagger/v1/court_case.json
alexdesi: Yes we have the we have a case number and a case id. We should check the doc or ask Matt for the nomis definition.

swagger/v1/court_case.json
willfish: 👍

app/models/court_case.rb
smoothcontract: Okay 👍

app/lib/nomis_client/court_cases.rb
```diff
@@ -0,0 +1,19 @@
...
+      ACTIVE_CASES_FILTER = 'activeOnly=true'
+
+      def get(booking_id)
+        court_cases_route = "/bookings/#{booking_id}/court-cases?#{ACTIVE_CASES_FILTER}"
+
+        begin
+          NomisClient::Base.get(court_cases_route).body
+        rescue StandardError => e
+          raise e
+        end
```
smoothcontract, line 15: I think you could remove this exception handler as it's not really doing anything

spec/swagger/api/v1/people_controller_court_cases_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+        run_test!
+
+      end
+
+      # response '404', 'not found' do
+      #   let(:id) { 'invalid-id' }
+      #
+      #   run_test!
+      # end
```
smoothcontract, line 64: Remove commented code unless planning to complete this test?

app/services/people/retrieve_court_cases.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module People
+  class RetrieveCourtCases
+    def self.call(person)
+      nomis_court_cases_response = NomisClient::CourtCases.get(person.latest_nomis_booking_id)
+
+      JSON.parse(nomis_court_cases_response).map do |court_case|
```
willfish, line 8: Do we want to order by start date or is the Nomis api always sorting correctly?

willfish: Talked to @alexdesi about this and we agreed that this isn't necessarily the responsibility of this api/nor necessary for this iteration.

spec/lib/nomis_client/court_cases_spec.rb
```diff
@@ -4,5 +4,16 @@
 
 RSpec.describe NomisClient::CourtCases, with_nomis_client_authentication: true do
   describe '.get' do
+    let(:nomis_client) { class_double(NomisClient::Base).as_stubbed_const }
+
+    let(:booking_id) { '1495077' }
+
+    it '#get' do
```
willfish, line 11:
```suggestion
    it 'calls the nomis client with the correct arguments' do
```

spec/lib/nomis_client/court_cases_spec.rb
```diff
@@ -4,5 +4,16 @@
 
 RSpec.describe NomisClient::CourtCases, with_nomis_client_authentication: true do
   describe '.get' do
+    let(:nomis_client) { class_double(NomisClient::Base).as_stubbed_const }
```
willfish, line 7:
```suggestion
    subject(:nomis_client) { class_double(NomisClient::Base).as_stubbed_const }
```

app/lib/nomis_client/court_cases.rb
alexdesi: yes, this kind of "reminder" to handle errors from nomis, but will handle them in a different ticket  (we need a general solution)

PR #355, opened 2020-04-02
P4-1213 Use StringIO rather than temp file to serialize image blob

smoothcontract:
Fixes P4-1213

## Proposed Changes

Previously this implementation made use of a Tempfile to write and then immediately read back the image content retrieved from Nomis API. That didn't feel like a great technique as the entire image blob is already held in memory at that point as a string. Wrapping the same string with a `StringIO` instance means it can be passed directly to active storage for attachment.

I was able to reproduce a consistent test failure on Circle CI prior to this change, and the same seed passes with the change so it looks very likely that using temp files on Circle CI is a little bit brittle.

## To test/demo change

Tested `RetrieveImage` service locally and ensured that prisoner images are successfully fetched from Nomis and persisted using active storage when calling `/people/{id}/images` endpoint.

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -15,15 +15,13 @@ class Person < VersionedModel
   end
 
   def attach_image(image_blob)
-    filename = id + '.jpg'
-
-    tempfile = Tempfile.new('temp.jpg').binmode
-    tempfile.write image_blob
-    tempfile.rewind
-
-    image.attach(io: tempfile, filename: filename, content_type: 'image/jpg')
-    tempfile.close
-
-    filename
+    "#{id}.jpg".tap do |filename|
+      image_io = StringIO.new(image_blob).binmode
+      begin
+        image.attach(io: image_io, filename: filename, content_type: 'image/jpg')
+      ensure
+        image_io.close
+      end
+    end
   end
 end


app/models/person.rb
```diff
@@ -15,15 +15,13 @@ def latest_nomis_booking_id
...
-    tempfile.rewind
-
-    image.attach(io: tempfile, filename: filename, content_type: 'image/jpg')
-    tempfile.close
-
-    filename
+    "#{id}.jpg".tap do |filename|
+      image_io = StringIO.new(image_blob).binmode
+      begin
+        image.attach(io: image_io, filename: filename, content_type: 'image/jpg')
```
martyn-w, line 21: Looks good - but just wondering on a wider question: how do we know the image is always a `jpeg` (and not say a `png` or `bmp` etc). Would it be possible to pass the mime_type as well as the blob?

smoothcontract: That's a good question - I didn't amend that aspect but did some digging. This method is only used to persist images that are retrieved from Nomis. I found the relevant API implementation which states that the return type is always `image/jpeg` so I think it's fine to bake in this assumption. Without this we would need some form of image parser as we just get a load of bytes back from this API call rather than a nice structured object:

```java
    @GetMapping(value = "/{bookingId}/image/data", produces = "image/jpeg")
    @ApiOperation(value = "Image data (as bytes).", notes = "Image data (as bytes).", nickname = "getMainBookingImageData")
```

From https://github.com/ministryofjustice/elite2-api/blob/master/src/main/java/net/syscon/elite/api/resource/BookingResource.java#L357

PR #356, opened 2020-04-07
P4-1263 Update date validation rules around proposed moves

smoothcontract:
Fixes P4-1263

## Proposed Changes

Proposed moves will not specify a date, but will specify a date_from and may specify a date to. This updates the date validation rules on Move model accordingly to make them conditional on the appropriate move status. This also extends the variables passed to notification emails to include date_from and date_to, with those and also date defaulting to `N/A` if nil. This then allows the notify email template to be updated in a subsequent ticket to support from/to dates while also ensuring that nothing breaks in the meantime.

### DIFF
--- a/app/mailers/move_mailer.rb
+++ b/app/mailers/move_mailer.rb
@@ -13,7 +13,10 @@ class MoveMailer < GovukNotifyRails::Mailer
         'from-location': move.from_location.title,
         # NB: to_location isn't set for prison_recall moves, so use N/A instead (GovUK Notify will error if nil is supplied)
         'to-location': move.to_location&.title || 'N/A',
-        'move-date': move.date.strftime(DATE_FORMAT),
+        # NB: date isn't set for proposed moves, so use N/A instead (GovUK Notify will error if nil is supplied)
+        'move-date': move.date&.strftime(DATE_FORMAT) || 'N/A',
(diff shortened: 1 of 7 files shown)


spec/requests/api/v1/moves_controller_create_spec.rb
```diff
@@ -252,12 +252,6 @@
 
       let(:errors_422) do
         [
-          {
-            'title' => 'Unprocessable entity',
-            'detail' => "Date can't be blank",
-            'source' => { 'pointer' => '/data/attributes/date' },
-            'code' => 'blank',
-          },
```
smoothcontract, line 260: This is the only test that broke as a result of making `date` optional for `proposed` moves. What's happening with this spec is that an invalid status value is passed as a property; that then means that a move instance is created but does not get the invalid status assigned. Instead the status value defaults to the first one defined in the enumerator, which is `proposed`. Hence as a proposed move the missing date attribute is not considered an error. Hmm.. as I'm writing this I'm thinking I might instead update the test to check that missing `Date from` throws an error instead 🤔

teneightfive:
Could we add something to the documentation to explain the different rules based on different statuses?

smoothcontract:
> Could we add something to the documentation to explain the different rules based on different statuses?

I've had a go at that, but it's a bit limiting really with the OpenAPI spec.

PR #359, opened 2020-04-08
P4-1263 Minor tweak for move date attribute API spec

smoothcontract:
Fixes P4-1263

## Proposed Changes

This is an optional field now (for proposed moves) so the API needs to allow values of either `nil` or a date. Updated API spec to reflect this behaviour and demonstrate successful creation of proposed move.

## To test/demo change

- See updated API spec

## Things to check & link
- [x] API docs has been updated if necessary

### DIFF
--- a/swagger/v1/move.json
+++ b/swagger/v1/move.json
@@ -74,8 +74,13 @@
             "readOnly": true
           },
           "date": {
-            "type": "string",
-            "format": "date",
+            "oneOf": [
+              {
+                "type": "string",
+                "format": "date"
+              },
+              { "type": "null" }
+            ],
             "example": "2020-05-17",
             "description": "Date on which the move is scheduled (mandatory unless status is proposed)"
           },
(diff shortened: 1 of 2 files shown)


swagger/v1/move.json
```diff
@@ -74,8 +74,13 @@
             "readOnly": true
           },
           "date": {
-            "type": "string",
-            "format": "date",
+            "oneOf": [
```
willfish, line 77: Isn't this equivalent to `"nullable": true`?

willfish: https://swagger.io/docs/specification/data-models/data-types/#null

smoothcontract: Possibly, but I've kept this consistent with other data types

smoothcontract: I checked the existing API schema, and it seems we only use `nullable` for optional relationships. So I'd prefer to keep this consistent with everything else.

PR #361, opened 2020-04-09
P4-1263 Updated unique date validation to exclude proposed moves

smoothcontract:
Fixes P4-1263

## Proposed Changes

Proposed moves will initially set a `nil` date on creation, and may then be updated to specify a potential fixed date. We should only enforce validation of a unique move date when the move changes to requested or completed state. Updated model and specs.

This was preventing more than one proposed move being created for the same person, from location and to location (was reported as working intermittently).

### DIFF
--- a/spec/factories/move.rb
+++ b/spec/factories/move.rb
@@ -13,10 +13,6 @@ FactoryBot.define do
     sequence(:created_at) { |n| Time.now - n.minutes }
     sequence(:date_from) { |n| Date.today - n.days }
 
-    trait :requested do
-      status { 'requested' }
-    end
-
     trait :cancelled do
       status { 'cancelled' }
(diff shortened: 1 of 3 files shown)


spec/factories/move.rb
```diff
@@ -13,10 +13,6 @@
     sequence(:created_at) { |n| Time.now - n.minutes }
     sequence(:date_from) { |n| Date.today - n.days }
 
-    trait :requested do
-      status { 'requested' }
-    end
-
```
smoothcontract, line 19: This is redundant as the factory already specifies a default status of `requested`. There's also a trait named `requested` further down the same factory. Tests pass without this so I think it's safe to remove and cleans up the factory although isn't entirely related to this PR.

PR #365, opened 2020-04-10
[P4-1165] Extrapolate court hearing controller from move controller

willfish:
P4-1165

### What

- [x] Adds controller for creating a court hearing
- [x] Removes creation of court hearing from move controller
- [x] Adds swagger documentation for court hearing
- [x] Update `POST /move` so that it creates the link with existing court hearing in a json api way
- [x] Update `POST /court_hearing` so that it creates the link with existing move in a json api way

### Why

- json api does not support creating nested attributes - it plans to in future implementations, though
- this implementation is actually cleaner

### AC

- [x] Existing `POST /move` and `GET /move/:id` still return expected relationships and includes
- [ ] Dom is happy with the swagger documentation we generate

### Future PRs

- Plumb posting court hearing to Nomis service into the `POST /court_hearings` endpoint


spec/requests/api/v1/court_hearings_controller_create_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+                  description: 'The court hearing to create',
+                  in: :body,
+                  attributes: {
+                    schema: '#/definitions/court_hearing/CourtHearing',
+                  }
+
+        let(:court_hearing) do
+          {
+            data: {
+              type: 'court_hearing',
```
teneightfive, line 37: I think `type` needs to be a plural

swagger/v1/court_hearing.json
```diff
@@ -7,65 +7,71 @@
...
-            "description": "Arbitrary comments that are useful for humans but not touched by computers."
-          },
-          "case_start_date": {
-            "type": "string",
-            "example": "2020-01-01",
-            "description": "ISO8601-compatible date indicating when the case started."
+          "type": { "type": "string", "enum": "court_hearings" },
+          "attributes": {
+            "type": "object",
+            "required": [
```
teneightfive, line 14: I’m not sure all of these values should be required at this stage. I think only:

- `start_time`

swagger/v1/court_hearing.json
```diff
@@ -7,65 +7,71 @@
...
+            ],
+            "properties": {
+              "id": {
+                "type": "string",
+                "format": "uuid",
+                "example": "22036201-4301-4021-8ee6-80bc415b8534",
+                "description": "The universally unique and sequential primary identifier for this entity"
+              },
+              "start_time": {
+                "type": "string",
```
teneightfive, line 33: Should this be a `datetime`?

swagger/v1/court_hearing.json
```diff
@@ -7,65 +7,71 @@
...
+                "description": "The nomis reference for the case of the current hearing."
+              },
+              "nomis_hearing_id": {
+                "type": "string",
+                "example": "4232424",
+                "nullable": true,
+                "description": "After a new hearing is created in Nomis backend it receives a unique id. This can be null and means we failed to create a corresponding hearing in Nomis or did not need too."
+              },
+              "saved_to_nomis": {
+                "type": "string",
```
teneightfive, line 54: Should the type be `boolean`?

swagger/v1/court_hearing.json
```diff
@@ -7,65 +7,71 @@
...
+                "example": false,
+                "description": "Indicates whether we successfully created a corresponding court hearing in Nomis' backend."
+              },
+              "case_type": {
+                "type": "string",
+                "example": "Adult",
+                "enum": ["Adult"],
```
teneightfive, line 61: Might be worth checking the other options as part of this so that we don’t get errors sending from the frontend with other types that are returning with court cases

swagger/v1/court_hearing.json
```diff
@@ -7,65 +7,71 @@
...
+                "description": "The type of the court that the court hearing is being held in."
+              },
+              "comments": {
+                "type": "string",
+                "example": "Witness for Joe Bloggs in Foo Bar court hearing.",
+                "description": "Arbitrary comments that are useful for humans but not touched by computers."
+              },
+              "case_start_date": {
+                "type": "string",
```
teneightfive, line 70: Should type be `date`?

spec/requests/api/v1/court_hearings_controller_create_spec.rb
willfish, line 37:
```suggestion
              type: 'court_hearings',
```

swagger/v1/court_hearing.json
willfish: We can use a formatting validator, actually.

swagger/v1/court_hearing.json
```diff
@@ -7,65 +7,71 @@
...
+            "properties": {
+              "id": {
+                "type": "string",
+                "format": "uuid",
+                "example": "22036201-4301-4021-8ee6-80bc415b8534",
+                "description": "The universally unique and sequential primary identifier for this entity"
+              },
+              "start_time": {
+                "type": "string",
+                "example": "2018-01-01T18:57Z",
```
willfish, line 34:
```suggestion
                "format": date-time,
                "example": "2018-01-01T18:57Z",
```

swagger/v1/court_hearing.json
willfish, line 54:
```suggestion
                "type": "boolean",
```

swagger/v1/court_hearing.json
willfish: Yeah, will need to do that before shipping

swagger/v1/court_hearing.json
```diff
@@ -7,65 +7,71 @@
...
+                "description": "The type of the court that the court hearing is being held in."
+              },
+              "comments": {
+                "type": "string",
+                "example": "Witness for Joe Bloggs in Foo Bar court hearing.",
+                "description": "Arbitrary comments that are useful for humans but not touched by computers."
+              },
+              "case_start_date": {
+                "type": "string",
+                "example": "2020-01-01",
```
willfish, line 71:
```suggestion
                "format": "date",
                "example": "2020-01-01",
```

swagger/v1/court_hearing.json
willfish: Using format here, too 👍 . Json doesn't support an actual date type. We use a string encoded using the ISO8601 standard and format validates that :)

swagger/v1/court_hearing.json
willfish: Will remove these as required.

teneightfive: Can we remove the required on all but `start_time` to begin with. Our journey currently has `comments` as an optional field so that will throw an error at this stage of all attributes are required. 

Not all hearings will have NOMIS related case information either when we do the second part of this so I think it would be better to start with minimal required attributes and add required ones as we need to.

willfish: Done.

app/controllers/api/v1/court_hearings_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+          :case_start_date,
+          :nomis_case_number,
+          :nomis_case_id,
+          :case_type,
+          :comments,
+        )
+      end
+
+      def move
+        return if params.require(:data)[:relationships].blank?
```
alexdesi, line 28: what if you are other relationship?

willfish: If we take the relationship params out we can handle each entity separately.

swagger/v1/court_hearing.json
willfish: From the documentation in nomis I can't see a list of the possible values. I think we can get away with not specifying an enum here?

alexdesi: Screen Shot 2020-04-10 at 16 55 05

app/controllers/api/v1/court_hearings_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+
+      def court_hearings_attributes
+        court_hearings_params.merge(move: move)
+      end
+
+      def court_hearings_params
+        params.require(:data).require(:attributes).permit(
+          :start_time,
+          :case_start_date,
+          :nomis_case_number,
```
teneightfive, line 20: Think this needs changing to `case_number` instead of `nomis_case_number` to match the docs?

app/controllers/api/v1/court_hearings_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+          :nomis_case_id,
+          :case_type,
+          :comments,
+        )
+      end
+
+      def move
```
teneightfive, line 27: Is it possible to add this relationship to the swagger docs?

swagger/v1/court_hearing.json
willfish: Yeah, no list of possible values :(.

app/controllers/api/v1/court_hearings_controller.rb
willfish: This is fixed I believe, yeah

PR #367, opened 2020-04-14
Call the service to create court hearings in Nomis

alexdesi:
Jira: P4-1308-nomis-prison-to-court

## Proposed Changes

This PR allows the hearing court to be created in nomis.

### DIFF
--- a/app/services/court_hearings/create_in_nomis.rb
+++ b/app/services/court_hearings/create_in_nomis.rb
@@ -1,6 +1,6 @@
 module CourtHearings
   class CreateInNomis
-    def call(move, court_hearings)
+    def self.call(move, court_hearings)
       booking_id = move.person.latest_nomis_booking_id
 
       body_locations = {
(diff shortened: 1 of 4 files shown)


spec/requests/api/v1/court_hearings_controller_create_spec.rb
```diff
@@ -59,6 +59,14 @@
 
         expect(move.court_hearings).to include(court_hearing)
       end
+
+      it 'creates the court hearing in Nomis' do
+        allow(CourtHearings::CreateInNomis).to receive(:call)
```
willfish, line 64: You'll want to hoist this to the top of the context for all specs that create a move.

app/services/court_hearings/create_in_nomis.rb
```diff
@@ -1,6 +1,6 @@
 module CourtHearings
   class CreateInNomis
-    def call(move, court_hearings)
+    def self.call(move, court_hearings)
```
willfish, line 3: You'll want to change the specs to reflect this change :)

spec/requests/api/v1/court_hearings_controller_create_spec.rb
alexdesi: 👍

PR #369, opened 2020-04-15
P4-1274 Added data model to support allocations

smoothcontract:
Fixes P4-1274

## Proposed Changes

This is the first pass that does not yet support associating people with the allocation (that will come in a later sprint). Eventually it's likely that an allocation will support a collection of moves, but that's also out of scope for the time being, as is allocation status. I've elected to use JSON storage for the complex case flags - the same approach is used elsewhere e.g. assessment answers to avoid over normalising the schema.

## To test/demo change

This PR doesn't do much yet, but I'm keen to get agreement on model & attribute names before proceeding further to add API docs etc.

### DIFF
--- a/app/models/allocation.rb
+++ b/app/models/allocation.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+class Allocation < VersionedModel
+  enum prisoner_category: {
+    b: 'B',
+    c: 'C',
+    d: 'D',
+  }
+
+  enum sentence_length: {
+    short: '16_or_less',
+    long: 'more_than_16',
+  }
+
+  belongs_to :from_location, class_name: 'Location'
+  belongs_to :to_location, class_name: 'Location'
+
+  validates :from_location, presence: true
+  validates :to_location, presence: true
+
+  validates :prisoner_category, inclusion: { in: prisoner_categories }, allow_nil: true
+  validates :sentence_length, inclusion: { in: sentence_lengths }, allow_nil: true
+
+  validates :moves_count, presence: true
+  validates :date, presence: true
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_04_14_104553) do
+ActiveRecord::Schema.define(version: 2020_04_15_091520) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
@@ -34,6 +34,21 @@ ActiveRecord::Schema.define(version: 2020_04_14_104553) do
     t.string "checksum", null: false
     t.datetime "created_at", null: false
     t.index ["key"], name: "index_active_storage_blobs_on_key", unique: true
+  end
+
+  create_table "allocations", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.uuid "from_location_id", null: false
+    t.uuid "to_location_id", null: false
+    t.date "date", null: false
+    t.string "prisoner_category"
+    t.string "sentence_length"
+    t.jsonb "complex_cases"
+    t.integer "moves_count", null: false
+    t.boolean "complete_in_full", default: false, null: false
+    t.text "other_criteria"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["date"], name: "index_allocations_on_date"
   end
 
   create_table "assessment_questions", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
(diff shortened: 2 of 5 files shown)


app/models/allocation.rb
```diff
@@ -0,0 +1,27 @@
...
+  belongs_to :from_location, class_name: 'Location'
+  belongs_to :to_location, class_name: 'Location'
+
+  validates :from_location, presence: true
+  validates :to_location, presence: true
+
+  validates :prisoner_category, inclusion: { in: prisoner_categories }
+  validates :sentence_length, inclusion: { in: sentence_lengths }
+
+  validates :moves_count, presence: true
```
smoothcontract, line 25: The naming of this column is deliberate - will likely use `:counter_cache` to populate this in future. For now it will be set explicitly in the front end.

martyn-w:
👍 Looks good to me - but best if reviewed by someone on the Allocations stream

app/models/allocation.rb
```diff
@@ -0,0 +1,27 @@
...
+
+class Allocation < VersionedModel
+  enum prisoner_category: {
+    b: 'B',
+    c: 'C',
+    d: 'D',
+  }
+
+  enum sentence_length: {
+    any: 'any',
```
teneightfive, line 11: What about `any` actually just being a null value for this attribute?

It feels like these (`prisoner_category`, `sentence_length`, etc) are essentially filters and maybe we could use the same approach as we use for filters on our endpoints - if there is no filter present, or an empty (`null` in this case) value for that filter, it returns all results. 

I'm just thinking about how these are going to be used in the future and if we have a standard pattern for filters it will allow us to consume and send them in a consistent way from the frontend client. It also feels cleaner from a filtering perspective as then we just don't pass a filter rather than having to pass `any` as a value for that filter when we don't want it to be applied.

smoothcontract: Yep, that makes sense to me. So `sentence_length` will permit null values to indicate no preference and is effectively an optional attribute. However in the front end the user does have to pick one of the mandatory radio buttons.

I think `prisoner_category` will need to be a mandatory attribute to match the proposed front end design (there's no `Any` option).

teneightfive: I think we can be more lenient with the API design here rather than requiring to tie it directly to the screens. We can have a separation between how we collect the data from users and how we display it back to them in the frontend client and the way we store the data in the DB and send to the API endpoints. 

I think it would be good to use a generic filter structure which could support `null` (no filtering needed), a single value, or an array of values for these attributes. 

I just had a chat with @Andy-Millar and basically these are just filters we're collecting at this stage to learn from the data and will evolve over time - we could add/remove/tweak options. So it might be better to have a more standard and flexible approach with how we store them in the API. 

And then we keep the design of what "type" a filter is (ie, radio vs checkbox vs any options) can just be left to the frontend rather than needing to validate in any way on the backend too.

I think I'm conscious of us not designing the data structure of API too tightly to specific components within the UI. The need of the frontend for this is to send a list of filters/criteria to store. This is then played back to the other user in this journey to allow them to pick suitable people to move, and in the long-term do some of that filtering for them.

smoothcontract: All good thoughts and useful feedback. I'm happy at this stage to also support `null` values for each of these filter types - we agreed above that `sentence_length` could support `null`, the `complex_cases` already allows `null` and we can allow `null` for `prisoner_category` too.

I think it's a little too early though to try and second guess other filtering options and try to devise a generic approach. I'd prefer to go with something simple for now and let the design evolve into something more complex as the requirements shift over time. YAGNI and KISS and all that :)

smoothcontract: One other thing - this does mean that the API will potentially have different validation rules to the front end - i.e. it would be possible for a pure API client to create an allocation without specifying a `sentence_length` or `prisoner_category`. That feels like it's probably okay (presumably an allocation for any prisoner category is a valid thing), but worth mentioning.

We will likely be using filtering in future to return the data, but it's important also to ensure that only valid/sensible data gets created. If the back end doesn't provide validation of these attributes being present then it changes the design a little with some front end validation being needed before submitting an API request and then potentially handling further errors. Hopefully that's already an established pattern.

db/schema.rb
```diff
@@ -36,6 +36,21 @@
     t.index ["key"], name: "index_active_storage_blobs_on_key", unique: true
   end
 
+  create_table "allocations", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.uuid "from_location_id", null: false
+    t.uuid "to_location_id", null: false
+    t.date "move_date", null: false
```
barbaracassani, line 42: In the FE for moves we use just `date`. Not sure if it's reflected on the BE as well? Anyway, not very important.

app/models/allocation.rb
teneightfive: Yep, all in favour of keeping it simple. I think just supporting `null` on those is a good approach for now and not passing values like `any` that suggest no filtering. In the case a user an "all" option  the frontend can just pass `null`.

Regarding the validation I think that's also okay for now as we're not actually doing anything with any of the filters. They are more a data gathering exercise for now and then played back to the users selecting people as information. We're not doing any validation that end yet so I think its acceptable to have them as optional in the API.

And if in future we start validating a person that is added against a move based on the allocation criteria we can define business rules around what also needs validating in the allocation model at that time.

Would it be easier in future to support multiple values for things like `prisoner_category` or `sentence_length` if they were changed from radios to checkboxes for example? I think that's definitely something work thinking now to make sure we're not in too much of a corner if that happens.

smoothcontract: Thanks for all the feedback @teneightfive - it's very useful 👍. I've updated the PR to allow `null` values for each of the "filter" attributes, but will keep them as separate attributes on the table for now. If we decide in future to allow filtering or populating multiple values for these then I suspect they can be moved to a `JSONB` storage - perhaps by merging all of them together. That wouldn't be hard to do if the need arises.

db/schema.rb
smoothcontract: No worries, happy to change to `date` 👍

PR #370, opened 2020-04-15
Save the CourtHearing.post response and Fix the time format

alexdesi:
This PR allow to Save the CourtHearing.post response and Fix the time format (UTC)

In case the response is not 201 it pushes the result to sentry to easily debug the error.

### DIFF
--- a/app/services/court_hearings/create_in_nomis.rb
+++ b/app/services/court_hearings/create_in_nomis.rb
@@ -9,12 +9,20 @@ module CourtHearings
       }
 
       court_hearings.each do |hearing|
-        NomisClient::CourtHearing.post(booking_id: booking_id,
+        response = NomisClient::CourtHearing.post(booking_id: booking_id,
                                        court_case_id: hearing.nomis_case_id,
                                        body_params: {
-                                           'courtHearingDateTime': hearing.start_time.iso8601,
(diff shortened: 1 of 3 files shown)


app/services/court_hearings/create_in_nomis.rb
```diff
@@ -9,12 +9,20 @@ def self.call(move, court_hearings)
       }
 
       court_hearings.each do |hearing|
-        NomisClient::CourtHearing.post(booking_id: booking_id,
+        response = NomisClient::CourtHearing.post(booking_id: booking_id,
                                        court_case_id: hearing.nomis_case_id,
                                        body_params: {
-                                           'courtHearingDateTime': hearing.start_time.iso8601,
+                                           'courtHearingDateTime': hearing.start_time.utc.iso8601,
```
willfish, line 15: I think this is okay to get it working but we should put something in the calendar to catchup with Matt when he gets back.

willfish: I also think we should write a test that validates that we're mutating the start time on the way out to nomis

alexdesi: yes! and we should develop a general solution (we might forget to add utc.iso8601 to the time fields)

PR #371, opened 2020-04-16
P4 1148 journey event models

martyn-w:
Fixes P4-1148
https://dsdmoj.atlassian.net/browse/P4-1148

## Proposed Changes

- Adds new journey and event models, and associated state machine. These will be used in future developments.

## To test/demo change

- rspec

- requires database to be migrated (happens automatically on deploy)


app/models/journey.rb
```diff
@@ -0,0 +1,42 @@
...
+  validates :to_location, presence: true
+  validates :client_timestamp, presence: true
+  validates :billable, exclusion: { in: [nil] }
+  validates :state, presence: true, inclusion: { in: states }
+
+  scope :default_order, -> { order(client_timestamp: :asc) }
+
+  delegate :complete, :uncomplete, :cancel, :uncancel, to: :state_machine
+
+  after_initialize :synchronise_state # NB there is an equivalent after(:build) callback used by FactoryBot in the journeys factory
```
alexdesi, line 25: Call backs are difficult to maintain and make things difficult to understand (a step towards spaghetti code)
but I am not sure about the alternative atm, 
Any idea from everyone?

app/state_machines/journey_state_machine.rb
```diff
@@ -0,0 +1,12 @@
+class JourneyStateMachine < FiniteMachine::Definition
```
alexdesi, line 1: It's great to have the SM in a different class, outside the models! 👍

spec/models/journey_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+RSpec.describe Journey, type: :model do
+  it { is_expected.to belong_to(:move) }
+  it { is_expected.to belong_to(:supplier) }
+  it { is_expected.to belong_to(:from_location) }
+  it { is_expected.to belong_to(:to_location) }
+  it { is_expected.to validate_presence_of(:move) }
+  it { is_expected.to validate_presence_of(:supplier) }
+  it { is_expected.to validate_presence_of(:from_location) }
+  it { is_expected.to validate_presence_of(:to_location) }
+  it { is_expected.to validate_presence_of(:client_timestamp) }
```
alexdesi, line 14: what do we mean with  `client`?

spec/models/journey_spec.rb
```diff
@@ -0,0 +1,95 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Journey, type: :model do
+  it { is_expected.to belong_to(:move) }
+  it { is_expected.to belong_to(:supplier) }
```
alexdesi, line 7: should not be the move linked with the supplier?

spec/models/journey_spec.rb
martyn-w: client usually means the supplier; but it could also be a police station or PMU, or indeed any thing which could trigger events on a move

spec/models/journey_spec.rb
martyn-w: as discussed, it probably should be linked, but not as part of this work :-)

martyn-w:
@alexdesi following our call I've renamed some of the methods and rspec examples - please could you re-review? thanks.

app/models/event.rb
```diff
@@ -0,0 +1,24 @@
+class Event < ApplicationRecord
+  belongs_to :move
+
+  enum event_types: {
```
alexdesi, line 4: it looks like these are more `event_names`

app/models/event.rb
```diff
@@ -0,0 +1,24 @@
...
+    move_redirected: 'move_redirected',
+    move_lockout: 'move_lockout',
+    journey_created: 'journey_created',
+    journey_updated: 'journey_updated',
+    journey_completed: 'journey_completed',
+    journey_uncompleted: 'journey_uncompleted',
+    journey_cancelled: 'journey_cancelled',
+    journey_uncancelled: 'journey_uncancelled',
+  }
+
```
alexdesi, line 18: I've noticed this just now, it looks like we got 2 only `event_types`: `move`, `juorney`
and we also have an `action`: `create, update, cancel, restore, ...`

What do you think? 
it might be worth to refactor this, if not what about renaming event_types to event_names?

Gemfile.lock
```diff
@@ -113,6 +113,8 @@ GEM
     faraday (0.15.4)
       multipart-post (>= 1.2, < 3)
     ffi (1.11.1)
+    finite_machine (0.12.1)
```
willfish, line 116: According to semantic versioning this isn't meant for production, yet?

martyn-w: hmm... apparently it was good enough for Marks & Spencer :-)

app/models/event.rb
```diff
@@ -0,0 +1,24 @@
+class Event < ApplicationRecord
```
willfish, line 1: This feels super coupled to a move. Do we want to rename this `MoveEvent`?

app/models/journey.rb
```diff
@@ -0,0 +1,42 @@
+class Journey < ApplicationRecord
+  belongs_to :move
+  belongs_to :supplier
+  belongs_to :from_location, class_name: 'Location'
```
willfish, line 4: Could we delegate to the move location or is it simpler/more flexible to duplicate these?

martyn-w: ah nope we can't delegate - a move from A->B may be composed of multiple journeys A -> C ->D -> B. 
So locations definitely need to be stored at both the Move level and at the Journey level.

app/state_machines/journey_state_machine.rb
```diff
@@ -0,0 +1,12 @@
...
+  initial :in_progress
+
+  event :complete, in_progress: :completed
+  event :uncomplete, completed: :in_progress
```
willfish, line 5: Wondering if the language here is confusing 🤔

spec/models/journey_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+    end
+  end
+
+  describe 'state_machine <--> journey state synchronisation' do
+    # These tests verify that the synchronisation between journey.state and the internal state_machine behave as expected
+    # for initialised, created, found and built records
+
+    let(:initial_state) { nil }
+    let(:state) { journey.state }
+    let(:state_machine_state) { journey.send(:state_machine).current }
```
willfish, line 68: Are we testing a private implementation detail, here?

martyn-w:
@willfish I have somehow lost your comments when trying to reply to them :-(  here are my responses:

> (Re: event model) This feels super coupled to a move. Do we want to rename this MoveEvent?

Good question... yes, Events are all about Moves. The same argument could be made about Journeys. But I think I prefer the more generic names "Event" and "Journey" as opposed to "MoveEvent" and "MoveJourney". 

> (Re state machine) Wondering if the language here is confusing 🤔

I am open to alternatives for states: "in progress", "completed", "cancelled"
And for alternatives for events: "complete", "uncomplete", "cancel", "uncancel"
- feel free to suggest something clearer?

spec/factories/events.rb
```diff
@@ -0,0 +1,55 @@
+FactoryBot.define do
+  factory :event do
+    association(:move)
+    event_type { 'move_created' }
+    client_timestamp { Time.now.utc + rand(-60..60).seconds } # NB: the client_timestamp will never be perfectly in sync with system clock
```
willfish, line 5: Is this an ecological validity of our fixtures thing? Doesn't seem necessary/tested/important for any of our examples?

spec/models/journey_spec.rb
martyn-w: in short, yes: I am testing that the state machine's state matches the model's state. The two are connected by code. The state machine's state is not accessible outside of the state machine so the only way I can see to test it matches the model is by inspection?

spec/factories/journeys.rb
```diff
@@ -0,0 +1,12 @@
+FactoryBot.define do
+  factory :journey do
+    association(:move)
+    association(:supplier)
+    association(:from_location, factory: :location)
+    association(:to_location, :court, factory: :location)
+    client_timestamp { Time.now.utc + rand(-60..60).seconds } # NB: the client_timestamp will never be perfectly in sync with system clock
+
+    # NB we need to initialize_state because FactoryBot fires the after_initialize callback before the attributes are initialised!
+    after(:build) { |object| object.send(:initialize_state) }
```
willfish, line 10: I'm wondering if there's a less jiggery pokery way of achieving the correct call order of callbacks 🤔

spec/factories/events.rb
martyn-w: It will become important as the client_timestamp is a critical field out with our control for assembling the order of events. We will rely on clients to accurately keep their clocks in sync. This code models the fact the `client_timestamp` will be a few seconds either side of the `created_at` timestamp, which will become important further down the line.

spec/factories/journeys.rb
martyn-w: I couldn't see a better way - I need to set the state for `new()`, for `build()`, for `create()` and for `find()`... Open to other suggestions though?

app/models/journey.rb
martyn-w: I'm afraid that the callbacks are necessary to initialise the state_machine for the various events in a model's lifetime (`new()`, `build()`, `find()` and `create()`).

app/models/event.rb
martyn-w: yep, I think `event_names` is a better term, I'll rename it 👍

app/models/event.rb
martyn-w: There will be lots more events in future: GPS tracking locations; lodging, pickup and handovers. I'll think about breaking the name into two: `event_type` + `action`. 🤔

willfish:
The state machine dsl feels clunky to me. Also, I think we can use something that's more battle tested? https://www.ruby-toolbox.com/categories/state_machines

martyn-w:
@willfish I sought advice about state machines and went with a recommended engine. Unless there are strong opinions about the unsuitability of `finite_machine`, I feel it would be quite wasteful to remove it and replace it with something else at this point.

spec/factories/journeys.rb
willfish: I'd need to read through how they've plumbed together finite_state_machine to understand this. You've got it working so I'm fine with this for the meanwhile

willfish:
I agree @martyn-w . You've clearly put a lot of thought/understanding into this implementation and I don't think I have a good enough reason to hold off this PR

app/models/event.rb
alexdesi: 👍

PR #373, opened 2020-04-16
Fix sentry for create court hearing

alexdesi:
This PR fixes the way sentry notifies when creating court hearings.


app/lib/nomis_client/court_hearing.rb
```diff
@@ -6,7 +6,20 @@ class << self
       def post(booking_id:, court_case_id:, body_params: {})
         court_cases_route = "/bookings/#{booking_id}/court-cases/#{court_case_id}/prison-to-court-hearings"
 
-        NomisClient::Base.post(court_cases_route, body: body_params.to_json).body
+        begin
```
willfish, line 9: I don't think you need the begin/end, here.

willfish:
Seems like genuine CI failure

PR #379, opened 2020-04-17
Reinstates pending court_hearings_spec examples

willfish:
## Proposed Changes

We recently made some specs pending so we could carry on validating an integration with nomis to create court hearings in staging. The way we were mocking was leaking into other tests in the suite

- [x] Reinstate pending specs with correct mocking
- [x] Adds extra spec after additional functionality

### DIFF
--- a/spec/lib/nomis_client/court_hearings_spec.rb
+++ b/spec/lib/nomis_client/court_hearings_spec.rb
@@ -13,6 +13,18 @@ RSpec.describe NomisClient::CourtHearing, with_nomis_client_authentication: true
     let(:response_status) { 201 }
     let(:response_body) { '{}' }
 
+    let(:raven_args) do
+      [
+        'CourtHearings:CreateInNomis success!',
+        extra: {
+          body_params: {},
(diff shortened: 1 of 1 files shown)


spec/lib/nomis_client/court_hearings_spec.rb
```diff
@@ -2,31 +2,71 @@
...
+RSpec.describe NomisClient::CourtHearing, with_nomis_client_authentication: true do
   describe '.post' do
     subject(:court_hearing_post) {
       described_class.post(booking_id: booking_id, court_case_id: court_case_id, body_params: {})
     }
 
     let(:booking_id) { 1111 }
     let(:court_case_id) { 2222 }
+    let(:response_status) { 201 }
+    let(:response_body) { '{}' }
```
alexdesi, line 14: I think we don't need these lines

willfish: I added them so we're explicit/have fewer things to know about outside of the test. You are correct, though

willfish: Oh, I wanted the change the response status to what we expect for the create, actually.

PR #380, opened 2020-04-17
Update swagger doc to explain need for move to create a nomis hearing

willfish:
## Proposed Changes

Add some clarity on description of move relationship and how it is required to create a Hearing in Nomis

### DIFF
--- a/spec/swagger/definitions/court_hearing.json
+++ b/spec/swagger/definitions/court_hearing.json
@@ -69,7 +69,7 @@
         "properties": {
           "move": {
             "$ref": "#/definitions/move_reference/MoveReference",
-            "description": "This is required for us to create a Court Hearing in nomis. Reference to the move entity that is associated with this court hearing."
+            "description": "Reference to the move entity that is associated with this court hearing. If it is present, Nomis will be updated with the new Court Hearing"
           }
         }
       }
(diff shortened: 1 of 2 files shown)


spec/swagger/definitions/court_hearing.json
```diff
@@ -69,7 +69,7 @@
         "properties": {
           "move": {
             "$ref": "#/definitions/move_reference/MoveReference",
-            "description": "Reference to the move entity that is associated with this court hearing."
+            "description": "This is required for us to create a Court Hearing in nomis. Reference to the move entity that is associated with this court hearing."
```
alexdesi, line 72: I am not sure about using the term "required",  this could make the users think that it is required, but actually you can call the endpoint without a move.
What about something like:

"Reference to the move entity that is associated with this court hearing. If it is present, Nomis will be updated with the new Court Hearing."

willfish, line 72:
```suggestion
            "description": "Reference to the move entity that is associated with this court hearing. If it is present, Nomis will be updated with the new Court Hearing"
```

swagger/v1/court_hearing.json
```diff
@@ -69,7 +69,7 @@
         "properties": {
           "move": {
             "$ref": "#/definitions/move_reference/MoveReference",
-            "description": "Reference to the move entity that is associated with this court hearing."
+            "description": "This is required for us to create a Court Hearing in nomis. Reference to the move entity that is associated with this court hearing."
```
willfish, line 72:
```suggestion
            "description": "Reference to the move entity that is associated with this court hearing. If it is present, Nomis will be updated with the new Court Hearing"
```

PR #386, opened 2020-04-20
P4-1435 Remove lots of duplicated swagger definitions and refactor specs

smoothcontract:
### Jira link

P4-1435

### What/Why?

This introduces a new approach to the parsing of JSON document fragments - previously these were loaded and parsed when needed for tests that work with JSON, via the `with_json_schema` annotation. Instead all of the JSON documents are parsed before running the spec suite, so they are loaded once regardless of which tests are then run. This may seem strange but has a number of benefits:

* Most of the tests we have are related to API behaviour somehow, so a large number of tests were already loading these definitions.
* Some tests should have been tagged with the `with_json_schema` context but weren't, which meant the tests only pass if some other test happened to include the context tag and was run first. Loading the schemas once for the whole suite removes the need to remember to include these tags.
* Many tests were using JSON fragments from `spec/swagger/definitions` rather than those in `swagger/v1` - meaning that definitions are duplicated (confusing and error prone). Now that the JSON fragments are preloaded we can use relative JSON references to the canonical swagger versions.
* No need to explicitly provide `definitions` key in swagger configuration - as we use relative JSON references everywhere now swagger can deduce these.

Also fixed a few minor issues with the swagger definitions -  `CourtHearing` was throwing some errors in the swagger UI (for `type`), and used the wrong syntax to define `MoveReference` association.

I've had to retain the `hand_coded_paths.json` file which is irritating; this is specifying quite a few endpoints that are not otherwise covered by rswag specs, so if these are omitted then lots of API endpoints don't get documented. We should work further to land on a single approach here - currently we have some specs under `spec/swagger/api/v1` (which are the only ones considered when running `rake rswag:specs:swaggerize`, some other request specs that are tagged with `rswag` (so do generate swagger docs, but only when explicitly running those tests) and other documentation that is manually expressed under `hand_coded_paths.json`.


swagger/v1/error_responses.json
```diff
@@ -1,15 +1,4 @@
...
-    "type": "object",
-    "required": [
-      "data"
-    ],
-    "properties": {
-      "data": {
-        "$ref": "#/definitions/document/Document"
-      }
-    }
-  },
```
smoothcontract, line 12: These were never used and are not error responses.

swagger/v1/court_hearing.json
```diff
@@ -10,7 +10,12 @@
         "example": "f0a91e16-1b0e-4e1f-93fe-319dda9933e6",
         "description": "The unique identifier (UUID) of this object"
       },
-      "type": { "type": "string", "enum": "court_hearings" },
+      "type": {
+        "type": "string",
+        "example": "court_hearings",
+        "enum": ["court_hearings"],
+        "description": "The type of this object - always `court_hearings`"
+      },
```
smoothcontract, line 18: Swagger-UI barfs when expanding the model node without the example and description.

spec/support/with_json_schema_context.rb
```diff
@@ -19,6 +19,8 @@ def load_json_schema(file_name)
   end
 end
 
-RSpec.configure do |rspec|
-  rspec.include_context 'with json schema', with_json_schema: true
+RSpec.configure do |config|
+  config.before(:suite) do
+    config.include_context 'with json schema'
+  end
 end
```
alexdesi, line 21: as discussed it looks like this only affect the 
`with json schema` (being inside the file with_json_schema_context.rb )

PR #387, opened 2020-04-20
Fix case_type enums

willfish:
### What?

I have added/removed/altered:

- [x] Fixes stringent enums - rely on nomis to define these

### Why?

I am doing this because:

- Duplicates source of truth for validation

### Have you? (optional)


spec/swagger/definitions/court_hearing.json
```diff
@@ -48,7 +61,6 @@
           "case_type": {
             "type": "string",
             "example": "Adult",
-            "enum": ["Adult"],
```
alexdesi, line 51: Is this the only change in this PR, apart the formatting?

willfish: Yeah, it's used to generate the swagger doc that @teneightfive is validating against.

PR #388, opened 2020-04-21
[P4-1264] Court case filters

willfish:
### Jira link

[P4-1264](https://dsdmoj.atlassian.net/browse/P4-1264)

### What?

- [x] Added filter for active court cases to court cases api

### Why?

- This is a required feature by the frontend in future and enables easily adding additional filters.

### DIFF
--- a/app/lib/nomis_client/court_cases.rb
+++ b/app/lib/nomis_client/court_cases.rb
@@ -4,8 +4,8 @@ module NomisClient
   class CourtCases < NomisClient::Base
     class << self
       DEFAULT_FILTER_PARAMS = {
-        active: 'true'
-      }
+        active: 'true',
+      }.freeze
 
       def get(booking_id, filter_params = DEFAULT_FILTER_PARAMS)
         query = filter_query(filter_params)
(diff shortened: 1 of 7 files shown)


app/lib/nomis_client/court_cases.rb
```diff
@@ -3,13 +3,21 @@
 module NomisClient
   class CourtCases < NomisClient::Base
     class << self
-      ACTIVE_CASES_FILTER = 'activeOnly=true'
+      def get(booking_id, filter_params)
```
alexdesi, line 6: Maybe you can add a default value to the filter, so you can avoid to specify one.

app/lib/nomis_client/court_cases.rb
```diff
@@ -3,13 +3,21 @@
...
 
         NomisClient::Base.get(court_cases_route).body
       end
+
+    private
+
+      def filter_query(filter_params)
+        return if filter_params.blank?
+
+        "?activeOnly=#{filter_params.fetch(:active, 'true')}"
```
alexdesi, line 19: it's good to have :active, 'true' as default value, 
anyway this behaviour is kind of hidden, it could be explicit if you put it as default params for this method.

willfish: We have both so if we pass filter params that don't include the active key this code will force it's default to be true.

PR #389, opened 2020-04-21
P4-1283 Add new reference endpoint for allocation complex cases

smoothcontract:
### Jira link

P4-1283

### What?

- [x] This adds a new model, controller and serializer to provide a list of complex cases.
- [x] Also includes a service to populate reference data as per front end designs.
- [x] Added API documentation.

### Why?

This will be used to populate the dynamic list of questions to ask when creating a new allocation (see Jira ticket screenshot).

### Have you?

### Deployment risks

- Adds a new endpoint `GET /api/v1/reference/complex_cases` but is not yet used
- New rake task `rake reference_data:create_complex_cases` will need to be run at some point after deployment on each environment before it can be used in the front end


app/controllers/api/v1/reference/complex_cases_controller.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    module Reference
+      class ComplexCasesController < ApiController
```
alexdesi, line 6: I am not sure about the name `ComplexCases` ,
I had to read the `Importer` to have get an idea.
Are there also `SimpleCases`?

Maybe it's just me, when I hear about cases for me it's about court cases.

app/services/complex_cases/importer.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module ComplexCases
+  class Importer
```
alexdesi, line 4: the name `importer`, make me think that we are importing data from an external source,
Actually we could store the "COMPLEX_CASES" in a yaml, as we did with `/data/supplier_locations.yml`

app/controllers/api/v1/reference/complex_cases_controller.rb
teneightfive: Yeah I think I agree with @alexdesi, outside of the context of an "allocation" a "complex case" is a little confusing. If this is an entity specifically tied to only allocations we could use a similar pattern to `reference/prison_transfer_reasons` - it started out as just `reasons` but on their own outside of the context of a prison transfer they weren't useful.

I think Alessandro is also right with the term "case" being associated more with courts whereas here is feels like these items are like complex criteria for selecting prisoners. From what I gathered from @Andy-Millar there are used to force conversation (or not) with receiving prisons.

app/services/complex_cases/importer.rb
```diff
@@ -0,0 +1,20 @@
...
+    COMPLEX_CASES = [
+      { key: 'segregated', title: 'Segregated prisoners' },
```
teneightfive, line 6: I wonder if we could use the same keys/terms we have used within the assessment sections to identify these scenarios. We worked a lot on those terms during early prototyping with our content designer and and to keep consistency in how we referred to scenarios across different organisations (Police/Prison/SCH/etc). This [spreadsheet](https://docs.google.com/spreadsheets/d/1f0JlJ6T1cL7fHxVQ2LOAokPacNPoFk_f35Pzucyc3UI/edit?ts=5e21635f) is a useful reference of all the mappings and terms we created across all the different systems.

For example I think we could swap these for some existing terms:
`segregated` -> `hold_separately`
`acct` -> `self_harm`
`mental` -> `mental_health_issue`
`drugs` -> `under_drug_treatment`

cc/ @Andy-Millar

Andy-Millar: Worth noting, the entire page is criteria for the allocation. Just to stress, we are not doing anything clever at the moment with filtering prisoners - just communicating to the OCAs what is the criteria for people to add to allocation. 

May be worth segmenting the criteria as follows:
- criteria-category
- criteria-sentence-length
- criteria-segregated
- criteria-self-harm AND criteria-acct [note: checkbox ticks both, but maybe later we want to seperate]
- criteria-mental-health
- criteria-integrated-drug-treatment-system
- fulfillment-status
- criteria-other

Hope this is helpful!

db/migrate/20200421105454_create_complex_cases.rb
```diff
@@ -0,0 +1,9 @@
+class CreateComplexCases < ActiveRecord::Migration[5.2]
+  def change
+    create_table :complex_cases, id: :uuid do |t|
+      t.string :key, null: false
```
martyn-w, line 4: might be good to add an index on `key`

app/services/complex_cases/importer.rb
smoothcontract: Thanks both, I'll amend the keys to match the suggestions used elsewhere for now.

app/controllers/api/v1/reference/complex_cases_controller.rb
smoothcontract: I've renamed throughout to `AllocationComplexCase`, lets go with that name for now as it does align with the UX and prescribed business terminology. We'll have to just ignore the passing similarity with other nouns including `case` :)

app/services/complex_cases/importer.rb
smoothcontract: This follows the same pattern used in other reference data endpoints e.g. https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/services/identifier_types/importer.rb

I don't see any advantage in adding another abstraction layer by moving the hash to a yaml file at this point. If the data changes in future we can revisit this and think about soft deletions and other complexities, but it's not needed yet.

db/migrate/20200421105454_create_complex_cases.rb
smoothcontract: There are only 4 rows, and I don't think I'll be joining against key in the immediate future so not sure it needs an index right now.

martyn-w: 👍

app/services/complex_cases/importer.rb
teneightfive: @smoothcontract after chatting with Andy again it does feel like there is a similarity between `sentence_length`, `prison_category` and `complex_cases`. And with the knowledge that these may change a lot in the future I wonder if it is worth thinking about mapping that relationship to this model? For example, having a `criteria` attribute or relationship that contains all the possible filtering criteria.

smoothcontract: I really want to avoid doing that now - it changes a fair bit of existing work and makes it challenging to deliver the stories for this sprint. I'd prefer to tackle that change at some point in the future when we revisit filtering, after getting a better idea of how the data is used.

app/services/complex_cases/importer.rb
alexdesi: I don't think it's good to have data hardcoded inside classes, but that's not an issue

smoothcontract: It would be hardcoded in a YAML file too, and would require a redeployment to update the contents regardless. I'm trying to keep this simple and then only add complexity later when the requirements dictate it (e.g. this data starts changing).

app/services/complex_cases/importer.rb
teneightfive: Happy to leave if for this sprint. I think we have a good understanding now that there is a concept of "criteria" on an allocation so when we look at filtering we can have a think about how we model these on the data layer.

PR #390, opened 2020-04-21
P4 1304 move api additions swagger

martyn-w:
### Jira link

[P4-1304](https://dsdmoj.atlassian.net/browse/P4-1304)

### What?

I have added/removed/altered:

- [X] Added new swagger definitions for:
* POST /moves/{move_id}/redirects
* POST /moves/{move_id}/lockouts
* POST /moves/{move_id}/complete

- [X] A little bit of tidying up and re-ordering of other swagger definitions

### Why?

I am doing this because:

- in preparation for implementing these endpoints, and as part of the Payment Mechanism epic P4-1175

### Have you? (optional)

- no code changes, minimal risk to services 👍
- could potentially break swagger documentation 👎


swagger/v1/location.json
```diff
@@ -57,5 +58,65 @@
         }
       }
     }
+  },
+  "Location2": {
```
smoothcontract, line 62: Why are there two Location types now?

swagger/v1/location_reference.json
```diff
@@ -1,5 +1,5 @@
 {
-  "LocationReference": {
+  "LocationReference1": {
```
smoothcontract, line 2: ..and two LocationReference types? 🤔

spec/swagger/definitions/hand_coded_paths.json
```diff
@@ -272,74 +272,110 @@
...
+            "application/vnd.api+json": {
+              "schema": {
+                "$ref": "put_move_responses.json#/422"
+              }
+            }
+          }
         }
+      }
+    },
+    "delete": {
```
smoothcontract, line 359: Can this endpoint and associated definition be deleted now?

swagger/v1/location.json
martyn-w: there are two locations (`Location1` and `Location2`, but only one type, `location`) - this is to make the examples a bit more realistic, so they can have a from_location and to_location which are different.

swagger/v1/location_reference.json
martyn-w: LocationReference1 refers to Location1, LocationReference2 refers to Location2 - to provide more realistic examples with distinct from_location and to_location

spec/swagger/definitions/hand_coded_paths.json
martyn-w: @smoothcontract yes, I think it probably should be - but would prefer to do that as separate PR for clarity

swagger/v1/location.json
smoothcontract: Right, got it - thanks for the explanation 👍

swagger/v1/location_reference.json
smoothcontract: Thankyou - makes sense 👍

swagger/v1/get_moves_responses.json
```diff
@@ -15,7 +15,8 @@
         "type": "array",
         "items": {
           "anyOf": [
-            { "$ref": "location.json#/Location" },
+            { "$ref": "location.json#/Location1" },
```
smoothcontract, line 18: Hmm... I still think this is confusing the `Location` type with an example location. This endpoint will always return`Location` types; and that type is defined in the schemas part of the swagger file.

smoothcontract:
It's no good, I'm still confused by the duplicate Location thing - there should one a single type definition in the swagger files, and an endpoint can return two different locations, but they would both be of type `Location` right?

PR #394, opened 2020-04-22
Simplify builds

willfish:
### Jira link

P4-<TODO>

### What?

This is our current deployment workflow:

image

This PR will remove the 3 build steps in the second phase of a deploy in the above image and replace them with 1.

- [ ]  Removes duplicated build steps that run the same code and produce the same output docker image
- [x] Removes redundant filter for a branch that is long dead

### Why?

I am doing this because:

- We want to kill unnecessary build/push to ecr steps
- Reduces complexity of our build process
- Reduce storage burden

- We need to validate that deploys still work as before. We can probably add a filter that runs deploys to dev/staging, etc, from this branch

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,14 +21,6 @@ aliases:
         only: /.*/
       branches:
         only: master
-  - &only_master_and_create_dev
-    filters:
-      tags:
-        only: /.*/
-      branches:
-        only:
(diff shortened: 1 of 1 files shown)


willfish:
This isn't quite finished and I don't have time right now to finish it. Closing and will reopen later

PR #396, opened 2020-04-22
Nomis Error handling

alexdesi:
Jira link: https://dsdmoj.atlassian.net/browse/P4-1261

- Fix error propagation from Nomis Api Calls
- Add Api:Json error in Get people#court_cases

Note: this PR is not yet handling the Timeout error, a new PR will follow soon

### What?

I have added/removed/altered:

- [x] Fix Nomis error propagation
- [x] Support Api:Json error
- [x] Add Nomis timeout error

### Why?
To give more info to the caller about Nomis failures.
To have a standard way to handle errors (Api:Json)

### Have you? (optional)

### DIFF
--- a/app/controllers/api/v1/people_controller.rb
+++ b/app/controllers/api/v1/people_controller.rb
@@ -38,12 +38,20 @@ module Api
       def court_cases
         person = Person.find(params[:person_id])
 
-        court_cases = People::RetrieveCourtCases.call(person, court_case_filter_params)
+        response = People::RetrieveCourtCases.call(person, court_case_filter_params)
 
-        render json: court_cases, each_serializer: CourtCaseSerializer, include: :location
+        if response.success?
+          render json: response.court_cases, each_serializer: CourtCaseSerializer, include: :location
+        else
+          render json: json_api_errors_for(response.error)
+        end
       end
 
     private
+
+      def json_api_errors_for(error)
+        { errors: [error.json_api_error] }
+      end
 
       def import_person_from_nomis prison_number
         # This prevents us from blaming the current user/application for the NOMIS sync
--- a/app/controllers/api_controller.rb
+++ b/app/controllers/api_controller.rb
@@ -13,7 +13,7 @@ class ApiController < ApplicationController
   rescue_from ActiveRecord::RecordInvalid, with: :render_unprocessable_entity_error
   rescue_from ActiveRecord::ReadOnlyRecord, with: :render_resource_readonly_error
   rescue_from CanCan::AccessDenied, with: :render_unauthorized_error
-  rescue_from Faraday::ConnectionFailed, Faraday::TimeoutError, with: :connection_error
+  rescue_from Faraday::ConnectionFailed, Faraday::TimeoutError, with: :render_connection_error
 
   def current_user
     doorkeeper_token&.application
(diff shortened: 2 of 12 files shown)


app/controllers/api/v1/people_controller.rb
```diff
@@ -38,13 +38,21 @@ def image
       def court_cases
         person = Person.find(params[:person_id])
 
-        court_cases = People::RetrieveCourtCases.call(person)
+        response = People::RetrieveCourtCases.call(person)
```
willfish, line 41: 🎉

willfish:
I think this PR covers handled http exceptions in the Nomis integration but misses known unhandled ones:

  1. Nomis api responds with 4xx # `OAuth2::Error`
  2. Nomis api responds with 5xx # `OAuth2::Error`
  3. Connection Issue / Gateway  # `Faraday::TimeoutError`, `Faraday::ConnectionError`

We could treat `Faraday::xxx` errors as Nomis errors in the same way you treat the handled errors currently?

I reckon we could rescue these globally here https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/46b3e229e96ead472cda2fe037dd7c161055d5c0/app/controllers/api_controller.rb#L11-L15

app/controllers/api_controller.rb
```diff
@@ -13,6 +13,7 @@ class ApiController < ApplicationController
   rescue_from ActiveRecord::RecordInvalid, with: :render_unprocessable_entity_error
   rescue_from ActiveRecord::ReadOnlyRecord, with: :render_resource_readonly_error
   rescue_from CanCan::AccessDenied, with: :render_unauthorized_error
+  rescue_from Faraday::ConnectionFailed, Faraday::TimeoutError, with: :connection_error
```
willfish, line 16:
```suggestion
  rescue_from Faraday::ConnectionFailed, Faraday::TimeoutError, with: :render_connection_error
```

app/controllers/api_controller.rb
```diff
@@ -108,6 +109,16 @@ def render_unauthorized_error(exception)
     )
   end
 
+  def connection_error(exception)
```
willfish, line 112:
```suggestion
  def render_connection_error(exception)
```

app/controllers/api_controller.rb
```diff
@@ -108,6 +109,16 @@ def render_unauthorized_error(exception)
...
+    render(
+      json: { errors: [{
+        title: 'Connection Error',
```
willfish, line 115: Should this be a Nomis connection Error?

spec/requests/api/v1/people_controller_court_cases_spec.rb
```diff
@@ -9,36 +9,58 @@
...
+        end
+      end
     end
 
-    context 'when we pass a filter in the query params' do
-      let(:query) { '?filter[active]=true' }
+    context 'when the court cases are NOT present in Nomis' do
+      let(:booking_id) { '123456789' }
+
+      let(:court_cases_from_nomis) { OpenStruct.new(success?: false, error: NomisClient::ApiError.new(status: 404, error_body: '{}')) }
```
willfish, line 56: 👍

willfish:
Oh, do we need to extend the documentation, actually?

app/controllers/api_controller.rb
alexdesi: good point, the reality is that we don't know, since we could use faraday to connect to different services ... 
to fix this I'm passing a message when I raise the exception, this is the response:
 
`{"errors"=>[{"title"=>"Connection Error", "detail"=>"Faraday::TimeoutError: Nomis Connection Error"}]}`

that specifies `Nomis`

PR #398, opened 2020-04-23
[P4-1095] Timetable feature

willfish:
### Jira link

P4-1095

### What?

I have added/removed/altered:

- [x] Adds ability to retrieve activities/scheduled events for a date range
- [x] Adds ability to retrieve hearings for a date range
- [x] Adds RetrieveTimetable service and spec to aggregate activities and hearings
- [x] Adds controller, route and request spec for retrieving timetable entries from Nomis
- [ ] Adds swagger documentation for new endpoint

### Why?

- This is part of a conflict resolution feature that enables booking moves and avoiding conflicts with timetable entries that exist with Nomis.

### Have you? (optional)


app/lib/nomis_client/activities.rb
```diff
@@ -0,0 +1,23 @@
...
+        activities_path = "/bookings/#{booking_id}/activities?fromDate=#{start_date.iso8601}&toDate=#{end_date.iso8601}"
+
+        activities = []
+
+        paginate_through(activities_path) do |activities_response|
+          activities_json = activities_response
+
+          activities_json .each do |activity_json|
+            activities << activity_json
+          end
```
alexdesi, line 14: I have the feeling this can be simplified,
is all of that just equivalent to 
`activities.join(activities_response)`
?

app/lib/nomis_client/base.rb
```diff
@@ -69,6 +79,41 @@ def update_json_headers(params)
           },
         }.deep_merge(params)
       end
+
+      def paginate_through(path, params = {}, verb = :get)
```
alexdesi, line 83: maybe we want to keep the same signature as 
`NomisClient::Base.send(verb, path, params) `

app/lib/nomis_client/base.rb
```diff
@@ -69,6 +79,41 @@ def update_json_headers(params)
...
+
+          if response.status == 200
+            collection_items = yield response.parsed
+
+            found_records += collection_items.count
+          end
+        end
+
+        responses
+      end
```
alexdesi, line 116: it looks a bit complex, it would be nice to simplify it, maybe extracting part of the functionality in another private method

app/lib/nomis_client/court_hearings.rb
```diff
@@ -1,17 +1,35 @@
...
+        court_hearings_path = "/bookings/#{booking_id}/court-hearings?fromDate=#{start_date.iso8601}&toDate=#{end_date.iso8601}"
+
+        court_hearings = []
+
+        paginate_through(court_hearings_path) do |court_hearings_response|
+          hearings_json = court_hearings_response['hearings']
+
+          hearings_json.each do |hearing_json|
+            court_hearings << hearing_json
+          end
```
alexdesi, line 16: is this not just a `map`?

app/services/people/retrieve_activities.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module People
+  class RetrieveActivities
+    def self.call(person)
+      activities = NomisClient::Activities.get(person.latest_nomis_booking_id)
+
+      activities.map do |activity_json|
```
alexdesi, line 8: not sure if the suffix `_json` add any info, it sounds that `activity_json` is a json string.

Maybe we could call it `nomis_activity`, or just `activity`

app/services/people/retrieve_court_hearings.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module People
+  class RetrieveCourtHearings
+    def self.call(person)
+      court_hearings = NomisClient::CourtHearings.get(person.latest_nomis_booking_id)
+
+      court_hearings.map do |court_hearing_json|
```
alexdesi, line 8: same thougth as before

spec/lib/nomis_client/court_hearings_spec.rb
```diff
@@ -2,7 +2,102 @@
...
+    it 'calls the NomisClient::Base.get with the correct path and params' do
+      court_hearings.get(booking_id, start_date, end_date)
+
+      expect(token).to have_received(:get).with(
+        "/elite2api/api/bookings/1495077/court-hearings?fromDate=#{start_date.iso8601}&toDate=#{end_date.iso8601}",
+        headers: { 'Page-Limit' => '20', 'Page-Offset' => '0' },
+      )
+    end
+
+    context 'when there are more than one page of court_hearings to retrieve' do
```
alexdesi, line 68: it sounds we are testing the logic for the pagination in various classes,
maybe we should extract that logic and test it separately?

spec/services/people/retrieve_activities_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    allow(NomisClient::Activities).to receive(:get).and_return(response_json)
+  end
+
+  it 'returns an array of CourtCase' do
+    response = described_class.call(person)
+
+    expect(response).to all(be_a(Activity))
+  end
+
+  it 'calls NomisClient::Activities with the correct args' do
```
alexdesi, line 19: you might want to be more specific  `.... with the correct booking_id`

app/services/people/retrieve_activities.rb
willfish: 🤷

willfish: activity makes me think of the model. nomis activity might work better

willfish, line 8:
```suggestion
      activities.map do |nomis_activity|
```

app/services/people/retrieve_activities.rb
```diff
@@ -0,0 +1,13 @@
...
+        Activity.new.build_from_nomis(activity_json)
```
willfish, line 9:
```suggestion
        Activity.new.build_from_nomis(nomis_activity)
```

app/services/people/retrieve_court_hearings.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module People
+  class RetrieveCourtHearings
+    def self.call(person)
+      court_hearings = NomisClient::CourtHearings.get(person.latest_nomis_booking_id)['hearings']
+
+      court_hearings.map do |court_hearing_json|
```
willfish, line 8:
```suggestion
      court_hearings.map do |nomis_court_hearing|
```

app/services/people/retrieve_court_hearings.rb
```diff
@@ -0,0 +1,13 @@
...
+        NomisCourtHearing.new.build_from_nomis(court_hearing_json)
```
willfish, line 9:
```suggestion
        NomisCourtHearing.new.build_from_nomis(nomis_court_hearing)
```

app/controllers/api/v1/people_controller.rb
```diff
@@ -47,6 +43,12 @@ def court_cases
         end
       end
 
+      def timetable
+        timetable = People::RetrieveTimetable.call(person)
```
martyn-w, line 47: this is a bit inconsistent: `People::RetrieveCourtCases.call()` returns a `response` object with `court_cases` as a property; whereas `People::RetrieveTimetable.call()` returns the actual timetables. 

Probably best to update `People::RetrieveTimetable.call()` to return a response object like `People::RetrieveCourtCases.call()`, then it can better handle Nomis errors

willfish: Yeah, I actually needed that to handle errors - see subsequent commits :)

app/models/activity.rb
```diff
@@ -0,0 +1,22 @@
+class Activity
+  include ActiveModel::Serialization
+
+  TYPE = 'Prison Activities'.freeze
+
+  attr_reader :id, :start_time, :reason, :type, :agency_id
+
+  def build_from_nomis(activity)
+    @id = activity['eventId']
+    # Nomis startTime does not include a zone so we shouldn't use our local zone with Time.zone.parse
```
martyn-w, line 10: @willfish can you clarify in what timezone Nomis times are provided? If it is local time (BST / GMT) **and the server runs in the same timezone**, then Time.zone.parse should be fine. Otherwise, if Nomis is in UTC (cf. `app/services/court_hearings/create_in_nomis.rb`) then I think we should be forcing the timezone in the parse.

E.g. consider the difference:

```
Time.zone.parse('19:00')
=> Mon, 27 Apr 2020 19:00:00 BST +01:00
Time.zone.parse('19:00 UTC')
=> Mon, 27 Apr 2020 20:00:00 BST +01:00
```

app/services/people/retrieve_timetable.rb
```diff
@@ -0,0 +1,18 @@
...
+  class RetrieveTimetable
+    def self.call(person)
+      activities_response = People::RetrieveActivities.call(person)
+      return activities_response unless activities_response.success?
+
+      court_hearings_response = People::RetrieveCourtHearings.call(person)
+      return court_hearings_response unless court_hearings_response.success?
+
+      timetable = court_hearings_response.content + activities_response.content
+      timetable.sort_by(&:start_time).reverse
```
martyn-w, line 13: this sort doesn't seem to be saved ?

willfish, line 13:
```suggestion
      timetable = timetable.sort_by(&:start_time).reverse
```

PR #401, opened 2020-04-24
P4-1286 get allocation

smoothcontract:
### Jira link

P4-1286

### What?

- [x] Refactored existing implementation for profile assessment answers and profile identifiers
- [x] Added new classes for allocation complex case answers
- [x] New endpoint to `GET /allocations/{allocation_id}`

### Why?

- The previous JSONB attribute classes were repetitive and not easy to extend for other attributes
- The new endpoint supports front end work to display an existing allocation

### Have you? (optional)

N/A


martyn-w:
@smoothcontract looks good 👍 but could you add rspecs for classes `BaseCollection` and for `JSONB`?

PR #402, opened 2020-04-27
DO NOT MERGE!!! Quick demonstration PR to bypass API authentication for local dev

smoothcontract:
This likely breaks things related to the correct locations returned for a given user.

### DIFF
--- a/app/controllers/api_controller.rb
+++ b/app/controllers/api_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class ApiController < ApplicationController
-  before_action :doorkeeper_authorize!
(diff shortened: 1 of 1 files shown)


app/controllers/api_controller.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class ApiController < ApplicationController
-  before_action :doorkeeper_authorize!
+  before_action :doorkeeper_authorize! unless Rails.env.development?
```
martyn-w, line 4: it would be handy to feature flag this (e.g. with an env var?), so in development we can turn on/off authorisation in development

PR #404, opened 2020-04-28
P4 1365 basic auth on webhooks

martyn-w:
### Jira link

[P4-1365](https://dsdmoj.atlassian.net/browse/P4-1365)

### What?

I have added/removed/altered:

- Added support for basic auth on webhooks

### Why?

I am doing this because:

- suppliers have requested it

### Have you? (optional)

- [X] Updated webhook docs


app/jobs/notify_webhook_job.rb
```diff
@@ -14,9 +14,13 @@ def perform(notification_id:)
     begin
       data = ActiveModelSerializers::Adapter.create(NotificationSerializer.new(notification)).to_json
       hmac = Encryptor.hmac(notification.subscription.secret, data)
```
smoothcontract, line 16: Maybe nice to assign `subscription = notification.subscription` here (or actually the line above) to simplify access throughout the method

app/jobs/notify_webhook_job.rb
```diff
@@ -29,10 +33,4 @@ def perform(notification_id:)
       raise e # re-raise the error to force the notification to be retried by sidekiq later
     end
   end
-
-private
-
-  def client
```
smoothcontract, line 35: I think I prefer this as a separate method TBH - makes the main part of the code easier to read.

smoothcontract: You could perhaps pass `subscription` as an argument to this method and move the basic auth header assignment stuff (lines 18-20) to here:

```ruby
def client(subscription)
  Faraday.new(headers: { 'Content-Type': 'application/vnd.api+json', 'User-Agent': 'pecs-webhooks/v1' }).tap do |client|
    if subscription.username.present? && subscription.password.present?
      client.headers['Authorization'] = "Basic #{Base64.strict_encode64(subscription.username + ':' + subscription.password)}"
    end
  end
end
```

app/services/encryptor.rb
```diff
@@ -20,7 +20,7 @@ def self.encrypt(value)
   end
 
   def self.decrypt(value)
-    new.decrypt_and_verify(value)
+    new.decrypt_and_verify(value) if value.present?
```
smoothcontract, line 23: Does this change have coverage?

app/jobs/notify_webhook_job.rb
martyn-w: 👍 yep that cleans it up a little, good suggestion

app/services/encryptor.rb
martyn-w: it does now :-)

app/jobs/notify_webhook_job.rb
martyn-w: Ok - i've done this minus the `tap`

PR #407, opened 2020-04-29
CourtHearings Post service: change error level + clean up

alexdesi:
### What?

Change the sentry error level and remove outdated comments


martyn-w:
👍 LGTM

PR #408, opened 2020-04-29
[P4-1413] Adds swagger documentation for new timetable api

willfish:
### Jira link

P4-1413

### What?

- [x] Added support for filter dates on the timetable api
- [x] Added swagger documentation for the new timetable api

### Why?

- Filters are required for flexibility by the frontend
- Swagger docs are needed to communicate endpoint semantics to our suppliers and frontend clients


spec/requests/api/v1/people_controller_timetable_spec.rb
```diff
@@ -91,11 +107,25 @@
...
       end
     end
+
+    context 'when filter params are missing' do
+      let(:nomis_activities) { [] }
+      let(:nomis_court_hearings) { [] }
+
+      let(:params) { { access_token: token.token } }
+
+      it 'return an empty data key' do
```
willfish, line 123:
```suggestion
      it 'returns an error' do
```

swagger/v1/timetable_entry.yaml
```diff
@@ -0,0 +1,42 @@
...
+    attributes:
+      type: object
+      required:
+        - start_time
+        - nomis_type
+        - reason
+      properties:
+        start_time:
+          type: string
+          example: "2020-04-22T08:30:00"
```
martyn-w, line 28: might be sensible to force a timezone on time parameters ? (though I know there are timezone problems further downstream in Nomis)

app/controllers/api/v1/people_controller.rb
```diff
@@ -44,7 +44,10 @@ def court_cases
       end
 
       def timetable
-        response = People::RetrieveTimetable.call(person)
+        start_date = Date.parse(timetable_filter_params[:date_from])
```
martyn-w, line 47: I'd suggest enforcing strict date validation before parsing - just in case some american software comes along and asks for `2/1/2020` (meaning 1st February, rather than 2nd January)

swagger/v1/timetable_entry.yaml
willfish: Oh good spot. So this is actually just an iso8601 time stamp so I'll update this. We infer it as London so will make it match that

app/controllers/api/v1/people_controller.rb
willfish: 🤮

willfish: Sounds good

swagger/v1/timetable_entry.yaml
willfish, line 28:
```suggestion
          example: "2020-04-30T13:01:57+01:00"
```

swagger/v1/timetable_entry.yaml
```diff
@@ -0,0 +1,42 @@
...
+          description: Time when the scheduled event is set to start
```
willfish, line 29:
```suggestion
          description: Time when the scheduled event is set to start
          format: date-time
```

app/controllers/api/v1/people_controller.rb
willfish: Added that now :+1:

PR #412, opened 2020-04-29
P4 1355 temporary redirect endpoint

martyn-w:
### Jira link

[P4-1355](https://dsdmoj.atlassian.net/browse/P4-1355)

### What?

I have added/removed/altered:

- Temporary implementation of the `/moves/:move_id/events` endpoint to allow a redirect, in order to update the `to_location` of an existing move.

### Why?

I am doing this because:

- The frontend needs it asap

### Have you? (optional)

- not dangerous, as this is not a currently used endpoint


config/routes.rb
```diff
@@ -24,6 +24,7 @@
       end
       resources :moves, only: %i[index show create destroy update] do
         resources :documents, only: %i[create destroy]
+        resources :events, only: %i[create], controller: 'move_events'
```
willfish, line 27: Are we using `controller: move_events` to avoid nesting the controller in a `Move` module?

spec/requests/api/v1/move_events_controller_create_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+RSpec.describe Api::V1::MoveEventsController do
+  include ActiveJob::TestHelper
+
+  let(:response_json) { JSON.parse(response.body) }
+
+  describe 'POST /moves/:move_id/events' do
+    let(:schema) { load_yaml_schema('post_move_events_responses.yaml') }
+
+    let(:supplier) { create(:supplier) }
+    let!(:application) { create(:application, owner_id: supplier.id) }
```
willfish, line 14:
```suggestion
    let(:application) { create(:application, owner_id: supplier.id) }
```

config/routes.rb
martyn-w: hmm... good question; don't mind either way - which is more railsy ?

spec/requests/api/v1/move_events_controller_create_spec.rb
```diff
@@ -0,0 +1,79 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Api::V1::MoveEventsController do
+  include ActiveJob::TestHelper
```
willfish, line 6: Is this necessary?

config/routes.rb
willfish: I think not overriding the controller (principle of convention over configuration) makes more sense.

willfish: But we're also doing it loads elsewhere in this file

spec/requests/api/v1/move_events_controller_create_spec.rb
martyn-w: it isn't, now removed - too much copy pasta

config/routes.rb
martyn-w: so I've experimented moving `MoveEventsController` to `EventsController` in the `Moves` module... however, this breaks a lot of other stuff reliant on the `Moves` module/namespace (`Moves::ParamsValidator`, `Moves::Finder`, Moves::ImportPeople` etc). 

I guess there is something wrong with the implementation of the moves namespace; but I think I'll leave that as a separate issue to resolve later and keep with MoveEventsController for now.

PR #415, opened 2020-04-30
[P4-1436] Fix timestamp coercion in court hearings POST

willfish:
### Jira link

P4-1436

### What?

We assumed that the Elite2 api supports utc time stamps and coerces them to whatever way it needs when it exploded in our integration to create Court Hearings.

The actual way it behaves is not to support time zones at all and expect everything to be received in a London time zone (i.e. GMT or BST depending on time of year).

This PR adds a standard Time format for Nomis and stops us formatting to UTC on creation of a court hearing in Nomis.

### Why?

To avoid breaking our integration with Nomis.


config/initializers/date_formats.rb
```diff
@@ -1 +1,2 @@
 Date::DATE_FORMATS[:nomis] = '%Y-%m-%dT%H:%M:%S'
+Time::DATE_FORMATS[:nomis] = '%Y-%m-%dT%H:%M:%S'
```
martyn-w, line 2: could you add a comment so we won't forget Nomis format is local London time:
`# NB: Nomis is in a London time zone (i.e. GMT or BST depending on time of year)`

willfish: Sure!

config/initializers/date_formats.rb
```diff
@@ -1 +1,2 @@
 Date::DATE_FORMATS[:nomis] = '%Y-%m-%dT%H:%M:%S'
```
willfish, line 1:
```suggestion
# NB: Nomis is in a London time zone (i.e. GMT or BST depending on time of year)
Date::DATE_FORMATS[:nomis] = '%Y-%m-%dT%H:%M:%S'
```

config/initializers/date_formats.rb
alexdesi, line 3: it looks a bit confusing having :nomis for date and time

willfish: I'm not sure of a better name, though? Happy to take suggestions 🙂

alexdesi: maybe `nomis_time` 
`nomis_full_time`, `nomis_extended_time`,  ... :)

martyn-w: Hi @alexdesi I think its OK as is:
```
Date.today.to_s(:nomis)
Time.zone.to_s(:nomis)
```

willfish: Going with as is as this is currently a bug in production.

alexdesi: ok, np, one point for me was it would be simpler to search in your editor for the time key :nomis would not gonna help you

PR #418, opened 2020-05-01
[P4-1422] Change to defaulting to creating a court hearing in Nomis

willfish:
### Jira link

P4-1422

### What?

Update default behaviour when creating a hearing in Nomis

### Why?

The original implementation breaks the frontend flow.

### Have you? (optional)

### DIFF
--- a/app/controllers/api/v1/court_hearings_controller.rb
+++ b/app/controllers/api/v1/court_hearings_controller.rb
@@ -41,7 +41,9 @@ module Api
       end
 
       def should_save_in_nomis?
-        move && params['should_save_in_nomis'] == 'true'
+        do_not_save_to_nomis = params['do_not_save_to_nomis'] == 'true'
+
+        move && !do_not_save_to_nomis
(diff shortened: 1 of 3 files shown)


app/controllers/api/v1/court_hearings_controller.rb
```diff
@@ -41,7 +41,9 @@ def move
       end
 
       def should_save_in_nomis?
-        move && params['should_save_in_nomis'] == 'true'
+        do_not_save_to_nomis = params['do_not_save_to_nomis'] == 'true'
```
teneightfive, line 44: Will this also support just passing the query param without a value?

It would be nice to have `/court_hearings?do_not_save_to_nomis` and `/court_hearings?do_not_save_to_nomis=true` behave the same.

No biggie if not

willfish: Only 'true' will change the behaviour. Everything else is noop

PR #423, opened 2020-05-01
P4 1357 patch people notifications

martyn-w:
### Jira link

[P4-1357](https://dsdmoj.atlassian.net/browse/P4-1357)

### What?

I have added/removed/altered:

- [X] triggers notifications (webhooks/emails) on patching to the /people endpoint
- [X] updated notification rules to prevent emails for back-dated moves (but allow webhooks for these)

### Why?

I am doing this because:

- the "update a move" workflow in the frontend actually patches the profile (/people) and currently doesn't trigger notifications

- Will send duplicate emails; this is a known and accepted limitation

### DIFF
--- a/app/controllers/api/v1/move_events_controller.rb
+++ b/app/controllers/api/v1/move_events_controller.rb
@@ -66,7 +66,7 @@ module Api
       end
 
       def move
-        @move ||= Move.accessible_by(current_ability).find_by(id: params.require(:move_id))
+        @move ||= Move.accessible_by(current_ability).find(params.require(:move_id))
       end
 
       def fake_event_object
--- a/app/controllers/api/v1/move_events_controller.rb
+++ b/app/controllers/api/v1/move_events_controller.rb
@@ -18,7 +18,13 @@ module Api
 
         case event_name
         when 'redirect'
-          move.update(to_location: to_location)
+          # NB: rather than update immediately, we need to detect whether the location has actually changed (or not) to
+          # prevent triggering duplicate webhook/email notifications
+          move.to_location = to_location
+          if move.to_location_id_changed?
+            move.save!
+            Notifier.prepare_notifications(topic: move, action_name: 'update')
+          end
           render json: fake_event_object, status: :created
         else
           render status: :bad_request, json: {
--- a/app/jobs/prepare_move_notifications_job.rb
+++ b/app/jobs/prepare_move_notifications_job.rb
@@ -40,14 +40,10 @@ private
   end
 
   def should_email?(move)
-    # NB: only email for current moves (not back-dated ones):
-    #   * move.status must be Requested or Cancelled (not Proposed), AND
-    #   * move.date is not in the past OR move.to_date is not in the past
-
-    [Move::MOVE_STATUS_REQUESTED, Move::MOVE_STATUS_CANCELLED].include?(move.status) &&
-      ((move.date.present? && move.date >= Time.zone.today) ||
-          (move.date.nil? && move.date_to.present? && move.date_to >= Time.zone.today)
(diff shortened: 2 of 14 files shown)


app/controllers/api/v1/move_events_controller.rb
```diff
@@ -60,7 +66,7 @@ def to_location
       end
 
       def move
-        @move ||= Move.accessible_by(current_ability).find_by(id: params.require(:move_id))
+        @move ||= Move.accessible_by(current_ability).find(params.require(:move_id))
```
willfish, line 69: 🎉

app/jobs/prepare_move_notifications_job.rb
```diff
@@ -38,9 +39,15 @@ def build_notification_id(subscription, type_id, topic, action_name)
...
 
-  def should_email?(move, action_name)
-    # NB: only notify by email on move requested and move cancellation
-    [Move::MOVE_STATUS_REQUESTED, Move::MOVE_STATUS_CANCELLED].include?(move.status) && %w(create update_status).include?(action_name)
+  def should_email?(move)
+    # NB: only email for current moves (not back-dated ones):
+    #   * move.status must be Requested or Cancelled (not Proposed), AND
+    #   * move.date is not in the past OR move.to_date is not in the past
+
+    [Move::MOVE_STATUS_REQUESTED, Move::MOVE_STATUS_CANCELLED].include?(move.status) &&
```
willfish, line 47: Would it be cleaner to move this to question to the Move model?

```ruby
move.current?
```

spec/jobs/prepare_person_notifications_job_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    describe 'it calls PrepareMoveNotificationsJob for the related moves' do
+      it { expect(PrepareMoveNotificationsJob).to have_received(:perform_now).with(topic_id: move1.id, action_name: action_name) }
+      it { expect(PrepareMoveNotificationsJob).to have_received(:perform_now).with(topic_id: move2.id, action_name: action_name) }
+      it { expect(PrepareMoveNotificationsJob).not_to have_received(:perform_now).with(topic_id: other_move.id, action_name: action_name) }
+    end
+  end
+
+  context 'when creating a person' do
+    let(:action_name) { 'create' }
```
willfish, line 27: What do you think of removing this context block? The action_name doesn't actually affect how the service behaves? It just gets passed through to the other service?

spec/requests/api/v1/move_events_controller_create_spec.rb
```diff
@@ -57,6 +65,13 @@
       it_behaves_like 'an endpoint that responds with error 401'
     end
 
+    context 'with a missing move_id' do
```
willfish, line 68: 👍

app/jobs/prepare_move_notifications_job.rb
willfish: It would be easier to unit test

spec/jobs/prepare_person_notifications_job_spec.rb
martyn-w: hmm, interesting question: so higher up (in the controller) there is a decision that creating a person should not email, but updating a person should email. I guess I wanted to capture in this test that the `PreparePersonNotificationsJob` behaves the same for both creates and updates.I think its important to test both contexts, but I'll refactor it a bit. 👍

app/jobs/prepare_move_notifications_job.rb
martyn-w: yes, good point 👍

PR #424, opened 2020-05-05
[P4-1444] Handle unparseable errors from Azure API Gateway

willfish:
### Jira link

P4-1444

### What?

When the Nomis app isn't routeable from the Azure API Gateway the API Gateway returns a html error and 502 status code. The elite2api is a json api so we hadn't previously added handling for non-json mime types.

This adds a special condition in the html case and returns a cleaner error message to the frontend/supplier.

- [x] Adds handling for non-json return body on error in Azure API Gateway

### Why?

This is quicker/simpler than getting an infrastructure change to update the default error templates.


willfish:
For more context https://sentry.service.dsd.io/mojds/bookasecuremove-staging/issues/48971/?referrer=slack

PR #426, opened 2020-05-05
P4-1376 Create moves when creating an allocation

cwrw:
### Jira link

[P4-1376](https://dsdmoj.atlassian.net/browse/P4-1376)

### What?

- [x] Associated Allocations to Moves, with relevant migrations 
- [x] Added move creation to the allocation endpoint. A move is created for each of the move count set in an allocation. Leaving move count for now on an allocation until we flush out if this is to be dynamically calculated or not. All moves are in a requested state, and inherit the date and locations from the Allocation.
- [x] Added a notification for each move created
- [x] Avoided validating a move before validating an allocation

### Why?

We would like to create an allocation with multiple empty moves, which allows an OCA to later add people to. Any amendments to serialisation for the create and get endpoints is being pushed till later tickets to flush out what we want to see when we return moves in an allocation

- Includes migratory actions for the database
- Sends notifications for creation of moves under an allocation


app/controllers/api/v1/allocations_controller.rb
```diff
@@ -65,6 +71,24 @@ def render_allocation(allocation, status)
...
+
+      def from_location
+        Location.find(allocation_params.dig(:relationships, :from_location, :data, :id))
+      end
+
+      def to_location
+        Location.find(allocation_params.dig(:relationships, :to_location, :data, :id))
+      end
+
+      def moves
```
cwrw, line 83: a bit on the fence around further refactoring this/moving it to a service, as felt that might be a bit overkill. Thoughts?

app/controllers/api/v1/allocations_controller.rb
```diff
@@ -15,8 +15,14 @@ def index
 
       def create
         allocation = Allocation.new(allocation_attributes)
+        allocation.moves = moves if allocation.valid?
+
         allocation.save!
 
+        allocation.moves.each do |move|
```
cwrw, line 22: amend to not send notifications unless allocation/moves are saved and valid

app/controllers/api/v1/allocations_controller.rb
smoothcontract: I think it would make sense if the majority of the create method was also moved into the service - it would be passed allocation_attributes and return an allocation instance, handling all the details of creating associated moves and firing the notifications. I think all of that responsibility belongs elsewhere than the controller - which perhaps it better added to one of the models, or given the conventions with this codebase a simple create allocation service. Resolving location Id's to locations could be done there too removing the need for lines 75-81.

That's the main change I'd suggest with this PR - it might make the service easier to test in isolation rather than having to go via a complete request/response.

cwrw: That's fair, I do agree with moving this out of the controller. My next question is around where this should live, my first thoughts would be to move it into the model, but the established pattern for services (as you've mentioned) in the app eg.`People` live in services under the class `People::Crud` and `People::Creator`, `People::Updater` etc. Are we happy to reuse this pattern for other models like `Allocation`, or move away from it, as Im keen on having this in a model instead. Maybe @martyn-w can weigh in regarding the services vs model approach and thoughts around the pattern and conventions we want to establish going forward?

app/controllers/api/v1/allocations_controller.rb
alexdesi: I was thinking that might be worth to move these calls to `Notifier` to a `after_action`, since are not the main goal of the actions.

app/controllers/api/v1/allocations_controller.rb
```diff
@@ -15,8 +15,14 @@ def index
 
       def create
         allocation = Allocation.new(allocation_attributes)
+        allocation.moves = moves if allocation.valid?
```
willfish, line 18: Could we fully validate and handle the propagated error in the controller with rich validation errors?

willfish: https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/controllers/api_controller.rb#L17

db/migrate/20200504081233_add_allocation_to_moves.rb
```diff
@@ -0,0 +1,5 @@
+class AddAllocationToMoves < ActiveRecord::Migration[5.2]
+  def change
+    add_column :moves, :allocation_id, :uuid, null: true
```
martyn-w, line 3: please add a FK and index

app/controllers/api/v1/allocations_controller.rb
cwrw: This was initially added to avoid validating on a `move` before actually validating the `allocation` itself. This was happening when we were missing a date, and the `move` validation was taking precedence in the error messages. I have now changed this to only look for the `date` attribute instead of performing the validation twice as that was confusing.

PR #427, opened 2020-05-05
Data migration to update profile ID with the correct profile.

alexdesi:
### Jira link

https://dsdmoj.atlassian.net/browse/P4-1449

### What?

Update the profile_in the moves, with the the profile that was most recently updated.

### Why?
This is the first step to change the relationship Move-Person to Move-Profile

run the rake task
`bundle exec rake moves:set_profiles`

### DIFF
--- a/lib/tasks/move_profile.rake
+++ b/lib/tasks/move_profile.rake
@@ -1,8 +1,24 @@
-namespace :move_profile do
-  desc 'backfill move profile ids'
-  task backfill: :environment do
-    Move.includes(person: :profiles).find_each.reject { |m| m.profile_id.present? }.each do |move|
-      move.update!(profile_id: move.person.latest_profile.id)
+namespace :moves do
+  desc 'Set all profile ids in all the moves'
+  task set_profiles: :environment do
+    moves = Move.where(profile_id: nil).includes(person: :profiles)
+
+    total = moves.count
+
+    if moves.empty?
+      puts 'All profiles IDs were already updated.'
+      return
     end
+
+    moves.find_each(batch_size: 200).with_index do |move, n|
+      puts "#{n}/#{total} moves processed ..." if (n % 200).zero? # Show progression
+      profile = move.person.profiles.order(:updated_at).last # Take the profile that was most recently updated
+
+      # update only profile and skip validations: some moves are invalid because of uniqueness of 'date', but that does
+      # not impact the correctness of this data migration.
+      move.update_attribute(:profile_id, profile.id)
+    end
+
+    puts "#{moves.count} profile IDs have been successfully updated."
   end
 end


lib/tasks/move_profile.rake
```diff
@@ -1,8 +1,24 @@
...
+
+    total = moves.count
+
+    if moves.empty?
+      puts 'All profiles IDs were already updated.'
+      return
     end
+
+    moves.find_each(batch_size: 200).with_index do |move, n|
+      puts "#{n}/#{total} moves processed ..." if (n % 200).zero? # Show progression
```
willfish, line 14: Reminds me of https://github.com/paul/progress_bar :). Not suggesting we add it. But have used in the past.

lib/tasks/move_profile.rake
```diff
@@ -1,8 +1,24 @@
...
+      profile = move.person.profiles.order(:updated_at).last # Take the profile that was most recently updated
```
willfish, line 15: I guess a couple of things are true here:

1. We never update profiles (so created_at/updated_at) are the same? I think it might be safer to just use created_at? since we can update the relationships of a profile but that doesn't make it the latest one?
2. We don't mind pulling the latest profile (as I understand it we create a profile in two places - when we get the summary of a person in PeoplesController#index (via the importer) and when the PeoplesController received a create?). Looking at it we can only ever have one profile for a person, currently, so there's no harm in doing this? 

I'm a little confused about the circumstances that would lead to not having a profile on a move, though 🤔.

willfish: I'm also wondering whether the ordering for needs to be updated to reflect how you've decided profiles should be ordered?

https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/models/person.rb#L9-L11

lib/tasks/move_profile.rake
alexdesi: :) it is just temporary, after a while we are going to remove it

lib/tasks/move_profile.rake
alexdesi: yes, your points are good in theory, but not in practice :)

Consider that the project is was started more year ago and we can have inconsistent data,
this migration is based on analyzing the real data:
- I found that the profile with the most updated `updated_at`  attribute, has also the `assessment_answer` associated.
- I found that `created_at` is often the same for different profile, and `updated_at` are different.

I agree that we should update the 
```
def latest_profile
    profiles.last
  end
```

alexdesi: PS. to compare 2 different profile I wrote this: https://gist.github.com/alexdesi/88df4bce3353d05ae0984572e9ab956f

PR #428, opened 2020-05-05
[P4-1433] Grape and netflix json:api serialiser spike

willfish:
### Jira link

P4-1433

### What?

This spike is part of an exploration to see how we can improve on jsonapi endpoint creation using:

1. [Netflix JSON:API serializer gem](https://github.com/Netflix/fast_jsonapi)
2. [Grape](https://github.com/ruby-grape/grape)
3. [Grape Netflix JSON:API integration layer](https://github.com/EmCousin/grape_fast_jsonapi)

- [x] Validate migration process for serialising an existing endpoint (old world - not grape). I chose CourtCases for this as it seemed like a simple enough implementation to replace.
- [x] Validate mounting and exposing a grape api that renders court cases in the same way as the `/api/v1/people/:person_id/court_cases` endpoint
- [ ] Validate swagger doc generation in the new api
- [x] Validate writing a spec test for the grape api
- [ ] Validate auth

I'm interested in:

1. Is it easier to use/generate json:api compliant endpoints?
2. Are there any gotchas we should know about?
3. Does it enable faster development of new apis?

### Why?

A whole bunch of conversations have happened historically - too many to go over here - that have highlighted a general frustration with the process of generating our apis with standard rails/AMS and rswag.

This is an example of an alternative approach that may or may not be easier - that's what this spike endeavours to find out.


willfish:
### Lesson 1

Migrating to Netflix serializers is dangerous if there aren't unit tests of the format of the existing serializer output - we should add fully document unit tests for all of our serializers.

app/serializers/fast_jsonapi/location_serializer.rb
```diff
@@ -0,0 +1,13 @@
...
+
+class FastJsonapi::LocationSerializer
+  include FastJsonapi::ObjectSerializer
+
+  set_type 'locations'
+
+  attributes :key, :title, :location_type, :nomis_agency_id, :can_upload_documents, :disabled_at
+
+  attribute :suppliers do |location|
+    location.suppliers.each { |supplier| FastJsonapi::SupplierSerializer.new(supplier) }
```
willfish, line 11: Validate me

willfish:
### Lesson 2

The way we manage supplier relationships on a location totally aren't json:api spec compliant and potentially work by accident.

PR #429, opened 2020-05-05
Small improvement for Moves::ImportPeople. More need to be done!

alexdesi:
### What?

`Moves::ImportPeople` is spaghetti 🍝  code,
I need to work on that part at the moment,  so this is just a mini step to clean it up with minimal risk.

### DIFF
--- a/app/controllers/api/v1/people_controller.rb
+++ b/app/controllers/api/v1/people_controller.rb
@@ -71,10 +71,10 @@ module Api
         { errors: [error.json_api_error] }
       end
 
-      def import_person_from_nomis prison_number
+      def import_person_from_nomis(prison_number)
         # This prevents us from blaming the current user/application for the NOMIS sync
         PaperTrail.request(whodunnit: nil) do
-          Moves::ImportPeople.new([person_nomis_prison_number: prison_number]).call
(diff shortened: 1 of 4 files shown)


app/services/moves/import_people.rb
```diff
@@ -2,10 +2,10 @@
 
 module Moves
   class ImportPeople
-    attr_accessor :items
+    attr_accessor :people_nomis_prison_numbers
```
martyn-w, line 5: this name doesn't roll of the tongue easily, how about `nomis_prisoner_numbers` ?

app/controllers/api/v1/people_controller.rb
```diff
@@ -74,7 +74,7 @@ def json_api_errors_for(error)
       def import_person_from_nomis prison_number
```
martyn-w, line 74: tiny point: could we use parentheses in method definitions with parameters, i.e.  `def import_person_from_nomis(prison_number)`

alexdesi: yes, I agree

app/services/moves/import_people.rb
alexdesi: :D yeah prison_numbers is enough

PR #430, opened 2020-05-05
[P4-1433] Removes unused supplier serializer code

willfish:
### Jira link

P4-1433

### What?

This [pr](https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/143/)
incorrectly added a suppliers relationship as one of the attributes of
the LocationSerializer.

This code never returns SupplierSerializer objects and actually returns the result of just calling Location#suppliers so I can't see any harm removing it. AMS eventually just calls to_json on instances of Supplier and never uses the SupplierSerializer.

In a separate PR/in future I propose we remove the attribute `suppliers` that are each getting `Supplier#to_json` called on them everytime we return a location and instead use includes/relationships in the way specified by json:api.

### Why?

- This code is not used

### DIFF
--- a/spec/serializers/location_serializer_spec.rb
+++ b/spec/serializers/location_serializer_spec.rb
@@ -45,6 +45,16 @@ RSpec.describe LocationSerializer do
   end
 
   it 'contains a suppliers attribute' do
-    expect(attributes[:suppliers]).not_to be_nil
+    expect(attributes[:suppliers]).to eq(
+      [
+        {
+          created_at: supplier.created_at.xmlschema,
+          id: supplier.id,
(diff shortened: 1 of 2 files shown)


spec/serializers/location_serializer_spec.rb
```diff
@@ -45,6 +45,16 @@
   end
 
   it 'contains a suppliers attribute' do
-    expect(attributes[:suppliers]).not_to be_nil
+    expect(attributes[:suppliers]).to eq(
```
willfish, line 48: This being an afterthought, I ran the same example on master and it passes

PR #431, opened 2020-05-06
P4-1273 Add importer class to populate regions reference data

smoothcontract:
### Jira link

P4-1273

### What?

- [x] Added new region model and corresponding migration including join table with locations (regions have many locations and locations can be part of many regions)
- [x] Added YML file containing details of the regions (key, name and NOMIS location id's) to populate data (reworked from original Excel spreadsheet)
- [x] Added rake task to call importer with the YML file

### Why?

- The reference data for this is mastered in the back end as the source of truth. Working with a YML file is much simpler and avoids the need to add extra gems (e.g. Roo) to process spreadsheets. We can modify the YML file as needed in future should the regions change.
- Future work will expose a reference data endpoint to return the locations associated with each region
- This is needed for the front end allocations work to allow user to select region of interest

- Rake task `rails reference_data:create_regions` can be run safely as the data is not yet linked to any API endpoints
- Included YML file does not contain any sensitive data

### DIFF
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -20,6 +20,7 @@ class Location < ApplicationRecord
   NOMIS_TYPES_WITH_DOCUMENTS = %w[STC SCH].freeze
 
   has_and_belongs_to_many :suppliers
+  has_and_belongs_to_many :regions
   # Deleting locations isn't really a thing in practice - so dependent: :destroy is a pragmatic choice
   has_many :moves_from, class_name: 'Move', foreign_key: :from_location_id, inverse_of: :from_location, dependent: :destroy
   has_many :moves_to, class_name: 'Move', foreign_key: :to_location_id, inverse_of: :to_location, dependent: :destroy
--- a/db/migrate/20200506105933_create_regions.rb
+++ b/db/migrate/20200506105933_create_regions.rb
@@ -1,0 +1,17 @@
+class CreateRegions < ActiveRecord::Migration[5.2]
+  def change
+    create_table :regions, id: :uuid do |t|
+      t.string :name, null: false
+      t.string :key, null: false, index: true
+      t.timestamps
+    end
+
+    create_join_table :locations, :regions do |t|
+      t.references :location, type: :uuid, null: false, index: true, foreign_key: true
+      t.references :region, type: :uuid, null: false, index: true, foreign_key: true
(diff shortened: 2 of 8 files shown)


smoothcontract:
Looks like there's an unrelated flaky spec causing the failure on Circle CI. Once this has been reviewed I'll rebase and hopefully that will pass.

db/migrate/20200506105933_create_regions.rb
```diff
@@ -0,0 +1,17 @@
...
+      t.string :name, null: false
+      t.string :key, null: false, index: true
+      t.timestamps
+    end
+
+    create_join_table :locations, :regions do |t|
+      t.references :location, type: :uuid, null: false, index: true, foreign_key: true
+      t.references :region, type: :uuid, null: false, index: true, foreign_key: true
+      t.index %i[location_id region_id], unique: true
+      t.index %i[region_id location_id], unique: true
```
martyn-w, line 13: just wondering if `t.index %i[location_id region_id], unique: true` translates to the same index as `t.index %i[region_id location_id], unique: true` in the database?

app/models/location.rb
```diff
@@ -20,6 +20,7 @@ class Location < ApplicationRecord
   NOMIS_TYPES_WITH_DOCUMENTS = %w[STC SCH].freeze
 
   has_and_belongs_to_many :suppliers
+  has_and_belongs_to_many :regions
```
martyn-w, line 23: does a location actually belong to more than one region? I can't seem to see an example of that in the yaml file?

smoothcontract: You have to look really closely. Annoyingly, there are only 2 locations that exist in 2 regions so it can't be just mapped as a simple `belongs_to` relationship. They're `FMI` and `PRI`.

martyn-w: jolly good 👍

lib/tasks/data/regions.yml
```diff
@@ -0,0 +1,151 @@
...
+    - WRI
+    - WHI
+  8:
+    name: Youth Custody Service
+    locations:
+    - FMI
+    - PRI
+    - CKI
+    - WNI
+    - WYI
```
smoothcontract, line 151: Curses! Will add an extra blank line.

PR #432, opened 2020-05-06
Add profile relationship to Move

alexdesi:
### Jira link
https://dsdmoj.atlassian.net/browse/P4-1450

### What?
This add the relationship Move-Profile
and remove the Move-Person

Also, this PR update the `set_profiles` rake task,
so it can run without using the `move.person` relationship.

### Important
We must keep the the API interface exactly the same. That is why I did not touch any expected swagger response.

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -2,7 +2,8 @@
 
 class Person < VersionedModel
   has_many :profiles, dependent: :destroy
-  has_many :moves, dependent: :destroy
+
+  has_many :moves, through: :profiles
 
   has_one_attached :image
 
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -8,7 +8,7 @@ class Person < VersionedModel
   has_one_attached :image
 
   def latest_profile
-    profiles.last
+    profiles.order(:updated_at).last
   end
 
   def latest_nomis_booking_id
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -7,7 +7,7 @@ class Profile < VersionedModel
   belongs_to :ethnicity, optional: true
   belongs_to :gender, optional: true
 
-  has_many :moves
+  has_many :moves, dependent: :destroy
 
   validates :person, presence: true
   validates :last_name, presence: true
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -7,6 +7,8 @@ class Profile < VersionedModel
   belongs_to :ethnicity, optional: true
   belongs_to :gender, optional: true
 
+  has_many :moves
+
   validates :person, presence: true
   validates :last_name, presence: true
   validates :first_names, presence: true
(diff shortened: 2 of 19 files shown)


lib/tasks/move_profile.rake
```diff
@@ -1,7 +1,7 @@
 namespace :moves do
   desc 'Set all profile ids in all the moves'
   task set_profiles: :environment do
-    moves = Move.where(profile_id: nil).includes(person: :profiles)
+    moves = Move.where(profile_id: nil)
```
jcavell, line 4: Could we have some better naming. For the file, maybe call it move_set_profile_from_person.rake.
Instead of the local variable moves, call it moves_with_nil_profile

lib/tasks/move_profile.rake
```diff
@@ -12,7 +12,11 @@ namespace :moves do
 
     moves.find_each(batch_size: 200).with_index do |move, n|
       puts "#{n}/#{total} moves processed ..." if (n % 200).zero? # Show progression
-      profile = move.person.profiles.order(:updated_at).last # Take the profile that was most recently updated
+
```
jcavell, line 15: Can't you use your new person.latest_profile method?

app/models/profile.rb
```diff
@@ -7,6 +7,8 @@ class Profile < VersionedModel
   belongs_to :ethnicity, optional: true
   belongs_to :gender, optional: true
 
+  has_many :moves, dependent: :destroy
+
```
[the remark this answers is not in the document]
alexdesi: Ok, I'm gonna change the relationship to `has_one`
and the `dependent` to  `:nullify`
that means, when we delete a profile, the relevant move will have the move_id wil be set to `nil` (will be orphan)

lib/tasks/move_profile.rake
alexdesi: np, I've called this `set_profile_from_person` since it is already inside the namespace `moves`.
Anyway, after a while this tasks will be removed.

lib/tasks/move_profile.rake
alexdesi: good pont!

app/models/person.rb
```diff
@@ -2,12 +2,13 @@
...
   has_many :profiles, dependent: :destroy
-  has_many :moves, dependent: :destroy
+
+  has_many :moves, through: :profiles
 
   has_one_attached :image
 
   def latest_profile
-    profiles.last
+    profiles.order(:updated_at).last
```
willfish, line 11: :tada:

app/services/moves/finder.rb
```diff
@@ -26,7 +26,7 @@ def call
     def apply_ordering(scope)
       case @order_by
       when :name
-        scope.joins(person: :profiles).merge(Profile.ordered_by_name(@order_direction))
+        scope.joins(:profile).merge(Profile.ordered_by_name(@order_direction))
```
willfish, line 29: :tada:

PR #436, opened 2020-05-07
Fix flaky specs

smoothcontract:
This sometimes fails for me but I found a seed (65033) that reliably reproduced the problem. Looks to just be (incorrectly) expecting documents to be returned in a specific order rather than just checking for the correct documents.

### DIFF
--- a/spec/requests/api/v1/moves_controller_update_spec.rb
+++ b/spec/requests/api/v1/moves_controller_update_spec.rb
@@ -146,9 +146,9 @@ RSpec.describe Api::V1::MovesController do
           end
 
           it 'updates the moves documents' do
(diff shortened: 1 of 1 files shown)


spec/requests/api/v1/moves_controller_update_spec.rb
```diff
@@ -160,7 +160,7 @@
 
             expect(
               response_json.dig('data', 'relationships', 'documents', 'data').map { |document| document['id'] },
-            ).to eq(after_documents.pluck(:id))
+            ).to match(after_documents.pluck(:id))
```
willfish, line 163:
```suggestion
            ).to contain_exactly(after_documents.pluck(:id))
```

PR #438, opened 2020-05-07
P4-1466 Surface an allocation associated to a move

cwrw:
jira-ticket: https://dsdmoj.atlassian.net/browse/P4-1466

### Jira link

[P4-1466](https://dsdmoj.atlassian.net/browse/P4-1466)

### What?

- [x] Add an allocation to the moves serializer
- [x] Fix bug in moves serializer regarding included objects vs fields (this will be deprecated when @willfish picks up moving to params specifying what should be included)

### Why?

We would like to surface allocations if they are associated with a move on the moves and move get endpoints. These allocations should appear as a relationship, as well as in detail in the  included section
I also fixed bug in the move serializer where we aren't utilising the correct format for only including fields vs objects. These need to be separated. The `include` key will handle what objects we expect to return, while the `fields` key determines which fields we only want to return per object. After speaking to Dom, we would like to start limiting fields as we go rather to not break current included objects. We are now only limiting allocations, but all other objects will have all their attributes returned otherwise. 

### Have you? (optional)

### DIFF
--- a/app/controllers/api/v1/moves_controller.rb
+++ b/app/controllers/api/v1/moves_controller.rb
@@ -8,7 +8,7 @@ module Api
         if moves_params.valid?
           moves = Moves::Finder.new(filter_params, current_ability, params[:sort] || {}).call
           # Excludes potentially many court hearing documents to reduce the request size. This was requested specifically by the frontend team.
-          paginate moves, include: MoveSerializer::INCLUDED_ATTRIBUTES.dup.except(:court_hearings)
+          paginate moves, include: MoveSerializer::INCLUDED_ATTRIBUTES.dup.except(:court_hearings), fields: MoveSerializer::INCLUDED_FIELDS
         else
           render json: { error: moves_params.errors }, status: :bad_request
         end
@@ -101,7 +101,7 @@ module Api
       end
 
       def render_move(move, status)
-        render json: move, status: status, include: MoveSerializer::INCLUDED_ATTRIBUTES
+        render json: move, status: status, include: MoveSerializer::INCLUDED_ATTRIBUTES, fields: MoveSerializer::INCLUDED_FIELDS
       end
 
       def move
--- a/app/serializers/move_serializer.rb
+++ b/app/serializers/move_serializer.rb
@@ -10,13 +10,19 @@ class MoveSerializer < ActiveModel::Serializer
   has_one :prison_transfer_reason, serializer: PrisonTransferReasonSerializer, if: -> { object.prison_transfer_reason.present? }
   has_many :documents, serializer: DocumentSerializer
   has_many :court_hearings, serializer: CourtHearingSerializer
+  belongs_to :allocation, serializer: AllocationSerializer
 
   INCLUDED_ATTRIBUTES = {
-    person: %i[first_names last_name date_of_birth assessment_answers indentifiers ethnicity gender reason_comment],
-    from_location: %i[location_type description],
-    to_location: %i[location_type description],
-    documents: %i[url filename filesize content_type],
-    prison_transfer_reason: %i[key title],
+    person: %i[ethnicity gender],
+    from_location: [],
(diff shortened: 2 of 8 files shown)


app/controllers/api/v1/moves_controller.rb
```diff
@@ -101,13 +101,13 @@ def patch_move_attributes
...
       def render_move(move, status)
-        render json: move, status: status, include: MoveSerializer::INCLUDED_ATTRIBUTES
+        render json: move, status: status, include: MoveSerializer::INCLUDED_ATTRIBUTES, fields: MoveSerializer::INCLUDED_FIELDS
       end
 
       def move
         @move ||= Move
           .accessible_by(current_ability)
-          .includes(:from_location, :to_location, person: { profiles: %i[gender ethnicity] })
+          .includes(:from_location, :to_location, :allocation, person: { profiles: %i[gender ethnicity] })
```
cwrw, line 110: actually don't think we need those in includes

app/services/moves/finder.rb
```diff
@@ -42,7 +42,7 @@ def apply_ordering(scope)
 
     def apply_filters(scope)
       scope = scope.accessible_by(ability)
-      scope = scope.includes(:from_location, :to_location, person: { profiles: %i[gender ethnicity] })
+      scope = scope.includes(:from_location, :to_location, :allocation, person: { profiles: %i[gender ethnicity] })
```
cwrw, line 45: same here

app/serializers/move_serializer.rb
```diff
@@ -10,13 +10,19 @@ class MoveSerializer < ActiveModel::Serializer
...
   has_many :court_hearings, serializer: CourtHearingSerializer
+  belongs_to :allocation, serializer: AllocationSerializer
 
   INCLUDED_ATTRIBUTES = {
-    person: %i[first_names last_name date_of_birth assessment_answers indentifiers ethnicity gender reason_comment],
-    from_location: %i[location_type description],
-    to_location: %i[location_type description],
-    documents: %i[url filename filesize content_type],
-    prison_transfer_reason: %i[key title],
+    person: %i[ethnicity gender],
```
martyn-w, line 16: I'm a bit worried that these serialization changes will have an impact on the front end and also on other api consumers... are they intended and have they been discussed?

smoothcontract: Discussed this offline with @cwrw - essentially the output of the endpoint hasn't changed (other than to include the allocation where it's related to a move). The previous implementation of included attributes did not work as suggested by the code - instead every relationship included all of the attributes. That was easy to see from testing with Postman.

I prefer that the code here now matches the actual behaviour. It's something we can revise further moving forwards, perhaps when we move to an alternate JSONAPI technique and allow the API client to specify attributes it wishes to receive.

app/serializers/move_serializer.rb
```diff
@@ -10,13 +10,19 @@ class MoveSerializer < ActiveModel::Serializer
...
+    from_location: [],
+    to_location: [],
+    documents: [],
+    prison_transfer_reason: [],
     court_hearings: [],
+    allocation: [],
+  }.freeze
+
+  INCLUDED_FIELDS = {
+    allocation: %i[to_location from_location moves_count created_at],
```
smoothcontract, line 26: Discussed this too as it wasn't immediately obvious. This does set an explicit list of attributes only for the allocation relationship (these are the ones that will be useful in the front end). For all other relationships, every attribute is implicitly included - this can be seen from the passing request specs.

app/serializers/move_serializer.rb
smoothcontract: Also `ethnicity` and `gender` are special cases - those are related models rather than attributes, so need to be explicitly included, otherwise the associated models don't get included in the response. Think of this as deep nesting of models 🤔

PR #439, opened 2020-05-07
Handle missing attributes when patching a moves documents

willfish:
### Jira link

P4-<TODO>

### What?

Adds support for patching a move without any attribute changes (i.e. only the documents)

### Why?

This is a genuine use case of this api.

### DIFF
--- a/spec/requests/api/v1/moves_controller_update_spec.rb
+++ b/spec/requests/api/v1/moves_controller_update_spec.rb
@@ -173,9 +173,9 @@ RSpec.describe Api::V1::MovesController do
             end
 
             it 'updates the moves documents' do
-              expect(move.reload.documents).to eq(before_documents)
+              expect(move.reload.documents).to match_array(before_documents)
               do_patch
-              expect(move.reload.documents).to eq(after_documents)
+              expect(move.reload.documents).to match_array(after_documents)
             end
 
             it 'does not affect other relationships' do
--- a/spec/requests/api/v1/moves_controller_update_spec.rb
+++ b/spec/requests/api/v1/moves_controller_update_spec.rb
@@ -162,6 +162,34 @@ RSpec.describe Api::V1::MovesController do
               response_json.dig('data', 'relationships', 'documents', 'data').map { |document| document['id'] },
             ).to eq(after_documents.pluck(:id))
           end
+
+          context 'when there are no attributes in the patch' do
+            let(:move_params) do
+              documents = after_documents.map { |d| { id: d.id, type: 'documents' } }
+              {
+                type: 'moves',
+                relationships: { documents: { data: documents } },
+              }
+            end
+
+            it 'updates the moves documents' do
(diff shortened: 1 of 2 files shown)


spec/requests/api/v1/moves_controller_update_spec.rb
```diff
@@ -162,6 +162,34 @@
               response_json.dig('data', 'relationships', 'documents', 'data').map { |document| document['id'] },
             ).to eq(after_documents.pluck(:id))
           end
+
+          context 'when there are no attributes in the patch' do
```
willfish, line 166:
```suggestion
          context 'when there are no attributes in the params' do
```

spec/requests/api/v1/moves_controller_update_spec.rb
```diff
@@ -162,6 +162,34 @@
...
+            it 'does not affect other relationships' do
+              expect { do_patch }.not_to change { move.reload.from_location }
+            end
+
+            it 'returns the updated documents in the response body' do
+              do_patch
+
+              expect(
+                response_json.dig('data', 'relationships', 'documents', 'data').map { |document| document['id'] },
+              ).to eq(after_documents.pluck(:id))
```
cwrw, line 190: if order is not an issue here could we use `match_array` to avoid the risk of flakiness?

willfish, line 190:
```suggestion
              ).to match_array(after_documents.pluck(:id))
```

spec/requests/api/v1/moves_controller_update_spec.rb
```diff
@@ -162,6 +162,34 @@
...
+            let(:move_params) do
+              documents = after_documents.map { |d| { id: d.id, type: 'documents' } }
+              {
+                type: 'moves',
+                relationships: { documents: { data: documents } },
+              }
+            end
+
+            it 'updates the moves documents' do
+              expect(move.reload.documents).to eq(before_documents)
```
willfish, line 176:
```suggestion
              expect(move.reload.documents).to match_array(before_documents)
```

spec/requests/api/v1/moves_controller_update_spec.rb
```diff
@@ -162,6 +162,34 @@
...
+              do_patch
+              expect(move.reload.documents).to eq(after_documents)
```
willfish, line 178:
```suggestion
              expect(move.reload.documents).to match_array(after_documents)
```

spec/requests/api/v1/moves_controller_update_spec.rb
willfish: So this is all very weird. It makes me think we're always relying on the natural ordering in the db which is a bit temperamental. We should fix that separately.

martyn-w: i've fixed this in my (unrelated) pr - I'd just changed `eq` to `match_array` as we can't assume an order from the database

PR #441, opened 2020-05-11
P4-1180 journey create, update, index and show endpoints

martyn-w:
### Jira link

[P4-1180](https://dsdmoj.atlassian.net/browse/P4-11800

### What?

I have added/removed/altered:

- [x] Adds Create, Update, Index and Show endpoints for Journeys
- [x] Changes Event model to be polymorphic (so journeys and moves can have events)
- [x] Logs create/update to event log
- [x] Lots of specs

### Why?

I am doing this because:
- to support creation of journeys, for billing

- Changes an api that is used in production: should be minimal risk as these are new and unused endpoints


app/controllers/api_controller.rb
```diff
@@ -64,10 +64,16 @@ def render_bad_request_error(exception)
   end
 
   def render_resource_not_found_error(exception)
+    # NB: cleaner error message without a long WHERE id=foo clause
+    detail = if exception.is_a?(ActiveRecord::RecordNotFound) && exception.id.present?
```
alexdesi, line 68: is it not redundant to check if 
`exception.is_a?(ActiveRecord::RecordNotFound`
since we have
`rescue_from ActiveRecord::RecordNotFound, with: :render_resource_not_found_error` ?

app/models/event.rb
```diff
@@ -1,23 +1,10 @@
 class Event < ApplicationRecord
-  belongs_to :move
+  belongs_to :entity, polymorphic: true
```
alexdesi, line 2: I understand we need a generic name here,
but what actually entity is? there is not a model for it, if I full search in the project you get tens of result and it's difficult to have an idea about this `entity`

app/services/journeys/params_validator.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module Journeys
+  class ParamsValidator
+    include ActiveModel::Validations
+
+    attr_reader :action_name, :timestamp, :billable, :vehicle, :from_location_id, :to_location_id
+
+    validates :billable, inclusion: { in: [true, false] }, allow_nil: true, on: :update # NB: billable is optional on update
```
alexdesi, line 9: It's not simpler just to set false (or true) as default?
`nil` for boolean is something we should whenever possible

config/routes.rb
```diff
@@ -25,6 +25,7 @@
       resources :moves, only: %i[index show create destroy update] do
         resources :documents, only: %i[create destroy]
         resources :events, only: %i[create], controller: 'move_events'
+        resources :journeys
```
alexdesi, line 28: As we did for move, better to be explicit here and specify all the possible actions.

app/controllers/api/v1/journeys_controller.rb
```diff
@@ -0,0 +1,104 @@
...
+      end
+
+    private
+
+      def move
+        @move ||= Move.accessible_by(current_ability).find(params.require(:move_id))
+      end
+
+      def journeys
+        @journeys ||= move.journeys.accessible_by(current_ability).default_order
```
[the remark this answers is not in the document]
smoothcontract: Not a fan of default scopes at all - they tend to be a pain when you want something other than the default scope behaviour, so this is fine I think.

app/controllers/api/v1/journeys_controller.rb
```diff
@@ -0,0 +1,104 @@
...
+        paginate journeys
+      end
+
+      def show
+        render json: journey, status: :ok
+      end
+
+      def create
+        authorize!(:create, journey)
+        journey.save!
```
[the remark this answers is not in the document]
martyn-w: nothing springs to mind - I think the validators are fairly comprehensive. If there are other (non validation) errors e.g. duplicate record they will be caught by the API controller

app/controllers/api/v1/journeys_controller.rb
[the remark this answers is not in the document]
martyn-w: agreed not to use a `default_scope` 😅

app/controllers/api_controller.rb
martyn-w: yep, you are right, I'll change it 👍

app/models/event.rb
martyn-w: agreed to rename to `related_entity`

config/routes.rb
martyn-w: good point 👍

app/controllers/api/v1/journeys_controller.rb
```diff
@@ -0,0 +1,104 @@
...
+      def index
+        paginate journeys
+      end
+
+      def show
+        render json: journey, status: :ok
+      end
+
+      def create
+        authorize!(:create, journey)
```
smoothcontract, line 29: I'm not entirely clear why this extra `authorize!` call is necessary? The `journey` method already chains `move` method, which in turn scopes to moves accessible to the supplier. So if a supplier attempts to create a journey relating to a move they don't have access to then I would expect the authorisation to fail within the `move` method. So presumably that all works even without this line?

app/services/journeys/params_validator.rb
martyn-w: hmm... in this case no, billable has a very real financial meaning, if its `false` the supplier will pay and if its `true` the MoJ will pay. So I want it to be explicitly set by the client and not to have an assumed/default value

app/controllers/api/v1/journeys_controller.rb
```diff
@@ -0,0 +1,104 @@
...
+      def move
+        @move ||= Move.accessible_by(current_ability).find(params.require(:move_id))
+      end
+
+      def journeys
+        @journeys ||= move.journeys.accessible_by(current_ability).default_order
+      end
+
+      def journey
+        @journey ||= if action_name == 'create'
```
smoothcontract, line 50: I think this would be nicer as two separate and explicit methods:

```ruby
def new_journey
  @new_journey ||= Journey.new(new_journey_attributes)
end

def journey
  @journey ||= Journeys.find(params.require(:id))
end
```

And then use the respective version above (just need to reference `new_journey` in create method, the other endpoints don't need changing):

```ruby
def create
  authorize!(:create, new_journey) # As mentioned above, I think this line is redundant anyway
  new_journey.save!
  render json: new_journey, status: :created
end
```

spec/requests/api/v1/journeys_controller_create_spec.rb
```diff
@@ -0,0 +1,147 @@
...
+        let(:journey_params) { nil }
+
+        it_behaves_like 'an endpoint that responds with error 400'
+      end
+
+      context 'with an invalid timestamp' do
+        let(:timestamp) { 'foo-bar' }
+
+        it_behaves_like 'an endpoint that responds with error 400' do
+          let(:errors_400) {
```
alexdesi, line 101: The invalid params error should not be express with 422 status?

spec/requests/api/v1/journeys_controller_create_spec.rb
```diff
@@ -0,0 +1,147 @@
...
+               'detail' => 'Validation failed: Timestamp must be formatted as a valid ISO-8601 date-time' }]
+          }
+        end
+      end
+
+      context 'with an invalid billable' do
+        let(:billable) { 'foo-bar' }
+
+        it_behaves_like 'an endpoint that responds with error 400' do
+          let(:errors_400) {
```
alexdesi, line 112: maybe 422 (as above)?

spec/requests/api/v1/journeys_controller_create_spec.rb
```diff
@@ -0,0 +1,147 @@
...
+      end
+
+      context 'when not authorized' do
+        let(:access_token) { 'foo-bar' }
+        let(:detail_401) { 'Token expired or invalid' }
+
+        it_behaves_like 'an endpoint that responds with error 401'
+      end
+
+      context 'with a missing move_id' do
```
alexdesi, line 126: this sounds like the the param move_id is missing,
maybe `when move is not found`, `when move does not exist`, ... up to you

spec/requests/api/v1/journeys_controller_create_spec.rb
```diff
@@ -0,0 +1,147 @@
...
+        let(:move_id) { 'foo-bar' }
+        let(:detail_404) { "Couldn't find Move with 'id'=foo-bar" }
+
+        it_behaves_like 'an endpoint that responds with error 404'
+      end
+
+      context 'with a reference to a missing relationship' do
+        let(:to_location_id) { 'foo-bar' }
+        let(:detail_404) { "Couldn't find Location with 'id'=foo-bar" }
```
alexdesi, line 135: this is a POST to the journey,
I am not sure if `404 not found` is the best status to return.
I've read that it should be used when we are trying to get a resource, the 422 (Unprocessable Entity) might be more appropriate, but double check with Jonny about this.

app/controllers/api_controller.rb
smoothcontract: Looks like there's one other case where this is called explicitly (and passes an instance of `Exception`) so the check is needed, but perhaps this calling code (from `PeopleController`) could be cleaned up instead, or possibly adapted to follow the standard pattern with populated 'model', `primary_key` and `id` attributes present):

```ruby
render_resource_not_found_error(Exception.new('Image not found'))
```

Better to change that to just throw an `ActiveRecord::RecordNotFound` error I think, and then let this method catch it - which would then allow the `detail` branching code here to be removed and replaced  with a consistent error message i.e.

```ruby
render(
  json: { errors: [{
    title: 'Resource not found',
    detail: "Couldn't find #{exception.model} with '#{exception.primary_key}'=#{exception.id}",
  }] },
  status: :not_found,
)
```

app/models/ability.rb
```diff
@@ -11,10 +11,12 @@ def initialize(application = nil)
       can :manage, Move, Move.served_by(application.owner) do |move|
         move.from_location.suppliers.include?(application.owner)
       end
+      can :manage, Journey, supplier_id: application.owner_id
     else
-      # momentarily give all permissions to not break compatibility with
+      # temporarily give all permissions to non-supplier accounts so as not to break compatibility with
```
smoothcontract, line 16: What's the longer term plan here?

app/models/event.rb
smoothcontract: Looks like they're events related to either moves or journeys for now - and potentially other things in future. It's unfortunate that `event` is a widely used term but not sure what else would make sense here.

spec/requests/api/v1/journeys_controller_index_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+    let!(:first_journey) { create(:journey, move: move, supplier: supplier, client_timestamp: '2020-05-04T08:00:00Z') }
+    let!(:other_move_journey) { create(:journey, supplier: supplier) } # another journey for a different move, same supplier
+    let!(:other_supplier_journey) { create(:journey, move: move) } # another journey for a different supplier, same move
+
+    let(:intermediate_journeys_count) { 1 }
+    let(:page) { 1 }
+    let(:per_page) { 20 }
+
+    before do
+      get "/api/v1/moves/#{move.id}/journeys?page=#{page}&per_page=#{per_page}", headers: headers, as: :json
```
alexdesi, line 28: Do we need specify the pagination here?
I was thinking that we could simplify the test removing it

spec/requests/api/v1/journeys_controller_index_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+        subject(:journey_ids) { response_json['data'].map { |x| x['id'] } }
+
+        it { expect(journey_ids.size).to be 3 }
+        it { expect(journey_ids.first).to eql(first_journey.id) } # NB: first_journey should come before last_journey
+        it { expect(journey_ids.last).to eql(last_journey.id) }
+        it { is_expected.not_to include(other_move_journey.id) } # NB: should not contain another move's journey
+        it { is_expected.not_to include(other_supplier_journey.id) } # NB: should not contain another supplier's journey in the same move
+      end
+
+      describe 'paginating results' do
```
alexdesi, line 46: maybe we should add the endpoint with the pagination params only inside this context

app/models/journey.rb
```diff
@@ -43,8 +45,26 @@ class Journey < ApplicationRecord
 
   after_initialize :initialize_state # NB there is an equivalent after(:build) callback used by FactoryBot in the journeys factory
 
+  def vehicle
+    get_metadata('vehicle')
```
smoothcontract, line 49: meta data seems like a strange term to use here - I would have thought that the vehicle used for the journey is more concrete than that, and makes sense to have as an explicit attribute on the model?

app/serializers/journey_serializer.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class JourneySerializer < ActiveModel::Serializer
+  include Rails.application.routes.url_helpers
```
smoothcontract, line 4: Why are the route helpers needed?

app/services/journeys/params_validator.rb
smoothcontract: I'd agree, this is a little strange. Booleans should only have two states, not three. If this needs three states then it probably makes more sense to use a different type.

spec/requests/api/v1/journeys_controller_show_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+      context 'when not authorized' do
+        let(:access_token) { 'foo-bar' }
+        let(:detail_401) { 'Token expired or invalid' }
+
+        it_behaves_like 'an endpoint that responds with error 401'
+      end
+
+      context "when attempting to access another supplier's journey" do
+        let(:journey) { create(:journey, move: move) } # another journey for a different supplier, same move
+        let(:detail_404) { "Couldn't find Journey with 'id'=#{journey.id}" }
```
alexdesi, line 74: should not be 401 unauthorized) when attempting to access another supplier's journey?

app/services/journeys/params_validator.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module Journeys
```
smoothcontract, line 3: Some of this code feels a little superfluous - the `ParamsValidator` pattern is used elsewhere to e.g. validate parseable dates which I think is useful to include. But the presence of specific attributes seems to replicate most of the logic in the controller where specific list of params are defined for each controller action. I would suspect some of this code could instead be moved to `params.require` checks in the controller instead? Just a thought and may not be something you want to change.

spec/requests/api/v1/journeys_controller_update_spec.rb
```diff
@@ -0,0 +1,178 @@
...
+    end
+
+
+    context 'when unsuccessful' do
+      let(:schema) { load_yaml_schema('error_responses.yaml') }
+
+      context 'with a bad request' do
+        let(:journey_params) { nil }
+
+        it_behaves_like 'an endpoint that responds with error 400'
```
alexdesi, line 125: it might be a 422

config/routes.rb
smoothcontract: Does the journeys controller implement all of the restful actions? If so I think it's fine to keep this as is.

spec/requests/api/v1/journeys_controller_update_spec.rb
alexdesi: this is a general things we should check for the other responses in these endpoints, maybe we should catch up with Jonny about this to confirm.
I'm telling this because I had the same issue in another endpoint I wrote.

config/routes.rb
martyn-w: it does not implement `delete`, so no, its not a completely restful

app/controllers/api/v1/journeys_controller.rb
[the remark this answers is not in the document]
alexdesi: 👍 good point, thanks guys, I've just changed my mind on this too 😄

app/controllers/api/v1/journeys_controller.rb
martyn-w: @smoothcontract good question: its to catch a user trying to create a journey assigned to a different supplier. Both moves and journeys have to be authorised; the move is authorised by its scope; but the journey is new and not yet persisted to the database - hence the authorisation check.

app/models/event.rb
alexdesi: `related_entity`, or `event_entity` are better name, at least they are not widely used.
As agree a new test helps to clarify that this entity can be

app/controllers/api/v1/journeys_controller.rb
martyn-w: @smoothcontract I did have something similar in an earlier version... but the `if` statement doesn't go away  - see the `create_event` method at the bottom of the controller. That needs to reference the `journey` and so will need to determine whether its `journey` or `new_journey`. On balance I decided always calling the journey `journey` simplified things.

app/models/ability.rb
martyn-w: the longer term plan is to use the external oauth provider and assign roles/rights via that, rather than the internal oauth/doorkeeper we currently have.

app/models/event.rb
martyn-w: These events will definitely relate to `moves` and journeys` and will likely relate to other things (people etc) in the future.
@smoothcontract is `related_entity` cleaner than `entity` ? It will be easier to search in the codebase.

app/models/journey.rb
martyn-w: @smoothcontract actually, it is not really that well defined at all... I did consider a more concrete implementation _a la_ the `Types::JSONB` approach. But I'm wary about being over prescriptive on the format of the vehicle record without more information from the suppliers. Allowing this to be any json seemed pragmatic for now.

app/serializers/journey_serializer.rb
martyn-w: oops, they aren't needed, I'll remove 👍

app/services/journeys/params_validator.rb
martyn-w: yes, perhaps, 🤔 
I will think about moving some of those `presence: true` validations back to the controller as `params.require`

app/services/journeys/params_validator.rb
martyn-w: the `nil` billable state in the validator in the `update` context means it is not provided, which means "don't update the billable flag"

the logic is:
* when creating a journey, `billable` *must* always be provided (only true or false)
* when updating a journey, `billable` *may* be provided. If it is provided it must only be true or false. If it is not provided, the validator will see its value as nil and ignore it.

The journeys controller will never allow a journey to be created or updated with billable=nil.

app/models/event.rb
smoothcontract: I don't mind `entity` at all - my comments were related to the overall name of the model (think I misunderstood @alexdesi original comment!). For polymorphic models, the normal Railsy convention is to somehow shoehorn `..able` on the end so that would make this `belongs_to :eventable, polymorphic: true`. Thoughts on that?

config/routes.rb
smoothcontract: In which case the list should be explicit I think

app/controllers/api/v1/journeys_controller.rb
smoothcontract: Cool, that makes sense - thanks for the explanation 👍

app/controllers/api/v1/journeys_controller.rb
smoothcontract: Ah, I didn't spot that extra use of this method. There is a work around you could use, although it might make Rubocop grumpy. If you memoized `new_journey` to the same `@journey` attribute then when `journey` is called later to build the event it would still add the event for the newly created journey.

```ruby
def new_journey
  @journey ||= Journey.new(new_journey_attributes)
end
```

That avoids the branching code, but it is a bit of a side effect I guess.

martyn-w: hmm... nope, I don't like `@journey` in new_journey as its not very clear what's going on. if its all the same, I think I'd prefer just to call the journey `journey` and have the if statement

app/services/journeys/params_validator.rb
smoothcontract: ..or perhaps you can make more assumptions in the controller about the list of params present knowing that the validator has passed. Fundamentally it would be good to define the list of required/permitted params in one place as much as possible.

app/models/journey.rb
martyn-w: actually, reflecting on this a bit more, I think you are right that vehicle should be an explicit attribute. At the time I envisaged lots of other metadata about the journey, so having one place to dump it all seemed sensible. But for now its probably just vehicle, so I'll change this 👍

app/controllers/api_controller.rb
martyn-w: @smoothcontract good spot re people_controller - I'll update that to throw an ActiveRecord::RecordNotFound as you suggest. In some tests I've found that the exception.id field is not always set, so I think an `if` statement is still required, e.g.

```
 if exception.id.present?
    "Couldn't find #{exception.model} with '#{exception.primary_key}'=#{exception.id}"
 else
    exception.to_s
 end
```

app/controllers/api/v1/journeys_controller.rb
smoothcontract: Cool 👍

app/models/event.rb
martyn-w: yep, happy with `eventable`, and I will also rename `entity` to `event_entity` to make searching for it easier

app/models/ability.rb
smoothcontract: Okay. I was slightly confused as to my knowledge we only have supplier accounts at present, so I'm not sure what there is that could break? The front end also uses an oauth token tied to a a supplier doesn't it? 🤔

martyn-w: the front end uses an api account which is not linked to a supplier and thus has these "super permissions"

app/services/journeys/params_validator.rb
smoothcontract: Okay, that's a good explanation - thanks this makes sense now 👍

app/models/journey.rb
smoothcontract: If the vehicle attributes might change over time then using a JSONB column might be a good choice - and gives the ability to wrap it in a typed class in future once we better understand the data structure. Having a dedicated `vehicle` attribute for this rather than merging it under `details` feels like a good design choice 🤞

spec/requests/api/v1/journeys_controller_index_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+            per_page: 20,
+            total_pages: 2,
+            total_objects: 21,
+            links: {
+              first: "/api/v1/moves/#{move.id}/journeys?page=1&per_page=#{per_page}",
+              last: "/api/v1/moves/#{move.id}/journeys?page=2&per_page=#{per_page}",
+              next: "/api/v1/moves/#{move.id}/journeys?page=2&per_page=#{per_page}",
+            },
+          )
+        end
```
smoothcontract, line 82: Appreciate this would be extra work, but there are a few request specs now that have the same pattern of pagination tests - perhaps worth adding a shared context for these, e.g. `it_behaves_like 'a paginated endpoint'`. See `allocations_controller_index_spec.rb`, `moves_controller_index_spec.rb`, `locations_controller_spec.rb`.

spec/requests/api/v1/journeys_controller_show_spec.rb
smoothcontract: I would imagine the cancancan authorisation checks would throw a 401 🤔

spec/services/journeys/params_validator_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+    it_behaves_like 'it validates timestamp'
+    it_behaves_like 'it validates from_location'
+    it_behaves_like 'it validates to_location'
+
+    # NB: there are no validations for vehicle
+  end
+
+  context 'when updating' do
+    let(:validation_context) { :update }
+    # NB: we cannot change a journey's location via update, so from_location and to_location should not be specified
```
smoothcontract, line 121: Is there a spec for this? What happens if `from_location` and `to_location` are specified for an update?

app/models/event.rb
smoothcontract: I think two things got confused here. I was suggesting leaving the model name as `Event`, with the `entity` attribute renamed to `eventable`.

martyn-w: ok, we are on the same page now: `entity_id` --> `eventable_id`, `entity_type` --> `eventable_type`

config/routes.rb
martyn-w: now fixed 👍

spec/requests/api/v1/journeys_controller_create_spec.rb
martyn-w: ok, I've changed this from 400 to 422 👍

spec/requests/api/v1/journeys_controller_create_spec.rb
martyn-w: ok, I've changed this from 400 to 422 👍

spec/requests/api/v1/journeys_controller_create_spec.rb
martyn-w: 👍 renamed to `when the move_id is not found`

app/models/ability.rb
smoothcontract: Interesting, I didn't know that the front end had a different type of auth! I'm glad I asked the question now 🤣

spec/requests/api/v1/journeys_controller_create_spec.rb
martyn-w: now returns 422 for missing referenced location 👍

spec/requests/api/v1/journeys_controller_update_spec.rb
martyn-w: so I'm going to leave a blank request `''` as 400 for now as I think that is a different sort of bad request to one where all the parameters are specified, but there is a mistake, e.g. `{"id": "123", "timestamp": "xxx"}`

spec/requests/api/v1/journeys_controller_show_spec.rb
martyn-w: now throws a 401 when unauthorised

app/controllers/api/v1/journeys_controller.rb
```diff
@@ -0,0 +1,124 @@
...
+        end
+      end
+
+      def create_event
+        # Logs the event for posterity and the immutable event log
+        Event.create!(
+          event_name: action_name,
+          eventable: journey,
+          client_timestamp: Time.zone.parse(params.dig(:data, :attributes, :timestamp)),
+          details: { data: data_params, supplier_id: current_user.owner.id },
```
smoothcontract, line 119: Does `owner` always get set for the current user? My understanding was that the owner would be a supplier linked to a specific oauth token, but if the front end also makes use of this API then that wouldn't have an associated supplier owner, and this could break 🤔 Maybe that's something we could guard against here?

spec/factories/events.rb
```diff
@@ -1,6 +1,6 @@
 FactoryBot.define do
   factory :event do
-    association(:move)
+    association(:entity)
```
smoothcontract, line 3: Should this be updated to `:eventable` ?

spec/requests/api/v1/journeys_controller_index_spec.rb
smoothcontract: Nicely done 👍

spec/requests/api/v1/journeys_controller_index_spec.rb
smoothcontract: This can be done in a future PR

swagger/v1/errors.yaml
```diff
@@ -56,10 +56,10 @@ UnsupportedMediaType:
 UnprocessableEntity:
   type: object
   required:
-  - code
-  - detail
-  - source
   - title
+  - detail
+#  - code
+#  - source
```
smoothcontract, line 62: Was this intentional? If so can commented lines be removed perhaps?

app/serializers/journey_serializer.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class JourneySerializer < ActiveModel::Serializer
+  attributes :id, :state, :billable, :vehicle
+  attribute :client_timestamp, key: :timestamp
+
+  has_one :from_location, serializer: LocationSerializer
```
smoothcontract, line 7: It doesn't seem to be necessary to explicitly state the serializer, this works just as well (see `AllocationSerializer` for example):

```ruby
has_one :from_location
has_one :to_location
```

app/controllers/api/v1/journeys_controller.rb
martyn-w: oops, you are quite right - good spot 👍 The front end shouldn't be creating any journeys; however, I'll update it so the FE can specify an optional supplier_id if required.

app/serializers/journey_serializer.rb
martyn-w: thanks, now fixed

spec/factories/events.rb
martyn-w: yes it should be eventable 😊  - now fixed

spec/requests/api/v1/journeys_controller_index_spec.rb
martyn-w: yep, I agree it would be good to tidy that up - will do as a separate PR

swagger/v1/errors.yaml
martyn-w: oops, they should have been deleted - now fixed

spec/requests/api/v1/journeys_controller_update_spec.rb
alexdesi: it's ok if you want to leave as it is, I don't think this is very important after all.
I am still not sure because 400 represents "The request could not be understood by the server due to malformed syntax" (https://stackoverflow.com/questions/16133923/400-vs-422-response-to-post-of-data)

app/controllers/api/v1/journeys_controller.rb
martyn-w: now fixed 👍

app/services/journeys/params_validator.rb
martyn-w: now updated to use require() in the controller 👍

spec/services/journeys/params_validator_spec.rb
martyn-w: adding a request spec to the journey update tests to verify you cannot change from/to location

PR #443, opened 2020-05-11
Log Nomis court hearing saving result

willfish:
### Jira link

P4-<TODO>

### What?

I have added/removed/altered:

- [x] Write an info log when a court hearing is saved/not saved in nomis
- [x] Remove sentry warning

### Why?

Reduce noisy and incorrect messaging when we do not save to Nomis.

### DIFF
--- a/app/controllers/api/v1/court_hearings_controller.rb
+++ b/app/controllers/api/v1/court_hearings_controller.rb
@@ -6,8 +6,9 @@ module Api
 
         if should_save_in_nomis?
           CourtHearings::CreateInNomis.call(move, move.court_hearings)
+          Rails.logger.info("Saved to nomis #{court_hearing.attributes.to_json}")
         else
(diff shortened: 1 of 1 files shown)


app/controllers/api/v1/court_hearings_controller.rb
```diff
@@ -41,9 +42,9 @@ def move
       end
 
       def should_save_in_nomis?
-        do_not_save_to_nomis = params['do_not_save_to_nomis'] == 'true'
+        save_to_nomis = params.fetch('do_not_save_to_nomis', 'false') != 'true'
```
martyn-w, line 45: just a sanity check: is the parameter name `do_not_save_to_nomis` or `save_to_nomis` ?

willfish: The query param is definitely `do_not_save_to_nomis`. I'm making the bit where we evaluate the business logic clearer as it was a bit garbled before

PR #444, opened 2020-05-11
[P4-1185] Updates default fallback question for building personal care need

willfish:
### Jira link

P4-1185

### What?

The initial implementation of the AssessmentAnswer documents for personal care needs was only for female prisons where the person might be pregnant.

This updates the default question to be more appropriate for a wider gamut of people we might move.

### Why?

This is principally a UX change that will make using our service make more sense.

### DIFF
--- a/spec/services/moves/import_people_spec.rb
+++ b/spec/services/moves/import_people_spec.rb
@@ -27,16 +27,16 @@ RSpec.describe Moves::ImportPeople do
      { prison_number: prisoner_two.nomis_prison_number, first_name: 'Bob' }]
   end
   let(:personal_care_needs_response) do
-    [{ offender_no: prisoner_one.nomis_prison_number, problem_type: 'MATSTAT', problem_code: 'ACCU9' },
-     { offender_no: prisoner_two.nomis_prison_number, problem_type: 'MATSTAT', problem_code: 'ACCU9' },
-     { offender_no: prisoner_two.nomis_prison_number, problem_type: 'MATSTAT', problem_code: 'ACCU4' }]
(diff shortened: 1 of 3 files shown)


spec/services/moves/import_people_spec.rb
```diff
@@ -27,16 +27,16 @@
      { prison_number: prisoner_two.nomis_prison_number, first_name: 'Bob' }]
   end
   let(:personal_care_needs_response) do
-    [{ offender_no: prisoner_one.nomis_prison_number, problem_type: 'MATSTAT', problem_code: 'ACCU9' },
-     { offender_no: prisoner_two.nomis_prison_number, problem_type: 'MATSTAT', problem_code: 'ACCU9' },
-     { offender_no: prisoner_two.nomis_prison_number, problem_type: 'MATSTAT', problem_code: 'ACCU4' }]
+    [{ offender_no: prisoner_one.nomis_prison_number, problem_type: 'FOO', problem_code: 'AA' },
```
willfish, line 30: These keys plumb together the correct fallback. Previously the key returned for the problem code happened to be pregnancy by accident.

alexdesi: it's not clear from the test why this is returning `FOO` as problem_type,
does it means that it is a problem type we are not handling?

alexdesi: maybe you want just to add a context to make it clear:

context "when the response has problem_type which ...." do

end
or anything that can clarify that `problem_type: 'FOO'`

willfish: Yeah, I use foo because internally this value isn't used to control the behaviour of anything

willfish: I can probably just remove the problem type attribute

PR #447, opened 2020-05-12
P4-1398 Surface associated moves in allocations

cwrw:
### Jira link

[P4-1398](https://dsdmoj.atlassian.net/browse/P4-1398)

### What?

- [x] Added moves `relationship` and `includes` in allocation serializer

### Why?

We would like to surface moves in the allocation endpoints. Moves should also surface people in `includes` if there are any associated with them.

There are two issues I'd like to address in the current implementation:
1) We are including a circular dependency in the `moves` include response where it will also surface the allocation in the relationship.
2) Inferring moves count rather than including an extra attribute

### Have you? (optional)

### DIFF
--- a/app/serializers/allocation_serializer.rb
+++ b/app/serializers/allocation_serializer.rb
@@ -5,6 +5,11 @@ class AllocationSerializer < ActiveModel::Serializer
 
   has_one :from_location
   has_one :to_location
+  has_many :moves
 
-  INCLUDED_ATTRIBUTES = %i[from_location to_location].freeze
+  INCLUDED_ATTRIBUTES = {
+    from_location: [],
+    to_location: [],
+    moves: %i[person],
+  }.freeze
 end
--- a/spec/requests/api/v1/allocations_controller_create_spec.rb
+++ b/spec/requests/api/v1/allocations_controller_create_spec.rb
@@ -7,7 +7,10 @@ RSpec.describe Api::V1::AllocationsController do
 
   let(:response_json) { JSON.parse(response.body) }
   let(:resource_to_json) do
-    JSON.parse(ActionController::Base.render(json: allocation, include: AllocationSerializer::INCLUDED_ATTRIBUTES))
+    resource = JSON.parse(ActionController::Base.render(json: allocation, include: AllocationSerializer::INCLUDED_ATTRIBUTES))
+    resource['data']['relationships']['moves']['data'] = UnorderedArray(*resource.dig('data', 'relationships', 'moves', 'data'))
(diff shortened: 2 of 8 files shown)


app/serializers/allocation_serializer.rb
```diff
@@ -5,6 +5,11 @@ class AllocationSerializer < ActiveModel::Serializer
 
   has_one :from_location
   has_one :to_location
+  has_many :moves
```
cwrw, line 8: When we include moves in allocations, we are currently introducing circular nested dependencies, this is because we also have an association in the `MoveSerializer` :
https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/serializers/move_serializer.rb#L13
This means that any move in the `included` section will have allocation defined in the relationships, even though we are seeing it under the allocation endpoint itself. This seems to be in conflict with the JSON:API specs as from the examples I've seen we don't usually return the relationship again if a model is included.

We currently have precedence of this in the codebase (around court hearings I believe). I could introduce either separate serialisers for moves to avoid this (inherit from a base and not include the `belongs_to` etc.), or keep it as is for now and add a ticket to tackle inconsistencies in the JSON:API spec epic. I have chosen the latter but would like to hear people's thoughts on this cc. @teneightfive

cwrw:
When creating an allocation, we are leveraging the `moves_count` attribute to create associated moves for an allocation, with also populating this field itself in an allocation. We would like to in the future infer the number of moves from the moves created rather than have a specific attribute on an allocation. However according to JSON:API specs, we should be always passing in the relationships for moves to the create allocation endpoint, so we will have to move away from `moves_count` completely and work with the FE team to send a different body for creating the associated moves. I think it's best to tackle this in a different ticket cc @teneightfive

spec/requests/api/v1/allocations_controller_create_spec.rb
```diff
@@ -7,7 +7,10 @@
 
   let(:response_json) { JSON.parse(response.body) }
   let(:resource_to_json) do
-    JSON.parse(ActionController::Base.render(json: allocation, include: AllocationSerializer::INCLUDED_ATTRIBUTES))
+    resource = JSON.parse(ActionController::Base.render(json: allocation, include: AllocationSerializer::INCLUDED_ATTRIBUTES))
+    resource['data']['relationships']['moves']['data'] = UnorderedArray(*resource.dig('data', 'relationships', 'moves', 'data'))
```
cwrw, line 11: pretty ugly but unfortunately comparing nested arrays in JSON is failing due to ordering in the array itself, would appreciate thoughts on a better way in doing this

PR #450, opened 2020-05-12
Fixes patching empty documents on move

willfish:
### Jira link

P4-<TODO>

### What?

Handle empty documents case for patching a moves documents.

### Why?

This is definitely something the frontend will need to do and was an oversight on our part.

### DIFF
--- a/app/controllers/api/v1/moves_controller.rb
+++ b/app/controllers/api/v1/moves_controller.rb
@@ -92,7 +92,7 @@ module Api
       end
 
       # 1. Frontend specifies empty docs: update documents to be empty
-      # 2. Frontend specifies no doc changes: don't update documents at all
+      # 2. Frontend does not include document relationship: don't update documents at all
       def patch_move_attributes
         attributes = patch_move_params.fetch(:attributes, {})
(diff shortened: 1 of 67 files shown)


app/controllers/api/v1/moves_controller.rb
```diff
@@ -90,13 +91,15 @@ def move_attributes
         )
       end
 
+      # 1. Frontend specifies empty docs: update documents to be empty
+      # 2. Frontend specifies no doc changes: don't update documents at all
```
willfish, line 95:
```suggestion
      # 2. Frontend does not include document relationship: don't update documents at all
```

spec/requests/api/v1/moves_controller_update_spec.rb
```diff
@@ -190,6 +190,36 @@
...
+
+          context 'when documents is nil' do
+            let(:move_params) do
+              {
+                type: 'moves',
+                relationships: { documents: { data: nil } },
+              }
+            end
+
+            it 'removes the documents from the move' do
```
willfish, line 217:
```suggestion
            it 'does not remove documents from the move' do
```

PR #451, opened 2020-05-13
P4-1469 new cancellation reason

martyn-w:
### Jira link

[P4-1469](https://dsdmoj.atlassian.net/browse/P4-1469)

### What?

I have added/removed/altered:

- [X] Added new cancellation reason 'rejected'
- [X] Add new validations on cancellation reason
- [X] Tidied up moves controller and moves model, renaming a few methods for claity

### Why?

I am doing this because:

- new reasons are required by the FE

### Have you? (optional)

### DIFF
--- a/app/controllers/api/v1/moves_controller.rb
+++ b/app/controllers/api/v1/moves_controller.rb
@@ -86,7 +86,7 @@ module Api
           documents: Document.where(id: (new_move_params.dig(:relationships, :documents, :data) || []).map { |doc| doc[:id] }),
           court_hearings: CourtHearing.where(id: (new_move_params.dig(:relationships, :court_hearings, :data) || []).map { |court_hearing| court_hearing[:id] }),
           prison_transfer_reason: PrisonTransferReason.find_by(id: new_move_params.dig(:relationships, :prison_transfer_reason, :data, :id)),
-            )
+        )
       end
 
       def update_move_params
--- a/app/controllers/api/v1/moves_controller.rb
+++ b/app/controllers/api/v1/moves_controller.rb
@@ -3,45 +3,44 @@
 module Api
   module V1
     class MovesController < ApiController
+      before_action :validate_filter_params, only: %i[index]
+
       def index
-        moves_params = Moves::ParamsValidator.new(filter_params, params[:sort] || {})
-        if moves_params.valid?
-          moves = Moves::Finder.new(filter_params, current_ability, params[:sort] || {}).call
-          # Excludes potentially many court hearing documents to reduce the request size. This was requested specifically by the frontend team.
-          paginate moves, include: MoveSerializer::INCLUDED_ATTRIBUTES.dup.except(:court_hearings), fields: MoveSerializer::INCLUDED_FIELDS
-        else
-          render json: { error: moves_params.errors }, status: :bad_request
-        end
+        moves = Moves::Finder.new(filter_params, current_ability, params[:sort] || {}).call
+        # Excludes potentially many court hearing documents to reduce the request size. This was requested specifically by the frontend team.
+        paginate moves, include: MoveSerializer::INCLUDED_ATTRIBUTES.dup.except(:court_hearings), fields: MoveSerializer::INCLUDED_FIELDS
       end
 
       def show
-        render_move(move, 200)
+        render_move(move, :ok)
       end
 
       def create
-        move = Move.new(move_attributes)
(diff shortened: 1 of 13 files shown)


app/controllers/api/v1/moves_controller.rb
```diff
@@ -3,46 +3,44 @@
...
+      before_action :validate_filter_params, only: %i[index]
+
       def index
-        moves_params = Moves::ParamsValidator.new(filter_params, params[:sort] || {})
-        if moves_params.valid?
-          moves = Moves::Finder.new(filter_params, current_ability, params[:sort] || {}).call
-          # Excludes potentially many court hearing documents to reduce the request size. This was requested specifically by the frontend team.
-          paginate moves, include: MoveSerializer::INCLUDED_ATTRIBUTES.dup.except(:court_hearings), fields: MoveSerializer::INCLUDED_FIELDS
-        else
-          render json: { error: moves_params.errors }, status: :bad_request
```
cwrw, line 13: with this change we are moving away from `:bad_request` 400 errors on validation and instead returning a 422, since we seem to be following this pattern in other controllers I was wondering why it was different here?

cwrw: Although I guess 422 is probably the correct error to be displaying for validations

lib/tasks/data_maintenance.rake
```diff
@@ -24,7 +24,7 @@ namespace :data_maintenance do
     Location.prisons.each do |prison|
       from_moves = prison.moves_from.requested.select(&:from_nomis?).each do |move|
         move.update!(status: Move::MOVE_STATUS_CANCELLED,
-                     cancellation_reason: Move::MOVE_CANCELLATION_REASON_MADE_IN_ERROR)
+                     cancellation_reason: Move::cancellation_reasons[:made_in_error])
```
cwrw, line 27: I think we might need to use `CANCELLATION_REASON_MADE_IN_ERROR` here instead as we are no longer using the hash?

martyn-w: oops, good spot 👍

app/controllers/api/v1/moves_controller.rb
martyn-w: so yes, starting with the journeys controller we're moving towards validation errors being 422 rather than 400 - cc @alexdesi 
we can look at other controllers as time allows

app/models/move.rb
```diff
@@ -100,6 +93,26 @@ def current?
       (self.date.nil? && self.date_to.nil? && self.date_from.present? && self.date_from >= Time.zone.today)
   end
 
+  def prison_recall?
+    move_type == Move::move_types[:prison_recall]
+  end
+
+  def proposed?
```
cwrw, line 100: I think `enum` might give us those for free? https://api.rubyonrails.org/v5.2.3/classes/ActiveRecord/Enum.html

martyn-w: yes it does, thanks for spotting this 👍

app/models/move.rb
```diff
@@ -46,27 +45,21 @@ class Move < VersionedModel
...
-            unless: -> { [MOVE_STATUS_PROPOSED, MOVE_STATUS_CANCELLED].include?(status) || person_id.blank? }
-  validates :date, presence: true,
-            unless: -> { status == MOVE_STATUS_PROPOSED }
-  validates :date_from, presence: true,
-            if: -> { status == MOVE_STATUS_PROPOSED }
-
-  validate :date_to_after_date_from
-
+            unless: -> { proposed? || cancelled? || person_id.blank? }
+  validates :date, presence: true, unless: -> { proposed? || cancelled? }
```
cwrw, line 57: could you add a test for the `cancelled` validation here?

spec/requests/api/v1/moves_controller_update_spec.rb
```diff
@@ -113,11 +113,11 @@
         end
 
         it 'updates the cancellation_reason of a move' do
```
cwrw, line 115: can we change the description to `does not update`?

app/models/move.rb
martyn-w: will do 👍

spec/requests/api/v1/moves_controller_update_spec.rb
martyn-w: now fixed

PR #455, opened 2020-05-13
P4-1428 Replicate review app database from main heroku API app

smoothcontract:
### Jira link

P4-1428

### What?

I have added/removed/altered:

- [x] Added extra configuration setting placeholders in `app.json` so that these are populated for new review apps deployed to Heroku
- [x] Configured values within Heroku on the main app and for the pipeline for each of the new config settings (mostly around NOMIS - these point to the T3 environment)
- [x] Populated `MASTER_DATABASE_URL` in the pipeline so that review apps know where the database is located for the main app
- [x] Added a post deploy script to replicate the database content from the main app when a review app is deployed
- [x] Populated all of the relevant reference data on the main Heroku review app (via running all of the rake tasks)
- [x] Added a new combined rake task to setup all of the reference data at once `rails reference_data:create_all` (makes it easier to setup a new environment)

### Why?

- Heroku main API app now works as expected and contains reference data so the API endpoints work. Authentication is currently disabled on the Heroku review apps but enabled for the main app. This protects the known good "reference" data - we need to keep that clean and lightweight and use the disposable copies on review apps for testing.

- In the near future I'm hoping that we'll be able to run the end to end test suite against a deployed Heroku review app and get that passing. This would then allow stronger testing of any API changes to ensure that no regressions are introduced.

### DIFF
--- a/app.json
+++ b/app.json
@@ -3,29 +3,61 @@
   "description": "",
   "repository": "https://github.com/ministryofjustice/hmpps-book-secure-move-api",
   "env": {
+	  "MASTER_DATABASE_URL": {
+      "description": "DATABASE_URL value from main API application deployed on Heroku",
+	    "required": true
+	  },
     "ENCRYPTOR_SALT": {
       "description": "Secret salt value for encryption/decryption",
-      "required": "true"
+      "required": true
     },
     "S3_ACCESS_KEY_ID": {
       "description": "Public access key for AWS S3 storage",
-      "required": "true"
+      "required": true
     },
     "S3_SECRET_ACCESS_KEY": {
       "description": "Private secrete access key for AWS S3 storage",
-      "required": "true"
+      "required": true
     },
     "S3_BUCKET_NAME": {
       "description": "Name of the storage bucket to use on AWS S3",
-      "required": "true"
+      "required": true
+    },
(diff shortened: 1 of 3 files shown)


lib/tasks/reference_data.rake
```diff
@@ -93,6 +93,21 @@ namespace :reference_data do
     puts_summary_of_relationships
   end
 
+  desc 'create all of the necessary reference data'
```
smoothcontract, line 96: This was just added as a convenience to make it easier to populate all of the reference data in the correct order with a single command.

martyn-w: could you update README.md lines 92-100 to use this task (`create_all`) instead

app.json
```diff
@@ -3,30 +3,62 @@
...
+      "value": true
     },
     "HEROKU_DISABLE_AUTH": {
       "description": "Disable authentication on Heroku review apps if true",
-      "required": "true"
+      "required": true
     }
   },
+  "scripts": {
+    "postdeploy": "pg_dump $MASTER_DATABASE_URL | psql $DATABASE_URL && bundle exec rails db:migrate"
```
martyn-w, line 60: good idea to copy the database 👍

PR #458, opened 2020-05-14
P4-1530 Allow person to be updated on a move

cwrw:
jira-ticket: https://dsdmoj.atlassian.net/browse/P4-1530

### Jira link

[P4-1530](https://dsdmoj.atlassian.net/browse/P4-1530)

### What?

* Move update logic to service
* Allow a person to be updated on a move
* If no person relationship is set on a move do not update the association
* If nil data is set on a move param remove any associated person

### Why?

To allow a person to be associated to a move on the allocation workflow, amend the PATCH endpoint to accept a person relationship to link/unlink people to that move

### Have you? (optional)

### DIFF
--- a/spec/requests/api/v1/moves_controller_update_spec.rb
+++ b/spec/requests/api/v1/moves_controller_update_spec.rb
@@ -222,6 +222,66 @@ RSpec.describe Api::V1::MovesController do
           end
         end
 
+        context 'when changing a moves person' do
+          let(:after_person) { create(:person) }
+          let(:move_params) do
+            {
+              type: 'moves',
+              attributes: {
+                status: 'requested',
+              },
+              relationships: { person: { data: { id: after_person.id, type: 'people' } } },
+            }
+          end
+
+          it 'updates the moves person' do
+            expect(move.reload.person).to eq(after_person)
+          end
+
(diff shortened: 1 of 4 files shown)


martyn-w:
is it definitely `person_id` that should be specified in the relationships ?

we are intending to remove the relationship between `person <--> move` soon and replace it with `profile <--> move`. If the FE has the profile_id already, it might simplify further work by passing that now, instead of `person_id` ?

martyn-w:
Otherwise, looks good 👍  - but needs tests in the request spec (`spec/requests/api/v1/moves_controller_update_spec.rb`) to confirm behaviour when a person/profile id is (a) passed, (b) not passed, and (c) when nil.

cwrw:
Just double checked with FE team and migration work has not been started to move to profile, so I think it's ok to keep person for now. Also added the tests for controller spec for the cases

spec/requests/api/v1/moves_controller_update_spec.rb
```diff
@@ -222,6 +222,66 @@
           end
         end
 
+        context 'when changing a moves person' do
```
martyn-w, line 225: good tests 👍

PR #459, opened 2020-05-15
P4-1441 Cancel allocations

cwrw:
### Jira link

[P4-1441](https://dsdmoj.atlassian.net/browse/P4-1441)

### What?

- Add a status to an allocation. An allocation can have 3 statuses: filled, unfilled and cancelled. We will now only implement cancelled
- Add events to allocation to allow for setting the status
- Add cancellation reason and comment to an allocation 
- Add allocation events create path
### Why?

To allow the cancellation of an allocation and the associated moves, add a new event endpoint which when set as cancelled will set the new "status" of allocation to cancelled as well as all moves.

The event endpoint is returning a fake response as it is still to be implemented in further work. 
I initially implemented cancellation reason but then it got de-scoped later on, so I kept the implementation to avoid having to redo this.

### Have you? (optional)

- Changes validation on cancellation reason move (to check with FE that empty cancellation reasons is ok)

### DIFF
--- a/app/controllers/api/v1/allocation_events_controller.rb
+++ b/app/controllers/api/v1/allocation_events_controller.rb
@@ -3,6 +3,7 @@
 module Api
   module V1
     class AllocationEventsController < ApiController
+      before_action :validate_params, only: :create
       after_action :send_move_notifications, only: :create
 
       def create
@@ -25,6 +26,10 @@ module Api
         :type,
         attributes: %i[timestamp event_name],
       ].freeze
+
+      def validate_params
+        AllocationEvents::ParamsValidator.new(event_params).validate!(action_name.to_sym)
+      end
 
       def event_params
         params.require(:data).permit(PERMITTED_EVENT_PARAMS).to_h
--- a/app/controllers/api/v1/allocation_events_controller.rb
+++ b/app/controllers/api/v1/allocation_events_controller.rb
@@ -1,0 +1,74 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class AllocationEventsController < ApiController
+      after_action :send_move_notifications, only: :create
+
+      def create
+        case event_name
+        when 'cancel'
+          allocation.cancel
+
+          allocation.save!
+          render json: fake_event_object, status: :created
+        else
+          render status: :bad_request, json: {
+            errors: [{ title: 'invalid event_name', detail: "#{event_name} is not supported" }],
+          }
+        end
+      end
+
+    private
+
+      PERMITTED_EVENT_PARAMS = [
+        :type,
+        attributes: %i[timestamp event_name],
+      ].freeze
+
+      def event_params
+        params.require(:data).permit(PERMITTED_EVENT_PARAMS).to_h
+      end
+
+      def allocation
+        @allocation ||= Allocation.find(params.require(:allocation_id))
+      end
+
+      def send_move_notifications
+        allocation.moves.each do |move|
+          Notifier.prepare_notifications(topic: move, action_name: 'update_status')
+        end
+      end
+
+      def event_name
+        @event_name ||= event_params.dig(:attributes, :event_name)
+      end
+
+      def timestamp
+        @timestamp ||= Time.zone.parse(event_params.dig(:attributes, :timestamp))
+      end
+
+      def fake_event_object
+        # NB: this is a temporarily implementation to simulate the creation of a new event
+        {
+          data: {
+            id: SecureRandom.uuid,
+            type: 'events',
+            attributes: {
+              event_name: event_name,
+              timestamp: timestamp,
+            },
+            relationships: {
+              allocation: {
+                data: {
+                  type: 'allocations',
+                  id: allocation.id,
+                },
+              },
+            },
+          },
+        }
+      end
+    end
+  end
+end
--- a/app/models/allocation.rb
+++ b/app/models/allocation.rb
@@ -23,6 +23,14 @@ class Allocation < VersionedModel
     cancelled: ALLOCATION_STATUS_CANCELLED,
   }
 
+  CANCELLATION_REASONS = [
+    MADE_IN_ERROR = 'made_in_error',
+    SUPPLIER_DECLINED_TO_MOVE = 'supplier_declined_to_move',
+    LACK_OF_SPACE = 'lack_of_space_at_receiving_establishment',
+    FAILED_TO_FILL_ALLOCATION = 'sending_establishment_failed_to_fill_allocation',
+    OTHER = 'other',
+  ].freeze
+
   belongs_to :from_location, class_name: 'Location'
   belongs_to :to_location, class_name: 'Location'
   has_many :moves, inverse_of: :allocation, dependent: :destroy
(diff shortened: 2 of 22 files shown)


app/models/allocation.rb
```diff
@@ -12,9 +16,25 @@ class Allocation < VersionedModel
...
+    SUPPLIER_DECLINED_TO_MOVE = 'supplier_declined_to_move',
+    LACK_OF_SPACE = 'lack_of_space_at_receiving_establishment',
+    FAILED_TO_FILL_ALLOCATION = 'sending_establishment_failed_to_fill_allocation',
+    OTHER = 'other',
+  ].freeze
+
   belongs_to :from_location, class_name: 'Location'
   belongs_to :to_location, class_name: 'Location'
-  has_many :moves, inverse_of: :allocation, dependent: :destroy
+  has_many :moves, inverse_of: :allocation, dependent: :destroy, autosave: true
```
cwrw, line 36: need to set autosave to allow updated to moved to be saved, as only create is set by default

app/controllers/api/v1/allocation_events_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+          }
+        end
+      end
+
+    private
+
+      PERMITTED_EVENT_PARAMS = [
+        :type,
+        attributes: %i[timestamp event_name],
+        relationships: { allocation: {} },
```
cwrw, line 27: remove relationship as we no longer have to set this

app/controllers/api/v1/allocation_events_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+      end
+
+      def allocation
+        @allocation ||= Allocation.find(params.require(:allocation_id))
+      end
+
+      def send_move_notifications
+        allocation.moves.each do |move|
+          Notifier.prepare_notifications(topic: move, action_name: 'update_status')
+        end
```
cwrw, line 41: on error test that we don't send notifications

swagger/v1/allocation_event_name_attribute.yaml
```diff
@@ -0,0 +1,6 @@
+AllocationEventName:
+  type: string
+  example: redirect
```
cwrw, line 3: change to cancel

swagger/v1/post_allocation_events_responses.yaml
```diff
@@ -0,0 +1,19 @@
+'201':
```
cwrw, line 1: add `PostAllocationEvent`

app/models/allocation.rb
```diff
@@ -25,5 +45,16 @@ class Allocation < VersionedModel
   validates :moves_count, presence: true, numericality: { only_integer: true, greater_than: 0 }
   validates :date, presence: true
 
+  # TODO: re-enable once cancellation reason is implemented in FE
```
cwrw, line 48: Cancellation reason and comment was de-scoped after I already implemented it, so I thought it might be better to keep here until the frontend team starts implementing it soon

spec/models/allocation_spec.rb
```diff
@@ -27,4 +31,56 @@
       expect(allocation.versions.map(&:event)).to eq(%w[create])
     end
   end
+
+  xdescribe 'cancellation_reason' do
```
cwrw, line 35: commenting out until this is back in scope

spec/models/allocation_spec.rb
```diff
@@ -27,4 +31,56 @@
...
+      expect(allocation.reload.moves.pluck(:status)).to contain_exactly(Move::MOVE_STATUS_CANCELLED)
+    end
+
+    it 'throws validation error if allocation invalid' do
+      allocation.from_location = nil
+
+      expect { allocation.cancel }.to raise_error(ActiveRecord::RecordInvalid)
+    end
+
+    it 'does not update moves if allocation invalid' do
```
cwrw, line 79: add test for invalid move

app/models/allocation.rb
```diff
@@ -25,5 +45,16 @@ class Allocation < VersionedModel
...
+  # validates :cancellation_reason, inclusion: { in: CANCELLATION_REASONS }, if: :cancelled?
+  # validates :cancellation_reason, absence: true, unless: :cancelled?
+
   attribute :complex_cases, Types::JSONB.new(Allocation::ComplexCaseAnswers)
+
+  def cancel
+    self.status = ALLOCATION_STATUS_CANCELLED
+    self.moves.each { |move| move.assign_attributes(status: Move::MOVE_STATUS_CANCELLED) }
```
martyn-w, line 56: cancelled moves _must_ have a `cancellation_reason` (otherwise validations will fail), which must be one of these values:
```
    CANCELLATION_REASON_MADE_IN_ERROR = 'made_in_error',
    CANCELLATION_REASON_SUPPLIER_DECLINED_TO_MOVE = 'supplier_declined_to_move',
    CANCELLATION_REASON_REJECTED = 'rejected',
    CANCELLATION_REASON_OTHER = 'other'
```

app/models/allocation.rb
martyn-w: this change may break other FE features, which now require and expect a cancellation reason

app/models/allocation.rb
```diff
@@ -12,9 +16,25 @@ class Allocation < VersionedModel
     long: 'more_than_16',
   }
 
+  # TODO: implement statemachine, for now just allow an allocation to be cancelled
```
martyn-w, line 19: 👍

db/migrate/20200514052159_add_status_to_allocations.rb
```diff
@@ -0,0 +1,5 @@
+class AddStatusToAllocations < ActiveRecord::Migration[5.2]
+  def change
+    add_column :allocations, :status, :string
```
martyn-w, line 3: probably should have `null: false, index: true`

cwrw: we are allowing `nil` allocations for now as we only set the cancelled flag, we need to iron out the other statuses post MVP

martyn-w: ok for null status for MVP 👍

martyn-w:
👍 this looks good. We've discussed the question of validation of reasons/comments and agreed with default values for now

app/models/move.rb
```diff
@@ -70,6 +72,14 @@ class Move < VersionedModel
   scope :served_by, ->(supplier_id) { where('from_location_id IN (?)', Location.supplier(supplier_id).pluck(:id)) }
   scope :not_cancelled, -> { where.not(status: MOVE_STATUS_CANCELLED) }
 
+  def cancel(reason: CANCELLATION_REASON_OTHER, comment: nil)
```
cwrw, line 75: might need to rename after `statemachine` implemented

app/controllers/api/v1/allocation_events_controller.rb
```diff
@@ -0,0 +1,74 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class AllocationEventsController < ApiController
```
martyn-w, line 5: minor (non-blocking) point: it would be best to validate the parameters (especially that `timestamp` is in an ISO-8601 format), to prevent taking liberties with date formats. See e.g.  `app/services/journeys/params_validator.rb`

spec/models/allocation_spec.rb
```diff
@@ -27,4 +31,80 @@
...
+      allocation.from_location = nil
+      allocation.cancel
+    rescue ActiveRecord::RecordInvalid
```
martyn-w, line 106: this test is dangerous: if `allocation.cancel` does not raise an exception, then there is no expectation and this will will pass. 

Instead test that an exception is raised, and also that the record has the correct state.

martyn-w: oh, sorry, I overlooked the test above which is doing that. instead, allow allocation to raise and catch, and put the expectation in the main block:

```
...
allocation.cancel rescue ActiveRecord::RecordInvalid
expect(foo).to eql (bla)
```

PR #460, opened 2020-05-15
P4-1384 GET /reference/regions API endpoint

smoothcontract:
### Jira link

P4-1384

### What?

- [x] New route, controller and serializer for GET /reference/regions
- [x] Rearranged routes in alphabetical order (makes it slightly easier to find them in the code)
- [x] Improved error reporting for invalid Swagger YAML files (now includes relevant filename to make tracking down issues easier)
- [x] Added specs and new region factory; refactored some existing specs to use factory
- [x] Updated swagger documentation

### Why?

- The front end work for allocations will soon allow user to choose a region to work with; this provides a list of locations that can then be used to scope the visible locations for the user, as well as filter the relevant list of prisons when creating an allocation.

### Have you? (optional)

- This is a new api endpoint, so no risk to production

### DIFF
--- a/app/controllers/api/v1/reference/regions_controller.rb
+++ b/app/controllers/api/v1/reference/regions_controller.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    module Reference
+      class RegionsController < ApiController
+        def index
+          render json: Region.all.includes(:locations)
+        end
+      end
+    end
+  end
+end
--- a/app/serializers/region_serializer.rb
+++ b/app/serializers/region_serializer.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class RegionSerializer < ActiveModel::Serializer
+  attributes :key, :name, :created_at, :updated_at
+
+  has_many :locations
+end
--- a/spec/requests/api/v1/reference/regions_controller_spec.rb
+++ b/spec/requests/api/v1/reference/regions_controller_spec.rb
@@ -1,0 +1,83 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Api::V1::Reference::RegionsController do
+  let!(:access_token) { create(:access_token).token }
+  let(:response_json) { JSON.parse(response.body) }
+  let(:content_type) { ApiController::CONTENT_TYPE }
+  let(:headers) { { 'CONTENT_TYPE': content_type }.merge('Authorization' => "Bearer #{access_token}") }
+
+  describe 'GET /api/v1/reference/regions' do
+    let(:schema) { load_yaml_schema('get_regions_responses.yaml') }
+
+    let(:location) { create(:location) }
+    let!(:region1) { create(:region, locations: [location]) }
+    let!(:region2) { create(:region, locations: [location]) }
+
+    let(:data) do
+      [
+        {
+          type: 'regions',
+          attributes: {
+            key: region1.key,
+            name: region1.name,
+          },
+          relationships: {
+            locations: {
+              data: [
+                {
+                  id: location.id,
+                  type: 'locations',
+                },
+              ],
+            },
+          },
+        },
+        {
+          type: 'regions',
+          attributes: {
+            key: region2.key,
+            name: region2.name,
+          },
+          relationships: {
+            locations: {
+              data: [
+                {
+                  id: location.id,
+                  type: 'locations',
+                },
+              ],
+            },
+          },
+        },
+      ]
+    end
+
+    before do
+      get '/api/v1/reference/regions', headers: headers
+    end
+
+    context 'when successful' do
+      it_behaves_like 'an endpoint that responds with success 200'
+
+      it 'returns the correct data' do
+        expect(response_json).to include_json(data: data)
+      end
+    end
+
+    context 'when not authorized', :with_invalid_auth_headers do
+      let(:headers) { { 'CONTENT_TYPE': content_type }.merge(auth_headers) }
+      let(:content_type) { ApiController::CONTENT_TYPE }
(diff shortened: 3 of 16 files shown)


app/controllers/api/v1/reference/regions_controller.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
```
cwrw, line 1: What is the reasoning behind the `reference` namespace as opposed to directly using `regions` etc., as I can see we are using it for `locations` as well?

app/controllers/api/v1/reference/regions_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+
+module Api
+  module V1
+    module Reference
+      class RegionsController < ApiController
+        def index
+          render json: Region.all.includes(:locations)
```
cwrw, line 8: do we need to paginate this as well?

app/serializers/region_serializer.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class RegionSerializer < ActiveModel::Serializer
+  attributes :key, :name, :created_at, :updated_at
+
+  has_many :locations
```
cwrw, line 6: Do we require the full location itself in the `included` section? As in to add something around the lines of `INCLUDED_ATTRIBUTES` and `include` in the controller? Or will we get that list in a different way maybe from a the single region?

spec/requests/api/v1/reference/regions_controller_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+  let(:response_json) { JSON.parse(response.body) }
+  let(:content_type) { ApiController::CONTENT_TYPE }
+  let(:headers) { { 'CONTENT_TYPE': content_type }.merge('Authorization' => "Bearer #{access_token}") }
+
+  describe 'GET /api/v1/reference/regions' do
+    let(:schema) { load_yaml_schema('get_regions_responses.yaml') }
+
+    let(:location) { create(:location) }
+    let!(:region1) { create(:region, locations: [location]) }
+    let!(:region2) { create(:region, locations: [location]) }
```
cwrw, line 16: can you add an example test that utilises the `has_and_belongs_to_many` relationship and how that looks in the response? (the 5a/5b example from the prototype)?

spec/support/json_schemas.rb
```diff
@@ -17,12 +16,17 @@ def load_json_schema(file_name)
 
   def load_yaml_schema(file_name)
     File.open("#{Rails.root}/swagger/v1/#{file_name}") do |file|
-      YAML.safe_load(file.read)
+      begin
+        YAML.safe_load(file.read)
+      rescue Psych::SyntaxError => e
+        # Include original filename in exception to make debugging a less cryptic affair
+        raise Psych::SyntaxError.new(file_name, e.line, e.column, e.offset, e.problem, e.context)
```
cwrw, line 23: nice 👍

app/controllers/api/v1/reference/regions_controller.rb
smoothcontract: It's just a convention - all of the reference data endpoints live in this module.

app/controllers/api/v1/reference/regions_controller.rb
smoothcontract: I don't think it's necessary, there are only 9 regions in total - and that's supposedly going to reduce to two in future. If instead it increases beyond 20 we can look at adding pagination then.

app/serializers/region_serializer.rb
smoothcontract: This will include only location id's as references - the front end already has full location details available via the `/reference/locations` endpoint and my understanding is that only the id's are needed to build region filtering against the `/allocations` endpoint. In future we can look at better supporting `includes` attribute to make this more flexible but that feels outside the scope of this PR/story.

spec/requests/api/v1/reference/regions_controller_spec.rb
smoothcontract: Hmm.. perhaps. Let me think about how best to do that. I've kept this as simple as possible with a single location so that I don't have to get into the messiness of ordering within relations but maybe it's something that could be covered with a specific test or two.

spec/serializers/region_serializer_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    context 'with locations' do
+      let(:location) { create(:location) }
+      let(:region) { create(:region, locations: [location]) }
+
+      it 'contains a locations relationship' do
+        expect(result_data[:relationships][:locations][:data]).to contain_exactly(id: location.id, type: 'locations')
+      end
+
+      it 'does not contain an included location' do
+        expect(result[:included]).to be_nil
```
smoothcontract, line 49: As per comment previously, we don't include the locations in the serializer, only references to them - so only the type `locations` and `id` attributes get included with the relation.

spec/support/json_schemas.rb
smoothcontract: This bugged me enough to fix it, spent too long scratching my head trying to figure out which file was causing issues.

spec/requests/api/v1/reference/regions_controller_spec.rb
smoothcontract: Added another test for this making ubiquitous use of `UnorderedArray` so shouldn't be subject to order based flakiness

app/controllers/api/v1/reference/regions_controller.rb
cwrw: spoke in person: we use `reference` for resources where data is static and won't be amended by any API calls

PR #463, opened 2020-05-18
Clean up rake tasks reference_data

alexdesi:
### What?
- Remove the long noisy output of rake `link_suppliers`
- Add some useful output to check the rake progression in `create_all`

### DIFF
--- a/lib/tasks/reference_data.rake
+++ b/lib/tasks/reference_data.rake
@@ -89,41 +89,29 @@ namespace :reference_data do
         puts "Locations for the Supplier '#{supplier.name}' have not changed."
       end
     end
-
-    puts_summary_of_relationships
   end
 
   desc 'create all of the necessary reference data'
   task create_all: :environment do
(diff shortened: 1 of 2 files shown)


lib/tasks/reference_data.rake
```diff
@@ -89,41 +89,29 @@ namespace :reference_data do
...
+    end
   end
 
 private
 
   def have_locations_changed?(locations1, locations2)
     ((locations1 - locations2) + (locations2 - locations1)).any?
   end
-
-  def puts_summary_of_relationships
```
martyn-w, line 117: I think this summary is still quite useful - perhaps move it to its own rake task ?

alexdesi: It might be, but I think it's not difficult to type `Supplier.find_by(key: 'serco').locations` in the rails console.
In case we miss that I'll happy to create a rake for it.

martyn-w: 👍

PR #464, opened 2020-05-18
Fix create application rake task

martyn-w:
### What?

I have added/removed/altered:

- [x] fixes the `rake auth:create_client_application` task

### Why?

I am doing this because:
- this task is currently broken in all non-development environments


lib/tasks/create_application.rake
```diff
@@ -1,5 +1,10 @@
 # frozen_string_literal: true
 
+# NB: there is an issue in Doorkeeper which will cause an "uninitialized constant Doorkeeper::Application" error when
+# running this rake task in production mode. It relates to the lazy-loading of ActiveRecord and the work-around is to call
+# ActiveRecord::Base.logger first. See https://github.com/doorkeeper-gem/doorkeeper/issues/1319
+ActiveRecord::Base.logger
```
cwrw, line 6: It seems that this has been fixed in the latest `doorkeeper` gem version, maybe if we bump up our gem to that version it might resolve this automatically? unless they haven't pushed that yet then this might be ok

martyn-w: Hi @cwrw, thanks for looking, I'll close this PR and do a separate PR to upgrade the gem: it will need a bit of testing to make sure nothing breaks.

martyn-w:
Closing to upgrade the gem in a separate PR

martyn-w:
Re-opening as upgrading the gem to 5.4.0 results in failures in the docker build process:
```
Step 17/22 : RUN SECRET_KEY_BASE=valuenotactuallyused bundle exec rails assets:precompile
 
 ---> Running in cb7ebdf7b012
 
rails aborted!
 
ActiveRecord::AdapterNotSpecified: 'production' database is not configured. Available: ["default", "development", "test"]
 
/usr/local/bundle/ruby/2.6.0/gems/activerecord-5.2.3/lib/active_record/connection_adapters/connection_specification.rb:260:in `resolve_symbol_connection'
 ```

martyn-w:
cc @mafonso this is a workaround whilst upgrading the gem is broken

cwrw:
I think this error is because we don't have `production` in our `database.yml` if run locally?

cwrw:
Also this might not be an issue if this gets merged, see https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/489

PR #465, opened 2020-05-18
Modernise rubocop

willfish:
### Jira link

P4-<TODO>

### What?

Our rubocop configuration is inherited from rubocop-govuk and is mega outdated.

- [x] Bumps rubocop
- [x] Automatically lints files after upgrade
- [x] Fixes namespace configuration changes

### Why?

We get such inconsistent linting behaviour currently. Newer versions of rubocop-govuk seemed less insane to me.

- This kind of thing can be quite risky without tests. I'm pretty comfortable that this is okay, but would like to run this through the end-to-end tests just to make sure we haven't affected what's getting returned by our apis


alexdesi:
Are we sure we want to introduce so many changes at once?
this will affect the git history with lot's of noise, and it'll be a bit more difficult to look for who made a certain change (and why)

willfish:
So I have two recommendations that are opposed to one another. If this was my project I'd have ripped out the gov rules and just move to https://github.com/thoughtbot/guides/blob/master/style/ruby/.rubocop.yml and lean towards what Abeer recommends.

I'm not worried about the size of the changes or the affect on code archeology personally, @alexdesi. However if that's valuable to you, I'm happy to consider not making these changes at all.

I think the moving to a different rubocop configuration entirely would need:

1. Agree which rubocop.yaml to use (I think you're leaning on the default one @cwrw?)
2. Make the large number of changes to get us there

We could also think of disabling the failure condition for rubocop infractions and make changes rule-by-rule.

TBH, this is all so low value that I just wanted to get something in and be done with it - validating using the e2e tests that we haven't broken the frontend.

app/models/prison_transfer_reason.rb
```diff
@@ -3,7 +3,10 @@
 class PrisonTransferReason < ApplicationRecord
   has_many :moves, dependent: :nullify
 
+  # rubocop:disable Rails/UniqueValidationWithoutIndex
   validates :key, presence: true, uniqueness: true
+  # rubocop:enable Rails/UniqueValidationWithoutIndex
```
smoothcontract, line 8: Maybe add a #TODO here to add an index in future?

app/models/region.rb
```diff
@@ -3,5 +3,7 @@
 class Region < ApplicationRecord
   has_and_belongs_to_many :locations
 
+  # rubocop:disable Rails/UniqueValidationWithoutIndex
   validates :name, :key, presence: true, uniqueness: true
+  # rubocop:enable Rails/UniqueValidationWithoutIndex
```
smoothcontract, line 8: Same here - maybe add a #TODO here to add an index in future?

lib/tasks/reference_data.rake
```diff
@@ -111,7 +113,22 @@ namespace :reference_data do
 
 private
 
+  # rubocop:disable all
```
smoothcontract, line 116: Maybe add a #TODO here so that we remember to clean this up in future?

PR #466, opened 2020-05-19
Clearer last_name sorting test

martyn-w:
### What?

I have added/removed/altered:

- [X] made last_names explicit

### Why?

I am doing this because:
- the test `rspec ./spec/requests/api/v1/moves_controller_sorting_spec.rb:214` fails occasionally without good reasons 🤔

### DIFF
--- a/spec/requests/api/v1/moves_controller_sorting_spec.rb
+++ b/spec/requests/api/v1/moves_controller_sorting_spec.rb
@@ -16,10 +16,15 @@ RSpec.describe Api::V1::MovesController do
     let(:location3) { create :location, title: 'LOCATION3' }
     let(:location4) { create :location, title: 'location4' }
 
-    let!(:move1) { create :move, to_location: location1 }
-    let!(:move2) { create :move, :prison_transfer, to_location: location2 }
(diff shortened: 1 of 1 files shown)


spec/requests/api/v1/moves_controller_sorting_spec.rb
```diff
@@ -192,27 +197,22 @@
...
 
         context 'when name' do
-          let(:people) { object_ids.map { |p_id| Person.find(p_id) } }
           let(:object_name) { 'person' }
-          let(:move_data) {
-            Move.all
-                                .sort_by { |move| move.profile.last_name }
-                                .map { |move| move.profile.person }
-          }
+          let(:last_names) { object_ids.map { |person_id| Person.find(person_id).latest_profile.last_name } }
```
alexdesi, line 201: We could clean up this line, maybe something like:
`Person.find(object_ids).include(:profile).pluck(:last_name)`

martyn-w: hmm.. not sure about that: `Person.find(object_ids)` won't guarantee the order and the test will sometimes fail

PR #469, opened 2020-05-19
P4-1406 move redirect event (part 1 - idempotency validator)

martyn-w:
### Jira link

[P4-1406](https://dsdmoj.atlassian.net/browse/P4-1406)

### What?

I have added/removed/altered:

- [ ] Added idempotency_key checker
- [ ] Added Redis gem (and Mock-Redis gem for tests)

### Why?

I am doing this because:

- to check for idempotency in move_events_controller (and soon, in other controllers)

- Redis already running on staging/preprod/production etc


Gemfile
```diff
@@ -40,6 +40,7 @@ gem 'pg', '~> 1.0.0'
 gem 'prometheus_exporter'
 gem 'puma', '~> 3.12.3', '> 3.12.3'
 gem 'rails', '~> 5.2.3'
+gem 'redis'
```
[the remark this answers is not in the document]
cwrw: Although `redis` is a good choice for storing keys and values and defining TTLs, and is quite fast, we are introducing a new dependency into our ecosystem which will require some setup and overhead. I think another option we can consider for the time being is to use our `postgresql` setup, where we can achieve the same idempotency key storage (have a look [here](https://brandur.org/idempotency-keys#idempotency-key)). If we feel there are performance implications we can then introduce the Redis dependency later on as it is the better choice in the long run.

app/controllers/api_controller.rb
```diff
@@ -28,6 +30,20 @@ def user_for_paper_trail
 
 private
 
+  # NB: for new controllers
+  def validate_required_idempotency_key
+    idempotence_validator.call(request.headers['IDEMPOTENCY_KEY'], required: true)
+  end
+
+  # NB: for legacy controllers
+  def validate_optional_idempotency_key
```
cwrw, line 39: Is there a future plan to move all controllers to use the `idempotency_key`, or just the state endpoints?

app/services/idempotence/validator.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+require 'redis'
+
+module Idempotence
+  class Validator
+    TTL = 43200 # 12 hours in seconds
+
+    def call(key, required: true)
```
cwrw, line 9: what is the format of the keys we are going to be accepting? We might need to add validation for those

app/controllers/api_controller.rb
```diff
@@ -28,6 +30,20 @@ def user_for_paper_trail
 
 private
 
+  # NB: for new controllers
+  def validate_required_idempotency_key
+    idempotence_validator.call(request.headers['IDEMPOTENCY_KEY'], required: true)
```
cwrw, line 35: Im slightly confused by the validator's function: it will set a new `idempotency_key` if it doesn't exist or throw an error otherwise. But if we were to send this request twice with the same key, I thought we should not see an error but rather return the existing key (and later on not perform the same action because it exists)?

Gemfile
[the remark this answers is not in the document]
martyn-w: Hi @cwrw, good point - however we are already running Redis (its used by sidekiq), so have already gone through the hardwork of provisioning and configuring etc. I think given the requirements: temporary, expiring keys, redis is a better match than the database.

app/controllers/api_controller.rb
martyn-w: Ultimately, I think all end points should expect an idempotency key. Enforcing that now however will break everything, hence the idea of an optional one.

app/services/idempotence/validator.rb
martyn-w: In the swagger docs I have specified it should be a uuid - I'll add a validation for that 👍

app/controllers/api_controller.rb
martyn-w: so, the function of the validator is: accept the request if it has not already been received before.

the test is: is the key in redis? 
if not, add the key and allow the request. 
If the key is in redis, reject the request with a conflict error.

Gemfile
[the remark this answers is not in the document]
cwrw: ah of course sidekiq is already using it! Ok makes sense then 👍 I guess we'll be using a different database index which will be set in the cloud platform repos

app/controllers/api_controller.rb
willfish: Could we return success but not perform any of the side effects? I thought that was the point of being idempotent?

martyn-w: @willfish no, the expected behaviour (as specified in supplier API specifications, swagger and other docs) is to return an HTTP 409 Conflict error.
Any change to this behaviour (to return success rather than error) would require discussion with suppliers as they will be coding with the current behaviour in mind.

martyn-w:
Closing this PR to implement a bit differently

PR #471, opened 2020-05-20
Remove skip_before patter and make tests clearer.

alexdesi:
### Jira link
no jira

### What?
Remove skip_before patter, making tests clearer and less spaghettish 🍝


martyn-w:
So, I'm not so keen on the repetitive `get_moves` requests in every test. Maybe this could be simplified along the lines of:
```
let(:extra_initialisation) { }
let(:get_moves) { do the GET etc }
before do
  extra_initialisation
  get_moves
end
it "tests something" do
  expect(result).to eql(foo)
end
context "when a special initial case" do
  let(:extra_initialisation) { create a move, bla bla bla}
  it "tests something else" do
    expect (response).to eql(foo)
  end
end
```

Or alternatively, use contexts and have `begin do .. end` in each different context ?

martyn-w:
In addition, ideally I think the `it` block of a test should be _testing_ something rather than _doing_ something: i.e. the call to `GET /moves` should run before the `it` block runs

PR #472, opened 2020-05-20
P4-1560 Add allocation filter to moves endpoint

cwrw:
To filter on moves with an allocation association, add a new filter
to either return all moves with an allocation if set to true, or to exclude
all moves associated to an allocation if set to false

### Jira link

P4- todo: create ticket 🥚 🐔 
https://dsdmoj.atlassian.net/browse/P4-1560
### What?

I have added/removed/altered:

- [x] Added a new filter to the moves endpoint

### Why?

To filter on moves with an allocation association, add a new filter
to either return all moves with an allocation if set to true, or to exclude
all moves associated to an allocation if set to false

### Have you? (optional)

### DIFF
--- a/app/services/moves/finder.rb
+++ b/app/services/moves/finder.rb
@@ -89,8 +89,8 @@ module Moves
     def apply_allocation_relationship_filters(scope)
       return scope unless filter_params.key?(:has_relationship_to_allocation)
 
-      scope = scope.where.not(allocation: nil) if filter_params[:has_relationship_to_allocation] == 'true'
-      scope = scope.where(allocation: nil) if filter_params[:has_relationship_to_allocation] == 'false'
+      scope = scope.where.not(allocation_id: nil) if filter_params[:has_relationship_to_allocation] == 'true'
+      scope = scope.where(allocation_id: nil) if filter_params[:has_relationship_to_allocation] == 'false'
       scope
     end
   end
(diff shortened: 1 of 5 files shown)


app/services/moves/finder.rb
```diff
@@ -84,5 +85,13 @@ def apply_supplier_filters(scope)
 
       scope.served_by(filter_params[:supplier_id])
     end
+
+    def apply_allocation_relationship_filters(scope)
+      return scope unless filter_params.key?(:has_relationship_to_allocation)
+
+      scope = scope.where.not(allocation: nil) if filter_params[:has_relationship_to_allocation] == 'true'
```
smoothcontract, line 92: Could this be changed to use `allocation_id` rather than `allocation`? I'm thinking it might be a tad faster...

spec/requests/api/v1/moves_controller_filter_spec.rb
```diff
@@ -142,6 +142,36 @@
...
+
+        it_behaves_like 'it returns the correct moves'
+      end
+
+      context 'when set to incorrect type' do
+        let(:expected_moves) { Move.where(id: [move_without_allocation.id, move_with_allocation.id]) }
+        let(:unexpected_moves) { Move.none }
+        let(:filter_value) { 'some-value' }
+
+        it_behaves_like 'it returns the correct moves'
```
smoothcontract, line 171: Does this need a test case for the key being excluded completely? Or set to `null` maybe?

spec/swagger/definitions/hand_coded_paths.yaml
```diff
@@ -320,6 +320,13 @@
         items:
           type: string
           format: uuid
+    - name: filter[has_relationship_to_allocation]
+      in: query
+      explode: false
+      description: Filters results to either exclude or include moves associated to an allocation
```
smoothcontract, line 326: Maybe add the word `optional` in here just to clarify it can be omitted.

PR #475, opened 2020-05-20
P4-1561 Add support for filtering GET /allocations by status

smoothcontract:
### Jira link

P4-1561

### What?

- [x] Add new filter by `status` within `GET /allocations` endpoint. This allows filtering results by each of the allocation statuses, with support for multiple statuses. If omitted all results are returned as before.

### Why?

- This supports front end work to exclude cancelled allocations in certain circumstances.

### Have you? (optional)

- Changes an api that is used in production, but in a non destructive way. Behaviour is unchanged if new filter parameter is not specified, so is backwards compatible.

### DIFF
--- a/spec/requests/api/v1/allocations_controller_index_spec.rb
+++ b/spec/requests/api/v1/allocations_controller_index_spec.rb
@@ -13,7 +13,7 @@ RSpec.describe Api::V1::AllocationsController do
   describe 'GET /allocations' do
     let(:schema) { load_yaml_schema('get_allocations_responses.yaml') }
 
-    let!(:allocations) { create_list :allocation, 21 }
+    let!(:allocations) { create_list :allocation, 2 }
     let(:params) { {} }
 
     before do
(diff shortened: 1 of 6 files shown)


spec/requests/api/v1/allocations_controller_index_spec.rb
```diff
@@ -13,7 +13,7 @@
   describe 'GET /allocations' do
     let(:schema) { load_yaml_schema('get_allocations_responses.yaml') }
 
-    let!(:allocations) { create_list :allocation, 21 }
+    let!(:allocations) { create_list :allocation, 2 }
```
smoothcontract, line 16: It feels silly to create 21 allocations for all of the specs here, it's only the pagination stuff that really needs quite so many so this should optimise the test performance a bit.

spec/requests/api/v1/allocations_controller_index_spec.rb
```diff
@@ -69,7 +69,25 @@
...
+        let(:filled_allocation) { create :allocation, :filled }
+        let(:cancelled_allocation) { create :allocation, :cancelled }
+        let!(:allocations) { [unfilled_allocation, filled_allocation, cancelled_allocation] }
+        let(:params) { { filter: { status: 'cancelled' } } }
+
+        it 'filters the results' do
+          expect(response_json['data'].size).to be 1
+        end
+
+        it 'returns the allocation that matches the filter' do
```
cwrw, line 83: can you add a test for a list of statuses?

spec/services/allocations/finder_spec.rb
```diff
@@ -106,7 +106,37 @@
...
+      context 'with multiple statuses' do
+        let(:filter_params) { { status: 'unfilled,cancelled' } }
+
+        it 'returns allocations matching status' do
+          expect(allocation_finder.call).to contain_exactly(unfilled_allocation, cancelled_allocation)
+        end
+      end
+
+      context 'with mis-matching status' do
+        let(:filter_params) { { status: 'foo' } }
```
cwrw, line 135: can you add a test for nil values as well?

spec/requests/api/v1/allocations_controller_index_spec.rb
smoothcontract: I think that's suitably covered off with the finder specs - this is really just to prove that the request params make it through the controller to the finder and would probably be overkill to test everything again. The test did fail until I remembered to whitelist the extra `status` param in the controller :)

spec/services/allocations/finder_spec.rb
smoothcontract: why not!

PR #479, opened 2020-05-21
[P4-1497] Prepare includes

willfish:
### Jira link

P4-1497

### What?

We are expecting includes to come from the query param in the following format:

`?include=gender,location,person.profile`

Which gets parsed by rails/rack as:

```ruby
"include" => [
  "gender",
  "location",
  "person.profile"
]
```

This format is supported by AMS in addition to the current format we're using.

This PR moves our current includes to the same format that comes from the query params to avoid doing things in more than one way and to simplify our setup.

### Why?

To simplify our setup

### DIFF
--- a/app/controllers/api/v1/moves_controller.rb
+++ b/app/controllers/api/v1/moves_controller.rb
@@ -124,7 +124,6 @@ module Api
         @updater ||= Moves::Updater.new(move, update_move_params)
       end
 
-      # ["foo.bar", "baz.bim"]
       def include_params
         params.permit(:include)
       end
--- a/app/controllers/api/v1/moves_controller.rb
+++ b/app/controllers/api/v1/moves_controller.rb
@@ -8,7 +8,8 @@ module Api
       def index
         moves = Moves::Finder.new(filter_params, current_ability, params[:sort] || {}).call
         # Excludes potentially many court hearing documents to reduce the request size. This was requested specifically by the frontend team.
-        paginate moves, include: MoveSerializer::INCLUDED_ATTRIBUTES.dup.except(:court_hearings), fields: MoveSerializer::INCLUDED_FIELDS
+        includes = MoveSerializer::INCLUDED_ATTRIBUTES.dup - %w[court_hearings]
+        paginate moves, include: includes, fields: MoveSerializer::INCLUDED_FIELDS
       end
 
       def show
@@ -121,6 +122,11 @@ module Api
 
       def updater
         @updater ||= Moves::Updater.new(move, update_move_params)
+      end
+
+      # ["foo.bar", "baz.bim"]
+      def include_params
+        params.permit(:include)
       end
     end
   end
--- a/app/controllers/api/v1/moves_controller.rb
+++ b/app/controllers/api/v1/moves_controller.rb
@@ -123,10 +123,6 @@ module Api
       def updater
         @updater ||= Moves::Updater.new(move, update_move_params)
       end
-
-      def include_params
-        params.permit(:include)
-      end
     end
   end
 end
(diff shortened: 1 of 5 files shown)


app/controllers/api/v1/moves_controller.rb
```diff
@@ -8,7 +8,8 @@ class MovesController < ApiController
       def index
         moves = Moves::Finder.new(filter_params, current_ability, params[:sort] || {}).call
         # Excludes potentially many court hearing documents to reduce the request size. This was requested specifically by the frontend team.
-        paginate moves, include: MoveSerializer::INCLUDED_ATTRIBUTES.dup.except(:court_hearings), fields: MoveSerializer::INCLUDED_FIELDS
+        includes = MoveSerializer::INCLUDED_ATTRIBUTES.dup - %w[court_hearings]
+        paginate moves, include: includes, fields: MoveSerializer::INCLUDED_FIELDS
```
alexdesi, line 12: what is `includes` ?

app/controllers/api/v1/moves_controller.rb
```diff
@@ -8,7 +8,8 @@ class MovesController < ApiController
       def index
         moves = Moves::Finder.new(filter_params, current_ability, params[:sort] || {}).call
         # Excludes potentially many court hearing documents to reduce the request size. This was requested specifically by the frontend team.
-        paginate moves, include: MoveSerializer::INCLUDED_ATTRIBUTES.dup.except(:court_hearings), fields: MoveSerializer::INCLUDED_FIELDS
+        includes = MoveSerializer::INCLUDED_ATTRIBUTES.dup - %w[court_hearings]
```
smoothcontract, line 11: `MoveSerializer::INCLUDED_ATTRIBUTES` constant is only reference here, so would be cleaner to simply remove `court_hearings` from that constant defined in `MoveSerializer`.

app/controllers/api/v1/moves_controller.rb
smoothcontract: See line 11 ☝️

app/serializers/move_serializer.rb
```diff
@@ -25,15 +25,16 @@ class MoveSerializer < ActiveModel::Serializer
...
-    allocation: [],
-  }.freeze
+  INCLUDED_ATTRIBUTES = %w[
+    person.ethnicity
+    person.gender
+    from_location
+    to_location
+    documents
+    prison_transfer_reason
+    court_hearings
```
smoothcontract, line 35: Seems like we don't need `court_hearings`, so could remove here.

cwrw: we are showing `court_hearings` in the single view endpoint though not the list one

app/controllers/api/v1/moves_controller.rb
alexdesi: 😄

app/serializers/move_serializer.rb
smoothcontract: Ah, bollocks. You're right - it's referenced in two places inside `moves_controller.rb`. Harrumph.

app/controllers/api/v1/moves_controller.rb
smoothcontract: Ignore this comment 🤦

willfish:
@cwrw What do you consider necessary and sufficient to validate this change?

I'm thinking that a simple test of the nested/non-nested dot-separated include forms in the move serializer will do?

cwrw:
I just double checked, and this work might be out of scope for this ticket, but I will come back to them and fix them up later, as was holding off till sparse fields work comes in, but feel free to merge

willfish:
@alexdesi is going to add specs for this. I've manually run these and am comfortable that the behaviour is the same.

PR #481, opened 2020-05-21
Adds strong migrations

willfish:
### Jira link

P4-<TODO>

### What?

Adds strong_migrations to our project.

This gem is essentially a linter for migrations - are they dangerous, are they slow, are they known wrong ways of doing stuff, etc.

It ignores our existing migrations and makes it so that all future migrations explode when a principle is breached.

### Why?

- Pedagogical - we learn some of the principles around safer migrations by exposure to failures
- Process - we get better at thinking more about how migrations are going to affect our production systems
- Safety - we cause fewer outages/exceptions and catch things earlier

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -43,6 +43,7 @@ gem 'rails', '~> 5.2.3'
 gem 'sassc-rails'
 gem 'sentry-raven'
 gem 'sidekiq'
+gem 'strong_migrations', '~> 0.6.8'
 gem 'tty-prompt'
 gem 'validate_url', '~> 1.0.8'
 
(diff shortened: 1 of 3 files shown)


willfish:
Example of running a migration to drop a column

```
== 20200521123716 DropColumnCreatedAtFromPeople: migrating ==================================
rails aborted!
StandardError: An error has occurred, this and all later migrations canceled:

=== Dangerous operation detected #strong_migrations ===

Active Record caches attributes which causes problems
when removing columns. Be sure to ignore the column:

class Person < ApplicationRecord
  self.ignored_columns = ["created_at"]
end

Deploy the code, then wrap this step in a safety_assured { ... } block.

class DropTablePeople < ActiveRecord::Migration[5.2]
  def change
    safety_assured { remove_column :people, :created_at }
  end
end
```

PR #482, opened 2020-05-21
[P4-1451] Support includes and return profile

alexdesi:
### Jira link

P4-1451

### What?

Support for the query attribute `?includes=profile`

### Why?

To support the Json:Api specification

No risk, since FE in not yet forced to use the new `include` query param


app/controllers/api/v1/moves_controller.rb
```diff
@@ -3,13 +3,17 @@
...
 
       def index
         moves = Moves::Finder.new(filter_params, current_ability, params[:sort] || {}).call
-        # Excludes potentially many court hearing documents to reduce the request size. This was requested specifically by the frontend team.
-        includes = MoveSerializer::INCLUDED_ATTRIBUTES.dup - %w[court_hearings]
-        paginate moves, include: includes, fields: MoveSerializer::INCLUDED_FIELDS
+
+        paginate moves, include: included_relationships, fields: MoveSerializer::INCLUDED_FIELDS
+
+        # paginate moves,
```
martyn-w, line 13: do these comments need to be here ?

app/controllers/api/v1/moves_controller.rb
```diff
@@ -123,6 +127,36 @@ def move
       def updater
         @updater ||= Moves::Updater.new(move, update_move_params)
       end
+
+      def included_relationships
+        # Excludes potentially many court hearing documents to reduce the request size. This was requested specifically by the frontend team.
+        MoveSerializer::SUPPORTED_RELATIONSHIPS + include_params - %w[court_hearings]
```
martyn-w, line 133: maybe there has already been discussion about this, but I think it would be better to:

* if include_params is specified in the query, then use that (filtered to all support relationships, so that you couldn't include e.g. users passwords)
* if include_params is not specified in the query then go with the current default: `MoveSerializer::SUPPORTED_RELATIONSHIPS - %w[court_hearings]`

That way we can start moving clients towards requesting what they need, rather than always give everything

app/serializers/profile_serializer.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+class ProfileSerializer < ActiveModel::Serializer
+  include Rails.application.routes.url_helpers
```
martyn-w, line 4: I don't think this line `include Rails.application.routes.url_helpers` is required

app/controllers/api/v1/moves_controller.rb
alexdesi: yes, I like that, I think it helps to support the include param and it's easy to explain.

alexdesi: my only concern is, the FE will be forced to pass all the `include` params at once

app/serializers/profile_serializer.rb
alexdesi: good point, I think we don't need it in PersonSerializer too.

app/serializers/profile_serializer.rb
```diff
@@ -0,0 +1,26 @@
...
+    :id,
+    :first_names,
+    :last_name,
+    :date_of_birth,
+    :assessment_answers,
+    :identifiers,
+    :gender_additional_information,
+  )
+
+  has_one :ethnicity, serializer: EthnicitySerializer, if: -> { object.ethnicity.present? }
```
willfish, line 14:
```suggestion
  has_one :ethnicity, serializer: EthnicitySerializer
```

app/serializers/profile_serializer.rb
```diff
@@ -0,0 +1,26 @@
...
+  has_one :gender, serializer: GenderSerializer
+
+  SUPPORTED_RELATIONSHIPS = %i[ethnicity gender].freeze
+
+  def assessment_answers
+    object.assessment_answers || []
```
willfish, line 20: Do we need these?

alexdesi: I think so, it's a jsonb not a rel.

app/serializers/profile_serializer.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def identifiers
+    object.profile_identifiers || []
+  end
+end
```
alexdesi, line 26: This serializer lack of a test, I'll add it.

alexdesi: done

app/controllers/api/v1/allocations_controller.rb
```diff
@@ -72,8 +72,8 @@ def send_move_notifications
         end
       end
 
-      def includes
-        IncludeParamHandler.new(params).call || AllocationSerializer::INCLUDED_ATTRIBUTES
+      def included_relationships
+        IncludeParamHandler.new(params).call || AllocationSerializer::SUPPORTED_RELATIONSHIPS
```
alexdesi, line 76: remove this

app/controllers/api/v1/moves_controller.rb
```diff
@@ -115,6 +116,26 @@ def move
       def updater
         @updater ||= Moves::Updater.new(move, update_move_params)
       end
+
+      def validate_include_params
+        # TODO: this is temporary, once FE uses the `include` params for all the endpoints/resources,
+        # supported_attributes will be set to to MoveSerializer::SUPPORTED_RELATIONSHIPS
+        supported_attributes = MoveSerializer::SUPPORTED_RELATIONSHIPS + %w[profile]
```
willfish, line 123: I think we can just return the profile and the person in the `MoveSerializer::SUPPORTED_RELATIONSHIPS` constant. The frontend will deserialize the json and not access the `profile` key so this is safe.

This means we can add:

```ruby
      class UnsupportedIncludeError < StandardError; end

      rescue UnsupportedIncludeError, with: :render_unsupported_includes

      def validate_include_params
        included_relationships.each do |relationship|
          raise UnsupportedIncludeError unless MoveSerializer::SUPPORTED_RELATIONSHIPS.include?(relationship)
        end
      end

      def included_relationships
        IncludeParamHandler.new(params).call || MoveSerializer::SUPPORTED_RELATIONSHIPS
      end

      def render_unsupported_includes(exception)
           render
                   status: :bad_request,
                   json: {
                     errors: [{ title: 'Bad request',
                                detail: "'#{resource}' is not supported. Valid values are: #{supported_attributes.join(', ')}" }],
                   }
      end
```

Notice I've renamed supported_attributes to supported_relationships and am propagating an error in the same way we will in the api base controller.

willfish: And just copy and paste the propagated exception

willfish: When we've migrated all of the includes

willfish: i.e. to the base controller

PR #483, opened 2020-05-21
[P4-1492] Allocation includes

willfish:
### Jira link

P4-1492

### What?

Enables clients of our allocations api to supply includes as a query parameter.

Removing the default behaviour and documentation to come later once the frontend has integrated successfully.

### Why?

This gives clients of our api significantly more flexibility and should reduce our heavy-handed approach of including everything on every call to the api.


spec/requests/api/v1/allocations_controller_index_spec.rb
```diff
@@ -167,6 +167,50 @@
...
+            from_locations + to_locations + moves
+          end
+
+          it 'returns the default includes' do
+            expect(response_json).to have_includes(included)
+          end
+        end
+
+        context 'when including the include query param' do
+          let(:query_params) { '?include=foo.bar,from_location' }
```
willfish, line 195: We've explicitly agreed to defer handling of invalid includes errors for the moment and AMS doesn't propagate any errors/leave any visible sign of some failure.

```markdown
# Inclusion of Related Resources
An endpoint MAY return resources related to the primary data by default.

An endpoint MAY also support an include request parameter to allow the client to customize which related resources should be returned.

If an endpoint does not support the include parameter, it MUST respond with 400 Bad Request to any requests that include it.

If an endpoint supports the include parameter and a client supplies it, the server MUST NOT include unrequested resource objects in the included section of the compound document.

The value of the include parameter MUST be a comma-separated (U+002C COMMA, “,”) list of relationship paths. A relationship path is a dot-separated (U+002E FULL-STOP, “.”) list of relationship names.

If a server is unable to identify a relationship path or does not support inclusion of resources from a path, it MUST respond with 400 Bad Request.
```

willfish:
We're going to do that in one PR at the end @cwrw

app/controllers/api/v1/allocations_controller.rb
```diff
@@ -71,6 +71,10 @@ def send_move_notifications
           Notifier.prepare_notifications(topic: move, action_name: 'create')
         end
       end
+
+      def includes
+        params.fetch(:include, nil)&.split(',') || AllocationSerializer::INCLUDED_ATTRIBUTES
```
cwrw, line 76: We don't require a `fetch` here as `params[:fetch]` will return `nil` if empty by default. Also this might be better suited in a service class, if we intend to further build handling errors and sparse fields and to allow this to be tested in isolation as well.

spec/requests/api/v1/allocations_controller_index_spec.rb
```diff
@@ -167,6 +167,50 @@
           expect(response.body).to eq('{"error":{"locations":["may not be used in combination with `from_locations` or `to_locations` filters."]}}')
         end
       end
+
+      describe 'included relationships', :skip_before do
+        let!(:allocations) { create_list :allocation, 2, :with_moves }
+
+        before do
+          get "/api/v1/allocations#{query_params}", params: params, headers: headers
```
cwrw, line 175: I think we only need to pass `params` here as those are converted to query params?

spec/requests/api/v1/allocations_controller_index_spec.rb
```diff
@@ -167,6 +167,50 @@
...
+
+          let(:included) { allocations.map(&:from_location) }
+
+          it 'returns the valid provided includes' do
+            expect(response_json).to have_includes(included)
+          end
+        end
+
+        context 'when including an empty include query param' do
+          let(:query_params) { '?include=' }
```
cwrw, line 205: can you also add a test for a nil value as well?

spec/requests/api/v1/allocations_controller_index_spec.rb
willfish: I just preferred the explicit query formulation. I know you're technically correct and that it all gets bundled together. If you feel strongly about it I can put the query into the params

spec/requests/api/v1/allocations_controller_index_spec.rb
willfish: `nil.to_s` is '' so is equivalent to https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/483/files#diff-ca0a918cadf1b72e0d225d2948b64e0eR178?

app/controllers/api/v1/allocations_controller.rb
willfish, line 76:
```suggestion
        params[:include]&.split(',') || AllocationSerializer::INCLUDED_ATTRIBUTES
```

spec/requests/api/v1/allocations_controller_index_spec.rb
willfish: So I validated this by passing `foo: bar` in the params hash and seeing whats loaded in the controller action. These values get merged in request specs properly :+1:.

spec/requests/api/v1/allocations_controller_index_spec.rb
willfish: Added a test.

spec/support/matchers/have_includes.rb
```diff
@@ -0,0 +1,32 @@
+RSpec::Matchers.define :have_includes do |*includes|
```
cwrw, line 1: This matcher somewhat ballooned in complexity and is making me rethink our decision around this 🤔 As this is all untested code really, maybe if we go back to something simple as this, maybe it's enough for now:
```
expect(result[:included].map { |r| r[:type] }).to contain_exactly(%w[people moves locations locations])
```
This ties into a bigger conversation around comparing nested JSON, and although `include_json` does resolve some of it I think there is something missing which we can add. Let me have a ponder and do a bit of research around that, but I guess it's outside the scope of this pr for now. Let me know your thoughts

spec/requests/api/v1/allocations_controller_index_spec.rb
cwrw: Not a blocker: I would prefer if we use `params` to avoid having to encode and format the query params ourselves. If we go with `query_params` however, can you get rid of `params: params` as we aren't using it in this context, and main hash is empty:
https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/483/files#diff-ca0a918cadf1b72e0d225d2948b64e0eL17
so something like this: 
```
get "/api/v1/allocations#{query_params}", headers: headers
```

spec/services/include_param_handler_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    let(:include) { nil }
+
+    it { expect(service.call).to be_nil }
+  end
+
+  context 'when the include param is a splittable string' do
+    let(:include) { 'foo.bar,baz.qux' }
+
+    it { expect(service.call).to eq(['foo.bar', 'baz.qux']) }
+  end
```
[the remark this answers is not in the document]
willfish: Done.

spec/support/matchers/have_includes.rb
willfish: I think that given I am specifying the exact ids and entities that are being included I don't need to validate the types? And actually am implicitly validating them?

willfish:
```
            from_locations = allocations.map(&:from_location)
            to_locations = allocations.map(&:to_location)
            moves = allocations.flat_map(&:moves)
            people = moves.map { |move| move&.profile&.person }
```

This bit ^^

willfish: This is also super visible in the expectation so is really obvious what we're testing around the types of things that are getting returned.

PR #484, opened 2020-05-22
P4-1406 redirect move endpoint

martyn-w:
### Jira link

[P4-1406](https://dsdmoj.atlassian.net/browse/P4-1406)

### What?

I have added/removed/altered:

- [x] Added /moves/id/redirect endpoint
- [x] Temporarily maintained existing /moves/id/events endpoint for now
- [x] Added event log runner for processing redirects

### Why?

I am doing this because:

- so clients can redirect moves

### Have you? (optional)


app/controllers/api/v1/journeys_controller.rb
```diff
@@ -78,9 +78,10 @@ def new_journey_params
 
       def new_journey_attributes
         @new_journey_attributes ||= new_journey_params.to_h[:attributes].tap do |attribs|
+          attribs.delete(:timestamp) # throw the timestamp away as we are using client_timestamp instead
```
smoothcontract, line 81: Something about this change isn't feeling right to me. In particular the change to call `timestamp` method below suggests use of `data_params` rather than the `new_journey_params` hash - this is confusing and might cause side effects. If this is the intent, perhaps this change makes the behaviour more explicit:

```ruby
timestamp = attribs.delete(:timestamp)

attribs.merge!(
  ...
  client_timestamp: Time.zone.parse(timezone)
  ...
)
```

app/controllers/api/v1/journeys_controller.rb
```diff
@@ -123,13 +124,19 @@ def update_journey_attributes
         end
       end
 
+      def timestamp
+        @timestamp ||= Time.zone.parse(data_params.dig(:attributes, :timestamp))
```
smoothcontract, line 128: This throws an error if the timestamp attribute is missing - might need a guard clause around that or catch the possible exception:

```ruby
Time.zone.parse(nil)
TypeError: no implicit conversion of nil into String
```

smoothcontract: Actually same could be true of the time zone parsing in my previous comment

app/controllers/api/v1/move_events_controller.rb
```diff
@@ -59,39 +75,28 @@ def timestamp
...
-      def to_location
-        @to_location ||= Location.find(event_params.dig(:relationships, :to_location, :data, :id))
+      def supplier_id
+        # NB: not all events will have a supplier_id so this could well be nil
+        current_user.owner&.id
       end
 
       def move
-        @move ||= Move.accessible_by(current_ability).find(params.require(:move_id))
+        @move ||= Move.accessible_by(current_ability).find(params.require(:id))
```
smoothcontract, line 84: Is this as a result of amending the routes structure? Might be worth double checking that this is the correct id (run `rails routes`) - I would expect this to still be named `:id` from the URL params.

app/models/event.rb
```diff
@@ -1,11 +1,30 @@
...
   validates :event_name, presence: true, inclusion: { in: EVENT_NAMES }
   validates :client_timestamp, presence: true
+  validates :details, presence: true
 
   scope :default_order, -> { order(client_timestamp: :asc) }
+
+  serialize :details, HashWithIndifferentAccessSerializer
+
+  def supplier_id
+    @supplier_id ||= details.dig(:supplier_id)
```
smoothcontract, line 16: I don't think `dig` is the best choice when only providing one argument - it's a very shallow hole :)

app/models/move.rb
```diff
@@ -42,7 +42,7 @@ class Move < VersionedModel
   has_many :notifications, as: :topic, dependent: :destroy # NB: polymorphic association
   has_many :journeys, -> { default_order }, dependent: :restrict_with_exception, inverse_of: :move
   has_many :court_hearings, dependent: :restrict_with_exception
-  has_many :events, as: :eventable, dependent: :destroy # NB: polymorphic association
+  has_many :events, as: :eventable, dependent: :destroy, class_name: 'MoveEvent' # NB: polymorphic association
```
smoothcontract, line 45: If this works I think it's slightly cleaner and more explicit (guessing here)...

```ruby
has_many :move_events, as: :eventable, dependent: :destroy
```

app/models/move_event.rb
```diff
@@ -0,0 +1,7 @@
+class MoveEvent < Event
+  # NB: this class exposes a few methods specific to moves to the Event model
+
+  def to_location
+    @to_location ||= Location.find(event_params&.dig(:relationships, :to_location, :data, :id))
```
smoothcontract, line 5: It seems a little odd for a model to be delving into a params hash. I may be misunderstanding the use case here but possibly this is a controller or service responsibility to find the relevant location and persist against this model as `to_location_id` - it feels more idiomatic to then replace this with:

```ruby
```

app/serializers/hash_with_indifferent_access_serializer.rb
```diff
@@ -0,0 +1,10 @@
+# Based on: https://nandovieira.com/using-postgresql-and-jsonb-with-ruby-on-rails
+class HashWithIndifferentAccessSerializer
```
smoothcontract, line 2: Not really sure what the purpose of this class is? It feels like this should have a more rigidly defined structure so that we can have appropriate swagger documentation, schemas, specs etc.

spec/factories/events.rb
```diff
@@ -1,8 +1,9 @@
 FactoryBot.define do
   factory :event do
-    association(:eventable)
+    eventable { association(:move) }
     event_name { 'create' }
     client_timestamp { Time.now.utc + rand(-60..60).seconds } # NB: the client_timestamp will never be perfectly in sync with system clock
+    details { { supplier_id: '1234', event_params: { attributes: { notes: 'foo' } }, data_params: { attributes: { notes: 'bar' } } } }
```
smoothcontract, line 6: This formatting is pretty hard to read - perhaps break over multiple lines inside a `do .. end` block. Same for line 42.

spec/swagger/definitions/hand_coded_paths.yaml
```diff
@@ -489,9 +489,10 @@
               $ref: put_move_responses.yaml#/422
 /moves/{move_id}/events:
   post:
-    summary: '[NOT YET IMPLEMENTED] Creates a move event'
+    deprecated: true
+    summary: '[DO NOT USE] Creates a move event'
     description: |
-      ## [NOT YET IMPLEMENTED]
+      ## [DO NOT USE]
```
smoothcontract, line 495: If this wasn't implemented previously, and is now deprecated is it realistic that anyone is actually using it? If not  used then could it be deleted now?

app/controllers/api/v1/journeys_controller.rb
martyn-w: the `before_action :validate_params` takes care of ensuring `timestamp` is the approved iso8601 format (and is not nil)

app/controllers/api/v1/journeys_controller.rb
martyn-w: happy to change as you suggest. NB further on in this file (line 135) timestamp is used again to create the event, and so it will be reparsed

app/controllers/api/v1/move_events_controller.rb
martyn-w: yes, it is the routes have changed: `events` used to be a nested resource inside of `/moves/{move_id}/events`, so it was necessary for rails to distinguish from move_id and (id = event_id). But now the events endpoint is just an additional method on a move resource, rather than a resource in its own right..

app/models/event.rb
martyn-w: 👍 good point

app/models/move.rb
martyn-w: 👍

app/models/move_event.rb
martyn-w: lets chat

app/serializers/hash_with_indifferent_access_serializer.rb
martyn-w: so, this serializer makes jsonb objects behave as hashes with indifferent access, to make it easier to work with.
The `details` field in an event contains the full event parameters used to create the event, which will vary from event to event.  A cancellation event must specify a cancellation_reason for example, whereas a redirect event must specify a `to_location` relationship. So a rigid structure with a field for to_location or a field for cancellation_reason is not a good fit in this case (IMHO).

spec/swagger/definitions/hand_coded_paths.yaml
martyn-w: alas, it is actually being used by the front end, as a temporary endpoint, so can't be deleted right away.

PR #487, opened 2020-05-22
Allocation `moves_count` now includes only non cancelled moves

smoothcontract:
### Jira link

P4-1554

### What?

- [x] Updated validation on allocation `moves_count` (now only applies on `:create`)
- [x] Updated allocation `cancel` method to correct `moves_count` to 0
- [x] Updated `Moves::Updater` service to correct associated allocation `moves_count`
- [x] Refactored existing `Moves::Updater` specs to remove `:skip_before` tag

### Why?

- Allocation `moves_count` attribute should always reflect the number of associated non-cancelled moves. This can then be used reliably for the front end dashboard stats. Note that this PR does **not** change the allocation moves relationship, which still includes cancelled moves.
- When the allocation is first created, the `moves_count` attribute is used to create a number of associated moves with `requested` status, so the `moves_count` attribute on the allocation is correct at that point. Subsequently, the allocation cannot be changed but may be cancelled. Additionally the associated moves may be changed - notably to amend their status.
- If the allocation is cancelled completely, the `moves_count` should be set to zero as all of the associated moves are now cancelled.
- If a move associated to an allocation changes status (via the `Moves::Updater` service) then the allocation `moves_count` is now updated to count only non cancelled moves.

### Have you? (optional)

- This changes the behaviour of `moves_count` attribute within a production API, but actually corrects the existing incorrect behaviour.

### DIFF
--- a/app/services/moves/updater.rb
+++ b/app/services/moves/updater.rb
@@ -14,8 +14,10 @@ module Moves
       # NB: rather than update directly, we need to detect whether the move status has changed before saving the record
       self.status_changed = move.status_changed?
 
-      move.save!
-      move.allocation&.refresh_moves_count! if status_changed
+      move.transaction do
+        move.save!
+        move.allocation&.refresh_moves_count! if status_changed
+      end
     end
 
   private
--- a/app/services/moves/updater.rb
+++ b/app/services/moves/updater.rb
@@ -15,6 +15,7 @@ module Moves
       self.status_changed = move.status_changed?
 
       move.save!
+      move.allocation&.refresh_moves_count! if status_changed
     end
 
   private
--- a/spec/requests/api/v1/moves_controller_update_spec.rb
+++ b/spec/requests/api/v1/moves_controller_update_spec.rb
@@ -288,6 +288,25 @@ RSpec.describe Api::V1::MovesController do
         end
 
         context 'when cancelling a move' do
+          let(:move_params) do
+            {
+              type: 'moves',
+              attributes: {
+                status: 'cancelled',
+                cancellation_reason: 'other',
+              },
+            }
+          end
+
+          context 'with an associated allocation' do
+            let!(:allocation) { create :allocation, moves_count: 1 }
+            let!(:move) { create :move, :requested, from_location: from_location, allocation: allocation }
+
+            it 'updates the allocation moves_count' do
+              expect(allocation.reload.moves_count).to eq(0)
+            end
+          end
+
           context 'when the supplier has a webhook subscription', :skip_before do
             let!(:subscription) { create(:subscription, :no_email_address, supplier: supplier) }
             let!(:notification_type_webhook) { create(:notification_type, :webhook) }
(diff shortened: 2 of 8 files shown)


app/services/moves/updater.rb
```diff
@@ -15,6 +15,7 @@ def call
       self.status_changed = move.status_changed?
 
       move.save!
+      move.allocation&.refresh_moves_count! if status_changed
```
cwrw, line 18: instead of attempting to save changes twice, is there a nice way we can include the `moves_count` update, maybe in a `after_validation` hook?

smoothcontract: This saves the move once and the allocation once - I also wanted to avoid needlessly updating the allocation if the status hasn't changed. I prefer this approach as it's explicit and simple, plus we can see all of the steps involved in updating a move in one place (here in the service).

cwrw: My main concern lies in ensuring data consistency. When we perform two saves and one of them fails, we might get into an inconsistent state. One way around this would be to do something similar to https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/487/files#diff-ef7828cdbd6b61227ed9a64910384e69R65 where we still keep the `refresh_moves_count` in the allocation, but it only sets the attribute and does not save the record:
```
move.assign_attributes(attributes)
self.status_changed = move.status_changed?
move.allocation&.refresh_moves_count if status_changed
move.save!
```
This way we perform all our updates in one transaction.
Or we could wrap this whole thing in a transaction, I prefer the former though.

smoothcontract: Added a transaction wrapper so that the state is consistent even if saving the allocation somehow fails after saving the move.

spec/requests/api/v1/moves_controller_update_spec.rb
```diff
@@ -288,6 +288,25 @@
...
+              type: 'moves',
+              attributes: {
+                status: 'cancelled',
+                cancellation_reason: 'other',
+              },
+            }
+          end
+
+          context 'with an associated allocation' do
+            let!(:allocation) { create :allocation, moves_count: 1 }
```
cwrw, line 302: can you use the `with_moves` trait here instead of creating another move below?

spec/services/moves/updater_spec.rb
```diff
@@ -48,16 +44,39 @@
...
+
+    context 'when status changes to non cancelled with an associated allocation' do
+      let!(:allocation) { create :allocation, moves_count: 5 }
+      let!(:move) { create :move, :cancelled, from_location: from_location, allocation: allocation }
+
+      it 'corrects allocation moves_count' do
+        expect { updater.call }.to change { allocation.reload.moves_count }.from(5).to(1)
+      end
+    end
+
```
cwrw, line 74: can you add a test for when there is no allocation associated with the move? unless that's covered in the existing tests

smoothcontract: The first one is covered by line 47. I'll add another test for a move with an allocation where the status doesn't change.

PR #489, opened 2020-05-22
Removes custom wiki

willfish:
### Jira link

P4-<TODO>

### What?

We handcrafted a custom wiki and added a whole bunch of dependencies (programming languages, packages, installations) to our project.

Removes all of the code/infrastructure configurations around this to significantly simplify our project.

### Why?

We can get 99% of the value of this work by simply using the github wiki and none of the dependencies in our api project by removing this and migrating to the github wiki.


config/routes.rb
```diff
@@ -6,9 +6,6 @@
     mount Rswag::Api::Engine => '/api-docs'
   end
 
-  get "/docs/*id" => 'pages#show', as: :page, format: false
```
martyn-w, line 9: maybe redirect this to https://github.com/ministryofjustice/hmpps-book-secure-move-api/wiki ?

PR #490, opened 2020-05-26
Surface allocation moves in move endpoint

cwrw:
When a move is associated to an allocation we want to see all the included
moves as well. Surface those values in included fields and attributes. Please
note that because all moves are included in the main GET moves endpoint we
won't show those as included, and those only appear in the single GET move
requests

### DIFF
--- a/app/serializers/move_serializer.rb
+++ b/app/serializers/move_serializer.rb
@@ -34,10 +34,11 @@ class MoveSerializer < ActiveModel::Serializer
     prison_transfer_reason
     court_hearings
     allocation
+    allocation.moves
   ].freeze
 
   INCLUDED_FIELDS = {
-    allocation: %i[to_location from_location moves_count created_at],
+    allocation: %i[to_location from_location moves_count moves created_at],
   }.freeze
 
   def person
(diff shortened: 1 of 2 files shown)


app/serializers/move_serializer.rb
```diff
@@ -34,10 +34,11 @@ class MoveSerializer < ActiveModel::Serializer
     prison_transfer_reason
     court_hearings
     allocation
+    allocation.moves
```
cwrw, line 37: these will not appear in the GET all moves endpoint so there is no worry we'll return very large include responses

willfish:
@cwrw If this is still a thing, do you mind if we catch up about it. I'm fine with the code, but I'm wondering whether we can prioritise this work with Alex to pull out the includes using the query first? Then this PR just becomes extend @alexdesi change which [validates supported include relationships](https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/serializers/move_serializer.rb#L30-L40) to also support `allocation.moves`. Happy to resolve the conflicts together too if that's helpful.

cwrw:
closing this pr until this is needed

PR #493, opened 2020-05-26
[P4-1435] Adds migration for extra person fields

willfish:
### Jira link

P4-1435

### What?

It makes way more sense to have relatively-infrequently changing `Profile` attributes on a `Person`.

A `Profile` is expected to change every `Move` potentially and a new one should be created for the `Move`.

The full list of `Profile` attributes are:

```ruby
    t.uuid "person_id", null: false
    t.string "last_name", null: false
    t.string "first_names", null: false
    t.date "date_of_birth"
    t.string "aliases", default: [], array: true
    t.uuid "gender_id"
    t.uuid "ethnicity_id"
    t.uuid "nationality_id"
    t.jsonb "assessment_answers"
    t.jsonb "profile_identifiers"
    t.string "gender_additional_information"
```

- [x] Migrates attributes
- [x] Splits out and promotes identifier attributes 

### Why?

A profile should contain frequently changing assessments and personal care needs and shouldn't be an audit trail of all changes to the person.

### DIFF
--- a/db/migrate/20200526152251_adds_fields_to_people.rb
+++ b/db/migrate/20200526152251_adds_fields_to_people.rb
@@ -13,6 +13,5 @@ class AddsFieldsToPeople < ActiveRecord::Migration[5.2]
 
     add_reference :people, :ethnicity, type: :uuid
     add_reference :people, :gender, type: :uuid
-    add_reference :people, :nationality, type: :uuid
   end
 end
--- a/db/migrate/20200526152251_adds_fields_to_people.rb
+++ b/db/migrate/20200526152251_adds_fields_to_people.rb
@@ -13,5 +13,9 @@ class AddsFieldsToPeople < ActiveRecord::Migration[5.2]
 
     add_reference :people, :ethnicity, type: :uuid
     add_reference :people, :gender, type: :uuid
+
+    add_index :people, :prison_number
+    add_index :people, :criminal_records_office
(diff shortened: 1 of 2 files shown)


willfish:
I'm wondering whether we want to extend the identifier types to include those referenced here @jcavell https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/b9a00c382bfe2384184b808f1ba9ba732a93e0b1/app/models/profile/profile_identifier.rb#L7-L9?

db/migrate/20200526152251_adds_fields_to_people.rb
```diff
@@ -0,0 +1,18 @@
+class AddsFieldsToPeople < ActiveRecord::Migration[5.2]
+  def change
```
martyn-w, line 2: maybe add indexes on:
* prison_number
* criminal_records_office
* police_national_computer
* last_name
* date_of_birth
?

willfish: So we're filtering on the identifiers so I don't think we need it for the rest. We should add indexes when we need them, I think

martyn-w: 👍

alexdesi:
do you want to add in this PR the  point 2) too?
> Migrate data from latest_profile

with a rake tasks?

PR #494, opened 2020-05-27
P4-1555 Enhance Allocation status

cwrw:
### Jira link

[P4-1555](https://dsdmoj.atlassian.net/browse/P4-1555)

### What?

I have added/removed/altered:

- [x] Added a state machine to allocation statuses
- [x] Updated move updater logic to include updating allocation status
- [x] Update existing allocation cancellation logic to use state machine cancellation

### Why?

To ensure that we update an allocation status with filled/unfilled correctly we need to update the status when a move is updated. When a move is linked/unlinked to a profile or when its status changes (all done through the PATCH endpoint for a move for now) a trigger to update the `moves_count` and the `status` is done. We are utilising a state machine for the allocation for all the different steps. 
- A status is filled if all **non cancelled moves** are associated with profiles
- A status is unfilled if no **non cancelled moves** are associated with profiles
- A status is unfilled if some **non cancelled moves** are associated with profiles
- A status is unfilled if all moves are cancelled

Since we have existing allocations with `nil` values in production, we are releasing the new statuses in stages. The first stage includes a rake task to be run when this is deployed. This ensures that all statuses will be updated. This deployment before running the rake task however still accepts `nil` statuses and won't break. The next step after the rake task is run on all environments is to add validations and restrictions.

- Appends different states of data to existing records in production

### DIFF
--- a/app/models/allocation.rb
+++ b/app/models/allocation.rb
@@ -68,8 +68,10 @@ class Allocation < VersionedModel
   end
 
   def refresh_status_and_moves_count!
-    moves.not_cancelled.empty? || moves.not_cancelled.unfilled? ? unfill : fill
-    self.moves_count = moves.not_cancelled.count
+    current_moves = moves.not_cancelled
+
+    current_moves.unfilled? ? unfill : fill
+    self.moves_count = current_moves.count
 
     save!
   end
--- a/app/models/allocation.rb
+++ b/app/models/allocation.rb
@@ -49,6 +49,7 @@ class Allocation < VersionedModel
   validates :cancellation_reason, absence: true, unless: :cancelled?
 
   attribute :complex_cases, Types::JSONB.new(Allocation::ComplexCaseAnswers)
+  after_initialize :initialize_state
 
   def refresh_moves_count!
     self.moves_count = moves.not_cancelled.count
@@ -67,5 +68,15 @@ class Allocation < VersionedModel
     )
 
     save!
+  end
+
+  private
+
+  def state_machine
+    @state_machine ||= AllocationStateMachine.new(self)
+  end
+
+  def initialize_state
+    self.status = state_machine.current
   end
 end
--- a/app/models/allocation.rb
+++ b/app/models/allocation.rb
@@ -51,18 +51,18 @@ class Allocation < VersionedModel
   attribute :complex_cases, Types::JSONB.new(Allocation::ComplexCaseAnswers)
 
   after_initialize :initialize_state
-  delegate :fill, :unfill, to: :state_machine
+  delegate :fill, :unfill, :cancel, to: :state_machine
 
-  def cancel
+  def cancel_with_moves
     comment = 'Allocation was cancelled'
 
     assign_attributes(
-      status: ALLOCATION_STATUS_CANCELLED,
       cancellation_reason: CANCELLATION_REASON_OTHER,
       cancellation_reason_comment: comment,
       moves: moves.each { |move| move.cancel(comment: comment) },
       moves_count: 0,
     )
+    cancel
 
     save!
   end
--- a/app/models/allocation.rb
+++ b/app/models/allocation.rb
@@ -49,12 +49,9 @@ class Allocation < VersionedModel
   validates :cancellation_reason, absence: true, unless: :cancelled?
 
   attribute :complex_cases, Types::JSONB.new(Allocation::ComplexCaseAnswers)
-  after_initialize :initialize_state
 
-  def refresh_moves_count!
-    self.moves_count = moves.not_cancelled.count
-    save!
-  end
+  after_initialize :initialize_state
+  delegate :fill, :unfill, to: :state_machine
 
   def cancel
     comment = 'Allocation was cancelled'
@@ -70,13 +67,24 @@ class Allocation < VersionedModel
     save!
   end
 
-  private
+  def refresh_status_and_moves_count!
+    moves.not_cancelled.empty? || moves.not_cancelled.unfilled? ? unfill : fill
+    self.moves_count = moves.not_cancelled.count
+
+    save!
+  end
+
+private
 
   def state_machine
     @state_machine ||= AllocationStateMachine.new(self)
   end
 
   def initialize_state
-    self.status = state_machine.current
+    if status.present?
+      state_machine.restore!(status.to_sym)
+    else
+      self.status = state_machine.current
+    end
   end
 end
--- a/app/models/allocation.rb
+++ b/app/models/allocation.rb
@@ -51,9 +51,9 @@ class Allocation < VersionedModel
   attribute :complex_cases, Types::JSONB.new(Allocation::ComplexCaseAnswers)
 
   after_initialize :initialize_state
-  delegate :fill, :unfill, :cancel, to: :state_machine
+  delegate :fill, :unfill, to: :state_machine
 
-  def cancel_with_moves
+  def cancel
     comment = 'Allocation was cancelled'
 
     assign_attributes(
@@ -62,7 +62,7 @@ class Allocation < VersionedModel
       moves: moves.each { |move| move.cancel(comment: comment) },
       moves_count: 0,
     )
-    cancel
+    state_machine.cancel
 
     save!
   end
--- a/app/models/allocation.rb
+++ b/app/models/allocation.rb
@@ -86,7 +86,9 @@ private
     if status.present?
       state_machine.restore!(status.to_sym)
     else
-      self.status = state_machine.current
+      # TODO: restore after data is backfilled on production
+      # self.status = state_machine.current
+      self.status = moves.not_cancelled.unfilled? ? :unfilled : :filled
     end
   end
 end
--- a/app/models/move.rb
+++ b/app/models/move.rb
@@ -81,6 +81,10 @@ class Move < VersionedModel
     )
   end
 
+  def self.unfilled?
+    pluck(:profile_id).any?(&:nil?)
+  end
+
   def nomis_event_id=(event_id)
     nomis_event_ids << event_id
   end
--- a/app/models/move.rb
+++ b/app/models/move.rb
@@ -82,7 +82,7 @@ class Move < VersionedModel
   end
 
   def self.unfilled?
-    pluck(:profile_id).any?(&:nil?)
+    none? || exists?(profile_id: nil)
   end
 
   def nomis_event_id=(event_id)
--- a/app/state_machines/allocation_state_machine.rb
+++ b/app/state_machines/allocation_state_machine.rb
@@ -1,8 +1,8 @@
 class AllocationStateMachine < FiniteMachine::Definition
   initial :unfilled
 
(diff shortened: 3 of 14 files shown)


app/state_machines/allocation_state_machine.rb
```diff
@@ -0,0 +1,11 @@
+class AllocationStateMachine < FiniteMachine::Definition
+  initial :unfilled
+
+  event :fill, %i[none unfilled filled] => :filled
```
martyn-w, line 4: what is the `none` state and how do you get to it? The initial state is `unfilled` and none of the events will transition to `none` ?

lib/tasks/refresh_allocation_data.rake
```diff
@@ -0,0 +1,6 @@
+namespace :allocations do
+  desc 'Refresh allocation moves count and status'
+  task refresh_data: :environment do
+    Allocation.all.each(&:refresh_status_and_moves_count!)
```
martyn-w, line 4: maybe better to use `find_each` rather than `.all.each`, unless there is a good reason not to ?

app/models/allocation.rb
```diff
@@ -50,22 +50,41 @@ class Allocation < VersionedModel
...
 
-  def refresh_moves_count!
-    self.moves_count = moves.not_cancelled.count
-    save!
-  end
+  after_initialize :initialize_state
+  delegate :fill, :unfill, :cancel, to: :state_machine
 
-  def cancel
+  def cancel_with_moves
```
martyn-w, line 56: by defining this, is there any danger with exposing `.cancel` and `.cancel_with_moves` ? in what circumstances should you call one over the other?

app/models/allocation.rb
```diff
@@ -50,22 +50,41 @@ class Allocation < VersionedModel
...
       moves: moves.each { |move| move.cancel(comment: comment) },
       moves_count: 0,
     )
+    cancel
 
     save!
   end
+
+  def refresh_status_and_moves_count!
+    moves.not_cancelled.empty? || moves.not_cancelled.unfilled? ? unfill : fill
```
martyn-w, line 71: this looks like it will mark all allocations with at least one cancelled move as "filled". Is that the correct behaviour ?

martyn-w: oh my bad, I misread the statement. Might be good to add an explanatory comment though to avoid any confusion

spec/models/allocation_spec.rb
```diff
@@ -32,16 +32,59 @@
...
+    context 'when updating moves count' do
+      let(:cancelled_move) { create :move, :cancelled }
+      let(:proposed_move) { create :move, :proposed }
+      let(:requested_move) { create :move, :requested }
+      let(:completed_move) { create :move, :completed }
+      let(:moves) { [cancelled_move, proposed_move, requested_move, completed_move] }
+      let!(:allocation) { create :allocation, moves: moves, moves_count: 1 }
+
+      it 'updates the number of non cancelled moves' do
+        expect { allocation.refresh_status_and_moves_count! }.to change { allocation.reload.moves_count }.from(1).to(3)
```
martyn-w, line 45: I don't think(??) there is any need to the `reload` in this statement

spec/models/allocation_spec.rb
```diff
@@ -32,16 +32,59 @@
...
+    end
+
+    context 'when updating status' do
+      it 'sets status to `unfilled` if no profiles associated with uncancelled moves' do
+        moves = create_list(:move, 2, profile: nil)
+        cancelled_move = create(:move, :cancelled)
+        allocation = create(:allocation, moves: moves + [cancelled_move])
+
+        allocation.refresh_status_and_moves_count!
+        expect(allocation.reload).to be_unfilled
```
martyn-w, line 56: don't `reload` - the allocation object should be in the right state after calling `refresh_status_and_moves_count!`

spec/models/allocation_spec.rb
```diff
@@ -32,16 +32,59 @@
...
+
+      it 'sets status to `unfilled` if not all uncancelled moves are associated to profiles' do
+        move_without_profile = create(:move, profile: nil)
+        move_with_profile = create(:move)
+        cancelled_move = create(:move, :cancelled)
+
+        allocation = create(:allocation, moves: [move_without_profile, move_with_profile, cancelled_move])
+
+        allocation.refresh_status_and_moves_count!
+        expect(allocation.reload).to be_unfilled
```
martyn-w, line 67: don't reload

spec/models/allocation_spec.rb
```diff
@@ -32,16 +32,59 @@
...
 
-    it 'updates the number of non cancelled moves' do
-      expect { allocation.refresh_moves_count! }.to change { allocation.reload.moves_count }.from(1).to(3)
+      it 'sets status to `unfilled` if all moves cancelled' do
+        cancelled_move = create(:move, :cancelled)
+
+        allocation = create(:allocation, moves: [cancelled_move])
+
+        allocation.refresh_status_and_moves_count!
+        expect(allocation.reload).to be_unfilled
```
martyn-w, line 86: don't reload

spec/models/allocation_spec.rb
```diff
@@ -68,66 +111,143 @@
     end
   end
 
-  describe '#cancel' do
+  describe '#cancel_with_moves' do
```
martyn-w, line 114: don't reload in any of these tests

app/models/move.rb
```diff
@@ -81,6 +81,10 @@ def cancel(reason: CANCELLATION_REASON_OTHER, comment: nil)
     )
   end
 
+  def self.unfilled?
+    pluck(:profile_id).any?(&:nil?)
```
martyn-w, line 85: this is a bit inefficient, instead use the database with: `exists?(profile_id: nil)`

martyn-w:
👍 looks good, a few minor comments

app/state_machines/allocation_state_machine.rb
```diff
@@ -0,0 +1,11 @@
...
+  event :unfill, %i[none unfilled filled] => :unfilled
+  event :cancel, %i[none unfilled filled] => :cancelled
```
smoothcontract, line 6: It doesn't feel idiomatic to have the transitions back to the same state - that would make the graph look kinda weird. This feels cleaner:

```ruby
event :fill, %i[none unfilled] => :filled
event :unfill, %i[none filled] => :unfilled
event :cancel, %i[none unfilled filled] => :cancelled
```

smoothcontract: Also might be worth adding a terminal state for `cancelled` as there's no way to transition out of that. Simply:

```ruby
terminal :cancelled
```

spec/models/allocation_spec.rb
smoothcontract: I suspect it is - I've used a similar pattern in recent specs and without this rspec doesn't pick up the change to the model. Good to test this though :)

spec/state_machines/allocation_state_machine_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+    end
+
+    context 'when the cancel event is fired' do
+      before { machine.cancel }
+
+      it_behaves_like 'state_machine target status', :cancelled
+    end
+  end
+
+  context 'when in the none status' do
```
smoothcontract, line 70: Is this the same as having a `nil` persisted state in the allocation? If so I would expect the line below to assign `:initial_state` to `nil` rather than `:none`. I'm not sure if `:none` is a magic value representing `nil` or would end up being a named state (which we don't really want).

app/state_machines/allocation_state_machine.rb
cwrw: Since `status` currently allows `nil` values as it was introduced a while back, we should allow the allocation to exist in this `none` state (https://github.com/piotrmurach/finite_machine#24-initial) until we update all values (with the refresh rake task) in the environments. After that happens we can get rid of the `none` state and only have the 3

martyn-w: 👍 thanks for explaining

smoothcontract:
I had a go at stripping out the `none` state and the extra transitions to the current state.. generated graph looks a lot cleaner now (I will push this graph generation code soon).

[allocation_state_machine.pdf](https://github.com/ministryofjustice/hmpps-book-secure-move-api/files/4687800/allocation_state_machine.pdf)

spec/models/allocation_spec.rb
cwrw: I think we should be using it here (and elsewhere in this test), as we are persisting the value, not just amending the attributes on the fly, and those aren't picked up sometimes

martyn-w: how about instead, one extra test to check for persistence (using reload), and then remove reload from these tests? by keeping it here there is an implication that the record must always be reloaded from the database for it to be correct, which doesn't seem like a good idea.

app/models/allocation.rb
cwrw: That approach didn't work, I've amended as previously suggested: removing delegation from state machine and calling that explicitly instead

spec/models/allocation_spec.rb
cwrw: So I've revisited this, as initially when I wrote tests there were random failures but that might have been related to the way it was implemented previously. I've removed them all and kept one reload now where the test fails otherwise.

app/models/move.rb
cwrw: ah nice this is exactly what I was looking for 👍

app/state_machines/allocation_state_machine.rb
cwrw: Actually on revisiting this I ended up removing it as it is there by default, and we don't have to explicitly define it

spec/state_machines/allocation_state_machine_spec.rb
cwrw: so on revisiting this is available by default and should not be specified this way, so I removed it

spec/models/move_spec.rb
```diff
@@ -375,6 +375,31 @@
...
+    end
+
+    it 'returns false if all moves are associated to a profile' do
+      create_list(:move, 2)
+
+      expect(described_class).not_to be_unfilled
+    end
+
+    it 'returns false if no moves are present' do
+      expect(described_class).to be_unfilled
```
martyn-w, line 399: `.unfilled?` is returning true: is that the expected behaviour - if so, update the test description

spec/requests/api/v1/moves_controller_update_spec.rb
```diff
@@ -443,6 +446,57 @@
...
+                type: 'moves',
+                attributes: {
+                  status: 'requested',
+                },
+                relationships: { person: { data: { id: profile.person.id, type: 'people' } } },
+              }
+            end
+
+            it 'updates the allocation status to filled' do
+              expect(move.reload.allocation).to be_filled
```
martyn-w, line 479: 👍 good tests

PR #491, opened 2020-05-26
P4-1481 complete move

martyn-w:
### Jira link

[P4-1481](https://dsdmoj.atlassian.net/browse/P4-1481)

~NB: currently based against `P4-1406-redirect-move-endpoint` for readability.~ PR https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/484 should be merged first. (This PR builds on #484, so will be simpler once it is merged.)

### What?

I have added/removed/altered:

- [x] Added `complete a move` event
- [x] Added supporting tests

### Why?

I am doing this because:

- so we can switch to an event-based approach for changing a move's status

### Have you? (optional)

- [X] Updated API docs if necessary	

### Deployment risks (optional)

- minimal - this is a new endpoint which no one is using yet

PR #495, opened 2020-05-27
P4-1482 lockout move

martyn-w:
### Jira link

[P4-1482](https://dsdmoj.atlassian.net/browse/P4-1482)

~NB: based against PR https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/491 for readability. PRs #484 and #491 should be merged first. (This PR builds on those, so will be simpler once they are merged.)~

### What?

I have added/removed/altered:

- [x] Added lockout move event

### Why?

I am doing this because:

- so a lockout event can be recorded

### Have you? (optional)

- new endpoint so minimal risk


app/controllers/api/v1/move_events_controller.rb
```diff
@@ -62,12 +68,25 @@ def validate_params
         MoveEvents::ParamsValidator.new(event_params).validate!
       end
 
+      def validate_relationships
```
cwrw, line 71: I don't think this is really validating relationships, but rather defining different strong params for each type of different event. I would suggest we pass in different `event_params` per action, this way we are explicit in which params we allow for different types of actions, and we don't tie ourselves down further? So we can define some shared params under `base_params` maybe and do something like this:
```
def lockout_params
 ...
end

def redirect_params
  ...
end

def create_event(event_name, event_params)
...
end
```

app/services/event_log/move_runner.rb
```diff
@@ -15,6 +15,8 @@ def call
           move.to_location = event.to_location
         when 'complete'
           move.status = Move::MOVE_STATUS_COMPLETED
+        when 'lockout'
```
cwrw, line 18: if there is no action for `lockout` maybe we can leave it out of here and not `run_event_logs` in the controller? (as this will appear in the event log itself?)

martyn-w: hmm... I like the explicitness here where it is clear that no action is required for lockouts. If it were absent it might appear to be a mistake.

As for removing the call to the event runner from the controller, I think its better for the event runner to decide how to process (or not) the events, rather than the controller.

swagger/v1/post_move_lockouts_responses.yaml
```diff
@@ -0,0 +1,13 @@
+'204': # no content
```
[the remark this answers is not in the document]
martyn-w: yep, fair point 👍

app/controllers/api/v1/move_events_controller.rb
martyn-w: interesting idea, I'll have a go with that

app/services/event_log/move_runner.rb
cwrw: Sorry meant removing `run_event_logs` from the action as in: 
```
def lockouts
   create_event('lockout')
   render status: :no_content
end
```
But if your keeping it to be explicit that's fine

martyn-w: yep, I see what you mean - but I think it should be for the event runner to decide whether it has work to do (or not), not for the controller.

app/controllers/api/v1/move_events_controller.rb
```diff
@@ -51,27 +58,34 @@ def events
...
-        @event_name ||= event_params.dig(:attributes, :event_name)
+      def redirect_params
+        @redirect_params ||= params.require(:data).permit(REDIRECT_PARAMS).to_h
       end
 
-      def timestamp
-        @timestamp ||= Time.zone.parse(event_params.dig(:attributes, :timestamp))
+      def depreciated_event_params
+        # TODO: deleteme once FE updated
+        @depreciated_event_params ||= params.require(:data).permit(DEPRECIATED_EVENT_PARAMS).to_h
```
smoothcontract, line 88: `deprecated` 😁

martyn-w: oops, schoolboy error - oh well, they will be deleted soon

PR #496, opened 2020-05-27
[P4-1435] Populate person fields

willfish:
### Jira link

P4-1435

### What?

This is throwaway migratory code to populate new fields for a Person. Once we're dynamically updating these fields and have successfully run the rake task we can completely gut this from the code.

- [x] Adds a rake task that pulls out [recently-added](https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/493) fields from a Profile and adds them to all Person rows.

### Why?

As part of a four stage (add, populate, support, remove) migration to move should-have-been Person fields to the Person model.

- I've run the rake task locally and - once we're dynamically populating/validating this (see future PR) - I'll run this in a staged approach from dev -> staging -> preprod -> production. I don't anticipate any issues given how much we're testing this population task.


lib/tasks/person_populate.rake
```diff
@@ -0,0 +1,13 @@
+# TODO: Remove this once we've migrated all profile attributes to a person and are updating these attributes dynamically
+namespace :person do
+  desc 'Populates a person with profile field values'
+  task populate: :environment do
+    total = Person.count
+
+    Person.find_each do |person|
+      PersonPopulator.new(person, person.latest_profile).call
```
alexdesi, line 8: It should be nice to some output here,
since there is not any output, we do not have any idea on the progression, 
this could take tens of min minutes.
And also we do not want too much output :)
You could use the trick

```
.find_each(batch_size: 200).with_index do |move, n|
      puts "#{n}/#{total} person processed ..." if (n % 200).zero? 

```
(see move_profile.rake)

app/services/person_populator.rb
```diff
@@ -0,0 +1,47 @@
+class PersonPopulator
```
alexdesi, line 1: Let's add a TODO here,
to specify this will be removed after the profile/person refator

app/services/person_populator.rb
```diff
@@ -0,0 +1,47 @@
...
+private
+
+  attr_reader :person, :profile
+
+  STANDARD_ATTRIBUTES = %w[
+    first_names
+    last_name
+    date_of_birth
+    gender_additional_information
+    latest_nomis_booking_id
```
alexdesi, line 22: should not `latest_nomis_booking_id` remain to profile?

spec/services/person_populator_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+      }
+    end
+
+    it 'populates a person with the correct attributes' do
+      populator.call
+
+      expect(person.slice(*populated_attributes)).to eq(expected_attributes)
+    end
+  end
+end
```
alexdesi, line 87: ⭐  for adding a test for a data migration!

alexdesi: If you can add a TODO to remind us to remove it would be great!

app/services/person_populator.rb
willfish: See https://dsdmoj.atlassian.net/browse/P4-1435 :). AFAIK a person and a profile and a booking are three different things. latest_nomis_booking_id is the same as the other identifiers except for the fact that it doesn't represent a person in other systems. It references a booking. Probably having the identifiers on a person makes more sense so I agree with Jonny in deciding to move this here.

willfish: Said another way - a person may have many bookings (times they were booked into prison)

app/services/person_populator.rb
willfish: Done

lib/tasks/person_populate.rake
willfish: Done :)

spec/services/person_populator_spec.rb
willfish: 🎉. Done

app/services/person_populator.rb
alexdesi: I had a chat with @jcavell , I was worried because we have the same field in the profile table, and that is confusing.
So we agreed on renaming the field in the profile (Jonny created a ticket for that)

PR #497, opened 2020-05-27
Rename person_id to person_id_backup

alexdesi:
### Jira link
https://dsdmoj.atlassian.net/browse/P4-1453

### What?
This PR aims to remove the `person_id` from the `Moves` table.

To proceed safely, this PR just rename the field to `person_id_backup`,
so in case of disaster, we can easily rollback.

### Why?
To refactor Person/Profile.

Small risk, but mitigated by renaming the field as the first step (instead of just removing it).


willfish:
Let's look at this as part of the final cleanup of person/profile - we'll have a session to talk about this

PR #498, opened 2020-05-27
Removing Rswag

alexdesi:
### Jira link
https://dsdmoj.atlassian.net/browse/P4-1235

### What?
@jcavell started this to refactor the clumsy way Rswag/Rspec generates the API docs.

### Why?
The main point is removing Rswag, and decouple the creation of the API docs from the spec run.


willfish:
I think we can look at this separately later - I might take a different approach, etc.

PR #499, opened 2020-05-27
[P4-1435] Populate person fields when updating/creating a Person/Profile

willfish:
### Jira link

P4-1435

### What?

This is part one of a two-part PR to make sure the fields on `Person` are dynamically updated in addition to those of the `Profile`. These two will need to be merged before we can move to running the [rake task](https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/496) to backfill all existing `Person` records.

- [x] Adds support for populating new [fields](https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/493/) when **updating** a `Person`
- [x] Adds support for populating new [fields](https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/493/) when **creating** a `Person`

### Why?

We want to update a person every time we update a `Profile` with the new
attributes that should be on a `Person`.

I'm piggybacking off of the existing mechanism, here, as I suspect when
we do the separation work for CRUDing a `Person` and a `Profile` we'll
potentially use a different pattern for updating attributes and eventually chuck the updaters/creators and inheritance hierarchies for a simpler process.

### Not included

- When we synchronise nomis records to create a Person/Profile we won't be populating the person fields - this will come in a follow on PR


app/controllers/api/v1/people_controller.rb
```diff
@@ -111,11 +111,11 @@ def creator
       end
 
       def updater
-        @updater ||= People::Updater.new(params[:id], person_params)
+        @updater ||= People::Updater.new(person, person_params)
       end
 
       def person
-        @person ||= Person.find(params[:person_id])
+        @person ||= Person.find(params[:person_id] || params[:id])
```
martyn-w, line 118: why does the id change names between `:person_id` and `:id`? is the routing configured OK?

app/services/people/crud.rb
```diff
@@ -24,12 +24,30 @@ def assessment_answers
...
+        attributes.slice(*Person::IDENTIFIER_TYPES)
+      else
+        {}
+      end
     end
 
     def profile_params
       params[:attributes].slice(*ATTRIBUTES)
     end
+    alias_method :person_params, :profile_params
```
martyn-w, line 45: is this a good idea to alias the two? we may get confused later down the line

app/controllers/api/v1/people_controller.rb
willfish: Nested routes 😢

willfish:
```                     api_v1_person_images GET    /api/v1/people/:person_id/images(.:format)                                               api/v1/people#image
                api_v1_person_court_cases GET    /api/v1/people/:person_id/court_cases(.:format)                                          api/v1/people#court_cases
                  api_v1_person_timetable GET    /api/v1/people/:person_id/timetable(.:format)                                            api/v1/people#timetable
                            api_v1_people GET    /api/v1/people(.:format)                                                                 api/v1/people#index
                                          POST   /api/v1/people(.:format)                                                                 api/v1/people#create
                            api_v1_person PATCH  /api/v1/people/:id(.:format)                                                             api/v1/people#update
                                          PUT    /api/v1/people/:id(.:format)                                                             api/v1/people#update```

File: app/services/people/crud.rb (hunk 2 above)
Comment (willfish) [reply app/services/people/crud.rb:45]: We plan to separate them later. Watch this space :)

File: app/controllers/api/v1/people_controller.rb (hunk 1 above)
Comment (martyn-w) [reply app/controllers/api/v1/people_controller.rb:118]: Boo :-(

File: app/models/person.rb [hunk 3]
```diff
@@ -1,12 +1,21 @@
...
 
+  has_many :profiles, dependent: :destroy
   has_many :moves, through: :profiles
 
+  belongs_to :ethnicity, optional: true
+  belongs_to :gender, optional: true
+
   has_one_attached :image
 
+  validates :last_name, presence: true
```
cwrw, line 16: before we run the backfilling task will this not cause existing records to be invalid?

app/services/people/creator.rb
```diff
@@ -2,29 +2,34 @@
...
+        create_person
+        create_profile
+        profile.save!
+      end
     end
 
   private
 
+    def create_person
+      @person = Person.new(
```
cwrw, line 18: How do we ensure that this person does not exist before, now we have migrated the new list of fields over? Will we risk creating duplicate profiles if they exist before? (maybe this is done somewhere else I've missed)

app/models/person.rb
willfish: Yes it will. I'll remove it, but leave the fixes for the test failures it caused by adding it.

app/services/people/creator.rb
willfish: So this is true before/after this change. I haven't affected the behaviour around uniqueness of either a Profile or Person. It seems like we could do with a dedicated chat/story around our approach to uniqueness of a person/profile. My suspicion is that the frontend workflows handle this currently.

cwrw: spoke in person but for posterity: this still isn't an issue, we will ensure uniqueness in the new endpoints

spec/factories/person.rb
```diff
@@ -2,8 +2,19 @@
...
+    association(:ethnicity)
+    association(:gender)
+
     profiles { build_list :profile, 1 }
 
+    first_names { Faker::Name.first_name }
+    last_name { Faker::Name.last_name }
+    date_of_birth { Date.new(1980, 10, 20) }
+
+    police_national_computer { 'AB/1234567' }
```
cwrw, line 14: it is worth using faker here as well for those three identifiers, as those ideally should be unique  in further work?

spec/factories/person.rb
```diff
@@ -12,4 +23,9 @@
       end
     end
   end
+
+  factory :person_without_profiles, class: 'Person' do
```
cwrw, line 27: don't think we need to define the class here anymore as it's under the correct factory?

willfish: I'm actually not sure whether the symbol or the file are used to know the class :). I'll have a play

willfish: Uses the symbol, sadly :(. Could do with refactoring our factories a bunch I think.

cwrw: ah your right, somehow in my head this was a trait not a factory, as you were :D

spec/factories/person.rb
willfish: Yeah, I'll use a random string sequence matching the number of chars for each identifier. Doesn't have to be perfect :) - just needs to be unique.

willfish:
```
 Running tests in:

 ✓ With unfound person
 ✓ With existing person

 ✓ With a new person

 ✓ Reason - `Made in error`
 ✓ Reason - `Another reason`

 New move from Prison to Court
 ✓ With unfound person
 ✓ With existing person

 ✓ With existing person

 Download moves
 ✓ As Police user
 ✓ As Prison user
 ✓ As STC user
 ✓ As Supplier user

 Smoke tests
 ✓ As Police user
 ✓ As STC user
 ✓ As Prison user
 ✓ As Supplier user
 ✓ As OCA user


 18 passed (2m 54s)
```

willfish:
Running the integration tests ^^

PR #502, opened 2020-05-28
[P4-1563] Improve docker image size

mafonso:
### Jira link

P4-1563

### What?

I have added/removed/altered:

- [x] Used alpine ruby as the base image
- [x] Installed minimum set of OS packages to satisfy dependencies.

### Why?

I am doing this because:

- The current docker image is too big ~1.6GB, which takes a longer than desirable to download, delaying pod start.
- Reduces the CVE footprint by not having a full ubuntu/OS distro running in the container

### Have you? (optional)

- This PR does not change anything on the application resources or ruby/gem version. It changes however the underling OS and uses different implementation of some system libraries, which might have an impact on natively built gems.  This is considered to be very low risk as alpine is generally used in production for a number of workloads, including ruby, node, java, etc. 

The current test suites don't validate a running image, so testing was done by manually poking some API endpoints.

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,18 @@
 FROM ruby:2.6.6-alpine as build-stage
 
 ENV RAILS_ENV=production
-ENV BUNDLE_APP_CONFIG="app/.bundle"
+ENV BUNDLE_WITHOUT="development:test"
+ENV BUNDLE_FROZEN="true"
 
 WORKDIR /app
-
-RUN apk add git build-base tzdata postgresql-dev
+RUN apk --update add git build-base postgresql-dev
 
 COPY . /app
 RUN gem update bundler --no-document
-RUN bundle install --without="development test"  --jobs 4 --retry 3
+RUN bundle install --jobs 4 --retry 3 \
+     && rm -rf /usr/local/bundle/cache/*.gem \
+     && find /usr/local/bundle/gems/ -name "*.c" -delete \
+     && find /usr/local/bundle/gems/ -name "*.o" -delete
 
 ############### End of Build step ###############
 FROM ruby:2.6.6-alpine
@@ -26,22 +29,18 @@ ENV APP_GIT_COMMIT ${APP_GIT_COMMIT}
 ENV APPUID 1000
 
 ENV RAILS_ENV production
-ENV BUNDLE_APP_CONFIG="app/.bundle"
 
 ENV PUMA_PORT 3000
 EXPOSE $PUMA_PORT
 
 RUN addgroup -g $APPUID -S appgroup && \
-    adduser -u $APPUID -S appuser -G appgroup
+    adduser -u $APPUID -S appuser -G appgroup -h /app
 
-RUN apk add tzdata postgresql-dev
+RUN apk add --no-cache tzdata postgresql-dev
 
 WORKDIR /app
-COPY --from=build-stage /usr/local/bundle /usr/local/bundle
-COPY --from=build-stage /app /app
-
-RUN  chown -R appuser:appgroup /app  && \
-     chown -R appuser:appgroup /home/appuser
+COPY --chown=appuser:appgroup --from=build-stage /app /app
+COPY --chown=appuser:appgroup --from=build-stage /usr/local/bundle /usr/local/bundle
 
 USER $APPUID
 ENTRYPOINT ["./run.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ ENV BUNDLE_WITHOUT="development:test"
 ENV BUNDLE_FROZEN="true"
 
 WORKDIR /app
-RUN apk --update add git build-base postgresql-dev
+RUN apk --update --no-cache add git build-base postgresql-dev
 
 COPY . /app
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -1,4 +1,18 @@
...
+ENV RAILS_ENV=production
+ENV BUNDLE_APP_CONFIG="app/.bundle"
+
+WORKDIR /app
+
+RUN apk add git build-base tzdata postgresql-dev
+
+COPY . /app
+RUN gem update bundler --no-document
+RUN bundle install --without="development test"  --jobs 4 --retry 3
```
cwrw, line 12: maybe if we also append:
```
 && rm -rf /usr/local/bundle/cache/*.gem \
     && find /usr/local/bundle/gems/ -name "*.c" -delete \
     && find /usr/local/bundle/gems/ -name "*.o" -delete
```
if will shave off a bit from the final size of the image

Dockerfile
```diff
@@ -1,4 +1,18 @@
...
+
+############### End of Build step ###############
+FROM ruby:2.6.6-alpine
```
cwrw, line 15: 👍 for separate build steps

Dockerfile
```diff
@@ -9,28 +23,26 @@ ENV APP_BUILD_TAG ${APP_BUILD_TAG}
...
-
-# ugrade bundler to 2.1.4
-RUN gem update bundler --no-document
-
-COPY . /app
+RUN addgroup -g $APPUID -S appgroup && \
+    adduser -u $APPUID -S appuser -G appgroup
 
-RUN bundle install --verbose --without="development test" --jobs 4 --retry 3
+RUN apk add tzdata postgresql-dev
```
cwrw, line 37: could we append `--update here, so it becomes `apk add --update` to get the latest list of packages?

cwrw: Also we might add `rm -rf /var/cache/apk/*` although not sure how much that shaves off the final package size here?

mafonso:
> I did find I had to run the `docker-compose up` command twice, as the first time it failed to create the database, as the postgres container was still initializing when the `rake db:create` command ran:

Good spot!  I was able to reproduce this by nuking the `.data` dir.   A fresh Postgres runs `pginit` once after first start taking additional time.  I'll implement one of the wait-for techniques.

Dockerfile
mafonso: Yes, it does shave 19MB irc.  It's still about 10% of the new size, so worth taking out.
I had `--no-cache` initially but I must have lost it in a rebase and it was causing deprecation messages as well.
I'll reintroduce a way to remove this bloat again.

Dockerfile
mafonso: Regarding the cache, I also had `--no-cache` here to the same effect. I must have dropped a commit instead of flatten it while rebasing.  This one saves 2MB as I recall. 

and 👍 for the --update. I'm a bit on two minds about this, as in one occasion it brought a broken updated µlibc. But on the other hand it usually get's rid of CVEs quite quickly, so it's worth have it.

Dockerfile
```diff
@@ -9,28 +26,22 @@ ENV APP_BUILD_TAG ${APP_BUILD_TAG}
...
-
-# ugrade bundler to 2.1.4
-RUN gem update bundler --no-document
-
-COPY . /app
+RUN addgroup -g $APPUID -S appgroup && \
+    adduser -u $APPUID -S appuser -G appgroup -h /app
 
-RUN bundle install --verbose --without="development test" --jobs 4 --retry 3
+RUN apk add --no-cache tzdata postgresql-dev
```
cwrw, line 39: ah is it that we can either have update or no cache?

mafonso: This just prevents caching so you don't have to delete it after. Haven't covered the update yet.

Dockerfile
```diff
@@ -1,15 +1,18 @@
 FROM ruby:2.6.6-alpine as build-stage
 
 ENV RAILS_ENV=production
-ENV BUNDLE_APP_CONFIG="app/.bundle"
+ENV BUNDLE_WITHOUT="development:test"
+ENV BUNDLE_FROZEN="true"
```
mafonso, line 6: Using ENV variables for bundle instead of the command line arguments, which are on a deprecation course. Other option is having bundle config file which I'm not so keen to use in this context.

Dockerfile
```diff
@@ -1,15 +1,18 @@
...
-RUN apk add git build-base tzdata postgresql-dev
+RUN apk --update add git build-base postgresql-dev
 
 COPY . /app
 RUN gem update bundler --no-document
-RUN bundle install --without="development test"  --jobs 4 --retry 3
+RUN bundle install --jobs 4 --retry 3 \
+     && rm -rf /usr/local/bundle/cache/*.gem \
+     && find /usr/local/bundle/gems/ -name "*.c" -delete \
+     && find /usr/local/bundle/gems/ -name "*.o" -delete
```
mafonso, line 15: Neither the deprecated `--no-cache` nor it's ENV equivalent seem to work, not does setting the cache location to a disposable path such as `/tmp`.  see https://github.com/rubygems/rubygems/issues/3225

Using the typical workaround for now.

PR #505, opened 2020-05-28
P4-1544 Add new endpoint for cancelling an allocation with reason and comment

smoothcontract:
### Jira link

P4-1544

### What?

- [x] New endpoint allows a POST to `/allocations/:id/cancel` and accepts a cancellation_reason and optional cancellation_reason_comment attribute.

### Why?

- We need to support providing a custom cancellation reason and optional cancellation reason comment in the front end (at the moment both of those are using hard coded values).
- The existing "events" endpoint for allocations has now been deprecated as we've decided to move towards explicit "verb" endpoints for dedicated actions.
- For now the existing event endpoint will be retained as this is used in the front end - but it doesn't support a custom cancellation reason or cancellation reason comment. Once the new endpoint is adopted within the front end then the deprecated events endpoint can be removed which will clean up quite a bit of code; left lots of TODO's to annotate where this needs to happen.

### Have you? (optional)

- Introduces a new api so minimal risk, no change to the existing "events" api


app/controllers/api/v1/allocation_events_controller.rb
```diff
@@ -3,15 +3,21 @@
...
     class AllocationEventsController < ApiController
-      before_action :validate_params, only: :create
-      after_action :send_move_notifications, only: :create
+      before_action :validate_params, only: :events
+      after_action :send_move_notifications
 
-      def create
+      def cancel
+        allocation.cancel(cancellation_details)
+
```
cwrw, line 11: currently in the move event controller, we are replaying the event logs using `EventLog::MoveRunner.new(move).call`. Here although we are cancelling an allocation, it is also cancelling moves as well. Is there a way we can integrate those into the move runner?

app/controllers/api/v1/allocation_events_controller.rb
```diff
@@ -23,21 +29,41 @@ def create
...
       ].freeze
 
+      PERMITTED_CANCEL_PARAMS = [
+        attributes: %i[cancellation_reason cancellation_reason_comment],
+      ].freeze
+
       def validate_params
-        AllocationEvents::ParamsValidator.new(event_params).validate!(action_name.to_sym)
+        # TODO: remove ParamsValidator completely when allocation `events` endpoint is no longer in use
+        AllocationEvents::ParamsValidator.new(event_params).validate!(:cancel)
```
cwrw, line 44: we should still be validating parameters, even under the new methods as well

spec/requests/api/v1/allocation_events_controller_cancel_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+          data: {
+            type: 'cancel',
+            attributes: {
+              cancellation_reason: cancellation_reason,
+            },
+          },
+        }
+      end
+
+      it 'updates the allocation cancellation reason comment' do
```
cwrw, line 80: I'm wondering if we should instead of falling back to the default return a validation error, as initially the default was set until we implemented this 🤔

smoothcontract: I don't think an error would be appropriate, as the cancellation reason comment is optional. Perhaps instead the reason comment param could be omitted and set as `nil` in both the allocation and move models.

app/controllers/api/v1/allocation_events_controller.rb
smoothcontract: The params validator isn't applicable to cancelling as it's checking the format of a timestamp which we don't need and another type parameter which isn't relevant. Instead I'm relying on the model validation rules to ensure that the cancellation reason is set to a valid enum value when the allocation is cancelled - no need for a validator service just for that.

app/controllers/api/v1/allocation_events_controller.rb
smoothcontract: I don't really know what this is or whether it's still relevant to the new approach. Maybe we can discuss?

smoothcontract: Added new event log goodness - see new `Eventable` concern.

app/controllers/api/v1/allocation_events_controller.rb
smoothcontract: Reinstated validation of timestamp for cancel event via the params validator.

app/controllers/concerns/eventable.rb
```diff
@@ -0,0 +1,30 @@
+module Eventable
+  extend ActiveSupport::Concern
+
+  def supplier_id
+    # NB: not all events will have a supplier_id so this could well be nil
+    @supplier_id ||= current_user.owner&.id
+  end
+
+  def process_event(moves, event_name, event_params)
```
smoothcontract, line 9: This all came from move events controller - I needed to add a `move` param so that it can be called for each move within an allocation. To make this nicer to use this supports passing either a single move or a move relationship - hence the array flattening.

app/models/allocation.rb
```diff
@@ -53,15 +53,17 @@ class Allocation < VersionedModel
   after_initialize :initialize_state
   delegate :fill, :unfill, to: :state_machine
 
-  def cancel
-    comment = 'Allocation was cancelled'
-
+  def cancel(cancel_moves: true, reason: CANCELLATION_REASON_OTHER, comment: 'Allocation was cancelled')
```
smoothcontract, line 56: This is pretty clunky now - but mostly temporary. With the new cancellation endpoint we don't actually want to cancel the moves here - instead we create events that trigger move cancellation and run those events in sequence. For the legacy endpoint we do want the original behaviour of cancelling moves here.

Also it's now possible to pass cancellation reason and comment as params (in the new endpoint) where previously (in the old endpoint) those were hardcoded values.

app/services/allocation_events/params_validator.rb
```diff
@@ -4,9 +4,10 @@ module AllocationEvents
   class ParamsValidator
     include ActiveModel::Validations
 
+    # TODO: remove event_name when allocation `events` endpoint is no longer in use
     attr_reader :event_name, :timestamp
 
-    validates :event_name, inclusion: %w[cancel], presence: true
+    validates :event_name, inclusion: %w[cancel], allow_nil: true
```
smoothcontract, line 10: `event_name` only really makes sense in the context of the previous `events` approach - so if the event name is blank this validator will pass the params, but we'll still get a bad request error raised in the `events` method of allocation events controller. Actually I'm pretty sure validation of `event_name` could be removed now if preferred as the validation occurs in two places. See: https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/505/files#diff-cf52449dc55f39cafad70842929410deR29

spec/models/allocation_spec.rb
```diff
@@ -150,7 +162,13 @@
       expect(allocation.moves.first.cancellation_reason).to eq(Move::CANCELLATION_REASON_OTHER)
```
smoothcontract, line 162: I need to fix this - it should be testing the move `cancellation_reason_comment`. Also need to agree the behaviour if no cancellation reason comment is passed to the new endpoint - this is an optional field so should we just allow and persist blank comments?

spec/requests/api/v1/allocation_events_controller_cancel_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+        it_behaves_like 'an endpoint that responds with error 422' do
+          let(:errors_422) { [{ 'title' => 'Unprocessable entity', 'detail' => 'Cancellation reason is not included in the list' }] }
+        end
+      end
+
+      context 'with a missing cancellation_reason' do
+        let(:cancellation_reason) { nil }
+
+        it_behaves_like 'an endpoint that responds with error 422' do
+          let(:errors_422) { [{ 'title' => 'Unprocessable entity', 'detail' => 'Cancellation reason is not included in the list' }] }
```
smoothcontract, line 143: In the move events stuff the params validator is checking that a cancellation reason is provided for a cancel event. I didn't add that, instead relying on the allocation model to validate that a cancellation reason is provided when the state is cancelled. That avoids adding validation in two places; perhaps the move params validator should do the same?

swagger/v1/post_allocation_cancel.yaml
```diff
@@ -0,0 +1,33 @@
...
+    type:
+      type: string
+      example: cancel
+      enum:
+      - cancel
+      description: The type of this object - always `cancel`
+    attributes:
+      type: object
+      required:
+      - cancellation_reason
```
smoothcontract, line 16: TODO: add timestamp here

app/controllers/api/v1/allocation_events_controller.rb
```diff
@@ -23,38 +35,69 @@ def create
...
+      end
+
+      def cancel_params
+        @cancel_params ||= params.require(:data).permit(PERMITTED_CANCEL_PARAMS).to_h
+      end
+
+      def cancel_move_params
+        cancel_params.tap do |params|
+          params[:attributes][:cancellation_reason] = Move::CANCELLATION_REASON_OTHER
```
smoothcontract, line 59: Open to better suggestions for this - there's a rule that says we should always provide the reason `other` for the cancelled underlying moves regardless of the reason chosen for cancelling the allocation.

martyn-w: maybe add a comment to record that fact:  `# NB: we should always provide the reason other for the cancelled underlying moves regardless of the reason chosen for cancelling the allocation`

app/controllers/api/v1/move_events_controller.rb
```diff
@@ -3,6 +3,8 @@
 module Api
   module V1
     class MoveEventsController < ApiController
+      include Eventable
```
martyn-w, line 6: good idea 👍

app/controllers/concerns/eventable.rb
martyn-w: 👍

app/controllers/concerns/eventable.rb
```diff
@@ -0,0 +1,30 @@
+module Eventable
```
martyn-w, line 1: possibly this should be named `MoveEventable` or `Moves::Eventable` as its focus is purely on `move` events. Coming down the line are `journey` events, `people` events (`allocation` events??).

app/controllers/api/v1/allocation_events_controller.rb
```diff
@@ -3,15 +3,27 @@
...
+        render status: :no_content
+      end
+
+      def events
+        # TODO: this method should be deleted, but kept here until the front end is updated
+        validate_params!(event_params)
         case event_name
         when 'cancel'
           allocation.cancel
+          send_move_notifications
```
martyn-w, line 25: any reason why this can't go via the event log, the same as the `cancel` method above ?
```
        allocation.transaction do
          allocation.cancel(cancellation_details)
          process_event(allocation.moves, Event::CANCEL, cancel_move_params)
        end
```
? I know it will be deleted soon, but would be nice if it did the same thing under the hood.

app/services/allocation_events/params_validator.rb
martyn-w: yep fine to remove as its being caught elsewhere 👍 
... however you may need a validator for `type` though?

spec/requests/api/v1/allocation_events_controller_cancel_spec.rb
martyn-w: I am a bit wary about pushing the validation of events too far down the stack to the point of running the event. We had a bug last week where an event describing an invalid object was successfully created and saved, and then run. On running it crashed as the event was invalid, but it meant that no new events could be added for the object as the invalid one broke the chain.

In this implementation the event creator and the event runner are wrapped in a transaction in the controller, so any errors will not be saved to the db, so all should be well. Nevertheless, it seems a bit unsafe to rely on the runner as the primary validator.

app/controllers/concerns/eventable.rb
smoothcontract: Cool I'll try that 👍

app/services/allocation_events/params_validator.rb
smoothcontract: There's only one event at the moment, so no need for `type` to be introduced yet. I'd prefer to add that when the next event is added, or at least after we get rid of `event_name`.

spec/requests/api/v1/allocation_events_controller_cancel_spec.rb
smoothcontract: Hmm.. it's an interesting point. As you say, it's safe in this implementation. My counter argument would be that it's not a great pattern to be replicating validation rules in two places (i.e. the model and the params validator). On balance I prefer this approach unless there's a strong preference for something else?

martyn-w: I appreciate the point (and agree) that duplication of code is not a good thing. However, I don't think that we can safely skip validation at this level.

How about instead moving the relevant validations which are needed on the event and on the move/allocation into their own concern, e.g.

```
module MoveEvents
  module CancellationReasonValidator
    extend ActiveSupport::Concern
    included do
      validates :cancellation_reason, inclusion: { in: Move::CANCELLATION_REASONS }, if: -> { type == 'cancel' }
    end
  end
end
```

then in the move events `param_validator.rb` we can include it:
```
module MoveEvents
  class ParamsValidator
    include CancellationReasonValidator
    ...
```

and likewise include it in the move model / wherever else it should be validated. That way there is only one validation rule and it can be used in multiple places. Will that do the job?

app/controllers/api/v1/allocation_events_controller.rb
smoothcontract: I'd prefer not to mess with this existing method - it's not creating events at present and is used in production so is safer to switch to the new API endpoint and then start using the events approach.

martyn-w: 👍

app/controllers/api/v1/allocation_events_controller.rb
cwrw: Ah I wasn't aware of this, this is new. I thought initially we would mirror all the reasons defined on the allocation cancellations here (and add the ones that don't conform with the defined moves cancellation reasons), unless Other was selected on allocation where we would add the comment.

spec/requests/api/v1/allocation_events_controller_cancel_spec.rb
smoothcontract: I've amended to allow a blank cancellation reason comment to be passed in the request and then set this against both the allocation and move.

spec/requests/api/v1/allocation_events_controller_cancel_spec.rb
smoothcontract: Moving the validation to a shared concern and then including it in the params validator and model still means it's in two places so needs two sets of specs. I'm not convinced it's a great pattern and prefer that the bulk validation work lives in the model - particularly the more complex stuff that depends on business rules or other state of the model. I think it's appropriate to check the existence of specific parameters in a request, and that date formats look sensible but we shouldn't take that approach too far and try to do everything in the params validator.

It's safe to apply this approach within a database transaction as if something goes boom the data is left in a consistent state. I'll add a spec to ensure that no events get created.

app/controllers/api/v1/allocation_events_controller.rb
smoothcontract: Discussed in person - unfortunately moves and allocations have a different set of cancellation reasons, so for simplicity the move cancellation reason is always set to `other`. We could explore a mapping approach or try to standardise the cancellation reasons across moves and allocations in future.

PR #507, opened 2020-05-28
Setup API V2 Swagger docs

alexdesi:
### What?

The goal of this is to set up the swagger doc example for  API V2.

In details, this PR:
- add new swagger for V2
- add docs for the endpoints **get, post and patch v2/people**

Screen Shot 2020-06-03 at 18 21 29

Disable endpoints:
Screen Shot 2020-06-04 at 16 56 42


cwrw:
I guess we can close this @alexdesi?

spec/swagger_helper.rb
```diff
@@ -166,6 +166,150 @@ def load_swagger_yaml(*relative_path)
       },
       paths: load_swagger_yaml('hand_coded_paths.yaml'),
     },
+
```
alexdesi, line 169: I've just realised that we do not need to change the spec,
I think we use this only for generating docs with Rswag.

spec/swagger/definitions/hand_coded_paths_v1.yaml
```diff
@@ -93,8 +93,8 @@
         enum:
         - asc
         - desc
-    - $ref: pagination_parameters.yaml#/Page
-    - $ref: pagination_parameters.yaml#/PerPage
+    - $ref: /pagination_parameters.yaml#/Page
```
willfish, line 96: I guess the prefix is necessary?

spec/swagger/definitions/hand_coded_paths_v2.yaml
```diff
@@ -0,0 +1,102 @@
+/people:
+  get:
+    deprecated: true
```
willfish, line 3: 👍

swagger/v1/get_genders_responses.yaml
```diff
@@ -6,7 +6,7 @@
     data:
       type: array
       items:
-        $ref: gender.yaml#/Gender
+        $ref: '../gender.yaml#/Gender'
```
willfish, line 9: Confused by this?

swagger/v2/get_people_responses.yaml
```diff
@@ -11,17 +11,15 @@
       type: array
       items:
         anyOf:
-        - $ref: gender.yaml#/Gender
-        - $ref: ethnicity.yaml#/Ethnicity
-        - $ref: profile.yaml#/Profile
```
willfish, line 16: I guess we'll implement this on the move

spec/swagger_helper.rb
```diff
@@ -136,7 +136,7 @@ def load_swagger_yaml(*relative_path)
             "$ref": 'nationality.yaml#/Nationality',
           },
           Pagination: {
-            "$ref": 'pagination.yaml#/Pagination',
+            "$ref": '../shared/pagination.yaml#/Pagination',
```
cwrw, line 139: 🔥

spec/swagger/definitions/hand_coded_paths_v2.yaml
```diff
@@ -0,0 +1,102 @@
...
+    summary: '[DO NOT USE] Returns a list of people'
```
cwrw, line 4: 👍

swagger/v2/get_people_responses.yaml
```diff
@@ -11,17 +11,15 @@
       type: array
       items:
         anyOf:
-        - $ref: gender.yaml#/Gender
```
cwrw, line 14: if we return nothing for those, do we need to add something else here like an empty array? as we can possibly not set them

alexdesi: for now, I think we can add Gender and Ethnicity, in one of the next PRs we can add profile too

alexdesi: in V2, it is possible to create a new person with empty Gender/Ethnicity,
so `nullable: true` seems to be a good solution

spec/swagger/definitions/hand_coded_paths_v1.yaml
alexdesi: Actually no, this was a "partial Undo" 😄

swagger/v2/get_people_responses.yaml
alexdesi: if you refer to the profile, I'm going to add it back, since it is a supported relationship

alexdesi:
I've squashed just squashed all those messy commits into just one.

swagger/v2/get_people_responses.yaml
```diff
@@ -11,17 +11,17 @@
...
-        - $ref: ethnicity.yaml#/Ethnicity
-        - $ref: profile.yaml#/Profile
-        - {}
+          - $ref: '../v1/gender.yaml#/Gender'
+          - $ref: '../v1/ethnicity.yaml#/Ethnicity'
+          - $ref: '../v1/profile.yaml#/Profile'
```
cwrw, line 16: this needs to not have v1 as we have a v2 version

spec/support/json_schemas.rb
```diff
@@ -4,8 +4,11 @@
   def load_schema(file_name, version: 'v1')
     return unless File.file?("#{Rails.root}/swagger/#{version}/#{file_name}")
 
-    schema = load_yaml_schema(file_name)
-    JSON::Validator.add_schema(JSON::Schema.new(schema, file_name))
+    schema = load_yaml_schema(file_name, version: version)
+
+    # TODO: when moving to fully support v1 and v2 stop loading v1 twice
```
cwrw, line 9: When adding documentation for `V2`, we have to reference some "shared" documentation in `V1`, example: `'../v1/pagination_parameters.yaml#/PerPage'`. This works well when we swaggerize our schemas, and is compliant with swagger. However when we add schemas to `JSON::Validator`, we need to add them under the correct path or it won't be able to parse them. For instance it would attempt to read `'../v1/pagination_parameters.yaml#/PerPage'` and try to go up a directory to `v1` which is not supported before. If we were to add paths for v1 and v2, this would mean we would have to reference things everywhere with `v1` to support it. For now until we fully support v2, load v1 under the root path and v1.

PR #508, opened 2020-05-28
Add GET /api/v2/people

alexdesi:
### Jira link

https://dsdmoj.atlassian.net/browse/P4-1493

### What?

Add api/V2/people endpoint

API docs are updated in a different PR, to keep this PR clean.

### Why?
We are going to have a new set of endpoint having a clear separation between People and Profile.

### Have you? (optional)

No risk, the new endpoint will not change the behaviour of the old endpoints.

### DIFF
--- a/app/controllers/api/v2/people_controller.rb
+++ b/app/controllers/api/v2/people_controller.rb
@@ -3,9 +3,10 @@
 module Api
   module V2
     class PeopleController < ApiController
+      before_action :validate_include_params
+
       def index
-        # people = People::Finder.new(filter_params).call
-        people = Person.where(filter_params)
+        people = ::V2::People::Finder.new(filter_params).call
 
         paginate people, include: included_relationships, each_serializer: ::V2::PersonSerializer
       end
@@ -20,6 +21,10 @@ module Api
 
       def included_relationships
         IncludeParamHandler.new(params).call
+      end
+
+      def validate_include_params
+        ::V2::People::IncludeParamsValidator.new(included_relationships).validate!
       end
     end
   end
--- a/app/controllers/api/v2/people_controller.rb
+++ b/app/controllers/api/v2/people_controller.rb
@@ -7,7 +7,7 @@ module Api
         # people = People::Finder.new(filter_params).call
         people = Person.where(filter_params)
 
-        paginate people, include: included_relationships
+        paginate people, include: included_relationships, each_serializer: ::V2::PersonSerializer
       end
 
     private
--- a/app/controllers/api/v2/people_controller.rb
+++ b/app/controllers/api/v2/people_controller.rb
@@ -4,88 +4,18 @@ module Api
   module V2
     class PeopleController < ApiController
       def index
-        prison_number = filter_params[:prison_number]
-
-        import_person_from_nomis(prison_number) if prison_number
-
-        people = People::Finder.new(filter_params).call
+        # people = People::Finder.new(filter_params).call
+        people = Person.all
 
         paginate people, include: PersonSerializer::INCLUDED_DETAIL
       end
 
     private
 
-      def json_api_errors_for(error)
-        { errors: [error.json_api_error] }
-      end
-
-      def import_person_from_nomis(prison_number)
-        # This prevents us from blaming the current user/application for the NOMIS sync
-        PaperTrail.request(whodunnit: nil) do
-          Moves::ImportPeople.new([prison_number]).call
-        end
-      rescue StandardError => e
-        Raven.capture_exception(e)
-      end
-
-      PERMITTED_COURT_CASE_FILTER_PARAMS = %i[active].freeze
-      PERMITTED_TIMETABLE_FILTER_PARAMS = %i[date_from date_to].freeze
-      PERMITTED_FILTER_PARAMS = %i[police_national_computer prison_number].freeze
-      PERSON_ATTRIBUTES = [
-        :first_names,
-        :last_name,
-        :date_of_birth,
-        :gender_additional_information,
-        assessment_answers: [%i[key
-                                date
-                                expiry_data
-                                category
-                                title
-                                comments
-                                assessment_question_id
-                                nomis_alert_type
-                                nomis_alert_type_description
-                                nomis_alert_code
-                                nomis_alert_description
-                                created_at
-                                expires_at
-                                imported_from_nomis]],
-        identifiers: [%i[value identifier_type]],
-      ].freeze
-      PERMITTED_PERSON_PARAMS = [:type, attributes: PERSON_ATTRIBUTES, relationships: {}].freeze
-
-      def creator
-        @creator ||= People::Creator.new(person_params)
-      end
-
-      def updater
-        @updater ||= People::Updater.new(params[:id], person_params)
-      end
-
-      def person
-        @person ||= Person.find(params[:person_id])
-      end
-
-      def render_person(person, status)
-        render json: person, status: status, include: PersonSerializer::INCLUDED_DETAIL
-      end
-
-      def person_params
-        params.require(:data).permit(PERMITTED_PERSON_PARAMS).to_h
-      end
+      PERMITTED_FILTER_PARAMS = %i[police_national_computer criminal_records_office prison_number].freeze
 
       def filter_params
         params.require(:filter).permit(PERMITTED_FILTER_PARAMS).to_h
-      end
-
-      def court_case_filter_params
-        return unless params[:filter]
-
-        params.require(:filter).permit(PERMITTED_COURT_CASE_FILTER_PARAMS).to_h
-      end
-
-      def timetable_filter_params
-        params.require(:filter).permit(PERMITTED_TIMETABLE_FILTER_PARAMS).to_h
       end
     end
   end
--- a/app/controllers/api/v2/people_controller.rb
+++ b/app/controllers/api/v2/people_controller.rb
@@ -5,8 +5,12 @@ module Api
     class PeopleController < ApiController
       def index
         # people = People::Finder.new(filter_params).call
-        people = Person.all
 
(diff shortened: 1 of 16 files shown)


spec/requests/api/v2/people_controller_index_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+          it 'returns the relevant ethnicity' do
+            returned_types = response_json['included'].map { |r| r['type'] }.uniq
+
+            expect(returned_types).to contain_exactly('ethnicities', 'genders', 'profiles')
+          end
+        end
+      end
+    end
+
+    xcontext 'when not authorized', :with_invalid_auth_headers do
```
martyn-w, line 158: any reason why this is disabled?

spec/requests/api/v2/people_controller_index_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+        it 'returns the correct number of people' do
+          expect(response_json['data'].size).to eq(1)
+        end
+
+        it 'returns the person that matches the filter' do
+          expect(response_json).to include_json(data: [{ id: person.id }])
+        end
+      end
+
+      xdescribe 'filtering results by multiple values per filter'
```
martyn-w, line 82: guess this is still wip ?

cwrw: yea still working through this, please hold off the review

app/controllers/api/v2/people_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+      def filter_params
+        params.fetch(:filter, {}).permit(PERMITTED_FILTER_PARAMS).to_h
+      end
+
+      def included_relationships
+        IncludeParamHandler.new(params).call
+      end
+
+      def validate_include_params
+        ::V2::People::IncludeParamsValidator.new(included_relationships).tap do |validator|
```
cwrw, line 27: Unfortunately if `include` has values that are not in relationships it should return a `400` rather than a validation error `422` as per the specification: https://jsonapi.org/format/#fetching-includes

app/serializers/v2/profile_serializer.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module V2
+  class ProfileSerializer < ActiveModel::Serializer
+    attributes(
+      :id,
+      :assessment_answers,
+      :latest_nomis_booking_id,
```
cwrw, line 8: leaving here until we remove those off the profile soon

app/controllers/api/v2/people_controller.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module Api
+  module V2
+    class PeopleController < ApiController
+      before_action :validate_include_params, only: :index
```
willfish, line 6:
```suggestion
      before_action :validate_include_params
```

willfish: Seems to me that you'll be removing this anyway :)

app/controllers/api/v2/people_controller.rb
willfish: I think we should keep consistent with our way of rendering errors: https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/5bba6f774d9a87ac69b07418d454d1f4d48a47f2/app/controllers/api_controller.rb#L11-L17

app/services/v2/people/include_params_validator.rb
```diff
@@ -0,0 +1,23 @@
...
+
+module V2
+  module People
+    class IncludeParamsValidator
+      include ActiveModel::Validations
+
+      attr_reader :relationships
+
+      validates_each :relationships, allow_blank: true do |record, _attr, values|
+        unsupported_values = values - ::V2::PersonSerializer::SUPPORTED_RELATIONSHIPS
```
willfish, line 11: Potentially for later, but the only thing that makes this non-generic is the dependency on the `PersonSerializer`

spec/requests/api/v2/people_controller_index_spec.rb
```diff
@@ -0,0 +1,213 @@
...
+
+RSpec.describe Api::V2::PeopleController do
+  let(:supplier) { create(:supplier) }
+  let!(:application) { create(:application, owner_id: supplier.id) }
+  let!(:access_token) { create(:access_token, application: application).token }
+  let(:headers) { { 'CONTENT_TYPE': content_type }.merge('Authorization' => "Bearer #{access_token}") }
+  let(:response_json) { JSON.parse(response.body) }
+  let(:content_type) { ApiController::CONTENT_TYPE }
+
+  describe 'GET /people' do
```
willfish, line 13:
```suggestion
  describe 'GET /v2/people' do
```

spec/requests/api/v2/people_controller_index_spec.rb
```diff
@@ -0,0 +1,213 @@
...
+    let(:schema) { load_yaml_schema('get_people_responses.yaml', version: 'v2') }
```
willfish, line 14: 👍

spec/requests/api/v2/people_controller_index_spec.rb
```diff
@@ -0,0 +1,213 @@
...
+    let!(:people) { create_list :person, 2 }
+    let(:params) { {} }
+
+    context 'when successful' do
+      context 'when there are no params' do
+        before { get '/api/v2/people', params: params, headers: headers }
```
willfish, line 20: Should we include the `:skip_before` option, here :trollface:

spec/requests/api/v2/people_controller_index_spec.rb
```diff
@@ -0,0 +1,213 @@
...
+            bar: 'bar',
+            police_national_computer: 'AB/1234567',
+            foo: 'foo',
+          }
+        end
+        let(:params) { { filter: filters } }
+
+        before { get '/api/v2/people', params: params, headers: headers }
+
+        it 'returns the correct number of people' do
```
willfish, line 55: Given we truncate the db between examples I think it makes sense for either of the setup blocks (`let!` and `before`) to create another person with a different pnc number.

spec/requests/api/v2/people_controller_index_spec.rb
```diff
@@ -0,0 +1,213 @@
...
+          expect(response_json['data'].size).to eq(1)
+        end
+
+        it 'returns the person that matches the filter' do
+          expect(response_json).to include_json(data: [{ id: person.id }])
+        end
+      end
+
+      describe 'filtering results by multiple filters' do
```
willfish, line 64: This definitely doesn't cover the possible options for the identifier filters. I'm wondering whether we need to do these tests here and whether we'd benefit from removing most of these specs and validating handling of outbound messages to the finder?

app/controllers/api/v2/people_controller.rb
alexdesi: I remember that I had a conflict with another  `render_bad_request_error`,
we are already using that for ParameterMissin, 
We'll have a look

app/controllers/api/v2/people_controller.rb
alexdesi: I think it's good to have 'only: :index', so when we  add more action (very soon),
we don't forget to add it.

spec/requests/api/v2/people_controller_index_spec.rb
alexdesi: 😱

spec/requests/api/v2/people_controller_index_spec.rb
alexdesi: Hey @willfish, we do want to test that is the same person, it's like there was a single test like this:
```
        it 'returns the person' do
          expect(response_json['data'].size).to eq(1)
          expect(response_json).to include_json(data: [{ id: person.id }])
        end
```

spec/requests/api/v2/people_controller_index_spec.rb
alexdesi: I think it's fine that it does not cover all the possible options, the unit test `spec/services/v2/people/finder_spec.rb` covers all the cases.
but I agree with you, it feels we are testing this again and again in any endpoints, they still have some values, we could spy that those filters are called (for faster tests), of we could use the shared example, something like `it behaves like an endpoint that supports filters`, 
this is something we should do, maybe not just now.

spec/serializers/v2/profile_serializer_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  it 'contains type property' do
+    expect(result[:data][:type]).to eql 'profiles'
+  end
+
+  it 'contains id property' do
+    expect(result[:data][:id]).to eql profile.id
+  end
+
+  it 'contains last_synced_with_nomis' do
```
willfish, line 20: This is removed.

spec/services/v2/people/finder_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+      context 'with matching police_national_computer' do
+        let(:filter_params) { { police_national_computer: 'AB/00006' } }
+        let!(:person) { create(:person, police_national_computer: 'AB/00006') }
+
+        it 'returns people matching police_national_computer' do
+          expect(finder.call).to contain_exactly(person)
+        end
+      end
+
+      context 'with multiple police_national_computeres' do
```
willfish, line 22:
```suggestion
      context 'with multiple police_national_computer values' do
```

app/controllers/api/v2/people_controller.rb
alexdesi: We agreed that we are going to handle this on the next PRs, since those are relevant to this (post and patch)

PR #509, opened 2020-05-28
P4-1479 cancel endpoint

martyn-w:
### Jira link

[P4-1479](https://dsdmoj.atlassian.net/browse/P4-1479)

NB: merge PR #495 first as this PR builds on it.

### What?

I have added/removed/altered:

- [x] Added cancel move endpoint

### Why?

I am doing this because:

- so you can cancel a move the "proper" way

### Have you? (optional)

- minimal risk, not yet used in production


app/controllers/api/v1/move_events_controller.rb
```diff
@@ -3,37 +3,45 @@
...
     class MoveEventsController < ApiController
+      CANCEL_PARAMS = [:type, attributes: %i[timestamp cancellation_reason cancellation_reason_comment notes]].freeze
       COMPLETE_PARAMS = [:type, attributes: %i[timestamp notes]].freeze
       LOCKOUT_PARAMS = [:type, attributes: %i[timestamp notes], relationships: { from_location: {} }].freeze
       REDIRECT_PARAMS = [:type, attributes: %i[timestamp notes], relationships: { to_location: {} }].freeze
-      DEPRECIATED_EVENT_PARAMS = [:type, attributes: %i[timestamp event_name notes], relationships: { to_location: {} }].freeze
+      DEPRECATED_EVENT_PARAMS = [:type, attributes: %i[timestamp event_name notes], relationships: { to_location: {} }].freeze
+
+      def cancel
+        validate_params!(cancel_params)
```
cwrw, line 13: I think we might need to validate requiring the cancellation reason and notes?

app/services/move_events/params_validator.rb
```diff
@@ -4,9 +4,10 @@ module MoveEvents
   class ParamsValidator
     include ActiveModel::Validations
 
-    attr_reader :timestamp, :type
+    attr_reader :timestamp, :type, :cancellation_reason
```
cwrw, line 7: can you add tests for the updated?

spec/services/event_log/move_runner_spec.rb
```diff
@@ -81,6 +80,36 @@
...
+        expect { runner.call }.to change(move, :cancellation_reason_comment).from(nil).to('computer says no')
+      end
+
+      it_behaves_like 'it calls the Notifier with an update_status action_name'
+    end
+
+    context 'when the move is already cancelled' do
+      let!(:move) { create(:move, :cancelled, cancellation_reason: 'supplier_declined_to_move', cancellation_reason_comment: 'computer says no') }
+
+      it_behaves_like 'it does not call the Notifier'
```
cwrw, line 105: 👍

cwrw: can you add something similar when the move is invalid to check that we send no notifications?

app/services/move_events/params_validator.rb
martyn-w: will do 👍

spec/services/event_log/move_runner_spec.rb
martyn-w: not sure I follow? lets chat when you have a moment

PR #512, opened 2020-05-29
[P4-1435] Populate `Person` attributes when we import from Nomis

willfish:
### Jira link

P4-1435

### What?

- [x] Adds code to populate new `Person` [attributes](https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/499) when importing from Nomis
- [x] Adds support for [new](https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/511) `last_synced_with_nomis` audit timestamp for both `Profile` and `Person` when importing from Nomis.
- [x] Adds missing specs for the `People::Import` service

### Why?

We want to update a person every time we update a profile with the new
attributes that should be on a Person.


spec/services/people/importer_spec.rb
```diff
@@ -0,0 +1,111 @@
+RSpec.describe People::Importer do
+  subject(:service) { described_class.new(nomis_attributes) }
+
+  let(:nomis_attributes) do
+    {
```
willfish, line 5: This is taken from a real response in staging so is as ecologically valid as possible.

app/services/people/importer.rb
```diff
@@ -2,60 +2,89 @@
...
         { value: nomis_attributes[nomis], identifier_type: local } if nomis_attributes[nomis]
       }.compact
     end
 
     def latest_nomis_booking_id
       nomis_attributes[:latest_booking_id]
     end
+
+    def last_synced_with_nomis
+      Time.zone.now
```
willfish, line 87:
```suggestion
      @last_synced_with_nomis ||= Time.zone.now
```

willfish:
```
 Running tests in:

 ✓ With unfound person
 ✓ With existing person

 ✓ With a new person

 ✓ Reason - `Made in error`
 ✓ Reason - `Another reason`

 New move from Prison to Court
 ✓ With unfound person
 ✓ With existing person

 ✓ With existing person

 Download moves
 ✓ As Police user
 ✓ As Prison user
 ✓ As STC user
 ✓ As Supplier user

 Smoke tests
 ✓ As Police user
 ✓ As STC user
 ✓ As Prison user
 ✓ As Supplier user
 ✓ As OCA user


 18 passed (3m 01s)
```

spec/services/people/importer_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+      prison_number: 'G8872UH',
+      latest_booking_id: 1_093_139,
+      last_name: 'JAVAS',
+      first_name: "EF'LIAICO",
+      middle_names: 'MATTHEWS BARNARDOS',
+      date_of_birth: '1989-10-01',
+      aliases: nil,
+      pnc_number: '05/886838E',
+      cro_number: '610592/05U',
+      gender: gender.nomis_code,
```
martyn-w, line 15: could `gender` and `ethnicity` be hard-coded values too, to catch any future accidental changes?

app/services/people/importer.rb
```diff
@@ -2,60 +2,89 @@
...
     def call
       person = Person.find_or_initialize_by(nomis_prison_number: nomis_attributes[:prison_number])
+      person.assign_attributes(person_attributes)
+
       profile = person.latest_profile || person.profiles.build
-      profile.assign_attributes(attributes)
-      # return profile rather than person so that save does both profile and person if both are new
+      profile.assign_attributes(profile_attributes)
+
       profile
```
martyn-w, line 22: Something isn't quite right here: this is the `People::Importer` but it is returning a `profile`.
If it is possible, I think we should return `person` rather than `profile` ?

spec/services/people/importer_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+        'updated_at' => nil,
+        'assessment_answers' => [],
+        'gender_additional_information' => nil,
+        'latest_nomis_booking_id' => 1_093_139,
+        'last_synced_with_nomis' => be_within(2.seconds).of(now),
+      }
+    end
+
+    it 'initializes a new Person' do
+      person = service.call.person
```
martyn-w, line 65: I think the service should return a person, not a profile.

app/services/people/importer.rb
```diff
@@ -2,60 +2,89 @@
...
-      gender_param ? Gender.find_by(nomis_code: gender_param).id : nil
+    def gender
+      Gender.find_by(nomis_code: nomis_attributes[:gender])
     end
 
-    def find_ethnicity
-      ethnicity_param = nomis_attributes[:ethnicity]
-      ethnicity_param ? Ethnicity.find_by(title: ethnicity_param).id : nil
+    def ethnicity
+      Ethnicity.find_by(title: nomis_attributes[:ethnicity])
```
martyn-w, line 57: just a sanity check here: we are finding the ethnicity by title, e.g. `Mixed: Any other background` as opposed to nomis code, e.g. `M9` ?? it seems a bit odd.

spec/services/people/importer_spec.rb
martyn-w: you may need to change `let(:gender) {...}` and `let(:ethnicity) {...}`  statements to `let!`

spec/services/people/importer_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+      expect(person.attributes).to include(expected_person_attributes)
+    end
+
+    it 'assigns the correct attributes for the `Profile`' do
+      profile = service.call
+
+      expect(profile.attributes).to include(expected_profile_attributes)
+    end
+
+    context 'when called for an already saved `Person` and `Profile`' do
```
martyn-w, line 90: this test isn't really checking that first_person == second_person ?

spec/services/people/importer_spec.rb
willfish: I thought the same. The caller saves the profile and due to relationships (i.e. a `Profile` instantiates a #person member variable) the Profile is the thing that needs saving against it.

willfish: We'll be redoing all this later on, too. Separating out person/profile

app/services/people/importer.rb
willfish: Yeah, is a bit strange.

spec/services/people/importer_spec.rb
martyn-w: 👍

app/services/people/importer.rb
cwrw: Although this is marked as resolved, I was wondering the same thing (maybe conversation happened offline)?

martyn-w: hi @cwrw see this conversation: https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/512/files#r433164600  - there are further updates to resolve this in the future

app/services/people/importer.rb
```diff
@@ -2,60 +2,89 @@
...
       'police_national_computer' => :pnc_number,
       'prison_number' => :prison_number,
       'criminal_records_office' => :cro_number,
     }.freeze
 
-    attr_accessor :nomis_attributes
-
     def initialize(nomis_attributes)
-      self.nomis_attributes = nomis_attributes
+      @nomis_attributes = nomis_attributes
```
alexdesi, line 12: 👍

app/services/people/importer.rb
```diff
@@ -2,60 +2,89 @@
...
 
   private
 
-    def attributes
+    attr_reader :nomis_attributes
+
+    def person_attributes
+      nomis_attributes.slice(:last_name, :date_of_birth).merge(
+        first_names: first_names,
+        gender: gender,
```
cwrw, line 32: we're calling a few of those attributes twice, in person and profile, maybe it's worth memoizing those values in the methods below?

app/services/people/importer.rb
```diff
@@ -2,60 +2,89 @@
...
+        ethnicity: ethnicity,
+        police_national_computer: police_national_computer,
+        prison_number: prison_number,
+        criminal_records_office: criminal_records_office,
+        last_synced_with_nomis: last_synced_with_nomis,
```
cwrw, line 37: I was wondering about `gender_additional_information` and if we will still be supporting that on people?

spec/services/people/importer_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+        'aliases' => nil,
+        'gender_id' => gender.id,
+        'ethnicity_id' => ethnicity.id,
+        'nationality_id' => nil,
+        'created_at' => nil,
+        'updated_at' => nil,
+        'assessment_answers' => [],
+        'gender_additional_information' => nil,
+        'latest_nomis_booking_id' => 1_093_139,
+        'last_synced_with_nomis' => be_within(2.seconds).of(now),
```
cwrw, line 60: can you add `profile_identifiers` here as well?

app/services/people/importer.rb
willfish: Done

spec/services/people/importer_spec.rb
willfish: Done

app/services/people/importer.rb
willfish: This value only gets updated in the controller methods for creating these entities

PR #514, opened 2020-05-29
[P4-1502] Adds support for query includes in PeopleController

willfish:
### Jira link

P4-1502

### What?

Support query includes params in PeopleController

### Why?

Reduce query burden on backend api and be consistent with json:api standards.

Docs are getting updated for all endpoints at the end.

### DIFF
--- a/spec/requests/api/v1/people_controller_index_spec.rb
+++ b/spec/requests/api/v1/people_controller_index_spec.rb
@@ -10,42 +10,39 @@ RSpec.describe Api::V1::PeopleController do
 
   describe 'GET /people' do
     let(:prison_number) { 'G5033UT' }
+    let(:params) { { filter: { police_national_computer: 'AB/1234567' }, access_token: token.token } }
 
-    describe 'filtering the results' do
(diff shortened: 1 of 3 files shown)


spec/requests/api/v1/people_controller_index_spec.rb
```diff
@@ -10,42 +10,39 @@
 
   describe 'GET /people' do
     let(:prison_number) { 'G5033UT' }
+    let(:params) { { filter: { police_national_computer: 'AB/1234567' }, access_token: token.token } }
 
-    describe 'filtering the results' do
```
willfish, line 14: All requests have to include the filter so I've removed this outer block without changing any of the inner block contents.

willfish: Review without whitespace changes by using https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/514/files?diff=split&w=1

willfish:
I've not added errors for unsupported includes to that PR because I intend to iterate and add the functionality for all includes at a later date

PR #516, opened 2020-05-29
P4-1585 add booked move status

martyn-w:
### Jira link

[P4-1585](https://dsdmoj.atlassian.net/browse/P4-1585)

### What?

I have added/removed/altered:

- [x] Adds new booked move status

### Why?

I am doing this because:

- because this status forms part of supplier assurance testing

### Have you? (optional)

- high risk: the FE will not be expecting this status


spec/models/move_spec.rb
```diff
@@ -61,6 +61,12 @@
     )
   end
 
+  it 'does NOT validates presence of `profile` if `status` is booked' do
```
willfish, line 64:
```suggestion
  it 'does NOT validate presence of `profile` if `status` is booked' do
```

spec/requests/api/v1/moves_controller_index_spec.rb
```diff
@@ -91,6 +91,25 @@
         # rubocop:enable RSpec/ExampleLength
       end
 
+      context 'with a booked move' do
```
willfish, line 94: AFAIK contexts are used to describe specific contexts that affect the behaviour of the code. My sense is that this test and context doesn't add any value as far as testing behaviour is concerned? I could be wrong, though.

willfish: I mean, nothing really varies that could change the behaviour of how we index?

willfish: I think you're capturing this in the finder, spec, too :).

lib/tasks/fake_data.rake
```diff
@@ -190,7 +190,7 @@ namespace :fake_data do
           profile: profile,
           from_location: from_location,
           to_location: to_location,
-          status: %w[proposed requested completed].sample,
+          status: %w[proposed requested booked completed].sample,
```
willfish, line 193: 🎉

willfish:
So this isn't dangerous to the frontend currently, right? Because we don't expose the ability to mark a move's status as booked?

martyn-w:
@willfish its possible a move status could be set to 'booked' via the API, then the key question is, how will the FE behave ?

willfish:
Ah, then probably running the e2e tests against this makes sense

willfish:
In conjunction with creating a booked move

martyn-w:
The FE will need updating to show booked moves; they are currently invisible if the status=booked

# requested move
Screenshot 2020-05-29 at 17 09 39

# booked move
Screenshot 2020-05-29 at 17 09 55

martyn-w:
Closing for now, will re-open when this work can be unblocked.

martyn-w:
Re-opening to allow suppliers to change a move's status to be `booked`, on the understanding that booked moves will not be visible until the front end is updated. This is necessary for the next round of API assurance. cc @jason-sims

app/models/move.rb
```diff
@@ -53,7 +55,7 @@ class Move < VersionedModel
   validates :from_location, presence: true
   validates :to_location, presence: true, unless: :prison_recall?
   validates :move_type, inclusion: { in: move_types }
-  validates :profile, presence: true, unless: -> { requested? || cancelled? }
+  validates :profile, presence: true, unless: -> { requested? || booked? || cancelled? }
```
smoothcontract, line 58: Are you sure about this change? i.e. can a move can be booked (confirmed by the supplier) without a person specified?

martyn-w: @smoothcontract hmm, that doesn't look correct - I'll fix it

PR #517, opened 2020-06-01
[P4-1435] Move nomis fields to person

willfish:
### Jira link

P4-1435

### What?

After a bit of a discussion with @jcavell and @cwrw  we agreed that it's sensible to store `latest_nomis_booking_id` on a `Person` rather than a `Profile`.

- Adds migration to move `latest_nomis_booking_id` on a `Person` rather than a `Profile`
- Updates populator to addl this information into the `Person` in the rake task.

### Why?

- We don't currently need to have a trail of the information about a booking for the current prisoner
- When we do synchronise with Nomis this value will get updated on a person anyway so we're always pulling out the relevant information for the current move re court cases/court hearings.


db/migrate/20200601091307_fix_fields_on_person_again.rb
```diff
@@ -0,0 +1,6 @@
+class FixFieldsOnPersonAgain < ActiveRecord::Migration[5.2]
+  def change
+    remove_column :profiles, :last_synced_with_nomis,  :datetime
+    add_column    :people,   :latest_nomis_booking_id, :datetime
```
martyn-w, line 4: I don't think `latest_nomis_booking_id` is a `datetime` ?

cwrw: ah yes it should be `integer`, also are we not removing `latest_nomis_booking_id ` from profiles (or maybe not yet)?

willfish: That's coming later @cwrw

PR #519, opened 2020-06-01
P4-1562 add proposed journey state

martyn-w:
### Jira link

[P4-1562](https://dsdmoj.atlassian.net/browse/P4-1562)

### What?

I have added/removed/altered:

- [x] Added `proposed` initial journey state
- [x] Added `start` and `reject` journey events

### Why?

I am doing this because:
- suppliers have requested this state

### Have you? (optional)

- minimal - not currently being used

### DIFF
--- a/swagger/v1/get_journey.yaml
+++ b/swagger/v1/get_journey.yaml
@@ -33,11 +33,12 @@ GetJourney:
           type: string
           enum:
           - proposed
+          - rejected
           - in_progress
           - cancelled
           - completed
           example: in_progress
-          description: the state of the journey (proposed / in_progress / cancelled / completed)
+          description: the state of the journey (proposed / rejected / in_progress / cancelled / completed)
         vehicle:
           type: object
           properties:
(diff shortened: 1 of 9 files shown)


swagger/v1/get_journey.yaml
```diff
@@ -32,11 +32,12 @@ GetJourney:
         state:
           type: string
           enum:
+          - proposed
```
alexdesi, line 35: are we missing the `rejected` state here?

swagger/v1/get_journey.yaml
```diff
@@ -32,11 +32,12 @@ GetJourney:
...
           - in_progress
           - cancelled
           - completed
           example: in_progress
-          description: the state of the journey (in_progress / cancelled / completed)
+          description: the state of the journey (proposed / in_progress / cancelled / completed)
```
alexdesi, line 40: and here?

swagger/v1/get_journey.yaml
martyn-w: oops - my bad 🤦

PR #520, opened 2020-06-01
Removes unused environment configuration files

willfish:
### Jira link

P4-<TODO>

### What?

This needs merging **after** https://github.com/ministryofjustice/hmpps-book-secure-move-api-deploy/pull/61 is applied in order not to explode 🤯 

- [x] Removes unused environment configuration files (see diff outputs as they're all the same)

### Why?

- Remove wtfs per minute
- Reduce unnecessary settings files in the api project

- **Must be deployed after the configuration change in the deploy repo**


Dockerfile
```diff
@@ -1,6 +1,8 @@
 FROM ruby:2.6.6-alpine as build-stage
 
 ENV RAILS_ENV=production
+ENV RACK_ENV=production
+
```
mafonso, line 5: We probably don't need either (I know I added the `RAILS_ENV` one) at the build stage, do we?
Are these ENV vars considered during `bundle install`?  🤔

Dockerfile
```diff
@@ -29,6 +31,7 @@ ENV APP_GIT_COMMIT ${APP_GIT_COMMIT}
 ENV APPUID 1000
 
 ENV RAILS_ENV production
+ENV RACK_ENV production
```
mafonso, line 35: 👍

Dockerfile
willfish: Maybe we should remove/investigate them separately.

PR #521, opened 2020-06-01
P4-1597 fake journeys generator

martyn-w:
### Jira link

[P4-1597](https://dsdmoj.atlassian.net/browse/P4-1597)

### What?

I have added/removed/altered:

- [x] Added `fake_data:create_journeys` rake task

### Why?

I am doing this because:

- we can have some journey records to report on in further developments

- do not run on production!!!!


lib/tasks/fake_data.rake
```diff
@@ -216,11 +216,60 @@ namespace :fake_data do
     end
   end
 
+  desc 'create fake journeys'
+  task create_journeys: :environment do
+    puts 'create_journeys...'
+    Move
+        .left_outer_joins(:journeys).where(journeys: { move_id: nil })
```
cwrw, line 223: some strange indentation here, wonder how rubocop didn't pick this up 🤔

lib/tasks/fake_data.rake
```diff
@@ -216,11 +216,60 @@ namespace :fake_data do
...
+        .where(status: %w[completed booked requested])
+        .where.not(from_location: nil)
```
cwrw, line 225: will we ever have a `from_location` nil? Both journey and move validate against this?

martyn-w: good point well made 👍

lib/tasks/fake_data.rake
martyn-w: maybe 4 indents to signify line continuation vs 2 indents for being inside a block?

lib/tasks/fake_data.rake
```diff
@@ -216,11 +216,59 @@ namespace :fake_data do
...
+        timestamp += rand(30..90).minutes
+        current_location = intermediate_location
+        journey_counter += 1
+      end
+      move.journeys.create!(
+        client_timestamp: timestamp,
+        supplier: supplier,
+        from_location: current_location,
+        to_location: to_location,
+        state: 'completed',
```
cwrw, line 260: Just wondering about moves in the current states: `booked, requested` how does a journey be completed if they are still requested, maybe booked is ok? Can we only set those journeys on completed moves instead?

martyn-w: Unfortunately, none of the moves created through the front end get beyond the requested stage (as suppliers are not yet fully integrated with us). So this will create simulated journeys for those records which aren't completed but probably should have been.

lib/tasks/fake_data.rake
```diff
@@ -216,11 +216,285 @@ namespace :fake_data do
...
+    Move
+        .left_outer_joins(:journeys).where(journeys: { move_id: nil })
+        .where(status: %w[completed booked requested])
+        .find_each do |move|
+      # Creates some random journeys between A and B
+      current_location = move.from_location
+      number_of_journeys = rand(1..5)
+      journey_counter = 1
+      supplier = move.from_location.suppliers.first || Supplier.all.sample # pick a supplier based on the from location or at random
+      to_location = move.to_location || Location.order('RANDOM()').where(location_type: 'prison').first # pick a random prison if to_location was not specified
```
cwrw, line 231: For moves without a `to_location` set, since we are creating journeys for them with a final destination, should we mirror that in the move and persist the `to_location` there as well?

martyn-w: good question; moves without a `to_location` are a special type of police(?) to prison move, where the supplier and police know that the person is going to the default local prison. We don't have (and don't want to maintain) the list of default prisons, so we leave `to_location` blank. Removing the blank could upset some of the logic for these types of move.

martyn-w:
@cwrw yes, fair point that this has outgrown its stay in a rake task. The same thought had crossed my mind, but was hoping to avoid the whole shebang for code which is likely to be run only once or twice... I'll see what I can do.

PR #522, opened 2020-06-02
Filter out logs that are garbage

willfish:
### Jira link

P4-<TODO>

### What?

Looking at our logs and all I can see is ping lines. None of these are useful for debugging the majority of problems.

### Why?

Make logs more valuable


willfish:
image

PR #523, opened 2020-06-02
[P4-1510] Support `person` in include validator

willfish:
### Jira link

P4-1510

### What?

Active Model Serializers (AMS) automagically returns a root object
when given the following `.` (dot) separated format include query `person.gender`.

This meant that we never needed to add `person` to our static include
list.

The validator we added does not support root objects implicitly and requires us to be explicit.

Since both the validator and AMS depend on the same list this means we
need to make sure we're supporting the explicit form for included
documents when doing validations.

### Why?

To prevent the frontend experience unexpected validation failures when they explicitly request a `person` in their include query.

### DIFF
--- a/app/serializers/move_serializer.rb
+++ b/app/serializers/move_serializer.rb
@@ -29,6 +29,7 @@ class MoveSerializer < ActiveModel::Serializer
 
   SUPPORTED_RELATIONSHIPS = %w[
     profile
+    person
     person.ethnicity
     person.gender
     from_location
--- a/spec/requests/api/v1/moves_controller_index_spec.rb
+++ b/spec/requests/api/v1/moves_controller_index_spec.rb
@@ -167,7 +167,7 @@ RSpec.describe Api::V1::MovesController do
         end
 
         it 'includes profile in the response' do
-          get '/api/v1/moves?include=profile,person.gender', params: params, headers: headers
+          get '/api/v1/moves?include=profile,person,person.gender', params: params, headers: headers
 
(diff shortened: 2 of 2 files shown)


spec/requests/api/v1/moves_controller_index_spec.rb
```diff
@@ -167,7 +167,7 @@
         end
 
         it 'includes profile in the response' do
-          get '/api/v1/moves?include=profile', params: params, headers: headers
+          get '/api/v1/moves?include=profile,person.gender', params: params, headers: headers
```
willfish, line 170:
```suggestion
          get '/api/v1/moves?include=profile,person,person.gender', params: params, headers: headers
```

app/serializers/move_serializer.rb
```diff
@@ -29,6 +29,7 @@ class MoveSerializer < ActiveModel::Serializer
 
   SUPPORTED_RELATIONSHIPS = %w[
     profile
+    person
```
cwrw, line 32: instead of having to explicitly support this everywhere going forward (as now we will be rolling this out to all relationships) can we fix the validation itself to check for anything before the `.` and deem it valid?

willfish: We definitely could. Maybe we should solve this in the generalisation work we can refine in today's refinement.

willfish: I guess it's recursive, too. So all of the levels - not just the root - get included. Simple test will prove/disprove this

willfish:
Closing in favour of a general solution for the validator

PR #524, opened 2020-06-02
[P4-1500] Adds profile reference to documents

willfish:
### Jira link

P4-1500

### What?

We are going to be referencing a profile and dereferencing a move when we talk about documents that are relevant to the risk profile of a person. This change adds the reference to profile in isolation of any other changes to be used later.

### Why?

- Needed to better reflect the profile/move concepts

### DIFF
--- a/db/migrate/20200602095941_adds_profile_id_to_documents.rb
+++ b/db/migrate/20200602095941_adds_profile_id_to_documents.rb
@@ -1,0 +1,5 @@
+class AddsProfileIdToDocuments < ActiveRecord::Migration[5.2]
+  def change
+    add_reference :documents, :profile, type: :uuid
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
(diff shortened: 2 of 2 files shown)


db/migrate/20200602095941_adds_profile_id_to_documents.rb
```diff
@@ -0,0 +1,5 @@
+class AddsProfileIdToDocuments < ActiveRecord::Migration[5.2]
+  def change
+    add_reference :documents, :profile, type: :uuid
```
cwrw, line 3: can you add `foreign_key: true` as well

martyn-w: do the models need updating with `has_many :documents` and `belongs_to :profile` etc?

willfish: I tend to prefer to isolate migrations in a single PR and do code changes separately.

martyn-w: 👍

willfish:
Spiked on this and this is an approach which will require migrating away from later. We're going to use polymorphism on the documents table to enable support for documents on multiple entities.

PR #525, opened 2020-06-02
[P4-1610] Move to included supplier instead of supplier attribute

willfish:
### Jira link

P4-1610

### What?

Moves to included supplier on a location rather than an attribute.

This is definitely going to break the frontend.

### Why?

So we're inline with the json:api specifications for relationships.

### Have you? (optional)

- This **will break the frontend** - I'm coordinating with them to sort this out


willfish:
TODO: run the e2e tests

willfish:
Update to use query param for locations

app/serializers/location_serializer.rb
```diff
@@ -3,5 +3,13 @@
 class LocationSerializer < ActiveModel::Serializer
   type 'locations'
 
-  attributes :id, :key, :title, :location_type, :nomis_agency_id, :can_upload_documents, :disabled_at, :suppliers
```
cwrw, line 6: since suppliers always came with locations and now will no longer will be available unless explicitly mentioned, we might need to be explicit to include it everywhere we include the location resource in other serialisers example moves etc.

cwrw: example https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/serializers/move_serializer.rb#L30

willfish: Having this conversation with Alex currently

willfish: Also, https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/525#issuecomment-637454876

willfish: Okay, had a chat with Alex. We're going to add support for the query params and static list as elsewhere.

app/serializers/location_serializer.rb
```diff
@@ -3,5 +3,13 @@
...
+  attributes :id,
```
willfish, line 6: Remove id

willfish:
This passes e2e tests against master of the frontend and https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/pull/559

```
 Running tests in:

 ✓ With unfound person
 ✓ With existing person

 ✓ With a new person

 ✓ Reason - `Made in error`
 ✓ Reason - `Another reason`

 New move from Prison to Court
 ✓ With unfound person
 ✓ With existing person

 ✓ With existing person

 Download moves
 ✓ As Police user
 ✓ As Prison user
 ✓ As STC user
 ✓ As Supplier user

 Smoke tests
 ✓ As Police user
 ✓ As STC user
 ✓ As Prison user
 ✓ As Supplier user
 ✓ As OCA user


 18 passed (6m 52s)
```

PR #527, opened 2020-06-02
Add new post /v2/person endpoint

alexdesi:
### Jira link
https://dsdmoj.atlassian.net/browse/P4-1583

### What?
This PR adds the endpoint `/v2/people` to create a new person

### Why?
This is part of the planned set of the new endpoints.
This allows creating a person with more attributes, and it is from a profile (when you create a person, the profile is not created)

### Have you? (optional)

The Swagger Doc will be updated in the next PR: https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/507/files

Very low risk, since this is a new endpoint.

### DIFF
--- a/app/controllers/api/v2/people_controller.rb
+++ b/app/controllers/api/v2/people_controller.rb
@@ -43,17 +43,17 @@ module Api
       end
 
       def person_attributes
+        person_params[:attributes].merge(ethnicity: ethnicity, gender: gender)
+      end
+
+      def ethnicity
         ethnicity_id = params.require(:data).dig(:relationships, :ethnicity, :data, :id)
-        gender_id = params.require(:data).dig(:relationships, :gender, :data, :id)
-
-        # Validate the existence of Ethnicity and Gender
         Ethnicity.find(ethnicity_id) if ethnicity_id
-        Gender.find(gender_id) if gender_id
+      end
 
-        person_params[:attributes].merge(
-          ethnicity_id: ethnicity_id,
-          gender_id: gender_id,
-        )
+      def gender
+        gender_id = params.require(:data).dig(:relationships, :gender, :data, :id)
+        Gender.find(gender_id) if gender_id
       end
     end
   end
--- a/app/controllers/api/v2/people_controller.rb
+++ b/app/controllers/api/v2/people_controller.rb
@@ -43,9 +43,16 @@ module Api
       end
 
       def person_attributes
+        ethnicity_id = params.require(:data).dig(:relationships, :ethnicity, :data, :id)
+        gender_id = params.require(:data).dig(:relationships, :gender, :data, :id)
+
+        # Validate the existence of Ethnicity and Gender
+        Ethnicity.find(ethnicity_id) if ethnicity_id
+        Gender.find(gender_id) if gender_id
+
         person_params[:attributes].merge(
-          ethnicity_id: params.require(:data).dig(:relationships, :ethnicity, :data, :id),
-          gender_id: params.require(:data).dig(:relationships, :gender, :data, :id),
+          ethnicity_id: ethnicity_id,
+          gender_id: gender_id,
         )
       end
     end
--- a/app/controllers/api/v2/people_controller.rb
+++ b/app/controllers/api/v2/people_controller.rb
@@ -6,7 +6,7 @@ module Api
       before_action :validate_include_params, only: %i[index create]
 
       def create
-        person = Person.create(new_person_attributes)
+        person = Person.create(person_attributes)
 
         render json: person, status: :created, include: included_relationships, serializer: ::V2::PersonSerializer
       end
@@ -44,7 +44,7 @@ module Api
         params.require(:data).permit(PERMITTED_PERSON_PARAMS).to_h
       end
 
-      def new_person_attributes
+      def person_attributes
         person_params[:attributes].merge(
           ethnicity_id: params.require(:data).dig(:relationships, :ethnicity, :data, :id),
           gender_id: params.require(:data).dig(:relationships, :gender, :data, :id),
--- a/app/controllers/api/v2/people_controller.rb
+++ b/app/controllers/api/v2/people_controller.rb
@@ -3,6 +3,14 @@
 module Api
   module V2
     class PeopleController < ApiController
+      before_action :validate_include_params, only: :index
+
+      def create
+        person = Person.last
+
+        render json: person, status: :created
+      end
+
       def index
         people = ::V2::People::Finder.new(filter_params).call
 
--- a/app/controllers/api/v2/people_controller.rb
+++ b/app/controllers/api/v2/people_controller.rb
@@ -3,8 +3,6 @@
 module Api
   module V2
     class PeopleController < ApiController
-      before_action :validate_include_params, only: %i[index create]
-
       def create
         person = Person.create(person_attributes)
 
--- a/app/controllers/api/v2/people_controller.rb
+++ b/app/controllers/api/v2/people_controller.rb
@@ -25,7 +25,6 @@ module Api
         first_names
         last_name
         date_of_birth
-        nomis_prison_number
         prison_number
         criminal_records_office
         police_national_computer
--- a/app/controllers/api/v2/people_controller.rb
+++ b/app/controllers/api/v2/people_controller.rb
@@ -25,6 +25,10 @@ module Api
         first_names
         last_name
         date_of_birth
+        nomis_prison_number
+        prison_number
+        criminal_records_office
+        police_national_computer
         gender_additional_information
       ].freeze
       PERMITTED_PERSON_PARAMS = [:type, attributes: PERSON_ATTRIBUTES, relationships: {}].freeze
--- a/app/controllers/api/v2/people_controller.rb
+++ b/app/controllers/api/v2/people_controller.rb
@@ -3,12 +3,12 @@
 module Api
   module V2
     class PeopleController < ApiController
-      before_action :validate_include_params, only: :index
+      before_action :validate_include_params, only: %i[index create]
 
       def create
         person = Person.create(new_person_attributes)
 
-        render json: person, status: :created, serializer: ::V2::PersonSerializer
+        render json: person, status: :created, include: included_relationships, serializer: ::V2::PersonSerializer
       end
 
       def index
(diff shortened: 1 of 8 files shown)


alexdesi:
Note: nomis_prison_number and prison_number are the same, and we should not support both.
From a chat with @jcavell, he is in favor of `prison_number`.
I've removed the `nomis_prison_number` from the serializer, 
also, the Post is not able to update that field.

app/controllers/api/v2/people_controller.rb
```diff
@@ -3,7 +3,13 @@
 module Api
   module V2
     class PeopleController < ApiController
-      before_action :validate_include_params, only: :index
+      before_action :validate_include_params, only: %i[index create]
```
willfish, line 6: I've removed this

app/controllers/api/v2/people_controller.rb
```diff
@@ -35,6 +52,17 @@ def validate_include_params
           end
         end
       end
+
+      def person_params
+        params.require(:data).permit(PERMITTED_PERSON_PARAMS).to_h
+      end
+
+      def new_person_attributes
```
willfish, line 60:
```suggestion
      def person_attributes
```

app/controllers/api/v2/people_controller.rb
```diff
@@ -3,7 +3,13 @@
 module Api
   module V2
     class PeopleController < ApiController
-      before_action :validate_include_params, only: :index
+      before_action :validate_include_params, only: %i[index create]
+
+      def create
+        person = Person.create(new_person_attributes)
```
willfish, line 9:
```suggestion
        person = Person.create(person_attributes)
```

db/schema.rb
```diff
@@ -207,10 +207,10 @@
     t.text "cancellation_reason_comment"
     t.integer "nomis_event_ids", default: [], null: false, array: true
     t.uuid "profile_id"
-    t.boolean "move_agreed", default: false, null: false
-    t.string "move_agreed_by"
```
cwrw, line 211: I think those changes have been committed by mistake?

db/schema.rb
```diff
@@ -416,7 +416,7 @@
   add_foreign_key "moves", "allocations"
   add_foreign_key "moves", "locations", column: "from_location_id", name: "fk_rails_moves_from_location_id"
   add_foreign_key "moves", "locations", column: "to_location_id", name: "fk_rails_moves_to_location_id"
-  add_foreign_key "moves", "people"
+  add_foreign_key "moves", "people", name: "fk_rails_moves_person_id"
```
willfish, line 419: How comes?

spec/requests/api/v2/people_controller_create_spec.rb
```diff
@@ -0,0 +1,174 @@
...
+    let(:risk_type_2) { create :assessment_question, :risk }
+
+    let(:person_params) do
+      {
+        data: {
+          type: 'people',
+          attributes: {
+            first_names: 'Bob',
+            last_name: 'Roberts',
+            date_of_birth: Date.civil(1980, 1, 1),
```
willfish, line 26: Not seen Date.civil before :)

spec/requests/api/v2/people_controller_create_spec.rb
```diff
@@ -0,0 +1,174 @@
...
+  let(:access_token) { create(:access_token).token }
+  let(:content_type) { ApiController::CONTENT_TYPE }
+  let(:headers) { { 'CONTENT_TYPE': content_type }.merge('Authorization' => "Bearer #{access_token}") }
+
+  describe 'POST /people' do
+    let(:schema) { load_yaml_schema('post_people_responses.yaml', version: 'v2') }
+
+    let(:ethnicity) { create :ethnicity }
+    let(:gender) { create :gender }
+    let(:risk_type_1) { create :assessment_question, :risk }
```
willfish, line 16: Is this used?

spec/requests/api/v2/people_controller_create_spec.rb
```diff
@@ -0,0 +1,174 @@
...
+                id: gender.id,
+                type: 'genders',
+              },
+            },
+          },
+        },
+      }
+    end
+
+    context 'when successful' do
```
willfish, line 50: I don't think successful describes anything/adds any value?

app/controllers/api/v2/people_controller.rb
alexdesi: 👍

db/schema.rb
alexdesi: indeed, when I recreate the DB, for some reasons, it recreates the schema that is  slightly different

db/schema.rb
alexdesi: commited by mistake (I've run recreated the DB in my local)

spec/requests/api/v2/people_controller_create_spec.rb
alexdesi: I've copied from another test, but actually it is equivalent to `Date.new1980, 1, 1)`
I'll change it.

swagger/v2/post_people_responses.yaml
```diff
@@ -0,0 +1,52 @@
+'201':
+  type: object
+  required:
+  - data
+  properties:
+    data:
+      $ref: person.yaml#/Person
+'400':
```
willfish, line 8: See this over and over. Is it possible to include `$ref: generic_errors.yaml` or something?

app/controllers/api/v2/people_controller.rb
```diff
@@ -35,6 +52,17 @@ def validate_include_params
...
         end
       end
+
+      def person_params
+        params.require(:data).permit(PERMITTED_PERSON_PARAMS).to_h
+      end
+
+      def new_person_attributes
+        person_params[:attributes].merge(
+          ethnicity_id: params.require(:data).dig(:relationships, :ethnicity, :data, :id),
```
cwrw, line 62: I'm wondering about this, the pattern in other controllers has been to find the records for validation reasons, maybe we can do the same here?

alexdesi: good point! I'll add that!

spec/requests/api/v2/people_controller_create_spec.rb
alexdesi: ops 😅

spec/requests/api/v2/people_controller_create_spec.rb
alexdesi: I know I hate it, it's just legacy stuff, that's not even a proper contest for me,  I agree, I am going to remove it.

swagger/v2/post_people_responses.yaml
alexdesi: In my next PR I'll do some work with swagger, let's have a chat about this, I can do it later.

app/controllers/api/v2/people_controller.rb
```diff
@@ -13,13 +19,42 @@ def index
...
+
+      def person_attributes
+        ethnicity_id = params.require(:data).dig(:relationships, :ethnicity, :data, :id)
+        gender_id = params.require(:data).dig(:relationships, :gender, :data, :id)
+
+        # Validate the existence of Ethnicity and Gender
+        Ethnicity.find(ethnicity_id) if ethnicity_id
+        Gender.find(gender_id) if gender_id
+
+        person_params[:attributes].merge(
```
cwrw, line 53: To simplify further we can do something like this:
```
person_params[:attributes].merge(
ethnicity: Ethnicity.find(person_params.dig(:relationships, :ethnicity, :data, :id)),
gender: Gender.find(person_params.dig(:relationships, :gender, :data, :id)
)
```
As we will automatically set the ids

willfish: Was just writing this

willfish: Or push finding the gender into a method `gender` which includes pulling out the params

```ruby
def ethnicity
  Ethnicity.find(person_params.dig(:relationships, :ethnicity, :data, :id))
end

def gender
  Gender.find(person_params.dig(:relationships, :gender, :data, :id))
end

def person_attributes
  person_params[:attributes].merge(ethnicity: ethnicity, gender: gender)
end
```

config/routes.rb
```diff
@@ -50,7 +50,7 @@
     end
 
     namespace :v2 do
-      resources :people, only: %i[index]
+      resources :people, only: %i[index create]
```
willfish, line 53:
```suggestion
      resources :people
```

swagger/v2/post_people_responses.yaml
willfish: Sounds good dude!

spec/requests/api/v2/people_controller_create_spec.rb
willfish: Yay \0/

spec/requests/api/v2/people_controller_create_spec.rb
willfish: Cool beans :)

app/controllers/api/v2/people_controller.rb
alexdesi: @cwrw  we need the conditionals because a requirement is that all the params are not mandatory, for now (my understanding is that we could create the person and then later syncs it with nomis).

config/routes.rb
alexdesi: at the moment we only have the index and create actions, I am missing something?

app/controllers/api/v2/people_controller.rb
alexdesi: I refactored it with the small ethnicity/gender helper
In the long term, we should probably find a generic solution for this kind of problem.

spec/requests/api/v2/people_controller_index_spec.rb
```diff
@@ -137,37 +136,37 @@
...
         context 'when including a non existing relationship in a query param' do
-          let(:params) { { include: 'gender,bar' } }
+          let(:params) { { include: 'gender,non-existent-relationship' } }
 
           it 'responds with error 400' do
             response_error = response_json['errors'].first
 
             expect(response_error['title']).to eq('Bad request')
-            expect(response_error['detail']).to include('["bar"] is not supported.')
+            expect(response_error['detail']).too include('["non-existent-relationship"] is not supported.')
```
willfish, line 169:
```suggestion
            expect(response_error['detail']).to include('["non-existent-relationship"] is not supported.')
```

PR #528, opened 2020-06-02
Reduce number of fixtures created for pagination tests

cwrw:
Set a different Kaminari page default in tests. Where previously it equated to 20, set it to 5 instead and amend tests to create less fixtures. 

Also move pagination tests to a shared example.

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -89,3 +89,7 @@ Shoulda::Matchers.configure do |config|
     with.library :rails
   end
 end
+
+Kaminari.configure do |config|
+  config.default_per_page = 5
+end
--- a/spec/requests/api/v1/journeys_controller_index_spec.rb
+++ b/spec/requests/api/v1/journeys_controller_index_spec.rb
@@ -24,9 +24,10 @@ RSpec.describe Api::V1::JourneysController do
     let(:page) { 1 }
     let(:per_page) { 5 }
     let(:url) { "/api/v1/moves/#{move.id}/journeys" }
+    let(:params) { {} }
 
     before do
-      get url, headers: headers, as: :json
+      get url, params: params, headers: headers
     end
 
     context 'when successful' do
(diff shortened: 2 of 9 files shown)


spec/requests/api/v1/journeys_controller_index_spec.rb
```diff
@@ -22,11 +22,12 @@
...
     let(:intermediate_journeys_count) { 1 }
     let(:page) { 1 }
-    let(:per_page) { 20 }
+    let(:per_page) { 5 }
     let(:url) { "/api/v1/moves/#{move.id}/journeys" }
+    let(:params) { {} }
 
     before do
-      get url, headers: headers, as: :json
+      get url, params: params, headers: headers
```
cwrw, line 30: when attempting to pass in `params`, the `as: :json` was changing the `GET` request to a `POST`

spec/rails_helper.rb
```diff
@@ -89,3 +89,7 @@
     with.library :rails
   end
 end
+
+Kaminari.configure do |config|
```
smoothcontract, line 93: Might be nice to add a comment explaining why this overrides the default only affects tests.

spec/requests/api/v1/moves_controller_index_spec.rb
```diff
@@ -107,29 +107,9 @@
           }
         end
 
-        it 'paginates 20 results per page as default' do
-          get_moves
-
-          expect(response_json['data'].size).to eq 20
-        end
-
-        it 'returns 1 result on the second page'  do
```
smoothcontract, line 116: Looks like this test (of the links) is now removed. Is there a way that could be added to the shared context, perhaps by reading the link from the returned metadata?

cwrw: talked in person but for posterity: this test is defined in the shared example but using `params` instead

PR #529, opened 2020-06-03
P4-1582 Allow profiles to be associated to moves on POST/PATCH endpoints

cwrw:
### Jira link

[P4-1582](https://dsdmoj.atlassian.net/browse/P4-1582)

### What?

I have added/removed/altered:

- [x] Added support for setting a profile in POST move endpoint
- [x] Added support for setting a profile in PATCH move endpoint

### Why?

Allow a move to be created or updated by either passing in a profile id or person id (where latest profile will be associated to a move) in relationships. The profile will take precedence unless a person id is passed in on creation. It will also take precedence with validation errors. On updates the ability has been added to either unlink an existing profile or update it to a new one. 

After migration of profiles/people is complete on moves, all person setting logic is to be cleaned up.

### Have you? (optional)

### DIFF
--- a/app/controllers/api/v1/moves_controller.rb
+++ b/app/controllers/api/v1/moves_controller.rb
@@ -98,11 +98,11 @@ module Api
       end
 
       def profile_or_person_latest_profile
-        person_id = new_move_params.dig(:relationships, :person, :data, :id)
-        # moves are always created against the latest_profile for the person if profile not provided
-        return Person.find(person_id).latest_profile if person_id
+        profile_id = new_move_params.dig(:relationships, :profile, :data, :id)
+        return Profile.find(profile_id) if profile_id
 
-        Profile.find(new_move_params.dig(:relationships, :profile, :data, :id))
+        # moves are always created against the latest_profile for the person if profile not provided
+        Person.find(new_move_params.dig(:relationships, :person, :data, :id)).latest_profile
       end
 
       def update_move_params
--- a/app/controllers/api/v1/moves_controller.rb
+++ b/app/controllers/api/v1/moves_controller.rb
@@ -87,16 +87,22 @@ module Api
       end
 
       def new_move_attributes
-        person = Person.find(new_move_params.dig(:relationships, :person, :data, :id))
-        # moves are always created against the latest_profile for the person
         new_move_params[:attributes].merge(
-          profile: person.latest_profile,
+          profile: profile_or_person_latest_profile,
           from_location: Location.find(new_move_params.dig(:relationships, :from_location, :data, :id)),
           to_location: Location.find_by(id: new_move_params.dig(:relationships, :to_location, :data, :id)),
           documents: Document.where(id: (new_move_params.dig(:relationships, :documents, :data) || []).map { |doc| doc[:id] }),
           court_hearings: CourtHearing.where(id: (new_move_params.dig(:relationships, :court_hearings, :data) || []).map { |court_hearing| court_hearing[:id] }),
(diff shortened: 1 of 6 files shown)


spec/requests/api/v1/moves_controller_update_spec.rb
```diff
@@ -496,6 +556,42 @@
               expect(move.reload.allocation).to be_unfilled
             end
           end
+
+          context 'when linking a profile' do
```
cwrw, line 560: Not sure about having those tests here as well as in the updater spec, but added them to mirror the person tests for allocation already defined. Maybe we should remove both 🤔

app/controllers/api/v1/moves_controller.rb
```diff
@@ -99,6 +97,14 @@ def new_move_attributes
         )
       end
 
+      def profile_or_person_latest_profile
+        person_id = new_move_params.dig(:relationships, :person, :data, :id)
```
cwrw, line 101: I intentionally didn't add logic to check if both are passed in as this will soon go away and only set profile, to avoid adding tests and complication.

spec/requests/api/v1/moves_controller_update_spec.rb
willfish: I would remove them both. Happy if you want to make it part of this PR

willfish: By both, I mean the duplication in the request spec

spec/services/moves/updater_spec.rb
```diff
@@ -181,6 +211,44 @@
       end
     end
 
+    context 'with profile' do
+      let(:before_profile) { create(:profile) }
+      let(:after_profile) { create(:profile) }
+      let!(:move) { create(:move, profile: before_profile) }
```
willfish, line 217: I know we do it all over the place, but I really don't like `let!`

jcavell:
Minor point, but on the description you say "The profile will take precedence unless a person id is passed in on creation" but this is a bit confusing because actually person takes precedence if a person and profile are sent.

app/controllers/api/v1/moves_controller.rb
martyn-w: would it make more sense to the specified profile rather than the person's latest profile if both profile_id and person_id are provided?

app/services/moves/updater.rb
```diff
@@ -43,11 +47,14 @@ def attributes
 
       attributes[:documents] = Document.where(id: document_ids) unless document_attributes.nil?
 
+      # TODO: to be removed once move profile migration complete
       unless person_attributes.nil?
         person = Person.find_by(id: person_attributes.dig(:data, :id))
         attributes[:profile] = person&.latest_profile
```
martyn-w, line 53: maybe this should not be `person&.latest_profile` if `profile_id` is also present in the attributes ?

cwrw: I wanted to avoid checking both to avoid adding more complexity and tests for something going away soon, but happy to add it if its a risk

martyn-w: 👍

app/controllers/api/v1/moves_controller.rb
cwrw: spoke in person: updated this to make sure suppliers don't accidentally pass both in

app/services/moves/updater.rb
cwrw: spoke in person: updated this to guard against suppliers don't accidentally pass both in

PR #530, opened 2020-06-03
[P4-1617] Generic include validations

willfish:
### Jira link

P4-1617

### What?

- [x] Render include validation failures with rescue in the api controller
- [x] Validate _all_ include query params in same way

### Why?

We've written a bunch of tests/functionality that is inconsistent. This makes it simpler to extend new include validations and specs.


app/services/include_params_validator.rb
```diff
@@ -0,0 +1,32 @@
...
+  def initialize(relationships, supported_relationships)
+    @relationships = relationships
+    @supported_relationships = supported_relationships
+  end
+
+  def fully_validate!
+    raise ValidationError, self if invalid?
+  end
+
+  class ValidationError < StandardError
```
cwrw, line 25: I find using this might be simpler, what do you think? 
`ValidationError = Class.new(StandardError)`

cwrw: update: doesn't work

PR #531, opened 2020-06-03
[P4-1617] Support explicit and implicit nested chains for includes

willfish:
### Jira link

P4-1617

### What?

Jira link: https://dsdmoj.atlassian.net/browse/P4-1617

We've added a generic validator, see:

https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/530

This validator shares a supported relationships list with Active Model
Serializers.

AMS supports implicit and explicit forms of include query param chains
and we want our validator to support both to match this.

For example:
  'foo.bar.baz' is equivalent to ['foo', 'foo.bar', 'foo.bar.baz'] in
  AMS and ['foo.bar.baz'] in our validator

### Why?

- Makes our api more flexible and consistent

### DIFF
--- a/app/services/include_params_validator.rb
+++ b/app/services/include_params_validator.rb
@@ -6,7 +6,7 @@ class IncludeParamsValidator
   attr_reader :relationships, :supported_relationships
 
   validates_each :relationships, allow_blank: true do |record, _attr, values|
-    unsupported_values = values - record.formatted_supported_relationships
+    unsupported_values = values - record.splatted_supported_relationships
 
     unless unsupported_values.empty?
       record.errors.add 'Bad request', "#{unsupported_values} is not supported. Valid values are: #{record.supported_relationships}"
@@ -29,17 +29,17 @@ class IncludeParamsValidator
   #
   # To reflect this behaviour in the validation we need to support all variations
   # of nested chains by preformatting them in this validator.
-  def formatted_supported_relationships
-    formatted_supported_relationships = Set.new
+  def splatted_supported_relationships
+    splatted_supported_relationships = Set.new
 
     supported_relationships.each do |chain|
       split_chain = chain.split('.')
       split_chain.each_with_index do |_, index|
-        formatted_supported_relationships << split_chain[0..index].join('.')
+        splatted_supported_relationships << split_chain[0..index].join('.')
       end
     end
 
-    formatted_supported_relationships.to_a
+    splatted_supported_relationships.to_a
   end
 
   class ValidationError < StandardError
--- a/app/services/include_params_validator.rb
+++ b/app/services/include_params_validator.rb
@@ -6,7 +6,7 @@ class IncludeParamsValidator
   attr_reader :relationships, :supported_relationships
 
   validates_each :relationships, allow_blank: true do |record, _attr, values|
-    unsupported_values = values - record.supported_relationships
+    unsupported_values = values - record.formatted_supported_relationships
 
     unless unsupported_values.empty?
       record.errors.add 'Bad request', "#{unsupported_values} is not supported. Valid values are: #{record.supported_relationships}"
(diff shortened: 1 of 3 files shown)


spec/requests/api/v1/moves_controller_index_spec.rb
```diff
@@ -151,9 +151,10 @@
         end
 
         context 'when including the include query param' do
-          let(:query_params) { '?include=profile' }
+          let(:query_params) { '?include=profile,person.ethnicity' }
 
           it 'includes the requested includes in the response' do
+            binding.pry
```
cwrw, line 157: 🔥

willfish: Derp

app/services/include_params_validator.rb
```diff
@@ -6,7 +6,7 @@ class IncludeParamsValidator
   attr_reader :relationships, :supported_relationships
 
   validates_each :relationships, allow_blank: true do |record, _attr, values|
-    unsupported_values = values - record.supported_relationships
+    unsupported_values = values - record.formatted_supported_relationships
```
alexdesi, line 9: I am not sure about the name,
*formatted* conveys a different meaning, like formatting a text,
this case is more related to: extended , expanded, spread, splitted, ... you got the point :)

app/services/include_params_validator.rb
```diff
@@ -22,6 +22,26 @@ def fully_validate!
...
+  #
+  # To reflect this behaviour in the validation we need to support all variations
+  # of nested chains by preformatting them in this validator.
+  def formatted_supported_relationships
+    formatted_supported_relationships = Set.new
+
+    supported_relationships.each do |chain|
+      split_chain = chain.split('.')
+      split_chain.each_with_index do |_, index|
+        formatted_supported_relationships << split_chain[0..index].join('.')
```
alexdesi, line 38: have you checked that this line does create duplication?
for example, if ['foo', 'foo.bar', 'foo.bar.baz'] is passed, we do not want any repetion of foo and foo.bar

willfish: Using Set.new :)

app/services/include_params_validator.rb
willfish: ['split', 'splat', 'explicitify'].product(['supported_relationships', 'nested_chains']).map { |t| t.join('_') }
=> ["split_supported_relationships",
 "split_nested_chains",
 "splat_supported_relationships",
 "splat_nested_chains",
 "explicitify_supported_relationships",
 "explicitify_nested_chains"]

willfish: Maybe pick one of ^^

willfish: Also, I think format is basically fine

```
verb
(especially in computing) arrange or put into a format.
"how to format a document"
```

willfish, line 9:
```suggestion
    unsupported_values = values - record.splatted_supported_relationships
```

app/services/include_params_validator.rb
```diff
@@ -22,6 +22,26 @@ def fully_validate!
...
   end
 
+  # Active Model Serializers return all members of a nested chain
+  #
+  # For example:
+  #   'foo.bar.baz' is equivalent to ['foo', 'foo.bar', 'foo.bar.baz']
+  #
+  # To reflect this behaviour in the validation we need to support all variations
+  # of nested chains by preformatting them in this validator.
+  def formatted_supported_relationships
```
willfish, line 32:
```suggestion
  def splatted_supported_relationships
```

app/services/include_params_validator.rb
```diff
@@ -22,6 +22,26 @@ def fully_validate!
...
+    formatted_supported_relationships = Set.new
```
willfish, line 33:
```suggestion
    splatted_supported_relationships = Set.new
```

app/services/include_params_validator.rb
willfish, line 38:
```suggestion
        splatted_supported_relationships << split_chain[0..index].join('.')
```

app/services/include_params_validator.rb
```diff
@@ -22,6 +22,26 @@ def fully_validate!
...
+
+    supported_relationships.each do |chain|
+      split_chain = chain.split('.')
+      split_chain.each_with_index do |_, index|
+        formatted_supported_relationships << split_chain[0..index].join('.')
+      end
+    end
+
+    formatted_supported_relationships.to_a
```
willfish, line 42:
```suggestion
    splatted_supported_relationships.to_a
```

PR #533, opened 2020-06-03
Rails 6

willfish:
### Jira link

P4-<TODO>

### What?

Happily our application is an api, only, and isn't that complex so this was actually surprisingly easy.

- Move us to Rails 6

### Why?

- Keep us continually up-to-date with latest patches and performance enhancements (especially support for garbage collection changes coming in ruby 2.7.1)


Gemfile
```diff
@@ -37,7 +37,7 @@ gem 'paper_trail'
 gem 'pg', '~> 1.0.0'
 gem 'prometheus_exporter'
 gem 'puma', '~> 3.12.6'
-gem 'rails', '~> 5.2.3'
+gem 'rails'
```
cwrw, line 40: Can you lock the rails version here?

PR #537, opened 2020-06-05
Return supplier on locations for moves

willfish:
### Jira link

P4-<TODO>

### What?

- [x] Adds support for explicit support includes
- [x] Adds support for returning suppliers on a from and to location in a move 

### Why?

To clarify supported relationships for the frontend/suppliers
To support the way moves are being consumed by the frontend

### DIFF
--- a/app/serializers/move_serializer.rb
+++ b/app/serializers/move_serializer.rb
@@ -32,7 +32,9 @@ class MoveSerializer < ActiveModel::Serializer
     person.ethnicity
     person.gender
     from_location
+    from_location.suppliers
     to_location
+    to_location.suppliers
     documents
     prison_transfer_reason
     court_hearings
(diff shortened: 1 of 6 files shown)


app/serializers/move_serializer.rb
```diff
@@ -32,7 +32,9 @@ class MoveSerializer < ActiveModel::Serializer
     person.ethnicity
     person.gender
     from_location
+    from_location.suppliers
```
cwrw, line 35: we have locations in other serializers, can we add those there as well, eg allocations

willfish: We're not using it currently in the frontend. I think we should be draconian in supporting only the includes that are needed.

willfish: Although on second thoughts, I'm not sure. I think we should have chat about what includes we support separately. Going to ship this to get the frontend working and we can chat separately.

PR #538, opened 2020-06-05
P4-1579 Add new endpoint to explicitly reject a move

smoothcontract:
### Jira link

P4-1579

### What?

- [x] Adds a new endpoint `POST /moves/:move_id/reject`
- [x] Adds a new event type of `reject` and appropriate behaviour in event runner

### Why?

- This will be called when the OCA rejects a singleton move - they must specify a reason for rejecting the move, and optionally include a comment (which is stored in the existing cancellation_reason_comment attribute.
- It's currently possible to reject a move by calling the `PATCH /moves/:move_id` endpoint but we want to deprecate that approach and instead use distinct endpoints which change move status.
- This is a special case of cancelling a move so is worthy of it's own endpoint that requires a rejection reason to be specified.

### Have you? (optional)

- Adds a new endpoint, doesn't affect existing production endpoints so minimal risk

### DIFF
--- a/spec/factories/events.rb
+++ b/spec/factories/events.rb
@@ -45,6 +45,10 @@ FactoryBot.define do
       event_name { 'lockout' }
     end
 
+    trait :reject do
+      event_name { 'reject' }
+    end
+
     # NB: move_event factory inherits from the event factory
     factory :move_event, class: 'MoveEvent' do
       details do
@@ -75,6 +79,18 @@ FactoryBot.define do
             attributes: {
               cancellation_reason: nil,
               cancellation_reason_comment: 'this is a broken event',
+            },
+          } }
+        end
+      end
+
+      trait :reject do
+        event_name { 'reject' }
+        details do
+          { event_params: {
+            attributes: {
+              rejection_reason: 'no_transport_available',
+              cancellation_reason_comment: 'computer says no',
             },
           } }
         end
--- a/spec/requests/api/v1/move_events_controller_reject_spec.rb
+++ b/spec/requests/api/v1/move_events_controller_reject_spec.rb
@@ -1,0 +1,115 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Api::V1::MoveEventsController do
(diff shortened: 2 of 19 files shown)


spec/factories/events.rb
```diff
@@ -45,6 +45,10 @@
       event_name { 'lockout' }
     end
 
+    trait :reject do
+      event_name { 'reject' }
```
martyn-w, line 49: there seem to be two `:reject` traits in this factory?

martyn-w: ah no, this is for events, not move events

spec/requests/api/v1/move_events_controller_reject_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+      context 'with a bad event type' do
+        let(:reject_params) { { data: { type: 'Foo-bar', attributes: { timestamp: '2020-04-23T18:25:43.511Z', rejection_reason: 'no_space_at_receiving_prison' } } } }
+
+        it_behaves_like 'an endpoint that responds with error 422' do
+          let(:errors_422) { [{ 'title' => 'Invalid type', 'detail' => 'Validation failed: Type is not included in the list' }] }
+        end
+      end
+
+      context 'with a bad rejection_reason' do
+        let(:reject_params) { { data: { type: 'reject', attributes: { timestamp: '2020-04-23T18:25:43.511Z', rejection_reason: 'Yo Momma' } } } }
```
martyn-w, line 107: 😆

spec/services/event_log/move_runner_spec.rb
```diff
@@ -114,6 +114,40 @@
     end
   end
 
+  context 'when event_name=reject' do
+    let!(:event) { create(:move_event, :reject, eventable: move) }
+
+    context 'when the move is requested' do
```
martyn-w, line 120: is it legitimate to reject a `requested` move? should the move be rejected only if it is `proposed`?

martyn-w: leaving this question until we have a well-defined state machine

PR #542, opened 2020-06-08
Add PUT /v2/people endpoint

alexdesi:
### Jira link

https://dsdmoj.atlassian.net/browse/P4-1583

### What?
This PR adds the endpoint to patch a person: PUT /v2/people

### Why?
to allow to change an existing person

### Have you? (optional)


spec/requests/api/v2/people_controller_update_spec.rb
```diff
@@ -0,0 +1,189 @@
...
+          expect(response_error['title']).to eq('Bad request')
+          expect(response_error['detail']).to include('["non-existent-relationship"] is not supported.')
+        end
+      end
+    end
+
+    describe 'webhook and email notifications' do
+      before do
+        allow(Notifier).to receive(:prepare_notifications)
+        put "/api/v1/people/#{person.id}", params: person_params, headers: headers, as: :json
```
martyn-w, line 131: this is going to the wrong URL - should be /v2/, not /v1/

spec/requests/api/v2/people_controller_update_spec.rb
```diff
@@ -0,0 +1,189 @@
...
+      end
+
+      it 'calls the notifier when updating a person' do
+        expect(Notifier).to have_received(:prepare_notifications).with(topic: person, action_name: 'update')
+      end
+    end
+
+    context 'with a bad request' do
+      before { put "/api/v1/people/#{person.id}", params: {}, headers: headers, as: :json }
```
martyn-w, line 140: v2

spec/requests/api/v2/people_controller_update_spec.rb
```diff
@@ -0,0 +1,189 @@
...
+
+      it_behaves_like 'an endpoint that responds with error 400'
+    end
+
+    context 'when not authorized', :with_invalid_auth_headers do
+      let(:content_type) { ApiController::CONTENT_TYPE }
+      let(:headers) { { 'CONTENT_TYPE': content_type } }
+      let(:detail_401) { 'Token expired or invalid' }
+
+      before { put "/api/v1/people/#{person.id}", params: person_params, headers: headers, as: :json }
```
martyn-w, line 150: v2

spec/requests/api/v2/people_controller_update_spec.rb
```diff
@@ -0,0 +1,189 @@
...
+          {
+            'title' => 'Unprocessable entity',
+            'detail' => "Last name can't be blank",
+            'source' => { 'pointer' => '/data/attributes/last_name' },
+            'code' => 'blank',
+          },
+        ]
+      end
+
+      before { put "/api/v1/people/#{person.id}", params: person_params, headers: headers, as: :json }
```
martyn-w, line 184: v2

spec/requests/api/v2/people_controller_update_spec.rb
```diff
@@ -0,0 +1,189 @@
...
+
+      it_behaves_like 'an endpoint that responds with error 401'
+    end
+
+    context 'with an invalid CONTENT_TYPE header' do
+      let(:content_type) { 'application/xml' }
+
+      before { put "/api/v1/people/#{person.id}", params: person_params, headers: headers, as: :json }
```
martyn-w, line 158: v2

swagger/v2/put_people_responses.yaml
```diff
@@ -0,0 +1,52 @@
...
+      $ref: '../v2/person.yaml#/Person'
+'400':
+  type: object
+  required:
+  - errors
+  properties:
+    errors:
+      type: array
+      items:
+        $ref: '../v1/errors.yaml#/BadRequest'
```
martyn-w, line 16: not sure if a v2 endpoint should refer to v1 models or not ? I don't know what the write behaviour is but might be worth making an explicit note that the v2 schema uses v1 models for errors

spec/requests/api/v2/people_controller_update_spec.rb
alexdesi: 😱 thanks to spot that!

app/controllers/api/v2/people_controller.rb
```diff
@@ -3,16 +3,23 @@
...
         render json: person, status: :created, include: included_relationships, serializer: ::V2::PersonSerializer
       end
 
-      def index
-        people = ::V2::People::Finder.new(filter_params).call
+      def update
+        person = Person.find(params[:id])
 
-        paginate people, include: included_relationships, each_serializer: ::V2::PersonSerializer
+        person.update(person_attributes)
```
cwrw, line 20: Can we use `update!` instead of `update` so an exception is raised if the record is invalid

app/controllers/api/v2/people_controller.rb
```diff
@@ -3,16 +3,23 @@
...
+
+        render json: person, status: :ok, include: included_relationships, serializer: ::V2::PersonSerializer
```
cwrw, line 22: since this is shared with the `create` method, maybe we can move it into it's own method `render_person(person, status)`?

spec/requests/api/v2/people_controller_update_spec.rb
```diff
@@ -0,0 +1,189 @@
...
+require 'rails_helper'
+
+RSpec.describe Api::V2::PeopleController do
+  let(:response_json) { JSON.parse(response.body) }
+  let(:access_token) { create(:access_token).token }
+  let(:content_type) { ApiController::CONTENT_TYPE }
+  let(:headers) { { 'CONTENT_TYPE': content_type }.merge('Authorization' => "Bearer #{access_token}") }
+
+  describe 'PUT /api/v2/people' do
+    let(:schema) { load_yaml_schema('put_people_responses.yaml') } # Note: V1 put_people_responses is equal V2 put_people_responses
```
cwrw, line 12: I would suggest we add a new `patch_people_responses.yaml` in v2, this way we fully separate the implementations, even if some yaml files are similar.

spec/swagger/definitions/hand_coded_paths_v2.yaml
```diff
@@ -110,3 +110,62 @@
           application/vnd.api+json:
             schema:
               $ref: post_people_responses.yaml#/422
+/people/{person_id}:
+  patch:
```
cwrw, line 114: can you mark this as deprecated/do not use?

spec/requests/api/v2/people_controller_update_spec.rb
martyn-w: good point, I think its PATCH if you can update some attributes of the object, or PUT if you are replacing the entire object. I'd suggest sticking with PATCH as we use that elsewhere.

app/controllers/api/v2/people_controller.rb
alexdesi: sure, but consider that there is not any validation now, I have verified with Jonny about this.

spec/requests/api/v2/people_controller_update_spec.rb
alexdesi: I agree with you that we need to be consistent,
in this case, I would stick to PUT, since the `/api/v1/people` is using PUT,
and would be strange to have v1 with PUT and v2 with PATCH.

(Also consider that both put and patch point to the same update action in controller.)

spec/swagger/definitions/hand_coded_paths_v2.yaml
alexdesi: 👍

spec/requests/api/v2/people_controller_update_spec.rb
martyn-w: So, reading over https://en.wikipedia.org/wiki/Patch_verb#:~:text=The%20main%20difference%20between%20the,instructions%20to%20modify%20the%20resource. I definitely think this should be PATCH - as you allow the user to update a single attribute rather than the whole object

app/controllers/api/v2/people_controller.rb
cwrw: I think this is slightly different, let's catchup in person

swagger/v2/put_people_responses.yaml
alexdesi: How would you make the explicit note?
This is transparent n the swagger docs, maybe we should move the shared elements to a /shared folder? not for now 😄, it would require a bit of work

spec/requests/api/v2/people_controller_update_spec.rb
alexdesi: Update: yes, actually I can see the Doc states `PATCH` but spec uses PUT (both are supported anyway),
I am happy to change the test to patch (since put will still continue to work)

alexdesi: update2: Changed PUT to PATCH, as suggested.

alexdesi:
@martyn-w @cwrw I've handled your comments, how does it look now?

PR #548, opened 2020-06-09
P4-1496 Add PATCH endpoint for Profile V2

cwrw:
### Jira link

[P4-1496](https://dsdmoj.atlassian.net/browse/P4-1496)

### What?

I have added/removed/altered:

- [x] Added a new update route for profiles nested under people
- [x] Refactored profile controller slightly

### Why?

To allow users to amend profile assessment answers, add a new endpoint under `/people/{person_id}/profiles/{profile_id}` which takes in assessment answers in the format of jsonb and updates them on the record. I intentionally didn't add any validation for now as this will come in later tickets

I've also updated the missing swagger documentation for POST profiles and added the new endpoint for patch.

### Have you? (optional)


app/controllers/api/v2/profiles_controller.rb
```diff
@@ -17,7 +24,6 @@ def create
             %i[
               key
               date
-              expiry_data
```
cwrw, line 20: This wasn't being used

willfish: 👍

app/controllers/api/v2/profiles_controller.rb
```diff
@@ -38,18 +44,26 @@ def profile_params
         params.require(:data).permit(PROFILE_ATTRIBUTES).to_h
       end
 
-      def profile_attributes
+      def new_profile_attributes
```
willfish, line 47: I guess we'll not separate attributes like this later?

spec/requests/api/v2/profiles_controller_create_spec.rb
```diff
@@ -130,6 +130,14 @@
       it_behaves_like 'an endpoint that responds with error 400'
     end
 
+    context 'when the person_id is not found' do
```
willfish, line 133: 👍

app/controllers/api/v2/profiles_controller.rb
cwrw: yea as when we revisit this we won't need the first name/last name

spec/requests/api/v2/profiles_controller_update_spec.rb
```diff
@@ -0,0 +1,205 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Api::V2::ProfilesController do
+  let(:response_json) { JSON.parse(response.body) }
```
willfish, line 6: I keep wondering whether we want to move to move this into the rails helper or something? It's so common in all of our request specs

app/controllers/api/v2/profiles_controller.rb
```diff
@@ -4,8 +4,15 @@ module Api
   module V2
     class ProfilesController < ApiController
```
martyn-w, line 5: Could you add a TODO to say that validation should be added

spec/requests/api/v2/profiles_controller_update_spec.rb
```diff
@@ -0,0 +1,205 @@
...
+  let(:access_token) { create(:access_token).token }
+  let(:content_type) { ApiController::CONTENT_TYPE }
+  let(:headers) { { 'CONTENT_TYPE': content_type }.merge('Authorization' => "Bearer #{access_token}") }
+  let(:risk_type_1) { create :assessment_question, :risk }
```
willfish, line 10: Don't usually feel like `<thing>_<n>` is that revealing as far as naming things is concerned. I guess we get value out of creating multiple assessment questions 🤷‍♂️

willfish: Maybe we could create one and call is `assessment_question` and validate against that?

willfish: There's nothing different for one or more assessment questions when updating jsonb?

spec/requests/api/v2/profiles_controller_update_spec.rb
```diff
@@ -0,0 +1,205 @@
...
+    context 'when the profile_id is not found' do
+      let(:move_id) { 'foo-bar' }
+      let(:detail_404) { "Couldn't find Profile with 'id'=foo-bar" }
+
+      before { patch "/api/v2/people/#{profile.person.id}/profiles/foo-bar", params: profile_params, headers: headers, as: :json }
+
+      it_behaves_like 'an endpoint that responds with error 404'
+    end
+
+    context 'when the person_id is not found' do
```
willfish, line 189: 👍

spec/swagger/definitions/hand_coded_paths_v2.yaml
```diff
@@ -110,3 +110,131 @@
           application/vnd.api+json:
             schema:
               $ref: post_people_responses.yaml#/422
+/people/{person_id}/profiles:
```
willfish, line 113: I guess /api/v2 is shared by a prefix in the overall document :)

spec/requests/api/v2/profiles_controller_update_spec.rb
cwrw: not that I'm aware of no, but I think we should add validation for those though, I've discussed this with Jonny, and we'll pick those up in later work next sprint

spec/swagger/definitions/hand_coded_paths_v2.yaml
cwrw: Yea its on the server name itself, and to view this you have to explicitly navigate to the v2 spec:
Screenshot 2020-06-09 at 09 31 21

PR #549, opened 2020-06-09
[P4-1500] Adds support for documentable relationship

willfish:
### Jira link

P4-1500

### What?

We want to be able to add documents to any type of model. This enables
that.

### Why?

This is needed for upcoming work on the PER and is being done as part of a migration from having documents on a Move to having them on a Profile for the STC flows.


db/migrate/20200609082714_adds_documentable_to_documents.rb
```diff
@@ -0,0 +1,5 @@
+class AddsDocumentableToDocuments < ActiveRecord::Migration[6.0]
+  def change
+    add_reference :documents, :documentable, polymorphic: true, index: true
```
willfish, line 3: Foreign key isn't an option for polymorphic relationships

db/schema.rb
```diff
@@ -91,8 +91,13 @@
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.datetime "discarded_at"
+    t.uuid "profile_id"
```
willfish, line 94: Remove me

martyn-w, line 94: where is the `profile_id` field coming from? its not mentioned in the migration above ^ is there another migration to add to this PR ?

db/migrate/20200609082714_adds_documentable_to_documents.rb
martyn-w: 👍

db/schema.rb
willfish: See subsequent commit. I ran a migration that was abanded/superseded by this :)

PR #551, opened 2020-06-10
P4-1432 Allow relationships to be surfaced in Serializers even if missing

cwrw:
### Jira link

[P4-1432](https://dsdmoj.atlassian.net/browse/P4-1432)

### What?

- [x] Removed checks for hiding missing relationships in Move Serializer
- [x] Removed checks for hiding missing relationships in Person Serializer

### Why?

To adhere to the JSON:API [specs](https://jsonapi.org/format/#document-resource-object-linkage) regarding relationships linkage where an empty relationship should be represented as:
- null for empty to-one relationships.
- an empty array ([]) for empty to-many relationships.

Allow relationships to be surfaced in serializers even if missing.
 
### Have you? (optional)

### DIFF
--- a/app/serializers/move_serializer.rb
+++ b/app/serializers/move_serializer.rb
@@ -22,8 +22,8 @@ class MoveSerializer < ActiveModel::Serializer
   has_one :profile, serializer: ProfileSerializer # <- TODO: update the serializer
 
   has_one :from_location, serializer: LocationSerializer
-  has_one :to_location, serializer: LocationSerializer, if: -> { object.to_location.present? }
-  has_one :prison_transfer_reason, serializer: PrisonTransferReasonSerializer, if: -> { object.prison_transfer_reason.present? }
+  has_one :to_location, serializer: LocationSerializer
+  has_one :prison_transfer_reason, serializer: PrisonTransferReasonSerializer
   has_many :documents, serializer: DocumentSerializer
   has_many :court_hearings, serializer: CourtHearingSerializer
   belongs_to :allocation, serializer: AllocationSerializer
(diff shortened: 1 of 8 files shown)


app/serializers/move_serializer.rb
```diff
@@ -22,8 +22,8 @@ class MoveSerializer < ActiveModel::Serializer
   has_one :profile, serializer: ProfileSerializer
 
   has_one :from_location, serializer: LocationSerializer
-  has_one :to_location, serializer: LocationSerializer, if: -> { object.to_location.present? }
-  has_one :prison_transfer_reason, serializer: PrisonTransferReasonSerializer, if: -> { object.prison_transfer_reason.present? }
+  has_one :to_location, serializer: LocationSerializer
```
cwrw, line 25: Confirming all these changes with the FE team

spec/serializers/move_serializer_spec.rb
```diff
@@ -129,6 +129,25 @@
     it 'contains an included from and to location' do
       expect(result[:included]).to(include_json(expected_json))
     end
+
+    context 'without a to_location' do
```
alexdesi, line 133: What about making the context explicit, adding `to_location: nil` when creating the move?

spec/serializers/move_serializer_spec.rb
```diff
@@ -185,4 +204,41 @@
...
+        },
+      ]
+    end
+
+    it 'contains an included prison_transfer_reason' do
+      expect(result[:included]).to(include_json(expected_json))
+    end
+
+    context 'without a prison_transfer_reason' do
+      let(:move) { create(:move) }
```
alexdesi, line 233:
```suggestion
      let(:move) { create(:move, prison_transfer_reason: nil) }
```

spec/serializers/move_serializer_spec.rb
cwrw: The only time a move can have a `to_location: nil` is when it's a prison recall, which is what is defined  here: https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/551/files#diff-5744dd109e1ed343bd07f744c7d8b00fR141

PR #552, opened 2020-06-10
P4-1650 Add an explicit endpoint to approve a move

smoothcontract:
P4-1650

### What?

- [x] Adds a new route to `POST /moves/:id/approve`
- [x] Adds a new `approve` event type including `date` attribute
- [x] Adds support for `approve` in the move runner service
- [x] Adds validation for date parameter within move events

### Why?

- Currently the front end is PATCHING a proposed move and changing the status to `requested` and updating the move date.
- We want to move away from this approach and have dedicated endpoints which explicitly change move status and ensure the relevant related attributes are correctly specified.

### Have you? (optional)

- Introduces a new API so minimal production risk

### DIFF
--- a/app/services/move_events/params_validator.rb
+++ b/app/services/move_events/params_validator.rb
@@ -4,12 +4,22 @@ module MoveEvents
   class ParamsValidator
     include ActiveModel::Validations
 
-    attr_reader :timestamp, :type, :cancellation_reason, :rejection_reason
+    attr_reader :timestamp, :type, :date, :cancellation_reason, :rejection_reason
 
-    validates :type, presence: true, inclusion: { in: %w[cancel complete lockouts redirects reject] }
+    validates :type, presence: true, inclusion: { in: %w[cancel complete lockouts redirects approve reject] }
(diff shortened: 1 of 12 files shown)


app/services/move_events/params_validator.rb
```diff
@@ -4,12 +4,22 @@ module MoveEvents
...
+    validates :date, presence: true, if: -> { type == 'approve' }
+    validates :timestamp, presence: true
+
+    validates_each :date, allow_nil: true do |record, attr, value|
+      Date.strptime(value, '%Y-%m-%d')
+    rescue ArgumentError
+      record.errors.add attr, 'is not a valid date.'
+    end
+
+    validates_each :timestamp, allow_nil: true do |record, attr, value|
```
smoothcontract, line 22: I tweaked this very slightly - the idea is that if we want to validate other `time` attributes they can be listed here, but we explicitly validate `presence` for each attribute depending on the event type. At the moment we don't have other `time` attributes but I still think it's a nice approach.

PR #553, opened 2020-06-10
[P4-1500] Populate document on profile

willfish:
### Jira link

P4-1500

### What?

- [x] Fixes a broken migration for polymorphic `Document` records
- [x] Altered DocumentsController so that new `Document` records belong to a `Profile` (on write)
- [x] Altered `Move#documents` to support `Document` records belonging to either a `Move` or a `Profile` and not both (on read) for backwards compatibility
- [x] Force suppliers to have permissions to a Move to update Profile and Move documents

### Why?

The created `Document` records relate to assessments done for a `Profile` not the `Move`. This is part of a wider clearing up of how we conceptualise these entities and support them via our apis.

### Future work

Removing support for the backwards compatibility after we've migrated the data with a rake task.
We will change the apis to enable explicit creation of a `Document` for a `Profile` in future PRs


app/controllers/api/v1/documents_controller.rb
```diff
@@ -24,25 +24,27 @@ def destroy
...
 
       def set_restricted_request_content_type
         @restricted_request_content_type = 'multipart/form-data'
       end
+
+      def profile
+        move_id = params[:move_id]
+
+        if move_id
+          Move.accessible_by(current_ability).find(move_id)&.profile
```
alexdesi, line 45: Do we need `&.` to get the profile?

that is part of the Move if it's empty you'd get nil,
it the move is not found it raises an exception

willfish: We did need it, yeah

app/controllers/api/v1/documents_controller.rb
```diff
@@ -24,25 +24,27 @@ def destroy
...
 
     protected
 
       def set_restricted_request_content_type
         @restricted_request_content_type = 'multipart/form-data'
       end
+
+      def profile
+        move_id = params[:move_id]
+
```
alexdesi, line 43: this is just a matter of style, you could use a guard clause
`return unless move_id.present?`

app/models/move.rb
```diff
@@ -124,6 +124,12 @@ def person
     raise 'Attempt to Access to person!!!'
   end
 
+  def documents
+    # We need to make sure that we're returning Documents that are either for the current
+    # moves profile or the current move to support backwards compatibility.
+    Document.kept.where(move_id: id).or(Document.where(documentable: profile))
+  end
```
alexdesi, line 131: Do we need the "backwards compatibility" once the document migration is complete?
if not we should add a reminder, like a TODO, right?

app/controllers/api/v1/documents_controller.rb
willfish: I know about that style. I did find this more readable. If this was a more complex method I'd probably go for an early return.

app/models/move.rb
willfish: I don't need the reminder as I have a ticket detailing that I'll do it later.

spec/requests/api/v1/documents_controller_create_spec.rb
```diff
@@ -78,17 +78,18 @@
     let(:access_token) { create(:access_token).token }
     let(:headers) { { 'CONTENT_TYPE': content_type }.merge('Authorization' => "Bearer #{access_token}") }
 
+    let(:file) do
+      Rack::Test::UploadedFile.new(
+        Rails.root.join('spec/fixtures/file-sample_100kB.doc'),
```
alexdesi, line 83:
```suggestion
        Rails.root.join('spec', 'fixtures', 'file-sample_100kB.doc'),
```

spec/requests/api/v1/documents_controller_create_spec.rb
```diff
@@ -99,17 +100,39 @@
...
+        expect(move.profile.documents.last.file).to be_attached
       end
 
       it 'adds the right file to the document' do
-        expect(move.documents.last.file.filename).to eq 'file-sample_100kB.doc'
+        expect(move.profile.documents.last.file.filename).to eq 'file-sample_100kB.doc'
+      end
+    end
+
+    context 'with an unauthorised supplier' do
```
alexdesi, line 121: why is the supplier unauthorised? maybe this could be clearer from the test

spec/requests/api/v1/documents_controller_destroy_spec.rb
```diff
@@ -29,35 +29,68 @@
...
+        it 'returns the correct data' do
+          expect(response_json).to eq resource_to_json
+        end
 
-      it 'returns the correct data' do
-        expect(response_json).to eq resource_to_json
+        it 'deletes the document from the profile', skip_before: true do
+          expect(move.profile.documents.count).to eq(1)
+          delete "/api/v1/moves/#{move_id}/documents/#{document_id}", headers: headers
+          expect(move.profile.documents.count).to eq(0)
```
alexdesi, line 92:
```suggestion
          expect{delete "/api/v1/moves/#{move_id}/documents/#{document_id}", headers: headers}.to change{Counter.count}.from(1).to(0)

```

app/controllers/api/v1/documents_controller.rb
willfish, line 45:
```suggestion
          Move.accessible_by(current_ability).find(move_id).profile
```

spec/requests/api/v1/documents_controller_create_spec.rb
willfish, line 121:
```suggestion
    context "with a supplier without access to the move's location" do
```

PR #557, opened 2020-06-11
[P4-586] Run migrations out of band

mafonso:
### Jira link

[P4-586](https://dsdmoj.atlassian.net/browse/P4-586)
[P4-1168](https://dsdmoj.atlassian.net/browse/P4-1168)

### What?

I have added/removed/altered:

- [ ] Added a k8s job template to run migrations on a given environment
- [ ] run migrations prior to deployment, testing for success

### Why?

I am doing this because:
- We want to stop running migration every time each instance of the container (re)starts

- It could prevent a deployment from being executed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -235,8 +235,12 @@ commands:
       - run:
           name: run db migrations
           command: |
-             false
-            
+            set -o errexit
+            sed 's/__ENV__/<< parameters.env >>/' k8s_migrate_job.yml | kubectl apply -f -
+            kubectl wait --for=condition=complete job/hmpps-book-secure-move-api-migrate-db-<< parameters.env >>
+            exit_code=$?
+            kubectl delete job/hmpps-book-secure-move-api-migrate-db-<< parameters.env >>
+            exit ${exit_code}
       - deploy:
           name: rolling update image to << parameters.env >>
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -232,7 +232,11 @@ commands:
           command: |
             setup-kube-auth
             kubectl config use-context << parameters.env >>
-
+      - run:
+          name: run db migrations
+          command: |
+             false
+            
       - deploy:
           name: rolling update image to << parameters.env >>
           command: |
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -232,7 +232,15 @@ commands:
           command: |
             setup-kube-auth
             kubectl config use-context << parameters.env >>
-
+      - run:
+          name: run db migrations
+          command: |
+            set +e
```
willfish, line 238: I know this is pedantic, but is there a long form name for this. Like `set -o errexit` or something. Not a blocking comment at all.

Dockerfile
```diff
@@ -46,5 +46,5 @@ COPY --chown=appuser:appgroup --from=build-stage /app /app
 COPY --chown=appuser:appgroup --from=build-stage /usr/local/bundle /usr/local/bundle
 
 USER $APPUID
-ENTRYPOINT ["./run.sh"]
+CMD ["./run.sh"]
```
willfish, line 49: 👍

.circleci/config.yml
mafonso: yeah, fair point on the verbosity :) 
It will still look ugly as CircleCI call the shebang with `#!/bin/bash -eo pipefail`  and I'm overriding it next.

mafonso:
> I'm guessing you've triggered this locally/remotely and tested the migrations do run

Can't really test locally because it depends on the cloud-platform-tools image, that does run fine locally. But it was triggered in a special temporary branch that triggered a dev deploy. 
(It might be useful to have a special branch prefix that deploys stuff to dev for these cases)

Anyway, you can see the actual run [here](https://app.circleci.com/pipelines/github/ministryofjustice/hmpps-book-secure-move-api/3364/workflows/666255a5-cee3-4311-82b0-5d3052d494c1). 

It required this [PR](https://github.com/ministryofjustice/cloud-platform-environments/pull/2548) on Cloud Platform environments be merged so that circle CI could do it.

run.sh
```diff
@@ -2,5 +2,4 @@
 
 # Sadly toml files don't support inline expansion of environment variables
 bundle exec erb config/application_insights.erb >config/application_insights.toml
-bundle exec rake db:create db:migrate
```
mafonso, line 5: Will the removal of `db:create` cause an issue on Heroku?

PR #560, opened 2020-06-11
Hotfix returned allocation relationships for index view

smoothcontract:
Not 100% sure if this is right, will be ideal if this can be tested against the front end.

### Jira link

P4-<TODO>

### What?

I have added/removed/altered:

- [ ] Added foo in bar
- [ ] Removed baz

### Why?

I am doing this because:

-
-
-

### Have you? (optional)

### DIFF
--- a/app/serializers/allocation_serializer.rb
+++ b/app/serializers/allocation_serializer.rb
@@ -19,5 +19,6 @@ class AllocationSerializer < ActiveModel::Serializer
   has_one :to_location
   has_many :moves
 
+  MINIMUM_RELATIONSHIPS = { from_location: [], to_location: [], moves: %i[person] }.freeze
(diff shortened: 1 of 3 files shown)


app/serializers/allocation_serializer.rb
```diff
@@ -19,5 +19,6 @@ class AllocationSerializer < ActiveModel::Serializer
   has_one :to_location
   has_many :moves
 
+  MINIMUM_RELATIONSHIPS = { from_location: [], to_location: [], moves: %i[person] }.freeze
```
willfish, line 22:
```suggestion
  MINIMUM_RELATIONSHIPS = %i[from_location to_location moves.person].freeze
```

smoothcontract: @willfish I'll add that in a later PR, was a bit too hasty on the merge button this time :)

PR #565, opened 2020-06-11
[P4-1665] Removes references to Profile fields

willfish:
### Jira link

P4-1665

### What?

Fixes all references to Profile fields.

Approach was to remove the fields from the database table profiles and
see what breaks

-    t.string "last_name", null: false
-    t.string "first_names", null: false
-    t.date "date_of_birth"
-    t.string "aliases", default: [], array: true
-    t.uuid "gender_id"
-    t.uuid "ethnicity_id"
-    t.uuid "nationality_id"
-    t.jsonb "profile_identifiers"
-    t.string "gender_additional_information"
-    t.integer "latest_nomis_booking_id"

Whatever tests broke as part of the field removal are fixed by this PR

### Why?

This enables us to move forward with separation of Person/Profile

**This PR is meant to be transparent to the api and the e2e tests must still pass**


spec/requests/api/v1/people_controller_update_spec.rb
```diff
@@ -131,7 +131,7 @@
 
       it 'updates an existing person' do
         put "/api/v1/people/#{person.id}", params: person_params, headers: headers, as: :json
-        expect(person.reload.latest_profile.gender_additional_information).to eq gender_additional_information
+        expect(person.reload.gender_additional_information).to eq gender_additional_information
```
willfish, line 134: This test duplicates test on 112

app/services/people/nomis_profile_validator.rb
```diff
@@ -6,8 +6,8 @@ class NomisProfileValidator
 
     validates :latest_nomis_booking_id, presence: true
 
-    def initialize(profile)
-      @latest_nomis_booking_id = profile.latest_nomis_booking_id
+    def initialize(person)
```
martyn-w, line 9: Maybe this class should be renamed to `NomisPersonValidator` as it is no longer validating a profile

app/services/people/crud.rb
```diff
@@ -9,24 +9,20 @@ def initialize(params)
     end
 
     ATTRIBUTES = %i[first_names last_name date_of_birth gender_additional_information].freeze
-    PROFILE_ASSOCIATIONS = %i[gender ethnicity].freeze
+    PERSON_ASSOCIATIONS = %i[gender ethnicity].freeze
```
alexdesi, line 12: Should this have `profiles` as well?
what about using V2::PersonSerializer::SUPPORTED_RELATIONSHIPS ?

app/services/court_hearings/create_in_nomis.rb
```diff
@@ -1,7 +1,7 @@
 module CourtHearings
   class CreateInNomis
     def self.call(move, court_hearings)
-      booking_id = move.profile.latest_nomis_booking_id
+      booking_id = move.profile.person.latest_nomis_booking_id
```
alexdesi, line 4: We can do this later, up to you:
`move.profile.person` is used in many places now, so we should remove the 

```
  def person
    raise 'Attempt to Access to person!!!'
  end
```
on the Move and allow to access the person through the profile.

Not high priority!

app/models/profile.rb
```diff
@@ -4,25 +4,15 @@ class Profile < VersionedModel
...
-  validates :last_name, presence: true
-  validates :first_names, presence: true
+
   validate :validate_assessment_answers
   attribute :assessment_answers, Types::Jsonb.new(Profile::AssessmentAnswers)
-  attribute :profile_identifiers, Types::Jsonb.new(Profile::ProfileIdentifiers)
-
-  scope :ordered_by_name, ->(direction) { order('last_name' => direction, 'first_names' => direction) }
-
-  IDENTIFIER_TYPES = %w[
```
martyn-w, line 23: Todo: triple check with Dom / Jason that `niche_reference` and `athena_reference` can be removed from the system, and check that they have no values on staging or on production.

app/services/people/finder.rb
```diff
@@ -15,25 +15,15 @@ def call
   private
 
     def apply_filters(scope)
-      scope = scope.joins(:profiles)
-
       if filter_params.key?(:police_national_computer)
-        scope = scope.where('profiles.profile_identifiers @> ?', police_national_computer.to_json)
+        scope = scope.where(police_national_computer: filter_params[:police_national_computer])
```
martyn-w, line 19: note that the previous search was a _range_ search (`@>`) whereas the new search is an exact match. Is this intended?

db/schema.rb
```diff
@@ -347,19 +347,19 @@
...
     t.jsonb "assessment_answers"
-    t.jsonb "profile_identifiers"
-    t.string "gender_additional_information"
     t.integer "latest_nomis_booking_id"
+    t.string "gender_additional_information"
+    t.jsonb "profile_identifiers"
+    t.uuid "nationality_id"
+    t.uuid "ethnicity_id"
+    t.uuid "gender_id"
+    t.string "aliases"
```
martyn-w, line 359: this was an array `t.string "aliases", default: [], array: true` but now it isn't? has something else changed?

spec/services/people/creator_spec.rb
```diff
@@ -93,30 +89,9 @@
...
-
     it 'sets the correct Person gender' do
       expect(new_person.gender_id).to eql gender.id
     end
-
-    it 'sets the correct Profile gender' do
-      expect(new_profile.gender_id).to eql gender.id
-    end
-
-    it 'does not set the identifiers for the Person' do
```
martyn-w, line 108: as Person is now the only holder for identifiers, should this test be updated rather than removed?

app/services/people/nomis_profile_validator.rb
willfish: I do agree. Alessandro mentioned this during pair review. I'll rename it and rerun the tests.

app/services/people/crud.rb
willfish: If I add profiles this would change the functionality of the controllers. Leaving for moment and we might update this as part of adding support to create/update a person with a profile in future.

app/services/court_hearings/create_in_nomis.rb
willfish: We've agreed to do this in a separate PR.

app/services/people/finder.rb
willfish: Yeah, the values are totally unique and a person only ever has one unique identifier for each type.

martyn-w:
I have a couple of concerns with these changes:
* moving the identifiers from profile to person changes the relationship from a person and their identifiers from 1:many to 1:1. This change is OK _if_ the reality of the book a secure move service is that a person will never have multiple identifiers (consider giving a false name during arrest for example). @jcavell has confirmed a person will never have multiple `prison_number` or `police_national_computer` identifiers. Unsure about `criminal_records_office`.
* a few old tests have been deleted: e.g. `spec/models/profile/profile_identifier_spec.rb` and `spec/models/profile/profile_identifiers_spec.rb`. These should probably be subsumed into person identifier specs.
* also to flag that a couple of identifiers disappear (`niche_reference`, `athena_reference`) - it has been confirmed by @jcavell  and @teneightfive this is OK
* this is a significant change and should be tested very thoroughly!

app/services/people/finder.rb
martyn-w: what if the front end wants to search with a partial identifier, e.g. `AB1` rather than `AB12345678` etc?

willfish:
```
 Running tests in:
 - Chrome 83.0.4103.97 / macOS 10.15.5

 ✓ With unfound person
 ✓ With existing person

 ✓ With a new person

 ✓ Reason - `Made in error`
 ✓ Reason - `Another reason`

 New move from Prison to Court
 ✓ With unfound person
 ✓ With existing person

 ✓ With existing person

 Download moves
 ✓ As Police user
 ✓ As Prison user
 ✓ As STC user
 ✓ As Supplier user

 Smoke tests
 ✓ As Police user
 ✓ As STC user
 ✓ As Prison user
 ✓ As Supplier user
 ✓ As OCA user


 18 passed (2m 26s)
```

spec/services/people/creator_spec.rb
willfish: Resolved in latest commit

willfish:
As regards the production use of the unmigrated identifiers ( `niche_reference`, `athena_reference`) we proved that the identifiers are not used in production (as they only exist in staging). They are also never filtered in the index action of PeopleController.

PR #568, opened 2020-06-12
[P4-1631] Bring back notification on release.

mafonso:
### Jira link

P4-1631

### What?

I have added/removed/altered:

- [ ] Reinstated a notification that was going when building a docker images in the scope of a release to production.

### Why?

I am doing this because:

- This functionality was lost during the refractoring of the pipeline, as we no longer have a dedicated production build.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -127,6 +127,10 @@ references:
       at: .
 
 executors:
+  basic-executor:
+    docker:
+      - image: cimg/base:2020.06
+
   cloud-platform-executor:
     docker:
       - image: ${ECR_ENDPOINT}/cloud-platform/tools:circleci
@@ -269,9 +273,14 @@ commands:
 
 jobs:
   notify_of_approval:
-    executor: cloud-platform-executor
+    executor: basic-executor
     steps:
       - *notify_slack_of_approval
+
+  notify_of_release:
+    executor: basic-executor
+    steps:
+      - *notify_slack_on_release_start
 
   setup_test_environment:
     executor: test-executor
@@ -374,6 +383,8 @@ workflows:
 
   test-build-deploy:
     jobs:
+      - notify_of_release:
+          <<: *only_deploy_tags
       - setup_test_environment:
           <<: *all_tags
       - api_docs:


.circleci/config.yml
```diff
@@ -166,6 +166,10 @@ commands:
   build_docker_image:
     description: "Builds a Docker image"
     steps:
+      - when:
+            equal: [ /^v.*/, << pipeline.git.tag >> ]
+            steps:
+              - *notify_slack_on_release_start
```
mafonso, line 172: Don't really like this approach, as it might not be in sync with the trigger filters.
Will look at other options.

.circleci/config.yml
```diff
@@ -339,6 +345,7 @@ jobs:
       - setup_remote_docker
       - retag_docker_image:
           env: "preprod"
+          notify_slack: "true"
```
mafonso, line 348: Placing it here result in a two minute delay for the notification, comparing with the previous implementation.
The reason is that we no longer have a dedicated production build but a generic one.

`filter` does not work within a step of a job and `when` conditions, and could not find a way to match a regex on the `conditions` stanza.

teneightfive: Just waned to check at what point during the pipeline will this get sent out? I think previously we had it sending pretty much at the start of the pipeline before the tests, etc have passed, so that we get made aware as soon as possible. 

Then we send 2 further notifications - 1) when the tests pass and the manual approval step is ready and 2) when the production deploy job completed.

I think step 2) could be improved to actually check that the correct version has been deployed because with our current workflow is there were any errors spinning up the pods we don't know about it

mafonso: It is being sent out a bit later than before.  
Before the change it was going out before the image build, and that happened early in the pipeline. This was because there was a build specifically for production. 

Currently the build is generic and it has no context of the environment it's on, so it's going out on the next step that has that context, which is immediately after the image was built. This is 2-3 minutes later than before, but I agree that it is really close to the actual deployment step, which now just takes a few seconds.

Although we have internal variable that surface the branch and tags, I could not find a way of comparing it with a regex, the same way you can do with jobs filters.

I could potentially create a dedicated job in the pipeline to send this notification out, but it would consume an executor.  But since it does not happen often it may be worth trying.

Regarding the test of the correct deployed version, it is planed and there is a indeed a branch for that. It will come in a separate PR.

teneightfive: >Regarding the test of the correct deployed version, it is planed and there is a indeed a branch for that. It will come in a separate PR.

Amazing! It would be good if we can mimic that on the frontend pipeline too once we've figured out a solution.

>I could potentially create a dedicated job in the pipeline to send this notification out, but it would consume an executor. But since it does not happen often it may be worth trying.

I think it would be good to bring the notification forward as much as possible if we can to give people a bit of a heads up that it is being released. I think otherwise the risk is that if the notification happens too close to the actual deploy there is no time to raise any concerns around that deployment if people have forgotten to mention anything.

Are there any other jobs that are only run for version tags that we can trigger this notification within? If not, maybe a separate job is best. I think a separate job would also prevent this regression as its clear that is is part of a specific part of the pipeline.

mafonso:
As per chat with @teneightfive we agreed that it was preferable to have a dedicated step for the notification really soon in the pipeline. 
It is consuming one more executor, but, it's using a very basic image and takes about 5 secs and does not block any other step.

PR #571, opened 2020-06-15
[P4-1500] Adds rake task for moving a Document to a Profile

willfish:
### Jira link

P4-1500

### What?

We are moving Documents off of the Move and onto the Profile, see: https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/553.

- [x] Adds a rake task and service for moving Documents from a Move to a Profile
- [x] Adds a service to test and encapsulate the behaviour
- [x] Adds generation of seed data to validate moving documents

### Why?

We need this to move historical documents where new documents have already been updated. We're doing this because the documents are about the PTR (aka assessment answers) and not about the Move itself.

### Future things

- Run this rake task on all environments
- Make sure we've tidied up the Move#documents method to only return the Profile's documents
- Remove move_id on a Document
- Remove seeding a document on a Move


spec/services/document_mover_spec.rb
```diff
@@ -0,0 +1,83 @@
+require 'rails_helper'
+
+# TODO: Remove this once we've migrated all profile attributes to a person and are updating these attributes dynamically
```
willfish, line 3:
```suggestion
# TODO: Remove this once we've migrated all moves to a profile
```

PR #574, opened 2020-06-15
Add support for automatically rebooking a move when rejecting

smoothcontract:
### Jira link

P4-1633

### What?

- [x] Adds an additional `rebook` parameter to the move reject request
- [x] Adds method to move model to rebook (create a new move) with pertinent details from original move
- [x] Added new `original_move` relationship to allow rebooked moves to relate to the original move
- [x] Included `original_move` relationship in Move serializer

### Why?

- This allows the front end to specify that a move should be automatically be rebooked in 7 days; the new move is a proposed move with allocation, profile, location and date details copied across from the original move. This provides a convenience for the PMU/OCA to create a new move in future.

### Have you? (optional)

- This API endpoint is not yet implemented in the front end, so low risk (story to implement is being worked on at present)

### DIFF
--- a/app/models/move.rb
+++ b/app/models/move.rb
@@ -90,8 +90,12 @@ class Move < VersionedModel
     )
   end
 
+  def rebooked
+    self.class.find_by(original_move_id: id)
+  end
+
   def rebook
-    self.class.new(
+    rebooked || self.class.create(
       original_move_id: id,
       from_location_id: from_location_id,
       to_location_id: to_location_id,
--- a/app/models/move.rb
+++ b/app/models/move.rb
@@ -42,6 +42,7 @@ class Move < VersionedModel
   belongs_to :profile, optional: true
   belongs_to :prison_transfer_reason, optional: true
   belongs_to :allocation, inverse_of: :moves, optional: true
+  belongs_to :original_move, class_name: 'Move', optional: true
   # using https://github.com/jhawthorn/discard for documents, so only include the non-soft-deleted documents here
   has_many :documents, -> { kept }, dependent: :destroy, inverse_of: :move
   has_many :notifications, as: :topic, dependent: :destroy # NB: polymorphic association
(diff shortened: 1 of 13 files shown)


app/services/event_log/move_runner.rb
```diff
@@ -23,6 +23,7 @@ def call
           move.rejection_reason = event.rejection_reason
           move.cancellation_reason = Move::CANCELLATION_REASON_REJECTED
           move.cancellation_reason_comment = event.cancellation_reason_comment
+          move.rebook.save if event.rebook?
```
martyn-w, line 26: Keep in mind that the runner can/will run a move's events multiple times. Will this change create a new move record whenever the runner runs the move's event queue? Probably better to only create a new record if there isn't already a rebooked move? And in that case you migt want to add an index on `original_move_id` ?

db/schema.rb
```diff
@@ -218,6 +218,7 @@
     t.date "date_to"
     t.uuid "allocation_id"
     t.string "rejection_reason"
+    t.uuid "original_move_id"
```
martyn-w, line 221: might need an index if being searched on (see earlier comment)

app/models/move.rb
```diff
@@ -89,6 +90,19 @@ def cancel(reason: CANCELLATION_REASON_OTHER, comment: nil)
     )
   end
 
+  def rebook
+    self.class.new(
```
martyn-w, line 94: I think this should only be `new` if it is not already rebooked - otherwise it should return the existing rebook record

spec/models/move_spec.rb
```diff
@@ -412,6 +412,50 @@
     end
   end
 
+  describe '#rebook' do
```
martyn-w, line 415: could you add a test for when the move is already rebooked?

db/schema.rb
smoothcontract: Added a foreign key for this (should have added that before really)

spec/models/move_spec.rb
smoothcontract: Done 👍

app/models/move.rb
smoothcontract: Done 👍

app/services/event_log/move_runner.rb
smoothcontract: Very good point - now refactored to make `rebook` idempotent, so re-running the `reject` event will not create multiple rebookings.

PR #575, opened 2020-06-15
lastest -> latest

mafonso:
### Jira link

P4-1631 fix

### What?

I have added/removed/altered:

- [ ] Fixed a mispelt docker tag in the re-taging job

### Why?

I am doing this because:

- It was using a untracked tag and causing deployments and jobs not to pick it up. 
-


.circleci/config.yml
```diff
@@ -239,6 +239,7 @@ commands:
             sed 's/__ENV__/<< parameters.env >>/' k8s_migrate_job.yml | kubectl apply -f -
             kubectl wait --for=condition=complete job/hmpps-book-secure-move-api-migrate-db-<< parameters.env >>
             exit_code=$?
+            kubectl logs job/hmpps-book-secure-move-api-migrate-db-< parameters.env >>
```
willfish, line 242: This puts logs for the job into the circle ci output

.circleci/config.yml
```diff
@@ -210,7 +210,7 @@ commands:
           command: |
             MANIFEST=$(aws ecr batch-get-image --repository-name "${GITHUB_TEAM_NAME_SLUG}/${CIRCLE_PROJECT_REPONAME}" \
             --image-ids imageTag=${CIRCLE_SHA1} --query 'images[].imageManifest' --output text)
-            aws ecr put-image --repository-name "${GITHUB_TEAM_NAME_SLUG}/${CIRCLE_PROJECT_REPONAME}" --image-tag << parameters.env >>.lastest \
+            aws ecr put-image --repository-name "${GITHUB_TEAM_NAME_SLUG}/${CIRCLE_PROJECT_REPONAME}" --image-tag << parameters.env >>.latest \
```
willfish, line 213: We reference the latest sha in migrations and the current sha on deployment which is why we didn't run the latest migrations with the latest code.

PR #582, opened 2020-06-16
Collect Alert and PersonalCareNeeds in create profile.

alexdesi:
### Jira link

https://dsdmoj.atlassian.net/browse/P4-1665

### What?

This PR adds the capability to collect the alerts and personal care needs from Nomis, whenever the `prison_number` is provided

### Why?

To create a new Profile with correct assessment_answers

### Have you? (optional)


spec/requests/api/v1/profiles_controller_create_spec.rb
```diff
@@ -56,6 +56,47 @@
...
+      let(:profile_params) do
+        {
+          data: {
+            type: 'profiles',
+            attributes: {},
+          },
+        }
+      end
+
+      context 'when the person has the prison_number' do
```
willfish, line 70:
```suggestion
      context 'when the person has a prison_number' do
```

spec/requests/api/v1/profiles_controller_create_spec.rb
```diff
@@ -56,6 +56,47 @@
...
+
+          post "/api/v1/people/#{person.id}/profiles", params: profile_params, headers: headers, as: :json
+        end
+
+        it 'creates the profile using NOMIS attributes' do
+          expect(Profiles::ImportAlertsAndPersonalCareNeeds).to have_received(:call)
+                                                                  .with(person.profiles.last, person.prison_number)
+        end
+
+        context 'when the person has NOT a prison_number' do
```
willfish, line 84:
```suggestion
        context 'when the person does NOT have a prison_number' do
```

spec/requests/api/v1/profiles_controller_create_spec.rb
```diff
@@ -56,6 +56,47 @@
...
+          let(:prison_number) { nil }
+
+          before do
+            allow(Profiles::ImportAlertsAndPersonalCareNeeds).to receive(:call)
+
+            post "/api/v1/people/#{person.id}/profiles", params: profile_params, headers: headers, as: :json
+          end
+
+          it 'the profile has NOT assessment answers' do
```
willfish, line 93:
```suggestion
          it 'the profile does NOT have assessment answers' do
```

app/services/profiles/import_alerts_and_personal_care_needs.rb
```diff
@@ -0,0 +1,13 @@
+module Profiles
+  class ImportAlertsAndPersonalCareNeeds
+    def self.call(profile, prison_number)
+      alerts = NomisClient::Alerts.get([prison_number]).group_by { |p| p.fetch(:offender_no) }
+      personal_care_needs = NomisClient::PersonalCareNeeds.get([prison_number]).group_by { |p| p.fetch(:offender_no) }
+
+      Alerts::Importer.new(profile: profile, alerts: alerts.fetch(prison_number, [])).call
+      PersonalCareNeeds::Importer.new(profile: profile, personal_care_needs: personal_care_needs.fetch(prison_number, [])).call
+
+      profile
```
willfish, line 10:
```suggestion
      profile.save
```

We currently don't save the attributes coming from Nomis, we just assign them:

https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/models/profile.rb#L33

That's one of the misnomer consequences of the importer services. They don't actually import anything 😢

willfish: See also https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/services/moves/import_people.rb#L35-L38

willfish:
```suggestion
      profile.save!
```

alexdesi: Good point! I'll update the tests since the function now returns true/false.

alexdesi:
I'm updating the Swagger docs too.

PR #588, opened 2020-06-16
Extract logs on migrations job

mafonso:
### Jira link
Related to:
P4-1631

### What?

I have added/removed/altered:

- [ ] Extract the logs of the migration kube job into CircleCI output.

### Why?

I am doing this because:

- So that we can see the output of the migrations job in the pipeline.
-
-


mafonso:
This change was reverted because we lacked permissions to fetch the logs.
The PR to fix those permissions was approved, so we can bring this back.

https://github.com/ministryofjustice/cloud-platform-environments/pull/2572

PR #591, opened 2020-06-16
Attempt to speed up the API docs generation job in CircleCI

smoothcontract:
### What?

- [x] Renamed the few request specs that include `rswag` tag to have a filename we can easily identify
- [x] Updated glob expression in Circle CI API docs job to looks specifically for these `_rswag` files only

### Why?

- I noticed that this job often lags behind the others in the pipeline. We can't use parallelism to speed this up because we need a single output file for the swagger docs YAML. However I noticed that the glob to run the `rswag` specs outside of the recommended `specs/swagger` location is actually running **all** of the request specs which is huge overkill.

- It turns out only 3 files live in `spec/requests` and have the `rswag` tag attached so only those three need to be run. I've renamed those files to add a `_rswag` convention to the file so we can isolate just those specs and amended the expression that locates these files within the CircleCI job. Also it turns out that we don't have any `*_helper` files living under `spec/swagger` so I've removed those too.

- We're moving away from rswag gem gradually so I don't expect any further specs to be introduced with `rswag` tag so this quick workaround to rename the current files should help run only the 3 relevant specs and make the job much faster. In future we can rewrite these three specs to avoid using the `rswag` tag completely, and eventually be able to run `rake swaggerize` instead and have it parse the remaining swagger specs very quickly without running them.


.circleci/config.yml
```diff
@@ -108,7 +108,7 @@ references:
     run:
       name: Generate API Docs
       command: |
-        TESTS=$(circleci tests glob "spec/swagger/*_helper.rb" "spec/requests/**/*_spec.rb" | circleci tests split )
```
willfish, line 111: We renamed the request specs but we are now no longer testing the swagger directory specs?

PR #592, opened 2020-06-16
Remove unused allocation events endpoint

smoothcontract:
# Jira link

P4-1667

### What?

- [x] Completely removed `POST /allocations/:id/events` endpoint

### Why?

- Previously we implemented `POST /allocations/:id/events` but this has been replaced by a new dedicated endpoint `POST /allocations/:id/cancel`.
- The original endpoint did not actually get implemented in the front end so can be safely removed.
- It's nice to clean the code up.

### Have you? (optional)

- Removes an api that is not used in production


spec/models/allocation_spec.rb
```diff
@@ -126,12 +126,6 @@
       expect(allocation.moves_count).to eq(0)
     end
 
-    it 'changes the status of all associated moves to cancelled' do
```
cwrw, line 129: just to double check that there are tests for this at the new endpoint to ensure move/allocation mirror the cancellation reason/comment etc?

smoothcontract: Yep, covered here: https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/spec/requests/api/v1/allocation_events_controller_cancel_spec.rb#L63
and
https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/spec/services/event_log/move_runner_spec.rb#L87

PR #595, opened 2020-06-17
Amend includes for an allocation

cwrw:
Since we now support profiles instead of person, change includes to
support profile as well in the allocation serializer. I've kept the old person associations to allow for the final cleanup that will be done by the people/profile team.

### DIFF
--- a/spec/serializers/allocation_serializer_spec.rb
+++ b/spec/serializers/allocation_serializer_spec.rb
@@ -126,8 +126,8 @@ RSpec.describe AllocationSerializer do
       expect(result_data[:relationships][:moves][:data]).to contain_exactly(id: move.id, type: 'moves')
     end
 
-    it 'contains an included move and person' do
-      expect(result[:included].map { |r| r[:type] }).to match_array(%w[people moves locations locations genders ethnicities])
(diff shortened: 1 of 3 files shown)


spec/serializers/allocation_serializer_spec.rb
```diff
@@ -126,8 +126,8 @@
       expect(result_data[:relationships][:moves][:data]).to contain_exactly(id: move.id, type: 'moves')
     end
 
-    it 'contains an included move and person' do
-      expect(result[:included].map { |r| r[:type] }).to match_array(%w[people moves locations locations genders ethnicities])
+    it 'contains an included move and profile' do
+      expect(result[:included].map { |r| r[:type] }).to match_array(%w[people profiles moves locations locations genders ethnicities])
```
smoothcontract, line 130: `locations` is listed twice here - is that deliberate?

cwrw: yea, its to capture both to location and from location

smoothcontract: Cool 👍

PR #598, opened 2020-06-17
Refactor state machine integration with models

smoothcontract:
### Jira link

P4-1657

### What?

- [x] Added a shared concern to be used where we want to include a state machine with a model
- [x] Refactored `Allocation` and `Journey` models to use shared concern for state machine inclusion
- [x] Remove legacy handling for nil/none state on allocations (all allocations now have a state)
- [x] Added a DB constraint and model validation rule to ensure allocation status is not null
- [x] Removed/reduced redundant specs

### Why?

- There was some divergence in the approach to integrating the state machine with each model, this cleans that up and means it's more straightforward to add a state machine to `Move` in future.
- The `Allocation` state machine no longer needs to handle `nil` status (the state machine and status attribute was added after we already had some production data, but all production data has been cleaned up to set an explicit status value).

### DIFF
--- a/app/models/allocation.rb
+++ b/app/models/allocation.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class Allocation < VersionedModel
+  include StateMachineable
+
   ALLOCATION_STATUS_UNFILLED = 'unfilled'
   ALLOCATION_STATUS_FILLED = 'filled'
   ALLOCATION_STATUS_CANCELLED = 'cancelled'
@@ -35,6 +37,8 @@ class Allocation < VersionedModel
   has_many :moves, inverse_of: :allocation, dependent: :destroy, autosave: true
   has_many :events, as: :eventable, dependent: :destroy
 
+  validates :status, presence: true
+
   validates :from_location, presence: true
   validates :to_location, presence: true
 
@@ -49,7 +53,8 @@ class Allocation < VersionedModel
 
   attribute :complex_cases, Types::Jsonb.new(Allocation::ComplexCaseAnswers)
 
-  after_initialize :initialize_state
+  has_state_machine AllocationStateMachine, on: :status
+
   delegate :fill, :unfill, to: :state_machine
 
   def cancel(reason: CANCELLATION_REASON_OTHER, comment: 'Allocation was cancelled')
@@ -71,21 +76,5 @@ class Allocation < VersionedModel
     self.moves_count = current_moves.count
 
     save!
-  end
-
-private
-
-  def state_machine
-    @state_machine ||= AllocationStateMachine.new(self)
-  end
-
-  def initialize_state
-    if status.present?
-      state_machine.restore!(status.to_sym)
-    else
-      # TODO: restore after data is backfilled on production
-      # self.status = state_machine.current
-      self.status = moves.not_cancelled.unfilled? ? :unfilled : :filled
-    end
   end
 end
--- a/app/models/allocation.rb
+++ b/app/models/allocation.rb
@@ -18,7 +18,7 @@ class Allocation < VersionedModel
     long: 'more_than_16',
   }
 
-  enum status: {
+  enum states: {
     unfilled: ALLOCATION_STATUS_UNFILLED,
     filled: ALLOCATION_STATUS_FILLED,
     cancelled: ALLOCATION_STATUS_CANCELLED,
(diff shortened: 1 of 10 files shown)


app/models/concerns/state_machineable.rb
```diff
@@ -0,0 +1,41 @@
...
+  included do
+    after_initialize :initialize_state
+
+    class_attribute :state_attribute
+    class_attribute :state_machine_class
+  end
+
+  def reload(options = nil)
+    # Override to ensure state machine state is correctly restored. Unfortunately there's not a callback for this.
+    super.tap { restore_state }
```
martyn-w, line 15: 👍

app/models/journey.rb
```diff
@@ -18,6 +18,8 @@
 # The supplier is paid for only the first journey A-->B but not B-->C.
 #
 class Journey < ApplicationRecord
+  include StateMachineable
```
martyn-w, line 21: very nice 😄

martyn-w:
👍 good work - very happy with the Journey side of things, as it removes all the ugly hackery from the journey model.
Maybe @cwrw could review the allocation side of things, as the changes there are more significant?

app/models/allocation.rb
```diff
@@ -35,6 +37,8 @@ class Allocation < VersionedModel
   has_many :moves, inverse_of: :allocation, dependent: :destroy, autosave: true
   has_many :events, as: :eventable, dependent: :destroy
 
+  validates :status, presence: true
```
cwrw, line 40: we could validate for inclusion here as well

app/models/concerns/state_machineable.rb
```diff
@@ -0,0 +1,37 @@
...
+  extend ActiveSupport::Concern
+
+  included do
+    after_initialize :initialize_state
+
+    class_attribute :state_attribute
+    class_attribute :state_machine_class
+  end
+
+  def reload(options = nil)
```
cwrw, line 13: My main concern with this is that we won't return what's persisted in the database but rather our state machine current state (if for instance the state is `nil`) This might lead to some unexpected behaviour. Just wondering why we needed to add this for journeys, was it for testing purposes?

app/models/allocation.rb
```diff
@@ -72,20 +77,4 @@ def refresh_status_and_moves_count!
...
-    @state_machine ||= AllocationStateMachine.new(self)
-  end
-
-  def initialize_state
-    if status.present?
-      state_machine.restore!(status.to_sym)
-    else
-      # TODO: restore after data is backfilled on production
-      # self.status = state_machine.current
-      self.status = moves.not_cancelled.unfilled? ? :unfilled : :filled
```
cwrw, line 88: 👍 for removing this and restoring the default

spec/services/allocations/finder_spec.rb
```diff
@@ -144,8 +143,8 @@
       context 'with nil status' do
         let(:filter_params) { { status: nil } }
 
-        it 'returns only allocations without a status' do
-          expect(allocation_finder.call).to contain_exactly(allocation)
+        it 'returns empty results set' do
```
cwrw, line 146: Just to double check this has been coordinated with the FE to make sure they pass status filters correctly just incase

smoothcontract: This is just updating the spec to reflect the fact that we don't have any allocations with a `nil` status - if the front end is passing this filter as `nil` currently then no allocations would be returned, so it's safe to conclude this filter is either being omitted when not required.

app/models/concerns/state_machineable.rb
smoothcontract: Yes, this is just to support the specs - we don't have any code that calls `reload` outside of the test suite. This was particularly affecting the event runner for journeys - the runner actions state machine changes correctly but when the spec calls e.g. `journey.reload` the object under test was fetching the updated state attribute but not updating the state machine accordingly.

ActiveRecord has an `after_find` callback that would seem suited for this, but it's not called as part of `reload`. The alternative to this would be explicitly finding objects again within the test suite which is a bit of a pain, so this seems an acceptable approach to mitigate that.

smoothcontract:
> really good refactor, and great to cleanup the allocation states as we no longer have `nil` statuses. I'm wondering if we can cleanup the allocation state machine itself further, now we longer have to cater for `nil` values?

Well reminded - I've cleaned it up a bit further 👍

PR #601, opened 2020-06-18
Remove tty-prompt from project

willfish:
### Jira link

P4-<TODO>

### What?

- Removes tty-prompt gem
- Remove references to tty-prompt

This 1 gem brings in 8 gems and adds very little value - it was giving us a fzf interface for choosing from 3 options when `gets` and a message saying which options are possible works just as well.

I've ran all three variations of this and it works as we expect.

### Why?

I generally prefer to maintain as few deps as possible to make our lives
easier/give us fewer things to worry about from a security point of view

### DIFF
--- a/lib/tasks/create_application.rake
+++ b/lib/tasks/create_application.rake
@@ -25,5 +25,6 @@ namespace :auth do
     puts "client_id: #{application.uid}"
     puts "client_secret: #{application.plaintext_secret}"
     puts "supplier: #{supplier}"
+    puts "supplier_id: #{application.owner_id}"
   end
 end
--- a/lib/tasks/create_application.rake
+++ b/lib/tasks/create_application.rake
@@ -1,40 +1,28 @@
 # frozen_string_literal: true
 
-# NB: there is an issue in Doorkeeper which will cause an "uninitialized constant Doorkeeper::Application" error when
-# running this rake task in production mode. It relates to the lazy-loading of ActiveRecord and the work-around is to call
-# ActiveRecord::Base.logger first. See https://github.com/doorkeeper-gem/doorkeeper/issues/1319
-ActiveRecord::Base.logger
-
 namespace :auth do
-  NO_OWNER = 'No owner'
-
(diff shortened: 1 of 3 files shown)


willfish:
```shell
rake auth:create_client_application

What's the application's name?
qux

Which application owner should your tokens be associated with? ["geoamey", "serco", "none"]
geoamey

Created OAuth2 client with (name: qux)
client_id: pIg9LJZqUWpyVyxchaSBBmBY0xuggCkI0-IMeM33ZFM
client_secret: 9SXJSFBGMSsqY2DI_-5O8w_0HOIWaIhvP0SDUOeDeWM
supplier: geoamey
```

willfish:
```
What's the application's name?
wibble
Which application owner should your tokens be associated with? ["geoamey", "serco", "none"]
none
Created OAuth2 client with (name: piss)
client_id: lp9o8vSQBi5HDFpCfIrGrxhuPwo4C1VFrKizl7BLXfM
client_secret: aa0sbwG0RIlednVafhFvsaKeCIkD5_EAW35PbUGemKM
supplier: none
supplier_id:
```

lib/tasks/create_application.rake
```diff
@@ -1,40 +1,30 @@
...
 
     suppliers = Supplier.all.each_with_object({}) { |supplier, accumulator|
-      accumulator[supplier.name] = supplier.id
-    }.merge(NO_OWNER => nil)
-    owner_id = prompt.select('Is this application owned by a supplier?', suppliers)
-    if owner_id != NO_OWNER
-      owner = Supplier.find(owner_id)
-      application.owner = owner
-    end
+      accumulator[supplier.name.downcase] = supplier.id
```
willfish, line 14: Downcasing to make it easier to type in the CLI

martyn-w: no need for downcase: you can do `supplier.key` instead, e.g. 
```
suppliers = Supplier.all.map{|s| [s.key,  s.id]}.to_h.merge('none' => nil)
```

PR #608, opened 2020-06-19
[P4-1666] Drop V1 module

willfish:
### Jira link

P4-1666

### What?

- Remove v1 module namespacing

### Why?

The v1 module was chosen for us despite not having versioned apis. How we roll versioning is to say that everything is unversioned expect (supported under `/api/<resource_name>`) and specific versions are loaded for a version header.

That makes the version module redundant and confusing.


willfish:
@cwrw 

```
 Running tests in:
 - Chrome 83.0.4103.106 / macOS 10.15.5

 Cancel allocation
 ✓ Cancel allocation

 New PMU allocation

 ✓ With unfound person
 ✓ With existing person

 ✓ With a new person

 ✓ Reason - `Made in error`
 ✓ Reason - `Another reason`

 New move from Prison to Court
 ✓ With unfound person
 ✓ With existing person

 ✓ With existing person

 Download moves
 ✓ As Police user
 ✓ As Prison user
 ✓ As STC user
 ✓ As Supplier user

 Smoke tests
 ✓ As Police user
 ✓ As STC user
 ✓ As Prison user
 ✓ As Supplier user
 ✓ As OCA
```

PR #609, opened 2020-06-19
Add 415 http error if version in not supported.

alexdesi:
### Jira link

https://dsdmoj.atlassian.net/browse/P4-1666

### What?

Add 415 error in case the API version in the header is wrong.

### Why?
To add error check in the version supported.
### Have you? (optional)

### DIFF
--- a/app/controllers/api_controller.rb
+++ b/app/controllers/api_controller.rb
@@ -1,9 +1,10 @@
 # frozen_string_literal: true
 
 class ApiController < ApplicationController
-  before_action :extend_versioned_controller_actions
   before_action :doorkeeper_authorize!, if: :authentication_enabled?
   before_action :restrict_request_content_type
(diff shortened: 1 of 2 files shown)


app/controllers/api_controller.rb
```diff
@@ -203,11 +212,12 @@ def supported_relationships
 
   def api_version
     res = request.headers['Accept'].match(REGEXP_API_VERSION)
-    res&.[](:version) || '1'
+    res&.[](:version)
   end
 
   def extend_versioned_controller_actions
-    version = "Api::V#{api_version}"
+    default_version = '1'
```
willfish, line 219: We could make this a constant rather than loading it on every request?

alexdesi: I agree

willfish:
We're going to add this error to the v2 swagger docs

PR #615, opened 2020-06-20
[P4-1500] Dereference Move#documents

willfish:
### Jira link

P4-1500

### What?

We've migrated away from having documents on a move to a polymorphic
version of documents on a profile. This PR removes all reference to
a document on a move.

The approach was to remove the relationship and overridden
Move#documents method and fix broken specs.

**Updates to swagger documentation are happening in v2 separately where we're dropping support for documents and people on moves**

### Why?

We're moving away from having `Documents` on a `Move`.


app/controllers/api/moves_controller.rb
```diff
@@ -19,6 +19,10 @@ def show
 
     def create
       move = Move.new(new_move_attributes)
+
+      move.profile.documents = Document.where(
+        id: (new_move_params.dig(:relationships, :documents, :data) || []).map { |doc| doc[:id] },
+      )
```
alexdesi, line 25: I don't want to push you to create a service, but we should simplify this,
maybe you can just create a private method for
`new_move_params.dig(:relationships, :documents, :data) || []).map { |doc| doc[:id] }`
or maybe you can move the

```Document.where(
        id: (new_move_params.dig(:relationships, :documents, :data) || []).map { |doc| doc[:id] }
```
into the model?

app/services/moves/updater.rb
```diff
@@ -55,6 +53,9 @@ def attributes
 
       attributes[:profile] = Profile.find_by(id: profile_attributes.dig(:data, :id)) if profile_attributes.present?
 
+      profile = attributes[:profile] || move.profile
+      profile.documents = Document.where(id: document_ids) unless document_attributes.nil?
```
alexdesi, line 57:
```suggestion
      profile.documents = Document.where(id: document_ids) if document_attributes.present?
```

spec/requests/api/moves_controller_update_spec.rb
```diff
@@ -129,7 +129,7 @@
           expect(response_json).to eq resource_to_json
         end
 
-        context 'when changing a moves documents', :skip_before do
+        context 'when changing a moves profile documents', :skip_before do
```
alexdesi, line 132: still infamouse :skip_before 😱 
I guess we got not time for now to fix it

spec/requests/api/moves_controller_update_spec.rb
```diff
@@ -150,10 +150,10 @@
             }
           end
 
-          it 'updates the moves documents' do
-            expect(move.reload.documents).to match_array(before_documents)
+          it 'updates the moves profile documents' do
```
alexdesi, line 153: should we call it just "profile documents" instead of "moves profile documents"

spec/requests/api/moves_controller_update_spec.rb
```diff
@@ -177,10 +177,10 @@
               }
             end
 
-            it 'updates the moves documents' do
-              expect(move.reload.documents).to match_array(before_documents)
+            it 'updates the moves profile documents' do
```
alexdesi, line 180: same comment as before

spec/requests/api/moves_controller_update_spec.rb
```diff
@@ -150,10 +150,10 @@
             }
           end
 
-          it 'updates the moves documents' do
-            expect(move.reload.documents).to match_array(before_documents)
+          it 'updates the moves profile documents' do
+            expect(move.profile.documents).to match_array(before_documents)
```
alexdesi, line 154: do you think we need to test something before the action is done?
that means that we do not trust our setup

spec/requests/api/moves_controller_update_spec.rb
```diff
@@ -177,10 +177,10 @@
               }
             end
 
-            it 'updates the moves documents' do
-              expect(move.reload.documents).to match_array(before_documents)
+            it 'updates the moves profile documents' do
+              expect(move.profile.documents).to match_array(before_documents)
```
alexdesi, line 181: expectation before action, as before

spec/requests/api/moves_controller_update_spec.rb
```diff
@@ -205,9 +205,9 @@
             end
 
             it 'removes the documents from the move' do
-              expect(move.reload.documents).to match_array(before_documents)
+              expect(move.profile.documents).to match_array(before_documents)
```
alexdesi, line 208: same

spec/requests/api/moves_controller_update_spec.rb
```diff
@@ -220,9 +220,9 @@
             end
 
             it 'does not remove documents from the move' do
-              expect(move.reload.documents).to match_array(before_documents)
+              expect(move.profile.documents).to match_array(before_documents)
```
alexdesi, line 223: same

spec/services/moves/updater_spec.rb
```diff
@@ -6,8 +6,9 @@
   subject(:updater) { described_class.new(move, move_params) }
 
   let(:before_documents) { create_list(:document, 2) }
-  let!(:from_location) { create :location }
-  let!(:move) { create :move, :proposed, move_type: 'prison_recall', from_location: from_location, documents: before_documents }
+  let!(:from_location) { create(:location) }
+  let!(:move) { create(:move, :proposed, move_type: 'prison_recall', from_location: from_location, profile: profile) }
```
alexdesi, line 10: Maybe in the meantime we could remove the bang method?
a solution could be put them in a `before do`

app/services/moves/updater.rb
willfish: Can't do that because nil and empty [] mean specific things (see specs)

spec/requests/api/moves_controller_update_spec.rb
willfish: We'll be replacing v1 moves with v2 moves soon so this will die.

spec/requests/api/moves_controller_update_spec.rb
willfish: I guess move is the resource under test so I felt it made sense to reference it. Happy to change if you feel strongly.

spec/requests/api/moves_controller_update_spec.rb
willfish: I think it's more robust, yeah

spec/services/moves/updater_spec.rb
willfish: All of this code is going away so I'm not keen to refactor it.

app/controllers/api/moves_controller.rb
willfish: Probably in v2 we can do that if it stays this complex.

app/services/moves/updater.rb
willfish: Alessandro suggested generalising pulling out relationships which is a great idea.

PR #617, opened 2020-06-20
Use YAML format for swagger_docs

alexdesi:
- Use YAML format for swagger_docs
- Fix wrong paths on V2

This still need to default to V1 the docs when V2 endpoint swagger doc is not present.


swagger/v2/get_people_responses.yaml
```diff
@@ -13,7 +13,7 @@
         anyOf:
           - $ref: "../v1/gender.yaml#/Gender"
           - $ref: "../v1/ethnicity.yaml#/Ethnicity"
-          - $ref: ../v2/profile.yaml#/Profile
+          - $ref: "../v1/profile.yaml#/Profile"
```
willfish, line 16: 🎉

spec/swagger_helper.rb
```diff
@@ -22,212 +22,13 @@ def load_swagger_yaml(*relative_path)
...
-          },
-          Profile: {
-            "$ref": '../v2/profile.yaml#/Profile',
-          },
-        },
-      },
-      paths: load_swagger_yaml('hand_coded_paths_v2.yaml'),
-    },
+    'v1/swagger.yaml' => swagger_doc_v1,
+    'v2/swagger.yaml' => swagger_doc_v2
```
willfish, line 31:
```suggestion
    'v2/swagger.yaml' => swagger_doc_v2,
```

PR #620, opened 2020-06-20
[P4-1666] Remove nationality

willfish:
### Jira link

P4-1666

### What?

Removes references to Nationaility in the code base

Nationality isn't used by the frontend or anywhere in the backend.

We are removing references to this since the only place it is related to
anything is on a profile which isn't using it.

At the moment it just pollutes/confuses the code base.

For safety I checked staging and production environments for reference
to Nationality in the database and there are none:

Production

```ruby
irb(main):001:0> Profile.count
=> 59643
irb(main):002:0> Profile.where(nationality_id: nil).count
=> 59643
```

Staging

```ruby
Profile.where(nationality_id: nil).count
=> 11454
Profile.count
=> 11454
```

### Why?

Keep our code clean
Avoid confusion


spec/requests/api/profiles_controller_update_spec.rb
```diff
@@ -157,8 +157,6 @@
         end
 
         it 'does not remove documents from the profile' do
-          expect(profile.documents).to match_array(before_documents)
```
willfish, line 160: Missed this in a previous PR.

PR #623, opened 2020-06-22
Add support for include `locations` when retrieving regions

smoothcontract:
### Jira link

P4-1766

### What?

- [x] Adds supported relationships for region serializer (only `locations`)
- [x] If `include` is specified then full location details are included through `GET /reference/regions`

### Why?

- The front end requires this to correctly deserialize location id's through the Devour JSONAPI library. Although only location id's are needed to subsequently build up location filters on calling `GET /allocations` to filter by region, Devour seems unable to return relationship id's unless the corresponding entities are included in the response.

- This does increase the size of the response payload significantly, and is a bit slower. Local testing shows response time changing from ~35ms with a 10KB payload to ~85ms with a 40KB payload. I've been able to introduce this without any N+1 queries - the ActiveRecord element is only 2.3ms.

- API behaviour is unchanged if no include parameter is specified (i.e. locations are listed as relationships within each region, but not included).

- Regions API is not yet used in production

### DIFF
--- a/app/controllers/api/reference/regions_controller.rb
+++ b/app/controllers/api/reference/regions_controller.rb
@@ -4,7 +4,13 @@ module Api
   module Reference
     class RegionsController < ApiController
       def index
-        render json: Region.all.includes(:locations)
+        render json: Region.all.includes(locations: :suppliers), include: included_relationships
+      end
+
+    private
+
+      def supported_relationships
+        RegionSerializer::SUPPORTED_RELATIONSHIPS
       end
     end
   end
--- a/app/serializers/region_serializer.rb
+++ b/app/serializers/region_serializer.rb
@@ -4,4 +4,6 @@ class RegionSerializer < ActiveModel::Serializer
   attributes :key, :name, :created_at, :updated_at
 
   has_many :locations
+
+  SUPPORTED_RELATIONSHIPS = %w[locations].freeze
 end
--- a/spec/serializers/region_serializer_spec.rb
+++ b/spec/serializers/region_serializer_spec.rb
@@ -7,46 +7,50 @@ RSpec.describe RegionSerializer do
 
   let(:region) { create(:region) }
   let(:result) do
-    JSON.parse(ActiveModelSerializers::Adapter.create(serializer).to_json).deep_symbolize_keys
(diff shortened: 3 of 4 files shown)


app/controllers/api/reference/regions_controller.rb
```diff
@@ -4,7 +4,13 @@ module Api
   module Reference
     class RegionsController < ApiController
       def index
-        render json: Region.all.includes(:locations)
+        render json: Region.all.includes(locations: :suppliers), include: included_relationships
```
smoothcontract, line 7: Annoyingly the `LocationSerializer` includes associated suppliers, so `suppliers` are eager fetched here to avoid N+1 query for every location.

willfish: Can't we use `includes` only when we have the `location.supplier`include in the query param? I don't think the location serializer includes suppliers by default. if you remove suppliers in this line:

https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/spec/serializers/location_serializer_spec.rb#L11

The included test fails because it was expecting a supplier.

app/serializers/region_serializer.rb
```diff
@@ -4,4 +4,6 @@ class RegionSerializer < ActiveModel::Serializer
   attributes :key, :name, :created_at, :updated_at
 
   has_many :locations
+
+  SUPPORTED_RELATIONSHIPS = %w[locations].freeze
```
willfish, line 8:
```suggestion
  SUPPORTED_RELATIONSHIPS = %w[locations locations.supplier].freeze
```

app/controllers/api/reference/regions_controller.rb
willfish: This does raise an interesting/potentially valuable optimisation question - we should match the includes in index actions for the database query to make sure we're not doing N+1 everywhere.

I think we can achieve this in a general way to reduce N+1 in the future.

smoothcontract: The `LocationSerializer` is somewhat strange and does cause lots of N+1 queries unless the suppliers are prefetched. The suppliers don't get included in the returned JSON payload, but are listed as relationships of the included locations - which due to the magic of AMS means running a DB query for each location :(

PR #624, opened 2020-06-22
Fix casing of the prison number when indexing by prison number

willfish:
### Jira link

P4-<TODO>

### What?

prison_number/offender_number only comes in upcase form so just upcase them when they come into the index action

### Why?

- To support any casing on a prison number

### DIFF
--- a/spec/requests/api/people_controller_index_spec.rb
+++ b/spec/requests/api/people_controller_index_spec.rb
@@ -55,13 +55,21 @@ RSpec.describe Api::PeopleController do
 
       before do
         allow(People::Finder).to receive(:new).and_return(people_finder)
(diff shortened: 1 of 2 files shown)


spec/requests/api/people_controller_index_spec.rb
```diff
@@ -55,14 +55,22 @@
 
       before do
         allow(People::Finder).to receive(:new).and_return(people_finder)
-        allow(Moves::ImportPeople).to receive(:new).with([prison_number])
+        allow(Moves::ImportPeople).to receive(:new).with([prison_number.upcase])
```
willfish, line 58: We implicitly test that a prison number is always received in it's upcased form.

PR #626, opened 2020-06-22
Finalise v2 swagger documentation

willfish:
### Jira link

P4-<TODO>

### What?

As part of the supplier release we need to provide them with v2 documentation for integrating with our api.

In this PR, Alessandro and I:

- [x] Add moves v2 to swagger docs
- [x] Add Accept: application/vnd.api+json; version=2 to swagger doc
- [x] Add profile document patching to swagger doc
- [x] Remove unused files for hand_coded_paths_{v1,v2}

### Why?

As part of the supplier assurance we need to provide them with access to documentation for the breaking changes in v2.


swagger/v1/content_type_parameter.yaml
```diff
@@ -5,5 +5,4 @@ ContentType:
   required: true
   schema:
     type: string
-    default: application/vnd.api+json
```
willfish, line 8: This is wrong, we never default to this mime type

willfish: Hence why I've removed it :)

PR #627, opened 2020-06-22
Default rswag ui to point at v2 docs first in list

willfish:
### Jira link

P4-<TODO>

### What?

We're moving away from v1 and don't want to encourage integration with v1 apis.

Let's default to v2 so that links to our swagger docs point to what we want people to integrate with by default.

### Why?

Reduce accidents, generally.


willfish:
### Before default:

Screenshot 2020-06-22 at 16 18 42

### After default:

Screenshot 2020-06-22 at 16 18 50

PR #631, opened 2020-06-23
Set service_urls_expire_in to 5 min

alexdesi:
### Jira link

https://dsdmoj.atlassian.net/browse/P4-1558

### What?

This is an attempt to make urls expire in n minutes.

### Why?

For security reasons we must not create permanent url.


config/initializers/active_storage.rb
```diff
@@ -0,0 +1 @@
+Rails.application.config.active_storage.service_urls_expire_in = 1.minutes
```
alexdesi, line 1:
```suggestion
Rails.application.config.active_storage.service_urls_expire_in = 15.minutes
```

PR #632, opened 2020-06-23
P4-1616 Support includes in all relevant controllers

cwrw:
### Jira link

[P4-1616](https://dsdmoj.atlassian.net/browse/P4-1616)

### What?
- Add swagger docs for all include endpoints
- Add Include params to court cases and timetable people endpoints

### Why?

To support includes everywhere we have associated resources in the serializer, add `include` param to allow the resources to be rendered with its associated resources under the `included` section.

In the people controller however, since we are supporting multiple endpoints, we had to disable the normal validation which happens at the API controller level, and add custom validations for the extra endpoints themselves.

### Have you? (optional)

### DIFF
--- a/spec/swagger/swagger_doc_v1.yaml
+++ b/spec/swagger/swagger_doc_v1.yaml
@@ -183,6 +183,24 @@
             enum:
               - asc
               - desc
+        - name: include
+          description: Returns a specific list of related resources to the allocation
+          in: query
+          style: form
+          explode: false
+          schema:
+            type: string
+            enum:
+              - from_location
+              - to_location
+              - moves.person
+              - moves.person.gender
+              - moves.person.ethnicity
+              - moves.profile
+              - moves.profile.person
+              - moves.profile.person.ethnicity
+              - moves.profile.person.gender
+          example: from_location
         - "$ref": "../v1/pagination_parameters.yaml#/Page"
         - "$ref": "../v1/pagination_parameters.yaml#/PerPage"
       responses:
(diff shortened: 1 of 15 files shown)


willfish:
Are we not going to be using reflections to understand and support possible includes in a recursive way?

spec/swagger/swagger_doc_v1.yaml
```diff
@@ -550,6 +636,31 @@
...
+              - prison_transfer_reason
+              - person
+              - profile
+              - from_location.suppliers
+              - to_location.suppliers
+              - person.gender
+              - person.ethnicity
+              - profile.person
+              - profile.person.gender
+              - profile.person.ethnicity
```
smoothcontract, line 662: These are repeated above - might be worth extracting out into something like `move_include_attribute.yaml` to DRY this.

spec/swagger/swagger_doc_v1.yaml
```diff
@@ -1322,6 +1459,19 @@
...
+          style: form
+          explode: false
+          schema:
+            type: string
+            enum:
+              - from_location
+              - to_location
+              - from_location.suppliers
+              - to_location.suppliers
+          example: from_location
```
smoothcontract, line 1474: Repeated above - would be nice to DRY this too.

cwrw:
@willfish I initially started going down that route as we discussed earlier, which required adding the relevant associations everywhere to all existing serializers to prepare for that, before having a generic reflection of a resource and it's associations used for validation. This started getting onerous especially performance-wise, where we were getting n+1 queries even though we were `include`ing the right associations in the parent queries. This is because of how the serializers themselves as well as the underlying further queries we are building work. I think this ballooned into a bigger task of cleaning up queries, rethinking our serializers as well as having to look into conditional logic around circular dependencies and also having extra logic around allowing associations to be surfaced vs not allowing them in some cases due to security issues. I think for now I'm happy to be very explicit in what we allow on each serializer until we fix the other issues separately, we can revisit this then. Unfortunately this means that if we want to add further associations there has to be some work on the BE but I think most resources have those now anyway.

PR #633, opened 2020-06-23
P4-1618 Remove default includes

cwrw:
### Jira link

[P4-1618](https://dsdmoj.atlassian.net/browse/P4-1618)
### What?

I have added/removed/altered:

- Removed default includes in people/allocation controllers into v2 version

### Why?

Since we have moved to using the include param, remove default includes from controllers everywhere in v2, but leave v1 the same

- Only change API in v2, v1 should remain exactly the same

### DIFF
--- a/app/controllers/api/allocations_controller.rb
+++ b/app/controllers/api/allocations_controller.rb
@@ -76,10 +76,6 @@ module Api
       end
     end
 
-    def included_relationships
-      IncludeParamHandler.new(params).call || AllocationSerializer::SUPPORTED_RELATIONSHIPS
-    end
-
     def supported_relationships
       AllocationSerializer::SUPPORTED_RELATIONSHIPS
     end
--- a/app/controllers/api/people_controller.rb
+++ b/app/controllers/api/people_controller.rb
@@ -92,11 +92,6 @@ module Api
       profile_validator.validate!
     end
 
-    def other_included_relationships
-      # Custom included relationships to avoid the people actions relationships
-      @other_included_relationships ||= IncludeParamHandler.new(params).call || OTHER_SUPPORTED_RELATIONSHIPS
-    end
-
     def validate_other_include_params
       # Custom validation for includes to avoid the default validation shared in API controller
       IncludeParamsValidator
--- a/app/controllers/api/v2/allocations_actions.rb
+++ b/app/controllers/api/v2/allocations_actions.rb
@@ -1,0 +1,7 @@
+module Api::V2
+  module AllocationsActions
+    def included_relationships
+      IncludeParamHandler.new(params).call
+    end
+  end
+end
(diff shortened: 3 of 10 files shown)


app/controllers/api/allocations_controller.rb
```diff
@@ -76,10 +76,6 @@ def send_move_notifications
       end
     end
 
-    def included_relationships
-      IncludeParamHandler.new(params).call || AllocationSerializer::SUPPORTED_RELATIONSHIPS
```
cwrw, line 80: Checking with FE team around making sure non of those break anything

app/controllers/api/people_controller.rb
```diff
@@ -92,15 +92,10 @@ def validate_nomis_profile
       profile_validator.validate!
     end
 
-    def other_included_relationships
-      # Custom included relationships to avoid the people actions relationships
-      @other_included_relationships ||= IncludeParamHandler.new(params).call || OTHER_SUPPORTED_RELATIONSHIPS
-    end
-
     def validate_other_include_params
```
smoothcontract, line 95: Is this method still necessary?

cwrw: Yea unfortunately we have to have this custom method  (now at-least after the change) as we can support both v1/v2

spec/requests/api/allocations_controller_create_v2_spec.rb
```diff
@@ -0,0 +1,335 @@
+# frozen_string_literal: true
```
cwrw, line 1: not sure about this, basically I copied v1 tests and fixed broken ones. I'm wondering what the strategy is regarding v1/v2 tests to avoid duplication cc @willfish ?

cwrw:
@smoothcontract I had to rejig this PR drastically to move those changes to V2 instead as they are breaking changes. Could you have another look please :)?

spec/requests/api/allocations_controller_create_v2_spec.rb
willfish: The strategy is to duplicate and remove everything v1 in 1 or 2 months time.

app/controllers/api/v2/allocations_actions.rb
```diff
@@ -0,0 +1,7 @@
+module Api::V2
+  module AllocationsActions
+    def included_relationships
+      IncludeParamHandler.new(params).call
+    end
+  end
+end
```
smoothcontract, line 7: I don't think you need this file - for V2 presumably the plan is to not return any included relationships by default, which is covered by the same method in `ApiController`. This module doesn't change that method implementation at all so looks redundant to me. For V1 the equivalent module is overriding the implementation from `ApiController` so that makes sense.

PR #636, opened 2020-06-24
Add support for URLs which expires after 5 min, for images and docs.

alexdesi:
### Jira link

https://dsdmoj.atlassian.net/browse/P4-1558

### What?

I have added/removed/altered:

- now, URL for the Image and document do not use ActiveStorage as a proxy, but direct S3 URL: https://cloud-platform-6ac35f1c447e16364485ad5fe3dfdc36.s3.eu-west-2.amazonaws.com

### Why?

I am doing this because:

ActiveStorage `url_for` provides permanent urls, but we need urls expiring in 5 minutes.


app/controllers/application_controller.rb
```diff
@@ -1,4 +1,9 @@
 # frozen_string_literal: true
 
 class ApplicationController < ActionController::API
+  before_action do
+    # This is to access S3 URLs with #service_url, which returns a url that expires in 5 Minutes
+    # In contrast, url_for returns permanent URL, which for privacy/security reasons, we do not want.
+    ActiveStorage::Current.host = request.base_url
+  end
```
smoothcontract, line 8: Could this be moved to `ApiController` instead, and perhaps made part of a dedicated method rather than including the code within a block pls?

spec/serializers/document_serializer_spec.rb
```diff
@@ -8,6 +8,11 @@
   let(:document) { create :document }
   let(:result) { ActiveModelSerializers::Adapter.create(serializer).serializable_hash }
 
+  before do
+    # This is needed because of use of #service_url, since these tests do not run the controllers.
+    ActiveStorage::Current.host = 'http://example.com'
+  end
```
smoothcontract, line 14: Would be preferable to move this to one of the spec helper files?

smoothcontract: ..or maybe `test.rb` 🤔

app/controllers/application_controller.rb
alexdesi: yes we can, `ApiController` seemed already growing a lot, but happy to move it there, np.

spec/serializers/document_serializer_spec.rb
alexdesi: maybe, but we need this only in this test.

alexdesi: added on test.rb

alexdesi: no, it does not work on test.rb, 

> ActiveStorage::Current.host makes use of CurrentAttributes, a type of attributes that are set/reset on every request and can be used everywhere in Rails

app/controllers/application_controller.rb
```diff
@@ -1,4 +1,9 @@
 # frozen_string_literal: true
 
 class ApplicationController < ActionController::API
+  before_action do
+    # This is to access S3 URLs with #service_url, which returns a url that expires in 5 Minutes
```
willfish, line 5: I'll be grateful for this comment one day

app/controllers/application_controller.rb
alexdesi: done

spec/requests/api/documents_controller_create_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+          },
+        },
+      }
+    end
+
+    let(:expected_data) do
+      {
+        type: 'documents',
+        attributes:
+           { url: a_string_starting_with('http://www.example.com/rails/active_storage/disk'), # "http://www.example.com/rails/active_storage/disk/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaDdDRG9JYTJWNVNTSWhlV0oxYXprNFkyUmtZV1UxTm1SNWFXWjZaelV3Wm01NU4zRXpiZ1k2QmtWVU9oQmthWE53YjNOcGRHbHZia2tpV1dGMGRHRmphRzFsYm5RN0lHWnBiR1Z1WVcxbFBTSm1hV3hsTFhOaGJYQnNaVjh4TURCclFpNWtiMk1pT3lCbWFXeGxibUZ0WlNvOVZWUkdMVGduSjJacGJHVXRjMkZ0Y0d4bFh6RXdNR3RDTG1Sdll3WTdCbFE2RVdOdmJuUmxiblJmZEhsd1pTSVhZWEJ3YkdsallYUnBiMjR2YlhOM2IzSmsiLCJleHAiOiIyMDIwLTA2LTI1VDEzOjE3OjIzWiIsInB1ciI6ImJsb2Jfa2V5In19--2b4628c3ac424585800cf86fbd36496820cbd54e/file-sample_100kB.doc?content_type=application%2Fmsword&disposition=attachment%3B+filename%3D%22file-sample_100kB.doc%22%3B+filename%2A%3DUTF-8%27%27file-sample_100kB.doc",
```
alexdesi, line 35:
```suggestion
           { url: a_string_starting_with('http://www.example.com/rails/active_storage/disk'),
```

spec/serializers/document_serializer_spec.rb
alexdesi, line 14:
```suggestion
  before { ActiveStorage::Current.host = 'http://example.com' }
```

app/controllers/api_controller.rb
```diff
@@ -226,4 +227,10 @@ def extend_versioned_controller_actions
       extend "#{version}::#{actions_module}".constantize
     end
   end
+
+  def set_active_storage_host
+    # This is required to access S3 URLs with #service_url, which returns a url that expires in 5 Minutes
+    # In contrast, url_for returns permanent URL, which for privacy/security reasons, we do not want to use.
+    ActiveStorage::Current.host = request.base_url
```
willfish, line 234: Can't we get this from the environment? I could spoof my host and have it change this url to whatever I want? I'm sure that's fine and won't ever be a problem but maybe it's easier to just set a static host, here?

willfish: We have an environment variable with the full publically-routable address but it is set under CORS_ALLOWED_ORIGINS. I think we may need to have PUBLIC_HOST as well as HOST environment variable set.

willfish:
So annoyingly, because of the e2e testing issues with uploading documents, I don't think we have a way of actually testing this properly locally. I think we need to merge it and then validate that the e2e tests still pass for documents on staging.

app/controllers/api_controller.rb
alexdesi: yes, we could, 
`request.base_url` does already the job,
is there any but what is the benefit of using an Env var?

app/controllers/api/people_controller.rb
```diff
@@ -23,7 +23,7 @@ def image
       success = People::RetrieveImage.call(person)
 
       if success
-        render json: Image.new(person.id, url_for(person.image))
+        render json: Image.new(person.id, person.image.service_url)
```
alexdesi, line 26:
```suggestion
        render json: Image.new(person.id, person.image.service_url)
```

spec/requests/api/documents_controller_create_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    {
+      'CONTENT_TYPE': content_type,
+      'Accept': 'application/vnd.api+json',
+      'Authorization' => "Bearer #{access_token}",
+    }
+  end
+
+  describe 'POST /people' do
+    let(:file) { fixture_file_upload('file-sample_100kB.doc', 'application/msword') }
+
```
alexdesi, line 20:
```suggestion
```

PR #343, opened 2020-03-25
P4-1127 Redis healthcheck

starswan:
Fixes P4-1127

## Proposed Changes

- Implement Redis healthcheck

PR #638, opened 2020-06-24
P4 1559 idempotency layer

martyn-w:
### Jira link

[P4-1559](https://dsdmoj.atlassian.net/browse/P4-1559)

### What?

I have added/removed/altered:

- [x] Added idempotency layer
- [x] Added mock controller tests for api controller functionality

### Why?

I am doing this because:

- to enforce idempotency
- as a framework to speed up request specs


app/controllers/api_controller.rb
```diff
@@ -35,6 +36,24 @@ def user_for_paper_trail
 
 private
 
+  def idempotent_action
```
smoothcontract, line 39: `ApiController` is already huge and filled with things - perhaps you would consider a composition pattern for this instead and move these methods plus the `before_action` and `around_action` calls in the relevant controller to a new `Idempotable` concern? That way all the idempotency related code is in one place.

app/services/idempotency/headers_validator.rb
```diff
@@ -0,0 +1,18 @@
...
+    include ActiveModel::Validations
+
+    attr_reader :idempotency_key
+
+    validates :idempotency_key, presence: true
+    # a relatively relaxed regex to allow for any uuid version
+    validates_format_of :idempotency_key, with: /\A[0-9A-F]{8}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{12}\Z/i
+
+    def initialize(headers)
+      # NB: we need to do a case-insensitive match for IDEMPOTENCY_KEY; rails does not do this automatically if the key contains an underscore
```
smoothcontract, line 14: Hmm.. not convinced this is a great idea. Headers are case sensitive by design and we don't do this for any of the other request headers. Would be better for suppliers to include the correct header 🤔

smoothcontract: ..and potentially raise an invalid request error if it's not present when we expect it.

app/services/idempotency/store.rb
```diff
@@ -0,0 +1,64 @@
...
+        @conflict_key = "conf|#{idempotency_key}"
+        @cache_response_key = "resp|#{idempotency_key}|#{request_hash(request)}"
+      end
+    end
+
+    def get
+      return if @idempotency_key.blank?
+
+      # Return the cached response if it matches the idempotency key and request
+      cached_response = redis.hgetall(cache_response_key)
```
smoothcontract, line 30: Does this really necessitate dropping down to native Redis? Rails has very good caching wrappers for this stuff and built in support for testing cache plus enabling/disabling caching in local development. If we can use those API's rather than direct Redis it would be nicer.

app/services/idempotency/headers_validator.rb
martyn-w: Hi @smoothcontract headers are supposed to be case insensitive; see [section 4.2 of RFC 2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2). So I think it is important that we support this.

app/controllers/api_controller.rb
martyn-w: Yep, that is a fair point, I'll have a go at moving this into a concern 👍

app/services/idempotency/store.rb
martyn-w: I did look at the rails caching to start off with, but for one reason or another (I forget 🤔 ) opted for native redis. I can revisit this though and get back to you.

app/services/idempotency/headers_validator.rb
willfish: I think rails follows that RFC so whatever we're doing elsewhere should be updated. Maybe separately, though

smoothcontract: Been testing this locally and Rails/Rack does a few strange things. First, any underscores in the header names are converted to dashes, and then using the `[]` method against `request.headers` is already case insensitive. So you can replace the line below with `@idempotency_key = headers['Idempotency_key']`.

smoothcontract: Found the docs for this: https://api.rubyonrails.org/classes/ActionDispatch/Http/Headers.html

app/services/idempotency/store.rb
```diff
@@ -0,0 +1,64 @@
...
+require 'redis'
+
+module Idempotency
+  class Store
+    # NB the caching vs conflict logic is:
+    # *  re-using the idempotency key with the same request for 10 mins will return the same cached response
+    # *  re-using the idempotency key with the same request after 10 mins to 12 hours will raise a conflict error
+    # *  re-using the idempotency key with a different request will raise a conflict error, for 12 hours
+
+    CACHE_RESPONSE_TTL = 600 # 10 minutes
```
willfish, line 12:
```suggestion
    CACHE_RESPONSE_TTL = 10.minutes.to_i
```

app/services/idempotency/store.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
```
cwrw, line 1: I think we might need to add general error handling when talking to Redis for instance when there are connection errors etc.

app/services/idempotency/store.rb
```diff
@@ -0,0 +1,64 @@
...
+      # NB: we need to do a case-insensitive match for IDEMPOTENCY_KEY; rails does not do this automatically if the key contains an underscore
+      @idempotency_key = request.headers.find { |key, _v| key =~ /\AIDEMPOTENCY[\_\-]KEY\Z/i }&.last
+      if @idempotency_key.present?
+        @conflict_key = "conf|#{idempotency_key}"
+        @cache_response_key = "resp|#{idempotency_key}|#{request_hash(request)}"
+      end
+    end
+
+    def get
+      return if @idempotency_key.blank?
```
cwrw, line 27: what are the scenarios we might not have an idempotancy key when we get here (as validation errors should stop this at the controller level), I'm wondering if we should raise an error instead of guarding against setting/getting cached responses?

app/services/idempotency/store.rb
```diff
@@ -0,0 +1,64 @@
...
+    def redis_url
+      @redis_url ||= ENV.fetch('REDIS_URL', nil)
+    end
+
+    def redis
+      @redis ||= Redis.new(url: redis_url)
+    end
+
+    def request_hash(request)
+      Digest::MD5.base64digest("#{request.method}|#{request.path}|#{request.raw_post}")
```
cwrw, line 61: 👍  for not storing sensitive data

app/services/idempotency/store.rb
```diff
@@ -0,0 +1,64 @@
...
+      # Otherwise, raise a conflict error of the idempotency key exists, or return nil if not
+      conflict = redis.get(conflict_key)
+      raise ConflictError, idempotency_key unless conflict.nil?
+    end
+
+    def set(response)
+      return if @idempotency_key.blank?
+
+      # cache the response for a short time
+      redis.hmset(cache_response_key, *response)
```
cwrw, line 43: Maybe we could leverage [pipelining](https://github.com/redis/redis-rb#pipelining) here?

app/services/idempotency/store.rb
cwrw: maybe we should guard against corrupt/missing requests and responses as well, but again not sure what the scenarios are for those

app/services/idempotency/store.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+
+  private
+
+    def redis_url
+      @redis_url ||= ENV.fetch('REDIS_URL', nil)
+    end
+
+    def redis
+      @redis ||= Redis.new(url: redis_url)
```
cwrw, line 57: maybe we can have a dedicated Redis client that lives outside of here and define a `redis.yml` file with different configs for different envs?

app/controllers/api/journeys_controller.rb
```diff
@@ -2,7 +2,9 @@
 
 module Api
   class JourneysController < ApiController
+    before_action :validate_required_idempotency_key
     before_action :validate_params, only: %i[create update]
+    around_action :idempotent_action
```
cwrw, line 7: I think we might need to only do this on create and update, as the other requests being returned cached might be a  problem and there isn't a risk in get/show actions

willfish:
Have you tried creating an idempotent request locally and validating the response outside of rspec fixtures, etc?

spec/services/idempotency/conflict_error_spec.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Idempotency::ConflictError do
+  it { expect(described_class).to be < StandardError }
```
willfish, line 6: Because this is returned to clients, how do you feel about adding a test for the message interpolating the key?

willfish, line 6:
```suggestion
  it { expect(described_class).to be_a(StandardError) }
```

app/services/idempotency/store.rb
martyn-w: So, I have thought further on this, and when I was thinking about the rails caching layer, I naively only considered a file-based approach: which wouldn't work as we have multiple pods. However a rails cache backed by redis (probably) will work, so I will see if I can implement that here.

app/controllers/api_controller.rb
martyn-w: now moved to a concern

app/services/idempotency/headers_validator.rb
martyn-w: now a case-insensitive header, `Idempotency-Key`

spec/services/idempotency/conflict_error_spec.rb
martyn-w: Hi @willfish unfortunately that doesn't work: the current test is checking that it inherits from StandardError, not that it is a StandardError. This is the output with the suggested change:
```
   Failure/Error: it { expect(described_class).to be_a(StandardError) }
       expected Idempotency::ConflictError to be a kind of StandardError
```

martyn-w: yep, sure 👍

app/controllers/api_controller.rb
```diff
@@ -22,6 +22,7 @@ class ApiController < ApplicationController
   rescue_from Faraday::ConnectionFailed, Faraday::TimeoutError, with: :render_connection_error
   rescue_from ActiveModel::ValidationError, with: :render_validation_error
   rescue_from IncludeParamsValidator::ValidationError, with: :render_include_validation_error
+  rescue_from Idempotency::ConflictError, with: :render_conflict_error
```
smoothcontract, line 25: You could move this line and the corresponding method below to the Idempotentable concern.

app/services/idempotency/store.rb
```diff
@@ -0,0 +1,61 @@
...
+    end
+
+    # def request_hash(request)
```
smoothcontract, line 57: Remove this method now?

app/services/idempotency/store.rb
smoothcontract: I suspect with Rails caching in the controller concern you likely won't need this class at all. And yes, we already have Redis setup and integrated in production for Sidekiq - maybe we need a few config lines in `production.rb` to enable it for caching too.

app/services/idempotency/store.rb
```diff
@@ -0,0 +1,45 @@
...
+      @idempotency_key = idempotency_key
+      @conflict_key = "conf|#{idempotency_key}"
+      @cache_response_key = "resp|#{idempotency_key}|#{request_hash}"
+    end
+
+    def read
+      return if @idempotency_key.blank?
+
+      # Return the cached response if it matches the idempotency key and request
+      cached_response = Rails.cache.read(cache_response_key)
```
smoothcontract, line 27: Yay 💯

Gemfile
```diff
@@ -25,6 +25,7 @@ gem 'doorkeeper'
 gem 'faraday'
 gem 'finite_machine'
 gem 'govuk_notify_rails', '~> 2.1.2'
+gem 'hiredis', '~> 0.6.3'
```
smoothcontract, line 28: Not come across this gem before.. reading through the docs I'm not immediately sure why it's needed? https://github.com/redis/hiredis-rb

app/controllers/concerns/idempotentable.rb
```diff
@@ -0,0 +1,46 @@
+module Idempotentable
+  extend ActiveSupport::Concern
+
+  included do
+    before_action :validate_idempotency_key
+    around_action :idempotent_action
+    rescue_from Idempotency::ConflictError, with: :render_conflict_error
```
smoothcontract, line 7: 👍

app/services/idempotency/store.rb
```diff
@@ -0,0 +1,45 @@
...
+      return cached_response if cached_response.present?
+
+      # Otherwise, raise a conflict error of the idempotency key exists, or return nil if not
```
smoothcontract, line 30: ...error `if` the...

Gemfile
martyn-w: It was suggested in the rails caching guide: go faster stripes for Redis - https://guides.rubyonrails.org/caching_with_rails.html#activesupport-cache-rediscachestore

spec/requests/api_controller_authentication_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    Rails.application.routes.draw do
+      get '/custom', to: 'mock#custom'
+    end
+  end
+
+  around do |example|
+    create_mock_controller
+    get '/custom', headers: headers
+    example.run
+    undefine_mock_controller # NB: it is necessary to undefine the mock controller
```
smoothcontract, line 28: Rather than doing this, why not add `MockController` class to the code base - including the idempotentable module? The routing to that can be restricted to the specs only with:

```ruby
  namespace :api do
    constraints(->(_req) { Rails.env.test? }) do
      get 'custom', to: 'mock#custom`
    end
  end
```

This lets us add further test endpoints to the mock controller if we need to in future. I think I prefer that to defining code on the the fly like this (and then having to undefine it after each spec). Probably worth getting other opinions on this though.

spec/requests/api_controller_authentication_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    it_behaves_like 'an endpoint that responds with error 401'
+  end
+
+  context 'without a valid access token' do
+    let(:access_token) { 'FOOBAR' }
+
+    it_behaves_like 'an endpoint that responds with error 401'
+  end
+end
```
smoothcontract, line 56: Really like the introduction of this - hopefully it means we can remove these repeated specs from the other request specs eventually. Nice one 👍

spec/requests/api_controller_idempotent_action_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+    class MockController < ApiController
+      include Idempotentable
+
+      def authentication_enabled?
+        false # NB: disable authentication to simplify tests (it is tested elsewhere)
+      end
+
+      def custom
+        render json: { hello: request.params[:data] }, status: :ok
+      end
```
smoothcontract, line 25: Following the approach described previously with `MockController` included in the code base but only routed to in test mode, these specs could potentially use a different endpoint without authentication.

spec/support/with_undefine_mock_controller.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+RSpec.shared_context 'with undefine mock controller' do
+  let(:undefine_mock_controller) do
+    # NB: it is important to undefine the mock controller after the test, otherwise other rspecs may fail
+    Object.send(:remove_const, :MockController) if Object.const_defined?(:MockController)
+    Rails.application.reload_routes!
+  end
+end
```
smoothcontract, line 9: See above comments - this feels a little funky, would be nice if it could be removed?

Gemfile
smoothcontract: Cool - fast is good 👍

app/services/idempotency/store.rb
martyn-w: Hi @cwrw I have tested the behaviour when Redis is unavailable: the Rails.cache object simply returns nil if it is unable to connect. So everything will continue to work, just we will not be able to detect idempotency conflicts. That sounds like a pragmatic behaviour?

```
[6] pry(main)> Rails.cache.read('foo')
=> "bar"
# (turn off redis)
[9] pry(main)> Rails.cache.read('foo')
=> nil
# (turn on redis)
[10] pry(main)> Rails.cache.read('foo')
=> "bar"
```

martyn-w: There is a separate piece of work to implement a Redis health check, see https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/343

app/controllers/api/journeys_controller.rb
martyn-w: interesting point, I had been steering the suppliers to provide an idempotency key for _all_ requests; however in the latest commit I no longer check the key for show or index on the journey events controller

martyn-w:
Hi @willfish
> Have you tried creating an idempotent request locally and validating the response outside of rspec fixtures, etc?

Yes, I have, and it works :-)

app/services/idempotency/store.rb
martyn-w: I'm no longer using lowlevel redis access - instead the rails.cache object will use the `hiredis` gem for high-performance redis

app/services/idempotency/store.rb
martyn-w: not relevant now having switched to rails.cache

app/services/idempotency/store.rb
martyn-w: fixed

spec/requests/api_controller_authentication_spec.rb
martyn-w: Hi @smoothcontract  I'm a bit wary about the suggested approach: it doesn't seem correct to add controllers to the codebase which are only there to facilitate testing: I think test code should belong with the specs, not with the application. I'll see if I can make the it work with a more concrete MockController defined outside of the app...

willfish: It does smell a bit affecting the production runtime for testing controllers. I don't have super strong opinions, though. As long as it is isolated and potentially namespaced, I'm pretty confident we won't run into any issues even with eager loading, etc, in production. I think on balance I prefer the existing approach that clearly says what's getting defined for the test. I don't think this discussion should block the PR from flying, either.

spec/requests/mock/authentication_controller_spec.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+module Mock
```
martyn-w, line 5: hi @smoothcontract I've refactored the mock controllers a bit so they are no longer being dynamic defined. Instead, each controller is defined at the top of its relevant rspec (nb: class names must be unique). Is this approach cleaner?

app/controllers/concerns/idempotentable.rb
```diff
@@ -0,0 +1,46 @@
+module Idempotentable
+  extend ActiveSupport::Concern
+
+  included do
+    before_action :validate_idempotency_key
```
cwrw, line 5: should those actions still live here if they are defined in the respective controllers now?

app/controllers/api/journeys_controller.rb
martyn-w: The code has been updated to only do the idempotency check for `:create` and `:update` in the `journeys_controller`

app/controllers/concerns/idempotentable.rb
martyn-w: Good spot - they definitely should not be there - now removed

spec/requests/mock/authentication_controller_spec.rb
smoothcontract: Yes! I like that much better, good stuff 👍

PR #639, opened 2020-06-25
Add support for searching within GET /allocations endpoint

smoothcontract:
### Jira link

P4-1327

### What?

- [x] Added new search[person] and search[location] params to GET /allocations endpoint
- [x] New params are propagated to revised Allocations::Finder service
- [x] Added scopes to Location and Person model to perform case insensitive substring match
- [x] Refactored request spec to avoid confusing `:skip_before` behaviour and remove redundant coverage

### Why?

- This supports adding a search feature in the front end. The original intent was to add a single search param covering both location titles and person name but this got very messy so I decided to separate the two searches for simplicity. This is an option we discussed during refinement and is likely to be more useful to the user.

- Tested locally with a copy of staging data and performance is good (searching across all allocations with no filters is around 400-500ms response time - the bulk of that being AMS rendering of included resources).

### Have you? (optional)

- Extends a production API with a non breaking change; original behaviour remains unchanged if search params are not passed in request

### DIFF
--- a/app/services/allocations/finder.rb
+++ b/app/services/allocations/finder.rb
@@ -2,13 +2,14 @@
 
 module Allocations
   class Finder
-    attr_reader :filter_params
+    attr_reader :filter_params, :search_params
 
-    def initialize(filter_params, order_params)
-      @filter_params = filter_params
-      @order_by = (order_params[:by] || 'date').to_sym
-      @order_direction = if order_params[:by]
-                           (order_params[:direction] || 'asc').to_sym
+    def initialize(filters: {}, ordering: {}, search: {})
+      @search_params = search
+      @filter_params = filters
+      @order_by = (ordering[:by] || 'date').to_sym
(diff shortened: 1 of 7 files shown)


app/services/allocations/finder.rb
```diff
@@ -35,8 +37,24 @@ def apply_ordering(scope)
...
+
+      scope.where(from_location_id: Location.search_by_title(search_params[:location])).or(
+        scope.where(to_location_id: Location.search_by_title(search_params[:location])),
+      )
+    end
+
+    def apply_person_search(scope)
+      return scope unless search_params.key?(:person)
+
+      scope.joins(moves: { profile: :person }).where('people.id' => Person.search_by_last_name(search_params[:person]))
```
cwrw, line 57: I think this might need to be an `includes` as the `joins` is ignoring the other scopes

PR #640, opened 2020-06-25
[P4-1786] Log court hearing params

willfish:
### Jira link

P4-1786

### What?

Adds more logging for debugging court hearing times

### Why?

More information about what the frontend is sending

### DIFF
--- a/app/controllers/api/court_hearings_controller.rb
+++ b/app/controllers/api/court_hearings_controller.rb
@@ -3,7 +3,9 @@ module Api
     def create
       court_hearing = CourtHearing.create!(court_hearings_attributes)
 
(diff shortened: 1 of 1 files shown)


app/controllers/api/court_hearings_controller.rb
```diff
@@ -3,6 +3,10 @@ class CourtHearingsController < ApiController
     def create
       court_hearing = CourtHearing.create!(court_hearings_attributes)
 
+      Rails.logger.info("Received court hearing #{request.body.read}")
+
+      request.body.rewind
+
```
alexdesi, line 9: as agreed, we should remove these once debug is done.

PR #648, opened 2020-06-26
P4 1765 allocation criteria

smoothcontract:
### Jira link

P4-1765

### What?

- [x] Added new `estate` and `sentence_length_other` attributes to allocation model
- [x] Extended values for existing `prisoner_category` and `sentence_length` attributes on allocation model
- [x] Cleaned up and refactored existing spec coverage for creating allocations
- [x] Improved spec performance when testing notifications
- [x] Added a rake task to back fill existing production allocation data

### Why?

- We want to support a new UI for specifying allocation criteria. Previously we only had `prisoner_category` of B/C/D - this now needs to include open and closed prison types. For `sentence_length` we now support `other` with a comment. Also we introduce `estate` for Adult/Juvenile/Youth Male/Female estate types. The new attributes are optional to avoid breaking the existing endpoint. Front end will need to be updated to pass these extra attributes when creating an allocation.

- The request spec for creating an allocation was slow and tested a few too many edge cases; these are better tested in the creator service spec which does all of the heavy lifting. Also removed use of the before/skip_before pattern - it's explicit now when the request needs to be fired for each spec.

- Improved specs for notifier; this previously fired a request to test each attribute individually. It's much faster to fire one request and then test 5 attributes at once. Applied this pattern across all affected specs. This did mean tweaking rubocop config slightly.

### Have you? (optional)

- Rake task `data_maintenance:fix_nil_allocations_estate` will need to be run on all environments once front end has been updated to support the new `estate` and `sentence_length_comment` attributes. Added this task in Jira so we don't forget: https://dsdmoj.atlassian.net/browse/P4-1848


lib/tasks/data_maintenance.rake
```diff
@@ -46,4 +46,9 @@ namespace :data_maintenance do
   task fix_move_agreed_for_non_prison_transfers: :environment do
     Move.where(move_agreed: false).where.not(move_type: 'prison_transfer').update_all(move_agreed: nil)
   end
+
+  desc 'fix nil estate for all existing production allocations data'
+  task fix_nil_allocations_estate: :environment do
+    Allocation.where(estate: nil).update_all(estate: :adult_male)
```
martyn-w, line 52: if allocation.estate is allowed to be nil, why are all existing nil values being updated to `:adult_male`?

spec/factories/allocation.rb
```diff
@@ -8,6 +8,7 @@
 
     prisoner_category { Allocation.prisoner_categories.values.sample }
     sentence_length { Allocation.sentence_lengths.values.sample }
+    estate { Allocation.estates.values.sample }
```
martyn-w, line 11: do you want to allow for a nil value too?

spec/requests/api/moves_controller_create_spec.rb
```diff
@@ -127,12 +127,14 @@
           end
         end
 
-        describe 'notification record' do
-          it { expect(notification.delivered_at).not_to be_nil }
-          it { expect(notification.topic).to eql(move) }
-          it { expect(notification.notification_type).to eql(notification_type_webhook) }
-          it { expect(notification.event_type).to eql('create_move') }
-          it { expect(notification.response_id).to be_nil }
+        it 'has correct attributes' do
```
martyn-w, line 130: 👍 good idea

spec/requests/api/allocations_controller_create_spec.rb
```diff
@@ -74,86 +77,48 @@
...
 
-      it 'sets the correct number of complex_cases' do
-        expect(response_json.dig('data', 'attributes', 'complex_cases').size).to match complex_cases_attributes.size
-      end
-
-      it 'sets the correct complex_cases attributes' do
-        expect(response_json.dig('data', 'attributes', 'complex_cases').first).to match complex_case1_attributes.stringify_keys
-      end
-
-      it 'sets the correct status' do
```
martyn-w, line 128: should this test be moved to the allocations creator spec?

spec/requests/api/allocations_controller_create_spec.rb
```diff
@@ -74,86 +77,48 @@
...
-        expect(response_json.dig('data', 'attributes', 'status')).to eq(Allocation::ALLOCATION_STATUS_UNFILLED)
-      end
-
-      context 'when omitting requested_by attribute' do
-        let(:allocation_attributes) { attributes_for(:allocation).except(:requested_by) }
```
martyn-w, line 133: move to allocations creator spec?

spec/requests/api/allocations_controller_create_spec.rb
```diff
@@ -74,86 +77,48 @@
...
-
-      context 'when specifying nil complex_cases attribute' do
-        let(:complex_cases_attributes) { nil }
-
-        it 'creates an allocation without complex cases' do
-          expect(allocation.complex_cases).to be_empty
-        end
-      end
-
-      context 'when omitting complex_cases attribute' do
```
martyn-w, line 148: move to allocations creator spec?

spec/requests/api/allocations_controller_create_spec.rb
```diff
@@ -255,74 +232,7 @@
...
-          end
-        end
-
-        it 'does not enqueue notifications' do
-          expect(subscription.notifications.count).to be_zero
-        end
-      end
-    end
-
-    context 'with incorrect moves_count type' do
```
martyn-w, line 296: an invalid moves_count would be good test for - either here or in a params validator

martyn-w: ignore ^ , I've found the test 👍

lib/tasks/data_maintenance.rake
smoothcontract: Good question. This API endpoint is already in use and the PR extends to add support for an extra `estate` attribute. We can't make it mandatory yet as that would break the existing front end. So the intent is to deploy the code changes, then make a front end change to populate the `estate` attribute when creating an allocation. Once that's done and deployed the rake task can be run to fix the existing data. This approach avoids a breaking change to the API.

 We may also need a follow up PR to make `estate` mandatory if the suppliers are using the same endpoint - if so that's a breaking change so will have to be done under v2.

spec/factories/allocation.rb
smoothcontract: The estate can be overridden if we want some different behaviour. These attributes all support nil values so I suppose there's no real need to set them in the factory 🤔

spec/requests/api/allocations_controller_create_spec.rb
smoothcontract: I deleted this case completely as the behaviour is the same as passing a nil value so is just duplication.

spec/requests/api/allocations_controller_create_spec.rb
smoothcontract: See below.

spec/requests/api/allocations_controller_create_spec.rb
smoothcontract: That's covered off by this spec:

```ruby
    it 'sets the correct attributes to an allocation' do
      call_creator
      expect(creator.allocation).to have_attributes(
        status: 'unfilled',
...
```

PR #649, opened 2020-06-28
[P4-1768] Moves report task

mafonso:
### Jira link

P4-1768

### What?

I have added/removed/altered:

- [ ] Saves generated JSON report to S3
- [ ] One report per supplier

### Why?

I am doing this because:
- It's the first step of the reporting tool, upload a basic the report to S3 to validate the workflow
-
-

### Have you? (optional)

- [x] Added environment variables to [deployment repository](https://github.com/ministryofjustice/hmpps-book-secure-move-api-deploy)

### DIFF
--- a/lib/tasks/journeys_report.rake
+++ b/lib/tasks/journeys_report.rake
@@ -1,65 +1,15 @@
 require 'csv'
 require 'json'
+require 'aws-sdk-s3'
+
 namespace :journeys do
-  desc 'Exports a CSV report of all journeys'
-  task csv_report: :environment do
-    puts 'exporting journeys_report.csv...'
-
-    CSV.open('journeys_report.csv', 'w') do |csv|
-      csv << ['SUPPLIER', 'DATE', 'MOVE REFERENCE', 'MOVE FROM', 'MOVE TO', 'MOVE EVENTS', 'TIMESTAMP', 'JOURNEY FROM (title)', 'JOURNEY FROM (nomis id)', 'JOURNEY TO (title)', 'JOURNEY TO (nomis id)', 'JOURNEY EVENTS', 'VEHICLE', 'BILLABLE']
-
-      current_reference = nil
-      Journey
-          .joins(:move).includes(:move)
-          .joins(:supplier).includes(:supplier)
-          .joins(:from_location).includes(:from_location)
-          .joins(:to_location).includes(:to_location)
-          .order('suppliers.key ASC, moves.date ASC, moves.reference ASC, journeys.client_timestamp ASC')
-          .each do |journey| # NB: cannot do find_each because we need to order
-        csv << if current_reference != journey.move.reference
-                 # new move
-                 [journey.supplier.key,
-                  journey.move.date,
-                  journey.move.reference,
-                  journey.move.from_location.title,
-                  journey.move.to_location&.title,
-                  journey.move.move_events.map { |event| "#{event.event_name}: #{event.notes}" }.join('; '),
-                  journey.client_timestamp.strftime('%d/%m %H:%M:%S'),
-                  journey.from_location.title,
-                  journey.from_location.nomis_agency_id,
-                  journey.to_location.title,
-                  journey.to_location.nomis_agency_id,
-                  journey.events.map { |event| "#{event.event_name}: #{event.notes}" }.join('; '),
-                  journey.vehicle['registration'],
-                  journey.billable]
-               else
-                 # journey within current move
-                 [nil,
-                  nil,
-                  nil,
-                  nil,
-                  nil,
-                  nil,
-                  journey.client_timestamp.strftime('%d/%m %H:%M:%S'),
-                  journey.from_location.title,
-                  journey.from_location.nomis_agency_id,
-                  journey.to_location.title,
-                  journey.to_location.nomis_agency_id,
-                  journey.events.map { |event| "#{event.event_name}: #{event.notes}" }.join('; '),
-                  journey.vehicle['registration'],
-                  journey.billable]
-               end
-        current_reference = journey.move.reference
-      end
-    end
-  end
-
   desc 'Exports a JSON report of all journeys'
   task json_report: :environment do
-    puts 'exporting journeys_report.json...'
+    timestamp = DateTime.now.utc.to_i
 
-    suppliers = []
     Supplier.order(:key).each do |supplier|
+      filename = "#{supplier.key}-#{timestamp}.json"
+      puts "Generating #{filename}..."
       moves = []
       Move
           .where("ID in (select move_id from journeys WHERE supplier_id = '#{supplier.id}')")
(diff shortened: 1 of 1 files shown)


lib/tasks/journeys_report.rake
```diff
@@ -1,65 +1,17 @@
...
-  end
+require 'aws-sdk-s3'
 
+namespace :journeys do
   desc 'Exports a JSON report of all journeys'
   task json_report: :environment do
-    puts 'exporting journeys_report.json...'
-
+    timestamp = DateTime.now.utc.strftime('%s')
     suppliers = []
```
martyn-w, line 9: `suppliers` probably isn't needed anymore as we are exporting one file per supplier: could you remove this variable and updates lines 63-69?

lib/tasks/journeys_report.rake
```diff
@@ -97,24 +49,36 @@ namespace :journeys do
...
-    File.open('journeys_report.json', 'w') do |file|
-      file.write(JSON.pretty_generate({ data: suppliers }))
+
+      File.open(filename, 'w') do |file|
+        file.write(JSON.pretty_generate({ data: suppliers })) # maybe change to .to_json for production
+      end
+
+      puts "Exporting #{filename} to S3..."
+      if ENV['S3_REPORTING_BUCKET_NAME']
+        bucket_name = ENV.fetch('S3_REPORTING_BUCKET_NAME')
```
martyn-w, line 74: no need to check the ENV twice, how about
```
bucket_name = ENV.fetch('S3_REPORTING_BUCKET_NAME')
if bucket_name.present?
   ...
```

lib/tasks/journeys_report.rake
mafonso: I was thinking the same, but then I realised that the file can eventually be just imported without any context and the data inside it needs to have some way of telling which supplier the data belongs to without looking at the filename.

martyn-w: ah yes, that is a good point. However, the way it is currently written the second supplier will include the data of the first supplier, so at the very least the `suppliers` variable needs to be re-initialised on every loop, and ideally renamed to `supplier`?

lib/tasks/journeys_report.rake
mafonso: Thanks! Will address that.

lib/tasks/journeys_report.rake
mafonso: Oh yes, that's a good point. Will fix that.

lib/tasks/journeys_report.rake
```diff
@@ -1,65 +1,17 @@
...
-    end
-  end
+require 'aws-sdk-s3'
 
+namespace :journeys do
   desc 'Exports a JSON report of all journeys'
   task json_report: :environment do
-    puts 'exporting journeys_report.json...'
-
+    timestamp = DateTime.now.utc.strftime('%s')
```
willfish, line 8:
```suggestion
    timestamp = Time.zone.now.utc.to_i
```

lib/tasks/journeys_report.rake
```diff
@@ -97,24 +49,35 @@ namespace :journeys do
...
       suppliers << {
         supplier: supplier.key,
         moves: moves,
       }
-    end
-    File.open('journeys_report.json', 'w') do |file|
-      file.write(JSON.pretty_generate({ data: suppliers }))
+
+      File.open(filename, 'w') do |file|
+        file.write(JSON.pretty_generate({ data: suppliers })) # maybe change to .to_json for production
```
willfish, line 69:
```ruby
data = {
  data: {
    supplier: supplier.key,
    moves: moves,
  }
}

json_data = JSON.pretty_generate(data)

File.write(filename, json_data)
```

willfish: Feels a bit cleaner with better labels for things in lexical scope

lib/tasks/journeys_report.rake
```diff
@@ -97,24 +49,35 @@ namespace :journeys do
...
+      end
+
+      puts "Exporting #{filename} to S3..."
+      bucket_name = ENV.fetch('S3_REPORTING_BUCKET_NAME')
+      if bucket_name.present?
+        s3 = Aws::S3::Resource.new
+        obj = s3.bucket(bucket_name).object(filename)
+        obj.upload_file(filename)
```
willfish, line 72: Can't we just upload what's in memory, here? Rather than reading the filename again?

lib/tasks/journeys_report.rake
mafonso: 👍

lib/tasks/journeys_report.rake
mafonso: I moved the supplier key to the "header" of the moves object.

```
  "moves": [
    {
      "supplier": "geoamey",
      "reference": "AHC5219N",
      "date": "2020-06-13",
      "from": "Medway STC",
      "to": "Colchester Crown Court",
      "events": [

      ],
      "journeys": [
        [....]
```
as I was thinking about the case where multiple files are just merged together

lib/tasks/journeys_report.rake
```diff
@@ -82,39 +32,47 @@ namespace :journeys do
...
+      File.open(filename, 'w') do |file|
+        file.write(JSON.pretty_generate({ moves: moves })) # maybe change to .to_json for production
+      end
+
+      puts "Exporting #{filename} to S3..."
+      bucket_name = ENV['S3_REPORTING_BUCKET_NAME']
+      if bucket_name.present?
+        s3 = Aws::S3::Resource.new
+        obj = s3.bucket(bucket_name).object(filename)
+        obj.upload_file(filename)
```
willfish, line 72:
```suggestion
        obj.put(body: StringIO.new(json_data))
```

lib/tasks/journeys_report.rake
mafonso: Me and @martyn-w discussed that and felt that for now is preferable to have it this way as we can just get the local file without uploading it (If you don't define a target bucket, it will just skip that step)

lib/tasks/journeys_report.rake
mafonso: Yes, that's a good suggestion.  
I was doing something similar with `obj.put` initially (although it was from the file, which kinda defeated the purpose).

This is an early stage and the file format will likely suffer many mutations, so it is convenient to have the intermediate file for now in pretty print format.

But we surely move to this approach when it's closer to it's final form.

lib/tasks/journeys_report.rake
mafonso: Moved the `supplier` key inside the moves, so that each move object is tagged with the supplier it belongs to.

willfish:
It seems to me like we should have tests for this behaviour. Especially if suppliers decide to integrate with our reports in s3.

lib/tasks/journeys_report.rake
martyn-w: yes, I agree the intermediate file format is clearer for now. Keep in mind this is still a proof of concept, where it is important to be able to debug and inspect at any stage, and not a finished system

lib/tasks/journeys_report.rake
martyn-w: @willfish @mafonso `Time.zone.now.utc.to_i` will return the minutes since the unix epoc, not the seconds. I think seconds is probably a safer choice.

martyn-w: Oh no, my bad, it _is_ seconds - I got my indentation wrong :-(

mafonso:
> It seems to me like we should have tests for this behaviour. Especially if suppliers decide to integrate with our reports in s3.

The use case is solely to be imported by the HUB, and the format is not closed yet.
The end-goal of this first ticket is just to generate a report and upload to S3, to validate the flow.

This is also the reason we went with a rake task extract as opposed to a real-time and more decoupled data stream.

willfish:
@mafonso Okay. Was reviewing as though this was something we'd be driving on a productionised basis :)

martyn-w:
@willfish i think its a bit premature for tests as the format is still in flux, and we don't have tests for other rake tasks. Suppliers will not be integrating with these reports as far as I know

PR #653, opened 2020-06-29
P4-1749 Import Frameworks

cwrw:
### Jira link

[P4-1749](https://dsdmoj.atlassian.net/browse/P4-1749)
### What?

I have added/removed/altered:

- [x] Added a new rake task and service to import frameworks
- [x] Added frameworks and framework question models to support the new questions
- [x] Added fixture files to support framework structures

### Why?

To populate PER questions from our [framework](https://github.com/ministryofjustice/hmpps-book-secure-move-frameworks/tree/master/frameworks/person-escort-record) add a rake task to populate the framework questions as well as version them. The task allows  a filepath to the frameworks (this will be changed later to read directly from the github repository and tags) as well as the version to set to those frameworks. For each framework a set of questions are built with the relevant sections, keys, options as well as dependent questions.


app/services/frameworks/question.rb
```diff
@@ -0,0 +1,66 @@
...
+        questions[followup] = questions[followup] || FrameworkQuestion.new(key: followup)
+        questions[followup].section = question.section
+        questions[followup].dependent_value = value
+        questions[followup].parent = question
+      end
+    end
+
+    def question
+      @question ||= begin
+        questions[key] = questions[key] || FrameworkQuestion.new(key: key)
```
cwrw, line 62: we might not have built the question yet, so define the question if it doesn't exist or extract it if it does

app/services/frameworks/question.rb
```diff
@@ -0,0 +1,66 @@
...
+    def build_followup_comments(followup_comment:, value:)
+      if followup_comment
+        question.followup_comment = true
+        question.followup_comment_options << value if required?(followup_comment.fetch('validations', []))
+      end
+    end
+
+    def build_followup_question(followups:, value:)
+      followups.each do |followup|
+        questions[followup] = questions[followup] || FrameworkQuestion.new(key: followup)
```
cwrw, line 53: we might not have built the question yet, so define the question if it doesn't exist or extract it if it does

app/models/framework_question.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class FrameworkQuestion < VersionedModel
```
cwrw, line 3: I went with framework questions as we already use assessment questions

app/models/framework_question.rb
```diff
@@ -0,0 +1,13 @@
...
+  validates :key, presence: true
+  validates :section, presence: true
+  validates :question_type, presence: true
```
smoothcontract, line 6: Should this be an enum?

db/migrate/20200624052728_create_frameworks.rb
```diff
@@ -0,0 +1,10 @@
+class CreateFrameworks < ActiveRecord::Migration[6.0]
+  def change
+    create_table :frameworks, id: :uuid do |t|
+      t.string :name, null: false
```
martyn-w, line 4: maybe name should be unique?

lib/tasks/populate_framework_data.rake
```diff
@@ -0,0 +1,11 @@
+namespace :frameworks do
+  desc 'Populate versioned questions from frameworks repository'
+  task :populate_data, %i[filepath version] => :environment do |_, args|
+    unless (filepath = args[:filepath]).present? && (version = args[:version]).present?
+      puts('No filepath and version provided')
```
martyn-w, line 5: how about: `abort('No filepath and version provided')`

spec/services/frameworks/importer_spec.rb
```diff
@@ -0,0 +1,83 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Frameworks::Importer do
+  describe '#call' do
+    let(:filepath) { Rails.root.join('spec/fixtures/files/frameworks/') }
+
+    context 'when creating frameworks' do
+      it 'persists multilpe frameworks' do
```
smoothcontract, line 10: multiple

db/schema.rb
```diff
@@ -121,6 +121,30 @@
...
+    t.uuid "parent_id"
+    t.datetime "created_at", precision: 6, null: false
+    t.datetime "updated_at", precision: 6, null: false
+    t.index ["framework_id"], name: "index_framework_questions_on_framework_id"
+    t.index ["parent_id"], name: "index_framework_questions_on_parent_id"
+  end
+
+  create_table "frameworks", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.string "name", null: false
+    t.decimal "version", null: false
```
smoothcontract, line 143: Decimal is a bit of an odd choice for a version number - this might get some freaky rounding issues, and also precludes the use of semver. Would you consider a string type instead?

db/migrate/20200624052728_create_frameworks.rb
cwrw: Names will be unique per *version*, I can add a check for those 👍

spec/services/frameworks/question_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+
+require 'rails_helper'
+
+RSpec.describe Frameworks::Question do
+  describe '#call' do
+    let(:fixture_path) { 'spec/fixtures/files/frameworks/person-escort-record-1/questions' }
+
+    it 'sets a question as required if required validation available' do
+      filepath = Rails.root.join(
+        "#{fixture_path}/medical-details-information.yml",
```
smoothcontract, line 11: Prefer to split these out into two arguments i.e.

```ruby
filepath = Rails.root.join(fixture_path, 'medical-details-information.yml')
```

spec/services/frameworks/question_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+      )
+      question = FrameworkQuestion.new(section: 'health', key: 'medical-details-information')
+      described_class.new(filepath: filepath, questions: { 'medical-details-information' => question }).call
+
+      expect(question.options).to be_empty
+    end
+
+    it 'allows folllowup comments to be set on a question' do
```
smoothcontract, line 49: follow

spec/services/frameworks/question_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+      filepath = Rails.root.join(
+        "#{fixture_path}/medical-professional-referral.yml",
+      )
+      question = FrameworkQuestion.new(section: 'health', key: 'medical-professional-referral')
+      described_class.new(filepath: filepath, questions: { 'medical-professional-referral' => question }).call
+
+      expect(question.followup_comment).to be_truthy
+    end
+
+    it 'allows folllowup comments to be required on an option' do
```
smoothcontract, line 59: follow

spec/services/frameworks/section_spec.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Frameworks::Section do
+  describe '#call' do
+    it 'returns a list of question keys mapped to questions' do
+      filepath = Rails.root.join(
+        'spec/fixtures/files/frameworks/person-escort-record-1/manifests/health-information.yml',
```
smoothcontract, line 9: I think there's a path helper for fixture files... can't recall the exact name off the top of my head though.

db/schema.rb
martyn-w: yep, I agree - string is preferable to decimal

app/services/frameworks/question.rb
```diff
@@ -0,0 +1,66 @@
...
+      validation_types = validations.flat_map { |validation| validation['type'] }
+      validation_types.any?('required')
+    end
+
+    def build_options(options)
+      options.each do |option|
+        question.options << option['value']
+
+        build_followup_questions(
+          followups: option['followup'].presence || [],
```
cwrw, line 33: change to fetch

app/models/framework_question.rb
cwrw: Indeed, Im just holding off until we have a full list of values before I define them which will come in following prs

db/migrate/20200624052728_create_frameworks.rb
cwrw: added an index/validation scoped to version

spec/services/frameworks/question_spec.rb
cwrw: not sure I understand? is this referring to the `followup` in the block?

spec/services/frameworks/section_spec.rb
cwrw: It's `fixture_path` but wouldn't help much here as its defined to `../spec/fixtures/`  https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/3d06c27f5cb6be777c8f96c511612161ed201e04/spec/rails_helper.rb#L55
That's why I went with this way instead. I tidied it up a bit more though so let me know your thoughts

spec/services/frameworks/question_spec.rb
smoothcontract: No, just a spelling mistake - three l's!

PR #654, opened 2020-06-29
Fix uniqueness for Move model

alexdesi:
### Jira link

https://dsdmoj.atlassian.net/browse/P4-1819

### What?

I have altered the Move validation.

### Why?

To make a move unique under the person not under the profile, in Api V2.

### Note: 
the API V1 and V2 require slightly different validations for the Move, 
to make it work I had to use this "hack":

```
def person_id
  attributes['person_id'] || person&.id
end
```

That is not a good piece of code!
In case we do NOT need FE to create a Move with a person and without a profile, we can remove it.

### DIFF
--- a/app/services/court_hearings/create_in_nomis.rb
+++ b/app/services/court_hearings/create_in_nomis.rb
@@ -1,7 +1,7 @@
 module CourtHearings
   class CreateInNomis
     def self.call(move, court_hearings)
-      booking_id = move.profile.person.latest_nomis_booking_id
+      booking_id = move.person.latest_nomis_booking_id
 
       body_locations = {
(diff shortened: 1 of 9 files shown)


willfish:
We validated in production/staging that we don't have any person_ids

app/services/court_hearings/create_in_nomis.rb
```diff
@@ -1,7 +1,7 @@
 module CourtHearings
   class CreateInNomis
     def self.call(move, court_hearings)
-      booking_id = move.profile.person.latest_nomis_booking_id
+      booking_id = move.person.latest_nomis_booking_id
```
willfish, line 4:
```suggestion
      booking_id = move.profile.person.latest_nomis_booking_id
```

PR #655, opened 2020-06-30
P4-1750 Add flags to frameworks

cwrw:
### Jira link

[P4-1750](https://dsdmoj.atlassian.net/browse/P4-1750)

### What?

I have added/removed/altered:

- [x] Added flags table and associated it to framework questions
- [x] Amended framework importer to persist flags if they appear under a question option
- [x] Updated framework question enums and added relevant validations

### Why?

I am doing this because:
To show users on the summary pages on the Frontend application certain alerts a person transported might have, flags are linked to questions if they are answered in a certain way. A question can have multiple flags, with a `type`, `name` and the question value they relate to.

### DIFF
--- a/app/models/framework_question.rb
+++ b/app/models/framework_question.rb
@@ -1,9 +1,16 @@
 # frozen_string_literal: true
 
 class FrameworkQuestion < VersionedModel
+  enum question_type: {
+    radio: 'radio',
+    checkbox: 'checkbox',
+    text: 'text',
+    textarea: 'textarea',
+  }
+
   validates :key, presence: true
   validates :section, presence: true
-  validates :question_type, presence: true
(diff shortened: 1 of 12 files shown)


app/models/framework_question.rb
```diff
@@ -1,13 +1,21 @@
 # frozen_string_literal: true
 
 class FrameworkQuestion < VersionedModel
+  enum question_type: {
+    radio: 'radio',
+    checkbox: 'checkbox',
+    text: 'text',
+    textarea: 'textarea',
```
martyn-w, line 8: dropdown?

cwrw: no instances of those yet :D

PR #656, opened 2020-06-30
[P4-1823] Handle profiles with no assessment answers in importer

willfish:
### Jira link

P4-1823

### What?

We noticed a bug during importing of people and profiles in v1 in production. Profiles (and therefore people) weren't being saved because the `changed?` method returns false for new profiles without any assessment_answers.

This PR fixes this issue by checking to see whether person attributes have changed in addition to profile ones and does the saving of Person/Profile separately.

### Why?

New Nomis prison numbers aren’t being created in production. This means that we can’t find them when calling GET /api/people?filter[prison_number]=:prison_number. This feature is needed.

### DIFF
--- a/app/services/moves/import_people.rb
+++ b/app/services/moves/import_people.rb
@@ -32,11 +32,13 @@ module Moves
 
         PersonalCareNeeds::Importer.new(profile: profile, personal_care_needs: personal_care_needs.fetch(offender_no, [])).call
 
-        if profile.changed?
-          changed_profile_count += 1
-          profile.save!
-        end
+        next unless profile.changed? || profile.person.changed?
+
+        changed_profile_count += 1
+        profile.person.save!
+        profile.save!
       end
+
       if new_person_count.positive? || changed_profile_count.positive?
         Rails.logger.info("[Moves::Importer] people new[#{new_person_count}] updated[#{changed_profile_count}]")
       end
--- a/spec/services/moves/import_people_spec.rb
+++ b/spec/services/moves/import_people_spec.rb
@@ -6,30 +6,25 @@ RSpec.describe Moves::ImportPeople do
   subject(:importer) { described_class.new(input_data) }
 
   let(:input_data) do
-    [prisoner_one.nomis_prison_number]
+    [nomis_prison_number_one]
   end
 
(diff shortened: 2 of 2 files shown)


spec/services/moves/import_people_spec.rb
```diff
@@ -50,4 +48,43 @@
...
+
+    it 'populates assessment answers' do
+      importer.call
+
+      assessment_answers = Person.last.profiles.first.assessment_answers
+
+      expect(assessment_answers.count).to eq(2)
+    end
+
+    context 'with no alerts or personal care needs' do
```
willfish, line 69: This verifies the original issue is now fixed

app/services/moves/import_people.rb
```diff
@@ -32,11 +32,13 @@ def import_people
...
         PersonalCareNeeds::Importer.new(profile: profile, personal_care_needs: personal_care_needs.fetch(offender_no, [])).call
 
-        if profile.changed?
-          changed_profile_count += 1
-          profile.save!
-        end
+        next unless profile.changed? || profile.person.changed?
+
+        changed_profile_count += 1
+        profile.person.save!
```
willfish, line 38: We save the person first to make sure there is a person_id for the profile to be valid

spec/services/moves/import_people_spec.rb
```diff
@@ -42,6 +37,9 @@
   end
 
   context 'with one existing record' do
+    let!(:prisoner_one) { create(:person, nomis_prison_number: nomis_prison_number_one, prison_number: nomis_prison_number_one) }
```
martyn-w, line 40: minor (non-blocking) point, `before` might be clearer than `let!`

```
before do
  create(:person, nomis_prison_number: nomis_prison_number_one, prison_number: nomis_prison_number_one)
  create(:person, nomis_prison_number: nomis_prison_number_two, prison_number: nomis_prison_number_two)
end
```

spec/services/moves/import_people_spec.rb
```diff
@@ -42,6 +37,9 @@
   end
 
   context 'with one existing record' do
```
martyn-w, line 39: two existing records

PR #658, opened 2020-06-30
P4 1827 faster specs

smoothcontract:
### Jira link

P4-1827

### What?

- [x] Sped up all the things
- [x] Use fewer build executors on Circle CI - shouldn't be needed now and will reduce build queues
- [x] Bypass authentication for most of the request specs, unless it's interesting to test something related (e.g. journeys)
- [x] Use a faster hashing mechanism for Doorkeeper secrets when running in test mode
- [x] Remove duplicated specs for authorization and content type headers that add no value in the various endpoints, and test these once
- [x] Remove various deprecation warnings when running test suite - these were hidden in the noisy test output so I've swapped that to progress (dots) indicator so we don't miss those in future. That also means less frantic scrolling to see what failed on Circle CI.
- [x] Use unlogged tables for the test database (this requires running `RAILS_ENV=test rails db:drop db:prepare` locally to pick up the change). This improves performance of Postgres significantly and your SSD will be pleased too.
- [x] Refactored a few specs that were passing auth details via request parameters rather than headers. I think it's nice to keep the approach consistent across the code base.

### Why?

- The test suite was irritatingly slow, and meant it was near impossible to perform TDD. When testing locally this drops the test suite time from almost 6 minutes to slightly over 1 minute. I've used a variety of techniques to achieve this with varying levels of controversy but the overall effect is pretty good.

### DIFF
--- a/app/controllers/api_controller.rb
+++ b/app/controllers/api_controller.rb
@@ -26,12 +26,12 @@ class ApiController < ApplicationController
   rescue_from IncludeParamsValidator::ValidationError, with: :render_include_validation_error
 
   def current_user
+    return Doorkeeper::Application.new unless authentication_enabled?
+
     doorkeeper_token&.application
   end
 
   def user_for_paper_trail
-    return unless authentication_enabled?
-
     current_user.owner_id
   end
 
@@ -41,6 +41,8 @@ private
     return false if Rails.env.development? && ENV['DEV_DISABLE_AUTH'] =~ /true/i
 
     return false if Rails.env.production? && ENV['HEROKU_DISABLE_AUTH'] =~ /true/i
+
+    return false if Rails.env.test? && request.headers['Authorization'] =~ /spoofed/i
 
     true
   end
--- a/app/jobs/prepare_move_notifications_job.rb
+++ b/app/jobs/prepare_move_notifications_job.rb
@@ -31,12 +31,11 @@ class PrepareMoveNotificationsJob < ApplicationJob
 private
 
   def build_notification_id(subscription, type_id, topic, action_name)
-    { notification_id:
-          subscription.notifications.create!(
-            notification_type_id: type_id,
-            topic: topic,
-            event_type: event_type(action_name),
-          ).id }
+    subscription.notifications.create!(
+      notification_type_id: type_id,
+      topic: topic,
+      event_type: event_type(action_name),
+    ).id
   end
 
   def should_email?(move)
--- a/spec/requests/api/allocation_events_controller_cancel_spec.rb
+++ b/spec/requests/api/allocation_events_controller_cancel_spec.rb
@@ -8,9 +8,7 @@ RSpec.describe Api::AllocationEventsController do
   describe 'POST /allocations/:id/cancel' do
     let(:schema) { load_yaml_schema('post_allocation_cancel_responses.yaml') }
 
-    let(:supplier) { create(:supplier) }
-    let(:application) { create(:application, owner_id: supplier.id) }
-    let(:access_token) { create(:access_token, application: application).token }
+    let(:access_token) { 'spoofed-token' }
     let(:headers) { { 'CONTENT_TYPE': content_type, 'Authorization': "Bearer #{access_token}" } }
     let(:content_type) { ApiController::CONTENT_TYPE }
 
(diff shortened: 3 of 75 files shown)


app/jobs/prepare_move_notifications_job.rb
```diff
@@ -31,12 +31,11 @@ def perform(topic_id:, action_name:)
...
-          subscription.notifications.create!(
-            notification_type_id: type_id,
-            topic: topic,
-            event_type: event_type(action_name),
-          ).id }
+    subscription.notifications.create!(
+      notification_type_id: type_id,
+      topic: topic,
+      event_type: event_type(action_name),
+    ).id
```
martyn-w, line 38: should this signature change: previously it was returning a hash, now it is return an id ?

app/controllers/api_controller.rb
```diff
@@ -42,6 +42,8 @@ def authentication_enabled?
 
     return false if Rails.env.production? && ENV['HEROKU_DISABLE_AUTH'] =~ /true/i
 
+    return false if Rails.env.test? && request.headers['Authorization'] =~ /spoofed/i
```
martyn-w, line 45: 👍 good idea

spec/requests/api/allocation_events_controller_cancel_spec.rb
```diff
@@ -8,9 +8,7 @@
   describe 'POST /allocations/:id/cancel' do
     let(:schema) { load_yaml_schema('post_allocation_cancel_responses.yaml') }
 
-    let(:supplier) { create(:supplier) }
-    let(:application) { create(:application, owner_id: supplier.id) }
-    let(:access_token) { create(:access_token, application: application).token }
+    let(:access_token) { 'spoofed-token' }
```
martyn-w, line 11: I think we may use this snippet everywhere - could you move it to a context so it can be included (`include 'spoof token'`)
```
let(:access_token) { 'spoofed-token' }
 let(:headers) { { 'CONTENT_TYPE': content_type, 'Authorization': "Bearer #{access_token}" } }
let(:content_type) { ApiController::CONTENT_TYPE }
```

app/jobs/prepare_move_notifications_job.rb
smoothcontract: Yep, that's intentional. I removed the single named argument on the job and just changed it to a regular parameter to fix the deprecation warning. It didn't seem necessary to name the argument when there's only one, so this removes the hash wrapper and just returns the id.

spec/requests/api/allocation_events_controller_cancel_spec.rb
smoothcontract: There's not enough consistency in the tests yet - there are lots of subtle variations on this and I didn't want to introduce even more noise. Happy to look at that again in a future PR though.

spec/support/supplier_with_access_token.rb
```diff
@@ -2,8 +2,7 @@
 
 RSpec.shared_context 'with supplier with access token' do
   let(:supplier) { create(:supplier) }
-  let(:application) { create(:application, owner_id: supplier.id) }
-  let(:access_token) { create(:access_token, application: application).token }
+  let(:access_token) { 'spoofed-token' }
```
martyn-w, line 5: I'm wary about this change: this specific context ties an access token to a supplier. In some tests we are (or should be) testing that supplier1 cannot access supplier2's moves. I suggest keeping this context to test specific permissions cases, and updating most specs to use spoofed authenticaiton.

smoothcontract: For tests that do need to check we get the correct details back (essentially authorisation rather than authentication) I've kept the real access token, or in some cases added a specific context to use the real token. For speed it's important to only use the full authentication stack where we absolutely have to, so I'd prefer to keep this as the default. I did get failing specs around suppliers seeing moves they weren't supposed to, so the test coverage is good for that - and is one example where it's not using the spoofed token.

spec/requests/api/allocation_events_controller_cancel_spec.rb
martyn-w: ok 👍

spec/support/supplier_with_access_token.rb
martyn-w: ok, in that case how about renaming this context 'with supplier with spoofed access token': to make it clearer that the access token is _not_ tied to supplier (as previously it was)

smoothcontract: Sure 👍

PR #659, opened 2020-06-30
[P4-1788] Further field re-mapping

mafonso:
### Jira link

P4-1788

### What?

I have added/removed/altered:

- [ ] Removed location title
- [ ] Remapped location id  to nomis_agency_id 
- [ ] Added `supplier` to the Journey object 
- [ ] Added a bookmark object to track last time the report run
- [ ] Only include moves between last report run and current time  

### Why?

I am doing this because:
- I'm converging the report functionality with the requirements.

### DIFF
--- a/lib/tasks/fake_data.rake
+++ b/lib/tasks/fake_data.rake
@@ -129,6 +129,7 @@ namespace :fake_data do
         key: title.parameterize(separator: '_'),
         title: title,
         location_type: :prison,
+        nomis_agency_id: "#{title.split.map { |i| i[0, 2] }.join.upcase}P",
         suppliers: [suppliers.sample],
       )
     end
@@ -142,6 +143,7 @@ namespace :fake_data do
       Location.create!(
         key: title.parameterize(separator: '_'),
         title: title,
+        nomis_agency_id: "#{title.split.map { |i| i[0, 2] }.join.upcase}T",
         location_type: :court,
       )
     end
--- a/lib/tasks/journeys_report.rake
+++ b/lib/tasks/journeys_report.rake
@@ -2,63 +2,82 @@ require 'csv'
 require 'json'
 require 'aws-sdk-s3'
 
+def get_last_run(bucket_name)
+  s3 = Aws::S3::Resource.new
+  # Amazon S3 lists objects in alphabetical order
+  objects = s3.bucket(bucket_name).objects().collect(&:key)
+  timestamp = objects.any? ? objects.last.split('/').first : '0'
+  DateTime.strptime(timestamp, '%s')
+end
+
 namespace :journeys do
   desc 'Exports a JSON report of all journeys'
   task json_report: :environment do
     timestamp = DateTime.now.utc.to_i
+    bucket_name = ENV['S3_REPORTING_BUCKET_NAME'] || raise('S3_REPORTING_BUCKET_NAME not defined')
+
+    if Rails.env.development?
+      Aws.config.update(
+        endpoint: 'http://localhost:4572',
+        credentials: Aws::Credentials.new('fakeid', 'fakesecret'),
+        region: 'eu-west-2',
+        force_path_style: true,
+      )
+    end
+
+    last_report_date = get_last_run(bucket_name)
+    puts last_report_date
 
     Supplier.order(:key).each do |supplier|
       filename = "#{supplier.key}-#{timestamp}.json"
       puts "Generating #{filename}..."
       moves = []
+      # Only include moves with journeys on them.
       Move
           .where("ID in (select move_id from journeys WHERE supplier_id = '#{supplier.id}')")
+          .where("moves.updated_at > '#{last_report_date}'")
+          .where("moves.updated_at <= '#{Time.at(timestamp).utc}'")
           .order('moves.date ASC, moves.reference ASC').each do |move|
         journeys = []
         Journey.where(move: move).default_order.each do |journey|
           journeys << {
-            timestamp: journey.client_timestamp.strftime('%d/%m %H:%M:%S'),
-            from: {
-              title: journey.from_location.title,
-              id: journey.from_location.id,
-              nomis_agency_id: journey.from_location.nomis_agency_id,
-            },
-            to: {
-              title: journey.to_location.title,
-              id: journey.to_location.id,
-              nomis_agency_id: journey.to_location.nomis_agency_id,
-            },
+            id: journey.id,
+            supplier: supplier.name,
+            timestamp: journey.client_timestamp.iso8601,
+            from: journey.from_location.nomis_agency_id,
+            to: journey.to_location.nomis_agency_id,
             vehicle: journey.vehicle['registration'],
             billable: journey.billable,
             events: journey.events.default_order.map do |event|
               {
-                timestamp: event.client_timestamp.strftime('%d/%m %H:%M:%S'),
+                timestamp: event.client_timestamp.iso8601,
                 event: event.event_name,
                 notes: event.notes,
-                from_location: event.from_location&.title,
-                to_location: event.to_location&.title,
+                from_location: event.from_location&.nomis_agency_id,
+                to_location: event.to_location&.nomis_agency_id,
               }
             end,
           }
         end
         moves << {
+          id: move.id,
           supplier: supplier.key,
           reference: move.reference,
-          request_date: move.created_at,
+          created_date: move.created_at.to_date,
           date: move.date,
-          from: move.from_location.key,
-          to: move.to_location&.key,
-          person_id: move.profile.id,
+          from: move.from_location.nomis_agency_id,
+          to: move.to_location&.nomis_agency_id,
+          person_id: move.profile.person_id,
+          gender: move.profile.person.gender.key,
           dob: move.profile.person.date_of_birth,
           age: ((move.date.to_date - move.profile.person.date_of_birth.to_date) / 365.25).to_i,
-          # age: move.date.year - move.profile.person.date_of_birth.year - ((move.date.month > move.profile.person.date_of_birth.month || (move.date.month == move.profile.person.date_of_birth.month && move.date.day >= move.profile.person.date_of_birth.day)) ? 0 : 1),
           events: move.move_events.default_order.map do |move_event|
             {
-              timestamp: move_event.client_timestamp.strftime('%d/%m %H:%M:%S'),
+              timestamp: move_event.client_timestamp.iso8601,
               event: move_event.event_name,
               notes: move_event.notes,
-              from_location: move_event.from_location&.title,
-              to_location: move_event.to_location&.title,
+              from_location: move_event.from_location&.nomis_agency_id,
+              to_location: move_event.to_location&.nomis_agency_id,
             }
           end,
           journeys: journeys,
@@ -69,11 +88,10 @@ namespace :journeys do
         file.write(JSON.pretty_generate({ moves: moves })) # maybe change to .to_json for production
       end
 
-      puts "Exporting #{filename} to S3..."
-      bucket_name = ENV['S3_REPORTING_BUCKET_NAME']
       if bucket_name.present?
+        puts "Exporting #{filename} to S3..."
         s3 = Aws::S3::Resource.new
-        obj = s3.bucket(bucket_name).object(filename)
+        obj = s3.bucket(bucket_name).object("#{timestamp}/#{filename}")
         obj.upload_file(filename)
       else
         puts 'S3_REPORTING_BUCKET_NAME not defined, skipping upload.'
--- a/lib/tasks/journeys_report.rake
+++ b/lib/tasks/journeys_report.rake
@@ -44,18 +44,23 @@ namespace :journeys do
         moves << {
           supplier: supplier.key,
           reference: move.reference,
+          request_date: move.created_at,
           date: move.date,
-          from: move.from_location.title,
-          to: move.to_location&.title,
+          from: move.from_location.key,
+          to: move.to_location&.key,
+          person_id: move.profile.id,
+          dob: move.profile.person.date_of_birth,
+          age: ((move.date.to_date - move.profile.person.date_of_birth.to_date) / 365.25).to_i,
+          # age: move.date.year - move.profile.person.date_of_birth.year - ((move.date.month > move.profile.person.date_of_birth.month || (move.date.month == move.profile.person.date_of_birth.month && move.date.day >= move.profile.person.date_of_birth.day)) ? 0 : 1),
           events: move.move_events.default_order.map do |move_event|
-                    {
-                      timestamp: move_event.client_timestamp.strftime('%d/%m %H:%M:%S'),
-                      event: move_event.event_name,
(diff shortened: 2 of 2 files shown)


lib/tasks/journeys_report.rake
```diff
@@ -44,18 +44,23 @@ namespace :journeys do
         moves << {
           supplier: supplier.key,
           reference: move.reference,
+          request_date: move.created_at,
```
martyn-w, line 47: I'd label this as `created_date`, not `requested_date` to avoid confusion

mafonso: Good point. Re-labeled.

lib/tasks/fake_data.rake
```diff
@@ -129,6 +129,7 @@ namespace :fake_data do
         key: title.parameterize(separator: '_'),
         title: title,
         location_type: :prison,
+        nomis_agency_id: "#{title.split.map { |i| i[0, 2] }.join.upcase}P",
```
martyn-w, line 132: simplify to: `title[0,2].upcase`

lib/tasks/journeys_report.rake
```diff
@@ -2,60 +2,97 @@ require 'csv'
...
+    bucket_name = ENV['S3_REPORTING_BUCKET_NAME'] || raise('S3_REPORTING_BUCKET_NAME not defined')
+
+    if ENV['RAILS_ENV'].to_s == 'development' || ENV['RAILS_ENV'].to_s == ''
+      Aws.config.update(
+        endpoint: 'http://localhost:4572',
+        credentials: Aws::Credentials.new('fakeid', 'fakesecret'),
+        region: 'eu-west-2',
+        force_path_style: true,
+      )
+    end
```
mafonso, line 36: This is useful during local development and the `endpoint` cannot yet be overridden with an ENV variable. 
We can get rid of it later.

lib/tasks/fake_data.rake
```diff
@@ -12,7 +12,7 @@ namespace :fake_data do
       person = Person.create!(
         first_names: Faker::Name.first_name,
         last_name: Faker::Name.last_name,
-        date_of_birth: Faker::Date.between(from: 80.years.ago, to: 20.years.ago),
+        date_of_birth: Faker::Date.between(from: 80.years.ago.to_date, to: 20.years.ago.to_date),
```
martyn-w, line 15: are the extra `.to_date`s necessary ? Faker seems to run fine without them:
```
pry(main)> Faker::Date.between(from: 80.years.ago, to: 20.years.ago)
=> Fri, 27 Oct 1950
```

lib/tasks/fake_data.rake
```diff
@@ -142,6 +143,7 @@ namespace :fake_data do
       Location.create!(
         key: title.parameterize(separator: '_'),
         title: title,
+        nomis_agency_id: "#{title.split.map { |i| i[0, 2] }.join.upcase}T",
```
martyn-w, line 146: simplify to `[0,2].upcase`

lib/tasks/fake_data.rake
```diff
@@ -179,7 +181,7 @@ namespace :fake_data do
     courts = Location.where(location_type: 'court').all
     file = StringIO.new(File.read('spec/fixtures/file-sample_100kB.doc'))
     1000.times do
-      date = Faker::Date.between(from: 10.days.ago, to: 20.days.from_now)
+      date = Faker::Date.between(from: 10.days.ago.to_date, to: 20.days.from_now.to_date)
```
martyn-w, line 184: no need for .to_date

lib/tasks/fake_data.rake
```diff
@@ -212,7 +214,7 @@ namespace :fake_data do
     puts 'create_allocations...'
     prisons = Location.where(location_type: 'prison').all
     50.times do
-      date = Faker::Date.between(from: 10.days.ago, to: 20.days.from_now)
+      date = Faker::Date.between(from: 10.days.ago.to_date, to: 20.days.from_now.to_date)
```
martyn-w, line 217: no need for .to_date

lib/tasks/journeys_report.rake
```diff
@@ -2,60 +2,97 @@ require 'csv'
...
+                from_location: event.from_location&.nomis_agency_id,
+                to_location: event.to_location&.nomis_agency_id,
               }
             end,
           }
         end
         moves << {
           supplier: supplier.key,
           reference: move.reference,
+          request_date: move.created_at.to_date,
```
martyn-w: add a comment/note/todo that this should be fixed at some point; otherwise we will forget and one day someone will ask why the requested date is the same as the created date

martyn-w: actually, I really don't think we should label this as `request_date` to avoid misunderstanding of data consumers.

lib/tasks/journeys_report.rake
```diff
@@ -64,15 +102,15 @@ namespace :journeys do
...
       if bucket_name.present?
+        puts "Exporting #{filename} to S3..."
         s3 = Aws::S3::Resource.new
         obj = s3.bucket(bucket_name).object(filename)
         obj.upload_file(filename)
       else
         puts 'S3_REPORTING_BUCKET_NAME not defined, skipping upload.'
       end
     end
+    update_bookmark(bucket_name)
```
martyn-w, line 114: 1. I think you should use the `timestamp` to label the bucket, as that is the time the process started, not the time when it finished.
2. you should only update the bookmark if you have uploaded files to S3, so the logic needs to change a little

lib/tasks/journeys_report.rake
martyn-w: 👍

lib/tasks/journeys_report.rake
```diff
@@ -2,60 +2,98 @@ require 'csv'
...
+    last_report_date = get_last_run(bucket_name)
+    puts last_report_date
 
     Supplier.order(:key).each do |supplier|
       filename = "#{supplier.key}-#{timestamp}.json"
       puts "Generating #{filename}..."
       moves = []
       Move
           .where("ID in (select move_id from journeys WHERE supplier_id = '#{supplier.id}')")
+          .where("moves.date > '#{last_report_date}'")
```
martyn-w, line 47: Watch out for three things: 
1. some pending moves don't have a date
2. move.date is only a date, not a datetime so what happens if you run the exporter twice on the same day (say 9am + 9pm) - you'll get duplicates
3. probably move.created_at is a better thing to segregate the date on, as all moves will have one and it will never change. Whereas move.date could be updated.

lib/tasks/journeys_report.rake
```diff
@@ -2,60 +2,98 @@ require 'csv'
...
+    if ENV['RAILS_ENV'].to_s == 'development' || ENV['RAILS_ENV'].to_s == ''
+      Aws.config.update(
+        endpoint: 'http://localhost:4572',
+        credentials: Aws::Credentials.new('fakeid', 'fakesecret'),
+        region: 'eu-west-2',
+        force_path_style: true,
+      )
+    end
+
+    last_report_date = get_last_run(bucket_name)
```
martyn-w, line 38: consider the logic for picking moves between certain times, and also the time it takes for the exporter to run. You'll need to ensure there are no duplicates and no missed moves between runs.

martyn-w: What happens if moves are created whilst the process is running ? I think it will be necessary to define a clear from_datetime and to_datetime and only pick moves within that range

martyn-w: Actually, thinking about this further; its possible for a move created last week to have a new journey added today: if we just filter on move.created_at time, we will miss the new journey. So, filtering needs more thought!

lib/tasks/fake_data.rake
mafonso: That only seems to give the first two chars of the whole string.
```
title = 'watford magistrates court'
nomis_id=title[0,2].upcase
=>"WA"
```

I wanted the first two of each word
```
=>nomis_agency_id= "#{title.split.map{|i| i[0,2]}.join.upcase}T"
```
As it is similar in format and length to what the real names are.

lib/tasks/fake_data.rake
mafonso: I just added it because I got a warning `Incompatible types Required:Date Passed:Time` on my pre-commit checks
But the project's rubocop config seems to ignore it. 

So I guess we can remove it.

lib/tasks/journeys_report.rake
mafonso: yes, those are very good points.   
It may take some time to run with large datasets.

lib/tasks/journeys_report.rake
mafonso: Yes, agreed.  I'll change to `created_at` for now, just to get the logic roughly working and we'll need to revisit to catch those edge cases.

lib/tasks/fake_data.rake
martyn-w: sorry, my mistake 👍

lib/tasks/fake_data.rake
martyn-w: lets go with the project's rubocop rules

lib/tasks/journeys_report.rake
mafonso: Not using a dedicated object for the bookmark.  Went with an approach between our timestamps and a typical S3 data partitioning format with a key prefix
but instead of:
```
PRE year=2020/month=06/day=01/
PRE year=2020/month=06/day=02/
```
I'm using a rather simpler
```
       PRE 1593773385/
       PRE 1593773505/
       PRE 1593773530/
```

lib/tasks/journeys_report.rake
mafonso: Using the effecting report time with the new approach.

lib/tasks/journeys_report.rake
mafonso: How about updated_at.  My reasoning is that a move might have events or journeys added to it between reports.
We would want to send a updated version of the move on the following report, so that the HUB can update their records.
my assumption here is that updated_at would reflect events or journeys being added to the move, which might not necessarily be true.

lib/tasks/journeys_report.rake
mafonso: > if we just filter on move.created_at time,

Would updated_at be a more suitable choice?

lib/tasks/journeys_report.rake
```diff
@@ -2,60 +2,85 @@ require 'csv'
...
-            to: {
-              title: journey.to_location.title,
-              id: journey.to_location.id,
-              nomis_agency_id: journey.to_location.nomis_agency_id,
-            },
+            id: journey.id,
+            supplier: supplier.name,
+            timestamp: journey.client_timestamp.iso8601,
+            from: journey.from_location.nomis_agency_id,
+            to: journey.to_location.nomis_agency_id,
```
mafonso, line 48: I've flattened the `to`and `from`to a string as we only seem to require the nomis_agency_id.

lib/tasks/journeys_report.rake
```diff
@@ -64,11 +89,10 @@ namespace :journeys do
         file.write(JSON.pretty_generate({ moves: moves })) # maybe change to .to_json for production
       end
 
-      puts "Exporting #{filename} to S3..."
-      bucket_name = ENV['S3_REPORTING_BUCKET_NAME']
       if bucket_name.present?
+        puts "Exporting #{filename} to S3..."
         s3 = Aws::S3::Resource.new
-        obj = s3.bucket(bucket_name).object(filename)
+        obj = s3.bucket(bucket_name).object("#{timestamp}/#{filename}")
```
mafonso, line 95: Using the `timestamp`as a prefix (ends up looking like a directory).
Kept the timestamp in the file regardless, as it it convenient when you pull a flat list of the keys (striping all the prefixes).

lib/tasks/journeys_report.rake
```diff
@@ -2,60 +2,85 @@ require 'csv'
 require 'json'
 require 'aws-sdk-s3'
 
+def get_last_run(bucket_name)
+  s3 = Aws::S3::Resource.new
+  # Amazon S3 lists objects in alphabetical order
+  objects = s3.bucket(bucket_name).objects().collect(&:key)
+  timestamp = objects.any? ? objects.last.split('/').first : '0'
+  last_run_date = DateTime.strptime(timestamp, '%s')
```
martyn-w, line 10:
```suggestion
  DateTime.strptime(timestamp, '%s')
```

lib/tasks/journeys_report.rake
```diff
@@ -2,60 +2,85 @@ require 'csv'
...
+  last_run_date
+end
+
 namespace :journeys do
   desc 'Exports a JSON report of all journeys'
   task json_report: :environment do
     timestamp = DateTime.now.utc.to_i
+    bucket_name = ENV['S3_REPORTING_BUCKET_NAME'] || raise('S3_REPORTING_BUCKET_NAME not defined')
+
+    if ENV['RAILS_ENV'].to_s == 'development' || ENV['RAILS_ENV'].to_s == ''
```
martyn-w, line 20:
```suggestion
    if Rails.env == 'development'
```

lib/tasks/journeys_report.rake
martyn-w: so, unfortunately, updating a journey attached to a move does not update the parent move record.

However, its pretty trivial to add this functionality: in the journey model (`app/models/journey`), change `belongs_to :move` to `belongs_to :move, touch: true`

Now, whenever a journey is updated, so will its parent move. Rails magic!

lib/tasks/journeys_report.rake
```diff
@@ -2,60 +2,85 @@ require 'csv'
...
+    puts last_report_date
 
     Supplier.order(:key).each do |supplier|
       filename = "#{supplier.key}-#{timestamp}.json"
       puts "Generating #{filename}..."
       moves = []
       Move
           .where("ID in (select move_id from journeys WHERE supplier_id = '#{supplier.id}')")
+          .where("moves.updated_at > '#{last_report_date}'")
+          .where('moves.updated_at < now()')
```
martyn-w, line 39: don't use `now()` -  that will increase as the process is running - instead, capture `current_time=DateTime.now` before the loop starts and use that value instead

mafonso: good point!

lib/tasks/journeys_report.rake
mafonso: Will do that in a separate PR, as it affects the main app as well.

lib/tasks/journeys_report.rake
```diff
@@ -17,7 +16,7 @@ namespace :journeys do
     timestamp = DateTime.now.utc.to_i
     bucket_name = ENV['S3_REPORTING_BUCKET_NAME'] || raise('S3_REPORTING_BUCKET_NAME not defined')
 
-    if Rails.env == 'development'
+    if Rails.env.development?
```
mafonso, line 19: Rubocop prefers it like this.

PR #663, opened 2020-07-01
Update swagger documentation 

cwrw:
Update swagger documentation to include the following endpoints:
`POST /court_hearings`
`GET /people/{person_id}/court_cases`
`GET /people/{person_id}/timetable`
`POST /documents`

Remove old `POST /moves/{move_id}/documents` endpoint
Remove duplication in `GET /reference/regions` endpoint that appears under region and locations

### DIFF
--- a/spec/swagger/swagger_doc_v1.yaml
+++ b/spec/swagger/swagger_doc_v1.yaml
@@ -384,6 +384,57 @@
             application/vnd.api+json:
               schema:
                 "$ref": "../v1/post_court_hearing_responses.yaml#/422"
+  "/documents":
+    post:
+      summary: Creates a new document
+      tags:
+        - Documents
+      consumes:
+        - application/vnd.api+json
+      parameters:
+        - name: body
+          in: body
+          required: true
+          description: The document object to be created
+          schema:
+            "$ref": "../v1/document.yaml#/Document"
+      responses:
+        "201":
+          description: created
+          content:
+            application/vnd.api+json:
+              schema:
+                "$ref": "../v1/post_documents_responses.yaml#/201"
+        "400":
+          description: bad request
+          content:
+            application/vnd.api+json:
+              schema:
+                "$ref": "../v1/post_documents_responses.yaml#/400"
+        "401":
+          description: unauthorized
+          content:
+            application/vnd.api+json:
+              schema:
+                "$ref": "../v1/post_documents_responses.yaml#/401"
+        "404":
+          description: resource not found
+          content:
+            application/vnd.api+json:
+              schema:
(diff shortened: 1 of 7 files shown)


spec/swagger/swagger_doc_v1.yaml
```diff
@@ -1141,59 +1243,6 @@
             application/vnd.api+json:
               schema:
                 "$ref": "../v1/error_responses.yaml#/422"
-  "/moves/{move_id}/documents":
```
martyn-w, line 1144: as a general rule, we should be cautious about (appearing to) remove features from v1 - suppliers could get upset; are you certain that this can and should be removed?

cwrw: Im happy to keep this even if it's been removed for the suppliers, as I couldn't find this path anywhere in routes: https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/config/routes.rb#L23

willfish:
I was having a look at this before production distractions yesterday.

We seem to have - in the past - added a mix of our rswag-style swagger with and without hand-coded paths (i.e. duplicating how we generate documentation). We checked the moves show to make sure we had all the documentation (which turns out is the only path that was duplicated in both hand_coded_paths and the rswag-style specs 🤦 .

```
spec/swagger/api/people_controller_image_spec.rb 
spec/swagger/api/moves_controller_show_spec.rb # duplicated generation of path
spec/swagger/api/people_controller_timetable_spec.rb
spec/swagger/api/court_hearings_controller_create_spec.rb
spec/swagger/api/people_controller_court_cases_spec.rb
spec/swagger/api/allocations_controller_show_spec.rb
spec/requests/api/moves_controller_role_access_rswag_spec.rb
spec/requests/api/reference/prison_transfer_reasons_controller_rswag_spec.rb
```

I think we're still missing `GET /api/reference/prison_transfer_reasons`

spec/swagger/swagger_doc_v1.yaml
teneightfive: Is there a way for us to check what endpoints they are/aren't using? 

I double checked we weren't using it on the FE and it hasn't used it for a while. We now use the preferred JSON:API way of creating resources at the resource endpoint of `/documents`.

martyn-w: Looking further, I see that this route was recently been removed in PR https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/612 and so far, there have been no complaints from suppliers. So I guess we have gotten away with this breaking change this time 😬

PR #664, opened 2020-07-02
Force to pull a new image on migrations

mafonso:
### Jira link

n/a

### What?

I have added/removed/altered:

- [ ] Changed the imagePullPolicy to **always**

### Why?

I am doing this because:

- It's possible that migrations can run with the existent version of the  container. 
We are tracking the `env.lastest` tag, that as been just recently updated when this runs.  If we don't force pull it for this job and the job is running in a node that is running any of the other pods we might end up using the cached one.

### DIFF
--- a/k8s_migrate_job.yml
+++ b/k8s_migrate_job.yml
@@ -16,6 +16,7 @@ spec:
         - name: app
           image: 754256621582.dkr.ecr.eu-west-2.amazonaws.com/book-a-secure-move/hmpps-book-secure-move-api:__ENV__.latest
           command: ["bundle", "exec", "rake", "db:migrate"]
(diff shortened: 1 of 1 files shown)


k8s_migrate_job.yml
```diff
@@ -16,6 +16,7 @@ spec:
         - name: app
           image: 754256621582.dkr.ecr.eu-west-2.amazonaws.com/book-a-secure-move/hmpps-book-secure-move-api:__ENV__.latest
```
mafonso, line 17: We probably don't want to use `:__ENV__.latest` here anyway. 
It's more robust to use the actual docker tag for the build, but we might want to address the deployment job at the same time for that.

PR #665, opened 2020-07-02
Improve Framework importer and error logging

cwrw:
* Don't attempt to persist already imported framework
* Add logging for framework import success
* Add logging for framework import failure, with additional logging of all errors coming from the framework and the questions specifically with keys to help point out the offending files coming from the framework

### DIFF
--- a/app/services/frameworks/importer.rb
+++ b/app/services/frameworks/importer.rb
@@ -1,10 +1,11 @@
 module Frameworks
   class Importer
-    attr_reader :filepath, :version
+    attr_reader :filepath, :version, :errors
 
     def initialize(filepath:, version:)
       @filepath = filepath
       @version = version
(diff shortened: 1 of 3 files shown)


app/services/frameworks/importer.rb
```diff
@@ -35,5 +39,31 @@ def build_questions(framework, questions)
         questions.merge!(Frameworks::Question.new(filepath: question, questions: questions.dup).call)
       end
     end
+
+    def persist_framework(name, questions)
+      framework = Framework.new(name: name, version: version, framework_questions: questions.values)
+
+      if framework.valid?
```
smoothcontract, line 46: Perhaps change to `if framework.save` - that will return false if the record isn't valid so combines two lines in one

PR #666, opened 2020-07-02
Remove person_id from moves

willfish:
### Jira link

P4-<TODO>

### What?

I have added/removed/altered:

- [x] Drop the moves.person_id column
- [x] Remove references to moves.person_id column

### Why?

Final bit of cleanup around the migration of a person/profile and the move.

### DIFF
--- a/app/serializers/move_serializer.rb
+++ b/app/serializers/move_serializer.rb
@@ -51,12 +51,7 @@ class MoveSerializer < ActiveModel::Serializer
   }.freeze
 
   def person
-    # TODO: Remove the support for person id in future
-    if object.profile_id
-      object&.profile&.person
-    elsif object.person_id
-      Person.find(object.person_id)
-    end
+    object&.profile&.person
   end
 
   def documents
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_07_01_123259) do
+ActiveRecord::Schema.define(version: 2020_07_02_083741) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
@@ -50,6 +50,7 @@ ActiveRecord::Schema.define(version: 2020_07_01_123259) do
     t.string "prisoner_category"
     t.string "sentence_length"
     t.jsonb "complex_cases"
+    t.integer "moves_count", null: false
     t.boolean "complete_in_full", default: false, null: false
     t.text "other_criteria"
     t.datetime "created_at", null: false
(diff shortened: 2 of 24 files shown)


app/serializers/move_serializer.rb
```diff
@@ -51,12 +51,7 @@ class MoveSerializer < ActiveModel::Serializer
   }.freeze
 
   def person
-    # TODO: Remove the support for person id in future
-    if object.profile_id
-      object&.profile&.person
-    elsif object.person_id
-      Person.find(object.person_id)
-    end
+    object&.profile&.person
```
smoothcontract, line 54: Don't think you need the guard against `object` being nil - just the one on the `profile`.

db/schema.rb
```diff
@@ -50,6 +50,7 @@
     t.string "prisoner_category"
     t.string "sentence_length"
     t.jsonb "complex_cases"
+    t.integer "moves_count", null: false
```
smoothcontract, line 53: Unrelated to PR

spec/requests/api/moves_controller_index_spec.rb
```diff
@@ -208,38 +208,6 @@
...
-      context 'when the move includes only a profile id' do
-        let(:person) { create(:person) }
-        let!(:move) { create(:move, person_id: nil, profile_id: person.latest_profile.id) }
-
-        it 'returns the correct person' do
-          get_moves
-          person_id = response_json['data'][0].dig('relationships', 'person', 'data', 'id')
-          expect(person_id).to eq(person.id)
-        end
-      end
```
smoothcontract, line 242: Do we still have coverage for the serializer `person` method both with and without a profile?

willfish: Move can't have a person_id anymore

app/serializers/move_serializer.rb
willfish: Feel free to use github code suggestions. Makes it easier

db/schema.rb
willfish: Weirdly moved, yeah.

spec/requests/api/moves_controller_index_spec.rb
willfish: We are validating that a person can be surfaced for a move, yeah.

app/serializers/move_serializer.rb
smoothcontract, line 54:
```suggestion
    object.profile&.person
```

db/schema.rb
smoothcontract: They do that all the time - I try to avoid adding those lines to the index but it is an annoying aspect of migrations.. seems over time everyone ends up with a slightly divergent column order in postgres.

PR #668, opened 2020-07-02
[P4-1852] Fix importing offenders from Nomis in v2

willfish:
### Jira link

P4-1852

### What?

The service we wrote and were mocking out in the controller for
importing a person from Nomis was built to operate against a single
person which is inconsistent with filter param possibilities described
by the json:api spec.

- [x] Enable multi-offender support in `People::ImportFromNomis` service

image

### Why?

This is critical functionality for the frontend and is currently broken.

### DIFF
--- a/app/services/people/import_from_nomis.rb
+++ b/app/services/people/import_from_nomis.rb
@@ -17,13 +17,12 @@ module People
     end
 
     def call
-      @prison_numbers.each do |prison_number|
+      people_from_nomis.each do |nomis_person|
+        prison_number = nomis_person[:prison_number]
         person = Person.find_or_initialize_by(prison_number: prison_number)
-        nomis_person = nomis_person(prison_number)
 
-        next unless nomis_person
-
-        person.assign_attributes(person_attributes(nomis_person))
+        attributes = person_attributes(nomis_person)
+        person.assign_attributes(attributes)
         person.save!
       end
     end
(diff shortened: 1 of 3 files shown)


app/services/people/import_from_nomis.rb
```diff
@@ -2,7 +2,7 @@
 
 module People
   class ImportFromNomis
-    MAPPING = {
+    NOMIS_OFFENDER_TO_PERSON_MAPPING = {
```
alexdesi, line 5: 👍

app/services/people/import_from_nomis.rb
```diff
@@ -12,43 +12,53 @@ class ImportFromNomis
...
-      person.assign_attributes(person_attributes)
-      person.save!
+      @prison_numbers.each do |prison_number|
+        person = Person.find_or_initialize_by(prison_number: prison_number)
+
+        nomis_person = people_from_nomis.find do |nomis_person|
+          nomis_person[:prison_number] == prison_number
+        end
+
+        next unless nomis_person
```
alexdesi: I think we can make this clearer, we could filter @prison_number before looping.

app/services/people/import_from_nomis.rb
```diff
@@ -12,43 +12,56 @@ class ImportFromNomis
...
 
     def call
-      person = Person.find_or_initialize_by(prison_number: @prison_number)
-      person.assign_attributes(person_attributes)
-      person.save!
+      @prison_numbers.each do |prison_number|
+        person = Person.find_or_initialize_by(prison_number: prison_number)
+        nomis_person = nomis_person(prison_number)
+
+        next unless nomis_person
```
alexdesi, line 24: we do not need this if we filter the prison numbers earlier, before the each, something like:
```
@prison_numbers.select do
  people_from_nomis.find { |nomis_person| nomis_person[:prison_number] == prison_number }.any?
end
```

PR #672, opened 2020-07-03
Removes references to moves.person_id

willfish:
### Jira link

P4-<TODO>

### What?

- [x]  Remove all references to `moves.person_id` field

### Why?

We migrated away from `Move#person` to `Move#profile#person`.

I've added the ignored_columns for person_id to soft-force our not referencing the person_id and to stop us loading it as an active record attribute on the model.

### DIFF
--- a/app/serializers/move_serializer.rb
+++ b/app/serializers/move_serializer.rb
@@ -50,15 +50,6 @@ class MoveSerializer < ActiveModel::Serializer
     allocation: %i[to_location from_location moves_count created_at],
   }.freeze
 
-  def person
-    # TODO: Remove the support for person id in future
-    if object.profile_id
-      object&.profile&.person
-    elsif object.person_id
-      Person.find(object.person_id)
-    end
-  end
-
   def documents
     object&.profile&.documents
   end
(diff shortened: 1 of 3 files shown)


app/serializers/move_serializer.rb
```diff
@@ -50,15 +50,6 @@ class MoveSerializer < ActiveModel::Serializer
     allocation: %i[to_location from_location moves_count created_at],
   }.freeze
 
-  def person
```
willfish, line 53: The move model defines a :through relationship so this isn't needed

See: 

https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/dbf4d75b6ccc0ca24ef14640f92394532c4e1653/app/serializers/move_serializer.rb#L21
https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/dbf4d75b6ccc0ca24ef14640f92394532c4e1653/app/models/move.rb#L50

spec/requests/api/moves_controller_index_spec.rb
```diff
@@ -208,38 +208,6 @@
           end
         end
       end
-
-      context 'when the move includes both a profile id and a person id' do
```
willfish, line 212: We validate the inclusion of a person, here: 

https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/dbf4d75b6ccc0ca24ef14640f92394532c4e1653/spec/serializers/move_serializer_spec.rb#L70-L106

PR #674, opened 2020-07-03
P4-1746 Build PER related tables

cwrw:
### Jira link

[P4-1746](https://dsdmoj.atlassian.net/browse/P4-1746)

### What?

- [x] Added a person escort record table and model
- [x] Added a framework response table and model
- [x] Added a framework response and flag join table
- [x] Added ability to set one value on framework responses

### Why?

To facilitate person escort record creation with responses, add the relevant tables as well as the validations and associations.
A Person escort record can be created with a framework and version, and will contain many responses which are initially empty. A response depending on the question it is for, can hold many answer types. To store the different types we have added two fields on the response table: text and jsonb. Those two will be able to store all the required values sent to us as answers, and will be hidden under a single field called value. According to the value type, the relevant answers will be set accordingly to the underlying tables.

### DIFF
--- a/app/models/framework_response.rb
+++ b/app/models/framework_response.rb
@@ -18,15 +18,18 @@ class FrameworkResponse < VersionedModel
   belongs_to :parent, class_name: 'FrameworkResponse', optional: true
   has_and_belongs_to_many :flags, join_table: 'framework_responses_flags'
 
+  validates :value_text, absence: true, if: -> { json? || array? }
+  validates :value_json, absence: true, if: -> { (string? || text?) && value_json != '{}' }
+
   def value
-    json? ? value_json : value_text
+    json? || array? ? value_json : value_text
   end
 
   def value=(answer)
-    if json?
-      value_json = answer
+    if json? || array?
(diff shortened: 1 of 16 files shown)


app/models/framework_response.rb
```diff
@@ -0,0 +1,45 @@
...
+  has_many :dependents, class_name: 'FrameworkResponse',
+                        foreign_key: 'parent_id'
+
+  belongs_to :parent, class_name: 'FrameworkResponse', optional: true
+  has_and_belongs_to_many :flags, join_table: 'framework_responses_flags'
+
+  validates :value_text, absence: true, if: -> { json? || array? }
+  validates :value_json, absence: true, if: -> { string? || text? }
+
+  def value
```
cwrw, line 24: Only set/get value on a `framework_response`. The value is responsible where to set the actual answers based on their type in the database

cwrw: On second thought this will probably change when I implement validations as we will be defining different classes for each type of question so those might move specifically to the defined radio/textbox etc.

PR #676, opened 2020-07-03
[P4-1797] Make logs parseable by fluentd

willfish:
### Jira link

P4-1797

### What?

This PR moves our logs (in production, only) to spit out json rather than unparseable strings.

### Why?

Fluentd defines a json parser for pod logs:

https://github.com/ministryofjustice/analytics-platform-fluentd/blob/eb7d529ba83abaa58c8253a39d2d11a86b91dc39/conf/kubernetes.conf#L8-L16

We think that if we move our logs to a json document per line we'll be able to filter our logs for useful things like request id in future (more on this later).

This will help us to debug issues in production more conveniently and generally make our logs more useable.

### E2E tests

```
 Running tests in:

 Cancel allocation
 ✓ Cancel allocation

 New PMU allocation

 ✓ With unfound person
 ✓ With existing person

 ✓ With a new person

 ✓ Reason - `Made in error`
 ✓ Reason - `Another reason`

 New move from Prison to Court
 ✓ With unfound person
 ✓ With existing person

 ✓ With existing person

 Download moves
 ✓ As Police user
 ✓ As Prison user
 ✓ As STC user
 ✓ As Supplier user

 Smoke tests
 ✓ As Police user
 ✓ As STC user
 ✓ As Prison user
 ✓ As Supplier user
 ✓ As OCA user


 21 passed
```

### DIFF
--- a/config/initializers/lograge.rb
+++ b/config/initializers/lograge.rb
@@ -1,5 +1,5 @@
 Rails.application.configure do
   config.lograge.base_controller_class = 'ActionController::API'
-  config.lograge.formatter = Lograge::Formatters::Logstash.new
+  config.lograge.formatter = Lograge::Formatters::Json.new
 end
 
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -1,3 +1,5 @@
+require 'log_formatter/json'
+
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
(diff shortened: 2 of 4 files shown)


config/initializers/lograge.rb
```diff
@@ -1,5 +1,5 @@
 Rails.application.configure do
   config.lograge.base_controller_class = 'ActionController::API'
-  config.lograge.formatter = Lograge::Formatters::Logstash.new
+  config.lograge.formatter = Lograge::Formatters::Json.new
```
willfish, line 3: Handles action_controller process_action subscription events (https://guides.rubyonrails.org/active_support_instrumentation.html#process-action-action-controller).

willfish: Lograge actually wipes out all default controller logging - including support for `log_tags` (a poorly named config option for pulling out attributes from an inflight request object).

willfish: This means we can't actually add request_ids without monkey patching various bits of our logging deps.

PR #680, opened 2020-07-03
[P4-1797] Removes application_insights

willfish:
### Jira link

P4-1797

### What?

Application insights is an Azure application which isn't particularly
compatible with ruby/and the gem is now unmaintained by Microsoft.

Requests its telemetry client makes to azure endpoints are really slow and this is causing issues with the responsiveness of our client applications.

### Why?

We get these insights from Kibana currently and are going to have a
conversation with suppliers/the wider team to see how we can replace
this broken implementation later.


willfish:
If we decide to do this later we can reopen this PR. I wasn't getting a lot of traction with it.

PR #681, opened 2020-07-03
[P4-2069] Remove appinsights

willfish:
### Jira link

P4-2069

### What?

- [x] Removes app insights gem

### Why?

Appinsights doesn't currently meet our audit requirements and slows down requests unpredictably.


martyn-w:
I don't think we can remove app insights without first switching to a viable alternative?

willfish:
We've got a meeting on the 7th of September to discuss alternatives/requirements :).

PR #688, opened 2020-07-06
Watch out for moves with null profiles

mafonso:
### Jira link

P4-1788

### What?

I have added/removed/altered:

- [ ] Handle moves with no profiles.

### Why?

I am doing this because:

- In staging we have some data quality issues. So this is to protect against those cases

### Have you? (optional)

### DIFF
--- a/lib/tasks/journeys_report.rake
+++ b/lib/tasks/journeys_report.rake
@@ -67,10 +67,10 @@ namespace :journeys do
           date: move.date,
           from: move.from_location.nomis_agency_id,
           to: move.to_location&.nomis_agency_id,
-          person_id: move.profile.person_id,
-          gender: move.person.gender&.key,
+          person_id: move.profile&.person_id,
+          gender: move.person.gender&.keys,
           dob: move.person.date_of_birth,
-          age: ((move.date.to_date - move.profile.person.date_of_birth.to_date) / 365.25).to_i,
+          age: ((move.date.to_date - move.person.date_of_birth.to_date) / 365.25).to_i,
           events: move.move_events.default_order.map do |move_event|
             {
               timestamp: move_event.client_timestamp.iso8601,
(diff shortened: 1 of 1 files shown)


lib/tasks/journeys_report.rake
```diff
@@ -67,7 +67,7 @@ namespace :journeys do
           date: move.date,
           from: move.from_location.nomis_agency_id,
           to: move.to_location&.nomis_agency_id,
-          person_id: move.profile.person_id,
+          person_id: move.profile&.person_id,
```
martyn-w, line 70: you'll need to check for move.profile.person throughout this file (e.g. gender, date_of_birth etc) and update accordingly

mafonso: Included all the references to `profile` in `moves`

lib/tasks/journeys_report.rake
```diff
@@ -67,10 +67,10 @@ namespace :journeys do
...
           from: move.from_location.nomis_agency_id,
           to: move.to_location&.nomis_agency_id,
-          person_id: move.profile.person_id,
-          gender: move.person.gender&.key,
+          person_id: move.profile&.person_id,
+          gender: move.person.gender&.keys,
           dob: move.person.date_of_birth,
-          age: ((move.date.to_date - move.profile.person.date_of_birth.to_date) / 365.25).to_i,
+          age: ((move.date.to_date - move.person.date_of_birth.to_date) / 365.25).to_i,
           events: move.move_events.default_order.map do |move_event|
```
mafonso, line 74: Some rebase needed,  `gender` and ` date_of_birth` we get from `persons`.

config/spring.rb
```diff
@@ -4,3 +4,11 @@
   tmp/restart.txt
   tmp/caching-dev.txt
 ].each { |path| Spring.watch(path) }
+Spring.after_fork do
```
martyn-w, line 7: oops - I don't think this debugger stuff is intended for here

mafonso: Oops, it ended up getting  pushed.

PR #689, opened 2020-07-06
P4-1746 Build PER related tables

cwrw:
### Jira link

[P4-1746](https://dsdmoj.atlassian.net/browse/P4-1746)

### What?

- [x] Added a person escort record table and model
- [x] Added a framework response table and model
- [x] Added STI on framework responses

### Why?

To facilitate person escort record creation with responses, add the relevant tables as well as the validations and associations.
A Person escort record can be created with a framework and version, and will contain many responses which are initially empty. A response depending on the question it is for, can hold many answer types. To store the different types we have added two fields on the response table: text and jsonb. Those two will be able to store all the required values sent to us as answers, and will be hidden under a single field called value. According to the value type, the relevant answers will be set accordingly to the underlying tables. 

The 4 types of responses are: string, array, object and collection

- **String**: This response is for questions with a text box, or a radio button which will require one string answer such as 'Yes'
- **Array**: This response is for questions which are a checklist, so users can select multiple options
- **Object**: This response is specifically for questions which are a radio, but could contain an optional/required comment or details box. To accommodate this client will send an object instead with an "option" and "details" field. 
- **Collection**: This response is for questions which are a checklist, but could contain an optional/required comment. 

Validations:
- If response is required based on attribute on question
- If option is included in list of options
- If radio response requires details 
- If checklist response requires details
- If question is a dependent question, response becomes required

### DIFF
--- a/app/models/framework_response.rb
+++ b/app/models/framework_response.rb
@@ -9,9 +9,4 @@ class FrameworkResponse < VersionedModel
                         foreign_key: 'parent_id'
 
   belongs_to :parent, class_name: 'FrameworkResponse', optional: true
-
-  def self.find_sti_class(type_name)
-    type_name = self.name
-    super
-  end
 end
--- a/app/models/framework_response.rb
+++ b/app/models/framework_response.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+class FrameworkResponse < VersionedModel
+  validates :type, presence: true
+
+  belongs_to :framework_question
+  belongs_to :person_escort_record
+  has_many :dependents, class_name: 'FrameworkResponse',
+                        foreign_key: 'parent_id'
+
+  belongs_to :parent, class_name: 'FrameworkResponse', optional: true
+
+  def self.find_sti_class(type_name)
+    type_name = self.name
+    super
+  end
+end
--- a/app/models/framework_response.rb
+++ b/app/models/framework_response.rb
@@ -9,8 +9,8 @@ class FrameworkResponse < VersionedModel
                         foreign_key: 'parent_id'
 
   belongs_to :parent, class_name: 'FrameworkResponse', optional: true
-  validates_each :value, on: :update do |record, _attr, _value|
-    if record.parent&.option_selected?(record.framework_question.dependent_value) && record.framework_question.required
+  validates_each :value, on: :update do |record, _attr, value|
+    if value.blank? && record.parent&.option_selected?(record.framework_question.dependent_value) && record.framework_question.required
       record.errors.add(:value, :blank)
     end
   end
--- a/app/models/framework_response.rb
+++ b/app/models/framework_response.rb
@@ -9,6 +9,11 @@ class FrameworkResponse < VersionedModel
                         foreign_key: 'parent_id'
 
   belongs_to :parent, class_name: 'FrameworkResponse', optional: true
+  validates_each :value, on: :update do |record, _attr, _value|
+    if record.parent&.option_selected?(record.framework_question.dependent_value) && record.framework_question.required
+      record.errors.add(:value, :blank)
+    end
+  end
 
   def self.find_sti_class(type_name)
     case type_name
--- a/app/models/framework_response.rb
+++ b/app/models/framework_response.rb
@@ -10,23 +10,14 @@ class FrameworkResponse < VersionedModel
 
   belongs_to :parent, class_name: 'FrameworkResponse', optional: true
   validates_each :value, on: :update do |record, _attr, value|
-    if value.blank? && record.parent&.option_selected?(record.framework_question.dependent_value) && record.framework_question.required
-      record.errors.add(:value, :blank)
-    end
+    record.errors.add(:value, :blank) if requires_value?(value, record)
   end
 
-  def self.find_sti_class(type_name)
-    case type_name
-    when 'object'
-      type_name = 'FrameworkResponse::Object'
-    when 'string'
-      type_name = 'FrameworkResponse::String'
-    when 'array'
-      type_name = 'FrameworkResponse::Array'
-    when 'collection'
-      type_name = 'FrameworkResponse::Collection'
-    end
+  def self.requires_value?(value, record)
+    return false if value.present? || !record.framework_question.required
 
-    super
+    return true if record.parent.blank?
+
+    record.parent.option_selected?(record.framework_question.dependent_value)
   end
 end
--- a/app/models/framework_response.rb
+++ b/app/models/framework_response.rb
@@ -9,4 +9,19 @@ class FrameworkResponse < VersionedModel
                         foreign_key: 'parent_id'
 
   belongs_to :parent, class_name: 'FrameworkResponse', optional: true
+
+  def self.find_sti_class(type_name)
+    case type_name
+    when 'object'
+      type_name = 'FrameworkResponse::Object'
+    when 'string'
+      type_name = 'FrameworkResponse::String'
+    when 'array'
+      type_name = 'FrameworkResponse::Array'
+    when 'collection'
+      type_name = 'FrameworkResponse::Collection'
+    end
+
+    super
+  end
 end
--- a/app/models/framework_response/array.rb
+++ b/app/models/framework_response/array.rb
@@ -1,8 +1,8 @@
 class FrameworkResponse
   class Array < FrameworkResponse
     validates :value_text, absence: true
-    validates :value_json, presence: true, on: :update, if: -> { framework_question.required }
-    validates :value_json, on: :update, inclusion: { in: ->(response) { response.framework_question.options }, if: ->(response) { response.framework_question.options.any? } }
+    validates :value_json, presence: true, on: :update, if: -> { framework_question.required && parent.nil? }
+    validates :value_json, on: :update, inclusion: { in: :question_options }, if: :question_options
 
     def self.sti_name
       'array'
@@ -14,6 +14,16 @@ class FrameworkResponse
 
     def value=(answer)
       self.value_json = answer.presence || []
+    end
+
+    def option_selected?(option)
+      value.include?(option)
+    end
+
+  private
+
+    def question_options
+      framework_question.options.presence
     end
   end
 end
--- a/app/models/framework_response/array.rb
+++ b/app/models/framework_response/array.rb
@@ -1,0 +1,19 @@
+class FrameworkResponse
+  class Array < FrameworkResponse
+    validates :value_text, absence: true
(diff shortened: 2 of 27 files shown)


app/models/framework_response.rb
```diff
@@ -0,0 +1,32 @@
...
+  belongs_to :parent, class_name: 'FrameworkResponse', optional: true
+  validates_each :value, on: :update do |record, _attr, _value|
+    if record.parent&.option_selected?(record.framework_question.dependent_value) && record.framework_question.required
+      record.errors.add(:value, :blank)
+    end
+  end
+
+  def self.find_sti_class(type_name)
+    case type_name
+    when 'object'
```
cwrw, line 20: This is required because we are defining our own STI classes. Divorcing the class from the type name will be beneficial when exposing those values through the API to clients

app/models/framework_response.rb
```diff
@@ -0,0 +1,32 @@
...
+  validates :type, presence: true
+
+  belongs_to :framework_question
+  belongs_to :person_escort_record
+  has_many :dependents, class_name: 'FrameworkResponse',
+                        foreign_key: 'parent_id'
+
+  belongs_to :parent, class_name: 'FrameworkResponse', optional: true
+  validates_each :value, on: :update do |record, _attr, _value|
+    if record.parent&.option_selected?(record.framework_question.dependent_value) && record.framework_question.required
```
willfish, line 13: Is it possible for the framework_question to be nil?

app/models/framework_response.rb
```diff
@@ -0,0 +1,32 @@
...
+class FrameworkResponse < VersionedModel
+  validates :type, presence: true
+
+  belongs_to :framework_question
+  belongs_to :person_escort_record
+  has_many :dependents, class_name: 'FrameworkResponse',
+                        foreign_key: 'parent_id'
+
+  belongs_to :parent, class_name: 'FrameworkResponse', optional: true
+  validates_each :value, on: :update do |record, _attr, _value|
```
cwrw, line 12: Move this validation to respective STI classes as we need to check if value is present

app/models/framework_response.rb
cwrw: We can never have a response without a question or a person escort record associated

db/migrate/20200702081808_create_framework_responses.rb
```diff
@@ -0,0 +1,16 @@
+class CreateFrameworkResponses < ActiveRecord::Migration[6.0]
+  def change
+    create_table :framework_responses, id: :uuid do |t|
+      t.references :person_escort_record, type: :uuid, null: false, index: true, foreign_key: true
+      t.references :framework_question, type: :uuid, null: false, index: true, foreign_key: true
+      t.text :value_text
+      t.jsonb :value_json
+      t.string :type, null: false
+      t.references :parent, type: :uuid, index: true
```
cwrw, line 9: investigating implications on performance for batch endpoints, might rejig

app/models/framework_response.rb
```diff
@@ -0,0 +1,32 @@
...
+  validates :type, presence: true
+
+  belongs_to :framework_question
+  belongs_to :person_escort_record
+  has_many :dependents, class_name: 'FrameworkResponse',
+                        foreign_key: 'parent_id'
+
+  belongs_to :parent, class_name: 'FrameworkResponse', optional: true
+  validates_each :value, on: :update do |record, _attr, value|
+    if value.blank? && record.parent&.option_selected?(record.framework_question.dependent_value) && record.framework_question.required
```
smoothcontract, line 13: This expression is a little too complicated to readily understand. Perhaps this could be split slightly, something like this....

```suggestion
      record.errors.add(:value, :blank) if value.blank? && record.requires_value?

      ...
      
  def requires_value?
      parent&.option_selected?(framework_question.dependent_value) && framework_question.required
  end
```

smoothcontract: I'm unsure what the behaviour of this would be if `parent` is `nil` too. Perhaps worth making that more explicit - I think this is how it will work:

```ruby
def requires_value?
  return false if parent.nil?

  parent.option_selected?(framework_question.dependent_value) && framework_question.required
end
```

app/models/framework_response.rb
smoothcontract: Rather than overriding this - and also the corresponding `sti_name` method in each descendant class could this let Rails do its default thing and then add a regex or something to clean up the exposed names when it comes to the API part? I'm always suspicious of overriding default Rails behaviour without good reason.

app/models/framework_response/array.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+
+    def option_selected?(option)
+      value.include?(option)
+    end
+
+  private
+
+    def question_options
+      framework_question.options.presence
```
smoothcontract, line 26: Memoize this? It's called twice in the `validates` line...

app/models/framework_response/array.rb
```diff
@@ -0,0 +1,29 @@
+class FrameworkResponse
+  class Array < FrameworkResponse
+    validates :value_text, absence: true
+    validates :value_json, presence: true, on: :update, if: -> { framework_question.required && parent.nil? }
```
smoothcontract, line 4: Maybe worth adding a method for this (in `FrameworkResponse` model) as it's repeated in a couple of these derived classes. Perhaps override this in descendant classes if needed?

```ruby
def requires_json_value?
  framework_question.required && parent.nil?
end
```

app/models/framework_response/details_collection.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+
+    def to_a
+      collection
+    end
+
+  private
+
+    def details_objects
+      collection.each do |item|
```
smoothcontract, line 28: Maybe use `collection.any?` instead...

```suggestion
      errors.add(:collection, 'One of more details objects is invalid') if collection.any?(&:invalid)
```

app/models/framework_response/collection.rb
```diff
@@ -0,0 +1,51 @@
...
+      )
+    end
+
+    def validate_presence
+      errors.add(:value_json, :blank) if value.empty?
+    end
+
+    def validate_details_collection
+      # TODO: Add proper validation messages
+      errors.add(:value, 'Details collection is invalid') if details_collection(value).invalid?
```
smoothcontract, line 44: Could these error messages be moved out to the locale file? Not essential but it's nice to keep the models free of these if possible.

app/models/framework_response.rb
```diff
@@ -0,0 +1,32 @@
...
+      record.errors.add(:value, :blank)
+    end
+  end
+
+  def self.find_sti_class(type_name)
```
willfish, line 18: Since this is implicitly overriding the method that lives in `ActiveRecord::Inheritance` I'd add a comment to save people time in understanding this.

`NB: Overrides mechanism used by ActiveRecord to instantiate objects in STI` or equivalent

app/models/framework_response/collection.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    def self.sti_name
+      'collection'
+    end
+
+    def value
+      value_json.presence || []
+    end
+
+    def value=(answer)
```
willfish, line 15: Feels weird mixing response/answer concepts, here. What about killing the word answer in this new implementation are just using response?

willfish:
```ruby
    def value=(response_value)
```

app/models/framework_response/collection.rb
willfish: I was wondering this. I think we shouldn't since we don't support localisation at all currently and we'd be creating new patterns in the app that don't currently exist

app/models/framework_response/collection.rb
```diff
@@ -0,0 +1,51 @@
+class FrameworkResponse
+  class Collection < FrameworkResponse
+    validates :value_text, absence: true
+    validate :validate_presence, on: :update, if: -> { framework_question.required && parent.nil? }
```
willfish, line 4:
```suggestion
    validate :validate_value_presence, on: :update, if: -> { framework_question.required && parent.nil? }
```

app/models/framework_response/collection.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    def details_collection(collection)
+      DetailsCollection.new(
+        collection: collection,
+        question_options: framework_question.options,
+        details_options: framework_question.followup_comment_options,
+      )
+    end
+
+    def validate_presence
```
willfish, line 38:
```suggestion
    def validate_value_presence
```

willfish:
So response is definitely a more loaded term than answer in the context of our app. We use response literally everywhere in the code to refer to the response of different apis. How do you feel about a rename of response to answer (or maybe result 🤢 )? I know it's a PITA, but I can imagine this causing some confusion in various conversations with the team moving forward, especially for new people/people not familiar with our code base.

app/models/framework_response.rb
cwrw: Removed custom STI names

app/models/framework_response.rb
cwrw: This implementation has now been removed

app/models/framework_response.rb
cwrw: I've tidied this up and moved it into a tested method

app/models/framework_response/array.rb
cwrw: I've moved all required validations to parent class

app/models/framework_response/collection.rb
cwrw: updated

app/models/framework_response/collection.rb
cwrw: I've gone with `raw_value`, let me know your thoughts

app/models/framework_response/collection.rb
cwrw: moved this validation to parent class as it was repeated everywhere

app/models/framework_response/collection.rb
cwrw: moved validation to parent class

app/models/framework_response/collection.rb
```diff
@@ -0,0 +1,41 @@
...
+    def value
+      value_json.presence || []
+    end
+
+    def value=(raw_value)
+      self.value_json =
+        if response_details
+          details_collection(raw_value).to_a
+        else
+          self.value_json = raw_value.presence || []
```
smoothcontract, line 15: Don't need `self.value_json` assignment here as it's covered on line 11.

PR #692, opened 2020-07-08
Add support for new video_remand_hearing move_type

smoothcontract:
### Jira link

P4-1816

### What?

- [x] Added new `video_remand_hearing` to `move_type` enum
- [x] Allow video remand hearing moves to be created without specifying a `to_location`
- [x] Enforce validation of `police` type for `from_location` when creating `video_remand_hearing`
- [x] Add new `MoveTypeValidator` class

### Why?

- This is the first of many new move types to be added, each introducing additional validation rules on the permitted from and to location types. As the first step to implementing support for these rules, validation of move type and associated location types is extracted to a new custom validator; this allows the logic to be shared across the move model and move runner.

### Have you? (optional)

- [x] Updated API docs if necessary - this will be updated further in future PR's to document the behaviour of the different move types

- Adds support for new move type, but does not change behaviour of existing move types

### DIFF
--- a/app/services/move_events/params_validator.rb
+++ b/app/services/move_events/params_validator.rb
@@ -11,7 +11,7 @@ module MoveEvents
     # TODO: in the future when pluralising types, existing move event records will need updating from singular types to plural types.
     ACCEPTABLE_PLURAL_VERBS = %w[cancels completes approves rejects].freeze
 
-    attr_reader :timestamp, :type, :date, :cancellation_reason, :rejection_reason, :from_location, :to_location
+    attr_reader :move, :timestamp, :type, :date, :cancellation_reason, :rejection_reason, :from_location, :to_location
 
     validates :type, presence: true, inclusion: { in: %w[accepts approve cancel complete lockouts redirects reject starts] }
     validates :cancellation_reason, inclusion: { in: Move::CANCELLATION_REASONS }, if: -> { type == 'cancel' }
@@ -34,8 +34,12 @@ module MoveEvents
 
     validates_with LocationValidator, locations: [:from_location], if: -> { type == 'lockouts' }
     validates_with LocationValidator, locations: [:to_location], if: -> { type == 'redirects' }
+    validates_with Moves::MoveTypeValidator, if: -> { type == 'redirects' }
 
-    def initialize(params)
+    delegate :move_type, to: :move
+
+    def initialize(move, params)
+      @move = move
       @timestamp = params.dig(:attributes, :timestamp)
       @type = params[:type]
       @date = params.dig(:attributes, :date)
--- a/app/services/move_events/params_validator.rb
+++ b/app/services/move_events/params_validator.rb
@@ -11,7 +11,7 @@ module MoveEvents
     # TODO: in the future when pluralising types, existing move event records will need updating from singular types to plural types.
     ACCEPTABLE_PLURAL_VERBS = %w[cancels completes approves rejects].freeze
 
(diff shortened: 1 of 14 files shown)


app/services/moves/move_type_validator.rb
```diff
@@ -0,0 +1,29 @@
...
+    def includes?(move_types)
+      move_types.include?(record.move_type.to_s)
+    end
+
+    def human_move_type
+      record.move_type.to_s.humanize(capitalize: false)
+    end
+
+    def validate_police_from_location
+      record.errors.add(:from_location, "must be a police location for #{human_move_type}") unless record.from_location&.police?
```
alexdesi, line 26: maybe I am wrong, when we add an error, we should specify
- the field,
- the type of error, for example :too_long and 
- the message, 
here the type is missing 🤔

swagger/v1/move.yaml
```diff
@@ -44,6 +44,7 @@ Move:
           - prison_recall
           - prison_transfer
           - police_transfer
+          - video_remand_hearing
```
willfish, line 47: Presumably we want this on v2 moves, too?

willfish:
I foresee that validator becoming complex and horrible ☠️

app/services/moves/move_type_validator.rb
smoothcontract: Not quite - the second argument can be a string or a symbol to be resolved in the locale file. We have plenty of examples of similar things, but I'll add an explicit test to check the returned error message text.

swagger/v1/move.yaml
smoothcontract: GPWM - I'll update the v2 docs too 👍

smoothcontract:
> I foresee that validator becoming complex and horrible ☠️ 🔮

You could well be right, but so far I'm only seeing the need to validate against from and to locations, so at least it's all in one place. We can erect suitable warning signs and barbed wire fences as needed.

app/services/move_events/params_validator.rb
```diff
@@ -32,17 +32,21 @@ class ParamsValidator
...
     end
 
-    validates_with LocationValidator, locations: [:from_location], if: -> { type == 'lockouts' }
-    validates_with LocationValidator, locations: [:to_location], if: -> { type == 'redirects' }
+    validates_with LocationValidator, locations: [:from_location_id], if: -> { type == 'lockouts' }
+    validates_with LocationValidator, locations: [:to_location_id], if: -> { type == 'redirects' }
+    validates_with Moves::MoveTypeValidator, if: -> { type == 'redirects' }
 
-    def initialize(params)
+    delegate :move_type, :from_location, to: :move
```
martyn-w, line 39: I think these names could potentially be confusing:
```
from_location --> move's from_location
from_location_id --> event's from_location
```

how about:
```
move_from_location --> move's from_location
event_from_location --> event's from_location(_id)
```
?

martyn-w:
Looks good so far.
1. I think the naming of `from_location` vs `from_location_id` is confusing in the move event; consider a different name instead. Sorry I didn't pick up the full implication of this earlier.
2. Add a test for move_type_validator - its a bit fidley as you'll need a new class, see e.g. spec/services/location_validator_spec.rb

app/services/move_events/params_validator.rb
smoothcontract: Discussed offline, this isn't trivial to change as the `MoveTypeValidator` needs to believe that this is really a `Move` model with the same `move_type` and `from_location` attributes. Agreed to have another look at that in the next PR for the other upcoming move types. In the meantime this is enough to unblock front end work.

PR #693, opened 2020-07-08
Prevent V2 patch move to update from_location and to_location

alexdesi:
### Jira link

P4-1904-prevent-updating-from-to

### What?
Fix the v2 move endpoint to ensure that from_location and to_location cannot be updated.

### Why?
This functionality was overlooked when creating the V2 endpoint. We still need it for V2.

### DIFF
--- a/app/controllers/api/v2/moves_actions.rb
+++ b/app/controllers/api/v2/moves_actions.rb
@@ -25,18 +25,13 @@ module Api::V2
 
     def update_and_render
       raise ActiveRecord::ReadOnlyRecord, "Can't change moves coming from Nomis" if move.from_nomis?
-      q
+
       @move.assign_attributes(update_move_attributes)
-
-      def notify
-        if @move.valid?
-          action_name = @move.status_changed? ? 'update_status' : 'update'
-          Notifier.prepare_notifications(topic: @move, action_name: action_name)
-        end
-      end
-
+      action_name = @move.status_changed? ? 'update_status' : 'update'
       @move.save!
       @move.allocation&.refresh_status_and_moves_count!
+
+      Notifier.prepare_notifications(topic: @move, action_name: action_name)
 
       render_move(@move, :ok)
     end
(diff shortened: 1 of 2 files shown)


spec/requests/api/moves_controller_update_v2_spec.rb
```diff
@@ -184,6 +185,35 @@
...
+            status: 'requested',
+          },
+          relationships: {
+            from_location: { data: { type: 'locations', id: new_from_location.id } },
+            to_location: { data: { type: 'locations', id: new_to_location.id } },
+          },
+        }
+      end
+
+      it 'does not affect from_location' do
```
alexdesi, line 205:
```suggestion
      it 'does not affect both from_location and to_location' do
```

spec/requests/api/moves_controller_update_v2_spec.rb
```diff
@@ -184,6 +185,35 @@
...
+            from_location: { data: { type: 'locations', id: new_from_location.id } },
+            to_location: { data: { type: 'locations', id: new_to_location.id } },
+          },
+        }
+      end
+
+      it 'does not affect both from_location and to_location' do
+        do_patch
+
+        move.reload
```
smoothcontract, line 208: I think you can remove these two lines? .. or is there a reason why the endpoint is called twice?

app/controllers/api/v2/moves_actions.rb
```diff
@@ -62,10 +62,15 @@ def move_params
...
     def move_attributes
-      move_params[:attributes].tap do |attributes|
-        attributes[:profile] = profile unless profile_attributes.nil?
+      update_move_attributes.tap do |attributes|
         attributes[:from_location] = from_location unless from_location_attributes.nil?
         attributes[:to_location] = to_location unless to_location_attributes.nil?
+      end
+    end
+
+    def update_move_attributes
```
martyn-w, line 71: how about `def common_move_attributes` ?

spec/requests/api/moves_controller_update_v2_spec.rb
alexdesi: ops, I thought I've removed it,
good spot thanks!

alexdesi, line 208:
```suggestion
```

app/controllers/api/v2/moves_actions.rb
alexdesi: yes, I agree 👍

smoothcontract: Good idea, I like that too 👍

PR #694, opened 2020-07-08
[P4-1878] Set supplier when creating a singleton move

willfish:
### Jira link

P4-1878

### What?

We're moving to setting a supplier when we create singleton and allocations `Move`.

This PR adds support for transparently inferring the supplier when **creating a singleton** `Move` via the doorkeeper application owner.

### Why?

Multiple suppliers will have the same location so the `Move#from_location#suppliers` mechanism is no longer sufficient for representing ownership on the move.

### DIFF
--- a/app/services/supplier_chooser.rb
+++ b/app/services/supplier_chooser.rb
@@ -1,7 +1,7 @@
 class SupplierChooser
-  def initialize(doorkeeper_application_owner, move)
+  def initialize(doorkeeper_application_owner, from_location)
     @doorkeeper_application_owner = doorkeeper_application_owner
-    @move = move
+    @from_location = from_location
   end
 
   def call
@@ -9,6 +9,6 @@ class SupplierChooser
 
     # TODO: Replace with the frontend supplied supplier
     #       There can only ever be one supplier per location, currently
-    @move.from_location.suppliers.first
+    @from_location.suppliers.first
   end
 end
--- a/app/services/supplier_chooser.rb
+++ b/app/services/supplier_chooser.rb
@@ -1,0 +1,14 @@
+class SupplierChooser
+  def initialize(doorkeeper_application_owner, move)
+    @doorkeeper_application_owner = doorkeeper_application_owner
+    @move = move
+  end
+
+  def call
+    return @doorkeeper_application_owner if @doorkeeper_application_owner
+
+    # TODO: Replace with the frontend supplied supplier
+    #       There can only ever be one supplier per location, currently
+    @move.from_location.suppliers.first
+  end
+end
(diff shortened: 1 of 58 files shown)


app/services/supplier_chooser.rb
```diff
@@ -0,0 +1,14 @@
+class SupplierChooser
```
alexdesi, line 1: maybe `SupplierSelector` is more appropriate?

actually I am not even sure about `SupplierSelector`, it just takes the first supplier.

spec/services/supplier_chooser_spec.rb
```diff
@@ -0,0 +1,25 @@
+RSpec.describe SupplierChooser do
+  subject(:service) { described_class.new(doorkeeper_application_owner, move) }
+
+  let(:move) { instance_double('Move', from_location: from_location) }
+  let(:from_location) { instance_double('Location', suppliers: suppliers) }
+  let(:suppliers) { [supplier_one, supplier_two] }
+  let(:supplier_one) { instance_double('Supplier') }
+  let(:supplier_two) { instance_double('Supplier') }
+
```
alexdesi, line 9: I was wondering if we could simplify this, and test with real objects, with a couple of factories, instead of 5 instance doubles.

app/services/supplier_chooser.rb
willfish: Selector makes me think of xml/html selections 🤔

spec/services/supplier_chooser_spec.rb
willfish: So I very specifically used instance doubles as we use factories/create rows in pretty much every test we write and this service felt like it definitely didn't need any complex functionality that's provided by ActiveRecord 🤔 .

willfish: Reduced the number of doubles

app/services/supplier_chooser.rb
alexdesi: 👍

willfish:
```
 Running tests in:

 Cancel allocation
 ✓ Cancel allocation

 New PMU allocation

 ✓ With unfound person
 ✓ With existing person

 ✓ With a new person

 ✓ Reason - `Made in error`
 ✓ Reason - `Another reason`

 New move from Prison to Court
 ✓ With unfound person
 ✓ With existing person

 ✓ With existing person

 Download moves
 ✓ As Police user
 ✓ As Prison user
 ✓ As STC user
 ✓ As Supplier user

 Smoke tests
 ✓ As Police user
 ✓ As STC user
 ✓ As Prison user
 ✓ As Supplier user
 ✓ As OCA user


 21 passed (4m 35
```

PR #695, opened 2020-07-08
[P4-1906] Fix v2 profile serialization

willfish:
### Jira link

P4-<TODO>

### What?

I have added/removed/altered:

- [x] Added v2 profile serializer and specs
- [x] Altered v2 move serializer to use v2 profile serializer

### Why?

We need to conform to the swagger documentation for v2 and make sure that we're not coupling to v1 people.

### DIFF
--- a/spec/serializers/v2/profile_serializer_spec.rb
+++ b/spec/serializers/v2/profile_serializer_spec.rb
@@ -88,4 +88,3 @@ RSpec.describe V2::ProfileSerializer do
     end
   end
 end
-
--- a/spec/serializers/v2/profile_serializer_spec.rb
+++ b/spec/serializers/v2/profile_serializer_spec.rb
@@ -99,4 +99,3 @@ RSpec.describe V2::ProfileSerializer do
     end
   end
 end
-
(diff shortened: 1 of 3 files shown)


spec/serializers/v2/profile_serializer_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+          type: 'profiles',
+          attributes: { assessment_answers: [] },
+          relationships: {
+            person: {
+              data: { id: profile.person.id, type: 'people' },
+            },
+            documents: { data: [{ id: profile.documents.first.id, type: 'documents' }] },
+          },
+        },
+        included: [{ id: profile.person.id, type: 'people' }, { id: profile.documents.first.id, type: 'documents' }],
```
martyn-w, line 80: could you include an attribute in this test that is only provided by the v2::person_serailizer, to confirm that it is a v2 person and not a v1 person?

willfish: Sure. Sorted in a recent commit.

PR #696, opened 2020-07-09
Prevent access to Nomis when assessment_answers are passed.

alexdesi:
### Jira link

https://dsdmoj.atlassian.net/browse/P4-1881

### What?

If the `assessment_answers` param is not present AND the person's profile has a `prison_number`
THEN the new profile will be updated with Alerts and Personal Care from NOMIS.

I have added/removed/altered:

the endpoint POST /people/:id/profiles

### Why?
This is because an individual may have been out (released) for a while and when they are remanded at court NOMIS would be out of date

### Have you? (optional)

### DIFF
--- a/spec/swagger/swagger_doc_v1.yaml
+++ b/spec/swagger/swagger_doc_v1.yaml
@@ -2283,10 +2283,10 @@
   "/people/{person_id}/profiles":
     post:
       summary: Creates a new profile
-      description: "This creates a new profile for a person. If the person has a
-        prison_number, the new profile is updated with Alerts and Personal Care
-        Needs from NOMIS.
-
+      description: "This creates a new profile for a person.<br>
+        Note: this endpoint accesses to Nomis:<br>
+        If the `assessment_answers` param is not present AND the person's profile has a `prison_number`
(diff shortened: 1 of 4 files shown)


spec/swagger/swagger_doc_v1.yaml
```diff
@@ -2283,10 +2283,10 @@
   "/people/{person_id}/profiles":
     post:
       summary: Creates a new profile
-      description: "This creates a new profile for a person. If the person has a
-        prison_number, the new profile is updated with Alerts and Personal Care
-        Needs from NOMIS.
-
+      description: "This creates a new profile for a person.<br>
+        Note: this endpoint accesses to Nomis:<br>
```
willfish, line 2287:
```suggestion
        Note: this endpoint retrieves Assessment Answers from Nomis:<br>
```

spec/swagger/swagger_doc_v1.yaml
```diff
@@ -2283,10 +2283,10 @@
...
+        If the `assessment_answers` param is not present AND the person's profile has a `prison_number`
```
willfish, line 2288:
```suggestion
        If the `assessment_answers` param is not present AND the person has a `prison_number`
```

PR #699, opened 2020-07-10
P4-1761 POST /person_escort_records endpoint

cwrw:
### Jira link

[P4-1761](https://dsdmoj.atlassian.net/browse/P4-1761)

### What?

- [x] Removed unstarted status from the PER state enums
- [x] Added ability to build framework responses from questions
- [x] Added ability to build a new PER from a framework and Profile
- [x] Added controller and new routes for POSTing the PER
- [x] Added swagger documentation for new endpoint

### Why?

To allow clients to create a new PER, add POST `/person_escort_records` endpoint which accepts a profile, and a version of the framework to build empty responses for all questions related to a framework. If the version of the framework is not passed in use the latest framework (versioning is going to be revised, this is only temporary).

To bulk create a PER record, active record import gem has been added. This allows us passing in an actual object rather than a hash, and helps with deeply nested record creation. However to associate a PER record to deeply nested responses, we have had to create it first then associate them after. 

### Have you? (optional)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,19 +16,17 @@ group :production do
   gem 'appinsights', github: 'ministryofjustice/appinsights'
 end
 
+gem 'activerecord-import', '~> 1.0', '>= 1.0.5'
 gem 'aws-sdk-s3', require: false
 gem 'bcrypt', require: false
 gem 'bootsnap', '>= 1.1.0', require: false
 gem 'cancancan'
-# explicit soft-deletes
 gem 'discard'
 gem 'doorkeeper'
 gem 'elastic-apm'
 gem 'faraday'
 gem 'finite_machine'
 gem 'govuk_notify_rails', '~> 2.1.2'
-# static page serving for extra API documentation
-gem 'activerecord-import', '~> 1.0', '>= 1.0.5'
 gem 'json-schema'
 gem 'kaminari'
 gem 'net-http-persistent'
--- a/Gemfile
+++ b/Gemfile
@@ -28,6 +28,7 @@ gem 'faraday'
 gem 'finite_machine'
 gem 'govuk_notify_rails', '~> 2.1.2'
 # static page serving for extra API documentation
+gem 'activerecord-import', '~> 1.0', '>= 1.0.5'
 gem 'json-schema'
 gem 'kaminari'
 gem 'net-http-persistent'
--- a/app/models/framework_question.rb
+++ b/app/models/framework_question.rb
@@ -34,7 +34,7 @@ class FrameworkQuestion < VersionedModel
     response
   end
 
-  private
+private
 
   def build_response(question, person_escort_record)
     klass =
--- a/app/models/framework_question.rb
+++ b/app/models/framework_question.rb
@@ -17,7 +17,36 @@ class FrameworkQuestion < VersionedModel
                         foreign_key: 'parent_id'
   belongs_to :parent, class_name: 'FrameworkQuestion', optional: true
 
-
   has_many :flags
   has_many :framework_responses
+
+  def build_responses(question: self, person_escort_record:)
+    response = build_response(question, person_escort_record)
+    return response unless question.dependents.any?
+
+    question.dependents.find_each do |dependent_question|
+      dependent_response = build_responses(question: dependent_question, person_escort_record: person_escort_record)
+      dependent_response_values = dependent_response.slice(:type, :framework_question_id, :dependents, :person_escort_record)
+
+      response.dependents.build(dependent_response_values)
+    end
+
+    response
+  end
+
+  private
+
+  def build_response(question, person_escort_record)
+    klass =
+      case question.question_type
+      when 'radio'
+        question.followup_comment ? FrameworkResponse::Object : FrameworkResponse::String
+      when 'checkbox'
+        question.followup_comment ? FrameworkResponse::Collection : FrameworkResponse::Array
+      else
+        FrameworkResponse::String
+      end
+
+    klass.new(framework_question_id: question.id, person_escort_record: person_escort_record)
+  end
 end
--- a/app/models/framework_question.rb
+++ b/app/models/framework_question.rb
@@ -15,8 +15,9 @@ class FrameworkQuestion < VersionedModel
   belongs_to :framework
   has_many :dependents, class_name: 'FrameworkQuestion',
                         foreign_key: 'parent_id'
-
   belongs_to :parent, class_name: 'FrameworkQuestion', optional: true
+
+
   has_many :flags
   has_many :framework_responses
 end
--- a/app/models/framework_question.rb
+++ b/app/models/framework_question.rb
@@ -20,14 +20,14 @@ class FrameworkQuestion < VersionedModel
   has_many :flags
   has_many :framework_responses
 
-  def build_responses(question: self, person_escort_record:)
+  def build_responses(question: self, person_escort_record:, questions:)
     response = build_response(question, person_escort_record)
-    return response unless question.dependents.any?
+    return response if question.dependents.empty?
 
-    question.dependents.find_each do |dependent_question|
-      dependent_response = build_responses(question: dependent_question, person_escort_record: person_escort_record)
+    question.dependents.each do |dependent_question|
+      # NB: to avoid extra queries use original set of questions
+      dependent_response = build_responses(question: questions[dependent_question.id], person_escort_record: person_escort_record, questions: questions)
       dependent_response_values = dependent_response.slice(:type, :framework_question_id, :dependents, :person_escort_record)
-
       response.dependents.build(dependent_response_values)
     end
 
--- a/app/models/person_escort_record.rb
+++ b/app/models/person_escort_record.rb
@@ -27,8 +27,11 @@ class PersonEscortRecord < VersionedModel
     ActiveRecord::Base.transaction do
       save!
 
-      framework_questions.includes(:parent, :dependents).where(framework_question_parents: { parent_id: nil }).find_each do |question|
-        response = question.build_responses(person_escort_record: self)
+      questions = framework_questions.includes(:dependents).index_by(&:id)
+      questions.values.each do |question|
+        next unless question.parent_id.nil?
+
+        response = question.build_responses(person_escort_record: self, questions: questions)
         framework_responses.build(response.slice(:type, :framework_question_id, :dependents))
       end
 
--- a/app/models/person_escort_record.rb
+++ b/app/models/person_escort_record.rb
@@ -7,6 +7,9 @@ class PersonEscortRecord < VersionedModel
 
   validates :state, presence: true, inclusion: { in: states }
   has_many :framework_responses, dependent: :destroy
+
   belongs_to :framework
+  has_many :framework_questions, through: :framework
   belongs_to :profile
+  validates :profile, uniqueness: true
 end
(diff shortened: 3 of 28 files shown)


app/models/person_escort_record.rb
```diff
@@ -1,13 +1,40 @@
...
+    # TODO: add state machine
+    record = new(profile: profile, framework: framework, state: 'in_progress')
+    record.build_responses!
+  end
+
+  def build_responses!
+    ActiveRecord::Base.transaction do
+      save!
+
+      framework_questions.includes(:dependents).where(framework_question_parents: { parent_id: nil }).find_each do |question|
```
cwrw, line 30: fix query to avoid n+1 and to include parents

app/models/framework_question.rb
```diff
@@ -15,8 +15,38 @@ class FrameworkQuestion < VersionedModel
   belongs_to :framework
   has_many :dependents, class_name: 'FrameworkQuestion',
                         foreign_key: 'parent_id'
-
   belongs_to :parent, class_name: 'FrameworkQuestion', optional: true
+
   has_many :flags
   has_many :framework_responses
+
+  def build_responses(question: self, person_escort_record:, questions:)
```
cwrw, line 23: maybe this method can live in `person_escort_record` class to avoid passing around `questions`/`person_escort_record` 🤔

app/serializers/person_escort_record_serializer.rb
```diff
@@ -0,0 +1,25 @@
+class PersonEscortRecordSerializer < ActiveModel::Serializer
+  belongs_to :profile, record_type: :profile
+  belongs_to :framework, record_type: :framework
+  has_many :framework_responses, serializer: FrameworkResponseSerializer, key: :responses
+
+  attributes :version, :status
+
+  def status
+    object.state
```
smoothcontract, line 9: Why is there a status and a state attribute, or is this just remapping things to the API view of the model?

If this is intentional, then perhaps use a more idiomatic AMS approach:

```
attribute :status, key: :state
```

app/serializers/person_escort_record_serializer.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  def version
+    object.framework.version
```
smoothcontract, line 13: Would prefer to add this delegation in the `PersonEscortRecord` model instead

app/serializers/person_escort_record_serializer.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  def framework_responses
+    object.framework_responses.includes(framework_question: :framework)
```
smoothcontract, line 17: Is there a way to move this to the `PersonEscortRecord` model? It feels slightly strange with the serializer resolving potential performance issues 🤔

spec/models/person_escort_record_spec.rb
```diff
@@ -4,8 +4,222 @@
...
+  end
+
+  describe '#build_responses' do
+    it 'persists the person_escort_record' do
+      framework = create(:framework)
+      create(:framework_question, framework: framework)
+      profile = create(:profile)
+      person_escort_record = build(:person_escort_record, framework: framework, profile: profile)
+
+      expect { person_escort_record.build_responses! }.to change(described_class, :count).from(0).to(1)
```
smoothcontract, line 112: Might be better to use `.by(1)` for increased robustness

spec/models/person_escort_record_spec.rb
```diff
@@ -4,8 +4,222 @@
...
+
+    it 'allows access to framework question through person escort record' do
+      profile = create(:profile)
+      framework = create(:framework)
+      create(:framework_question, :checkbox, framework: framework)
+      person_escort_record = described_class.save_with_responses!(profile_id: profile.id)
+
+      questions = person_escort_record.framework_questions
+
+      expect(questions.count).to eq(1)
```
smoothcontract, line 101: I think it's fine to combine these two lines and is slightly easier to read

```suggestion
      expect(person_escort_record.framework_questions.count).to eq(1)
```

spec/models/person_escort_record_spec.rb
```diff
@@ -4,8 +4,222 @@
...
+    end
+
+    it 'creates responses for multiple questions' do
+      framework = create(:framework)
+      create(:framework_question, framework: framework)
+      create(:framework_question, :checkbox, framework: framework)
+      profile = create(:profile)
+      person_escort_record = build(:person_escort_record, framework: framework, profile: profile)
+
+      expect { person_escort_record.build_responses! }.to change(FrameworkResponse, :count).from(0).to(2)
```
smoothcontract, line 136: `.by(2)` instead?

spec/requests/api/person_escort_records_controller_create_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+      }
+    end
+
+    before do
+      post '/api/v1/person_escort_records', params: person_escort_record_params, headers: headers, as: :json
+    end
+
+    context 'when successful' do
+      let(:application) { create(:application, owner: supplier) }
+      let(:access_token) { create(:access_token, application: application).token }
```
smoothcontract, line 40: Is there a need to use a real access token here? If a spoofed token will work then that's preferable for faster tests. Generally we only need to use a real token if authorisation depends on the supplier access to models. Swapping to a spoofed token also means no need to create the application or supplier records.

spec/requests/api/person_escort_records_controller_create_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+      let(:schema) { load_yaml_schema('post_person_escort_record_responses.yaml') }
+      let(:data) do
+        {
+          "id": PersonEscortRecord.first&.id,
```
smoothcontract, line 44: Perhaps `PersonEscortRecord.last` is more robust? Are there any circumstances where the record would not be created? I'm wondering if the `&` is strictly necessary.

spec/requests/api/person_escort_records_controller_create_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+            "framework": {
+              "data": {
+                "id": framework.id,
+                "type": 'frameworks',
+              },
+            },
+            "responses": {
+              "data": [
+                {
+                  "id": FrameworkResponse.first&.id,
```
smoothcontract, line 66: As above

swagger/v1/framework_response_reference.yaml
```diff
@@ -0,0 +1,27 @@
+FrameworkResponseReference:
+  type: object
+  required:
+    - data
+  properties:
+    data:
+      oneOf:
+      - type: object
+      - type: 'null'
+      - type: array
```
smoothcontract, line 10: An array of what? May be worth specifying the type of array members.

swagger/v1/person_escort_record.yaml
```diff
@@ -0,0 +1,62 @@
...
+      type: string
+      example: person_escort_records
+      enum:
+      - person_escort_records
+      description: The type of this object - always `person_escort_records`
+    attributes:
+      type: object
+      required:
+      - version
+      - status
```
smoothcontract, line 23: Hmm.. do we really want the POST PER endpoint to require status to be sent? We've moved away from that pattern with moves and now have explicit event endpoints to change status. Maybe the status should be read only?

app/serializers/person_escort_record_serializer.rb
cwrw: Yea it's just remapping things for the API view, I liked keeping state on the model. I'll amend to use key instead.

app/serializers/person_escort_record_serializer.rb
cwrw: This is a temporary implementation till we finalise the way we accept/persist versions. In the future this will be persisted so didn't want to add too much before we flushed that out soon and just wanted to return a value for now

app/serializers/person_escort_record_serializer.rb
cwrw: maybe as a scope? Although it would be somewhat similar just calling on the scope instead. I'm wary of making this the default scope. I tried having a look at optimisations and this seemed like the cleanest way, let me know your thoughts

swagger/v1/framework_response_reference.yaml
cwrw: so this reference could be an array of responses or a single response depending on the relationship. But not sure how I could add it as a nested item as its referencing itself?

app/serializers/person_escort_record_serializer.rb
smoothcontract: Fair enough 👍

app/serializers/person_escort_record_serializer.rb
smoothcontract: A scope would be slightly nicer - it's really just frustration with AMS I suppose. The way they work under the hood really isn't performant for relationships and it's good to try and address the inevitable N+1 issues. We need to look at a better gem at some point. A scope would be nice but it's not necessary to add right now, particularly if the design for this is still in flux.

swagger/v1/framework_response_reference.yaml
smoothcontract: Erm... maybe note something to that effect in the description then?

Gemfile
```diff
@@ -21,13 +21,13 @@ gem 'bcrypt', require: false
 gem 'bootsnap', '>= 1.1.0', require: false
 gem 'cancancan'
 # explicit soft-deletes
```
smoothcontract, line 23: this comment can be removed too - it relates to the `discard` gem but I really don't see much value in the comment. Apologies it's unrelated to your PR but the placement of the new gem makes it really apparent!

app/models/framework_question.rb
smoothcontract: That's worth looking at if it makes the implementation cleaner. This already violates the single responsibility slightly by building a bunch of responses inside the question class... not that it matters, so unless the PER class is already doing lots of things it wouldn't be any worse to move it. If we wanted to get really OOP and gang of five then it could be moved to a service so that it just does one thing but it feels like overkill to do that.

spec/models/person_escort_record_spec.rb
```diff
@@ -4,8 +4,220 @@
...
+    it 'creates responses for framework questions' do
+      profile = create(:profile)
+      framework = create(:framework)
+      create(:framework_question, :checkbox, framework: framework)
+      create(:framework_question, :checkbox, framework: framework)
+      person_escort_record = described_class.save_with_responses!(profile_id: profile.id)
+
+      responses = person_escort_record.framework_responses
+
+      expect(responses.count).to eq(2)
```
smoothcontract, line 76: Could combine these two lines

swagger/v1/person_escort_record.yaml
```diff
@@ -0,0 +1,61 @@
...
+      - version
+      properties:
+        status:
+          example: in_progress
+          type: string
+          enum:
+            - in_progress
+            - completed
+            - confirmed
+          description: Determines the current status of the `person_escort_record`
```
smoothcontract, line 31: Should this be tagged as `readOnly` attribute?

cwrw: no we will be able to mark something as completed later on

swagger/v1/framework_response_reference.yaml
cwrw: done

swagger/v1/person_escort_record.yaml
smoothcontract: Okay, but I'm hoping that will be via a dedicated `POST /per/:id/complete` endpoint ;-)

PR #700, opened 2020-07-10
Small optimisation to dockerfile

martyn-w:
### Jira link

P4-<TODO>

### What?
Small optimisation of the dockerfile

I have added/removed/altered:

- [x] install the gems before coping in the app

### Why?

I am doing this because:

- so that it can build faster (the gem installation can be cached even for code changes)

- hopefully OK

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,6 @@ RUN bundle install --jobs 4 --retry 3 \
      && find /usr/local/bundle/gems/ -name "*.c" -delete \
      && find /usr/local/bundle/gems/ -name "*.o" -delete
 
-COPY . /app
 ############### End of Build step ###############
 FROM ruby:2.7.0-alpine
 
@@ -44,9 +43,8 @@ RUN addgroup -g $APPUID -S appgroup && \
 RUN apk add --update --no-cache tzdata postgresql-dev
 
 WORKDIR /app
-COPY --chown=appuser:appgroup --from=build-stage /app /app
 COPY --chown=appuser:appgroup --from=build-stage /usr/local/bundle /usr/local/bundle
+COPY --chown=appuser:appgroup . /app
 
 USER $APPUID
 CMD ["./run.sh"]
-
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,14 +8,16 @@ ENV BUNDLE_FROZEN="true"
 
 WORKDIR /app
 RUN apk --update --no-cache add git build-base postgresql-dev
-
-COPY . /app
 RUN gem update bundler --no-document
+
+# NB: its more efficient not to copy the full app folder until after the gems are installed (reduces unnecessary rebuilds)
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -8,14 +8,16 @@ ENV BUNDLE_FROZEN="true"
 
 WORKDIR /app
 RUN apk --update --no-cache add git build-base postgresql-dev
-
-COPY . /app
 RUN gem update bundler --no-document
+
+# NB: its more efficient not to copy the full app folder until after the gems are installed (reduces unnecessary rebuilds)
+COPY Gemfile Gemfile.lock .ruby-version /app/
```
mafonso, line 14: Thanks!  :)

Dockerfile
```diff
@@ -8,14 +8,16 @@ ENV BUNDLE_FROZEN="true"
...
 RUN bundle install --jobs 4 --retry 3 \
      && rm -rf /usr/local/bundle/cache/*.gem \
      && find /usr/local/bundle/gems/ -name "*.c" -delete \
      && find /usr/local/bundle/gems/ -name "*.o" -delete
 
+COPY . /app
```
mafonso, line 20: If this is not needed on the build stage, would it make sense to copy it straight into the runtime image?

Dockerfile
mafonso: Having a bit more though of why this is.   If we are building our gems in `/usr/local/bundle` and not in  `app/vendor` would it make a difference? 

I did test both cases locally and couldn't see a difference.
(both starting from a fresh all-images-deleted docker engine) 
```
build -t master .                       3.06s user 5.41s system 5% cpu 2:29.94 total
build -t optimise-dockerfile .  3.24s user 5.84s system 5% cpu 2:38.57 total
```

martyn-w: Hi @mafonso: try building master, making a code change (add ` puts "hello world"` somewhere), then rebuilding - I think that will be significantly slower than this dockerfile as it will re-trigger all the gems to be installed

Dockerfile
martyn-w: hmm.. not sure... I'll try it and see

martyn-w: Hi @mafonso I've tweaked the dockerfile a bit more and delayed copying in the app until right to the end. (We do need to copy in the app folder BTW as all the code is needed - maybe the tests are not, but will leave that for another day.)

Dockerfile
```diff
@@ -42,9 +43,8 @@ RUN addgroup -g $APPUID -S appgroup && \
 RUN apk add --update --no-cache tzdata postgresql-dev
 
 WORKDIR /app
-COPY --chown=appuser:appgroup --from=build-stage /app /app
 COPY --chown=appuser:appgroup --from=build-stage /usr/local/bundle /usr/local/bundle
+COPY --chown=appuser:appgroup . /app
```
mafonso, line 48: Yes, this is even cleaner since we don't even needed at the build stage.

PR #701, opened 2020-07-10
[P4-1663] Fix missing notifications when updating a profile

willfish:
### Jira link

P4-1663

### What?

Fixes notifications that update suppliers when changes to the PTR are made.

### Why?

We had a live production issues where a person on the road wasn't updated when the PTR was changed.

### DIFF
--- a/app/controllers/api/profiles_controller.rb
+++ b/app/controllers/api/profiles_controller.rb
@@ -16,6 +16,7 @@ module Api
       profile = person.profiles.find(params.require(:id))
 
       profile.update!(profile_attributes)
+      Notifier.prepare_notifications(topic: profile.person, action_name: 'update')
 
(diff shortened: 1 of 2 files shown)


app/controllers/api/profiles_controller.rb
```diff
@@ -16,6 +16,7 @@ def update
       profile = person.profiles.find(params.require(:id))
 
       profile.update!(profile_attributes)
+      Notifier.prepare_notifications(topic: profile.person, action_name: 'update')
```
alexdesi, line 19: should we always call this?
You can call the service even without passing in new assessment answers

martyn-w: @alexdesi good question; yes I think we should. Its important to notify the suppliers of _any_ changes.

PR #703, opened 2020-07-10
Add an enum for location_type on Location model

smoothcontract:
### Jira link

P4-1830

### What?

- [x] Added an enum for `location_type` on `Location` model
- [x] Fixed some specs that were setting incorrect data
- [x] Removed redundant code that comes for free with the enum

### Why?

- Just about to add a check for `high_security_hospital?` location type as part of this story and it feels wrong to be doing all of this the hard way. Idiomatic Rails FTW!

- Locations are only created in production via the `reference_data:create_locations` rake task so it's safe to retrofit this (it doesn't change the underlying persistence). But just to be sure I ran this check against a snapshot of staging:

Screenshot 2020-07-09 at 13 58 21


app/models/location.rb
```diff
@@ -41,24 +57,9 @@ class Location < ApplicationRecord
...
-    location_type.to_s == LOCATION_TYPE_PRISON
-  end
-
-  def police?
-    location_type.to_s == LOCATION_TYPE_POLICE
-  end
-
-  def court?
-    location_type.to_s == LOCATION_TYPE_COURT
-  end
```
alexdesi, line 63: nice cleaning 👍

spec/services/locations/importer_spec.rb
```diff
@@ -25,34 +25,29 @@
...
-        title: "Don't import me",
-        location_type: 'FOO',
-        can_upload_documents: false },
     ]
   end
 
   context 'with no existing records' do
     it 'creates all the input items' do
-      expect { importer.call }.to change(Location, :count).by(6)
+      expect { importer.call }.to change(Location, :count).by(5)
```
alexdesi, line 50: I think this test is not testing anymore the state of "with no existing records", since the `nomis_agency_id: 'FOO1' ` has been removed, right?

app/models/location.rb
```diff
@@ -15,6 +15,22 @@ class Location < ApplicationRecord
...
+    secure_training_center: LOCATION_TYPE_SECURE_TRAINING_CENTER,
+    secure_childrens_home: LOCATION_TYPE_SECURE_CHILDRENS_HOME,
+    youth_offending_institute: LOCATION_TYPE_YOUTH_OFFENDNG_INSTITUTE,
+    approved_premises: LOCATION_TYPE_APPROVED_PREMISES,
+    probation_office: LOCATION_TYPE_PROBATION_OFFICE,
+    community_rehabilitation_company: LOCATION_TYPE_COMMUNITY_REHABILITATION_COMPANY,
+    foreign_national_prison: LOCATION_TYPE_FOREIGN_NATIONAL_PRISON,
+    voluntary_hostel: LOCATION_TYPE_VOLUNTARY_HOSTEL,
+    high_security_hospital: LOCATION_TYPE_HIGH_SECURITY_HOSPITAL,
+    immigration_detention_centre: LOCATION_TYPE_IMMIGRATION_DETENTION_CENTRE,
```
alexdesi, line 31: Adding the `enum` is great  👍 
also, what about making the constants less verbose? considering they are already inside the `Location` class, would be nice to remove the `LOCATION_` path, and put TYPE at the end, for example:
`COURT_TYPE`,
`IMMIGRATION_DETENTION_CENTRE_TYPE`

smoothcontract: Hmm.. would rather not change all of the constants, they're used in too many places and would make the PR noisy, plus this is a convention used in other models.

spec/services/locations/importer_spec.rb
smoothcontract: Updated spec description and added another spec to more explicitly demonstrate the behaviour

app/models/location.rb
```diff
@@ -15,6 +15,22 @@ class Location < ApplicationRecord
   LOCATION_TYPE_HIGH_SECURITY_HOSPITAL = 'high_security_hospital'
   LOCATION_TYPE_IMMIGRATION_DETENTION_CENTRE = 'immigration_detention_centre'
 
+  enum location_type: {
+    court: LOCATION_TYPE_COURT,
+    police: LOCATION_TYPE_POLICE,
+    prison: LOCATION_TYPE_PRISON,
+    secure_training_center: LOCATION_TYPE_SECURE_TRAINING_CENTER,
```
cwrw, line 22: ✂️ centre/center

app/models/location.rb
```diff
@@ -15,6 +15,22 @@ class Location < ApplicationRecord
...
+    secure_childrens_home: LOCATION_TYPE_SECURE_CHILDRENS_HOME,
+    youth_offending_institute: LOCATION_TYPE_YOUTH_OFFENDNG_INSTITUTE,
```
cwrw, line 24: ✂️ offendng/offending

spec/models/location_spec.rb
```diff
@@ -9,6 +9,7 @@
 
   it { is_expected.to validate_presence_of(:title) }
   it { is_expected.to validate_presence_of(:location_type) }
+  it { is_expected.to define_enum_for(:location_type).backed_by_column_of_type(:string) }
```
cwrw, line 12: not sure if we want a test for inclusion or if this is enough?

PR #708, opened 2020-07-13
[P4-1907] Fix relationship patching on people v2

willfish:
### Jira link

P4-1907

### What?

Previously we were wiping out the gender and ethnicity and not
conforming to the json:api specification around patching relationships
in the people_actions v2.

This updates the people_actions behaviour so that we:

1. Update relationships to nil when they are passed but empty
2. Update relationships to the passed type if they exist and are passed
3. Do not change relationships if they are not passed

### Why?

Conform to json api specs

### DIFF
--- a/app/controllers/api/v2/people_actions.rb
+++ b/app/controllers/api/v2/people_actions.rb
@@ -47,16 +47,38 @@ module Api::V2
     end
 
     def person_attributes
-      person_params[:attributes].merge(ethnicity: ethnicity, gender: gender)
+      attributes = {}
+      attributes.merge!(person_params.fetch(:attributes, {}))
+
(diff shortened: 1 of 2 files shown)


app/controllers/api/v2/people_actions.rb
```diff
@@ -47,16 +47,38 @@ def person_params
     end
 
     def person_attributes
-      person_params[:attributes].merge(ethnicity: ethnicity, gender: gender)
+      attributes = {}
+      attributes.merge!(person_params.fetch(:attributes, {}))
```
willfish, line 51: Simplified this to handle the case where we don't have any attributes (and only relationships or neither) to patch.

PR #709, opened 2020-07-13
Further report field mapping

mafonso:
### Jira link

[P4-1876](https://dsdmoj.atlassian.net/browse/)

### What?

I have added/removed/altered:

- [ ] Fetch the supplier notification date from the notifications table
- [ ] Added PNC Number field
- [ ] Added NOMIS Prison ID field

### Why?

I am doing this because:
- Addressing a requirements 

### Have you? (optional)

### DIFF
--- a/lib/tasks/journeys_report.rake
+++ b/lib/tasks/journeys_report.rake
@@ -32,7 +32,7 @@ namespace :journeys do
         Journey.where(move: move).default_order.each do |journey|
           journeys << {
             id: journey.id,
-            supplier: supplier.name,
+            supplier: supplier.key,
             timestamp: journey.client_timestamp.iso8601,
             from: journey.from_location.nomis_agency_id,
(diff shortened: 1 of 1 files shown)


lib/tasks/journeys_report.rake
```diff
@@ -50,16 +50,19 @@ namespace :journeys do
...
           reference: move.reference,
-          notification_date: move.created_at, # TODO: use notifications table
+          notified_at: move_notification&.delivered_at,
           updated_at: move.updated_at,
           move_date: move.date,
           from: move.from_location.nomis_agency_id,
           to: move.to_location&.nomis_agency_id,
           person_id: person&.id,
+          pnc_number: person&.police_national_computer,
+          prison_id: person&.nomis_prison_number,
```
mafonso, line 65: We also have both `prison.number` and `nomis_prison_number`, this one seems to more explicit map the requirement

PR #712, opened 2020-07-13
Added support for new 'hospital' move type

smoothcontract:
### Jira link

P4-1830, P4-1832, P4-1835, P4-1836

### What?

- [x] Refactored and improved coverage for move_type_validator - now moved to a dedicated spec.
- [x] Added the new enum member and associated validation rules (to_location must be a hospital).
- [x] Added some comments to explain validation approach within ParamsValidator
- [x] Added additional request specs to ensure new move type can be created successfully

### Why?

- We want to support hospital moves (moves from any location to a high security hospital) in the near future.

### Have you? (optional)

- [x] Updated API docs if necessary (v1 and v2 Move schema)

- Adds support for a new move type when creating a move, but is not a breaking change

### DIFF
--- a/app/services/move_events/params_validator.rb
+++ b/app/services/move_events/params_validator.rb
@@ -11,7 +11,7 @@ module MoveEvents
     # TODO: in the future when pluralising types, existing move event records will need updating from singular types to plural types.
     ACCEPTABLE_PLURAL_VERBS = %w[cancels completes approves rejects].freeze
 
-    attr_reader :move, :timestamp, :type, :date, :cancellation_reason, :rejection_reason, :from_location_id, :to_location_id
+    attr_reader :timestamp, :type, :date, :cancellation_reason, :rejection_reason, :from_location_id, :to_location_id
 
     validates :type, presence: true, inclusion: { in: %w[accepts approve cancel complete lockouts redirects reject starts] }
     validates :cancellation_reason, inclusion: { in: Move::CANCELLATION_REASONS }, if: -> { type == 'cancel' }
@@ -34,13 +34,8 @@ module MoveEvents
 
     validates_with LocationValidator, locations: [:from_location_id], if: -> { type == 'lockouts' }
     validates_with LocationValidator, locations: [:to_location_id], if: -> { type == 'redirects' }
-    validates_with Moves::MoveTypeValidator, if: -> { type == 'redirects' }
 
-    # These attributes are validated against the existing move - i.e. before applying this event
-    delegate :move_type, :from_location, to: :move
-
-    def initialize(move, params)
-      @move = move
+    def initialize(params)
       @timestamp = params.dig(:attributes, :timestamp)
       @type = params[:type]
       @date = params.dig(:attributes, :date)
@@ -49,11 +44,6 @@ module MoveEvents
       @from_location_id = params.dig(:relationships, :from_location, :data, :id)
       @to_location_id = params.dig(:relationships, :to_location, :data, :id)
       singularize_acceptable_plural_types
-    end
-
-    def to_location
-      # Determine the new Location defined by passed param to validate against existing move_type and from_location attributes
-      @to_location ||= Location.find_by(id: to_location_id)
     end
 
   private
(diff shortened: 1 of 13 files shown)


app/services/move_events/params_validator.rb
```diff
@@ -50,6 +51,11 @@ def initialize(move, params)
       singularize_acceptable_plural_types
     end
 
+    def to_location
```
martyn-w, line 54: I think location validation should be a function of the LocationValidator. How about passing in the move type or move record to the location validator in order to do more comprehensive validations?

app/services/moves/move_type_validator.rb
```diff
@@ -23,7 +24,11 @@ def human_move_type
     end
 
     def validate_police_from_location
-      record.errors.add(:from_location, "must be a police location for #{human_move_type}") unless record.from_location&.police?
+      record.errors.add(:from_location, "must be a police location for #{human_move_type} move") unless record.from_location&.police?
```
martyn-w, line 27: consider moving to location validator

app/services/moves/move_type_validator.rb
```diff
@@ -23,7 +24,11 @@ def human_move_type
...
+    end
+
+    def validate_hospital_to_location
+      record.errors.add(:to_location, "must be a high security hospital location for #{human_move_type} move") unless record.to_location&.high_security_hospital?
```
martyn-w, line 31: consider moving to location validator

martyn-w:
Hi Andrew, I'm afraid I'm finding this difficult to follow because of all the locations. There are:
* move.from_location => where the move is currently going from
* move.to_location  => where the move is currently going to (which can be nil)
* move_events/params_validator to_location  => the to_location the move_event references, e.g. for redirects
* move_type_validator.record.from_location => the from_location of the current move ??

Could you try moving these new location validations (is police, is hospital) to the location validator, passing in any extra params as required to the location validator to support that?

smoothcontract:
> Hi Andrew, I'm afraid I'm finding this difficult to follow because of all the locations. There are:
> 
> * move.from_location => where the move is currently going from
> * move.to_location  => where the move is currently going to (which can be nil)
> * move_events/params_validator to_location  => the to_location the move_event references, e.g. for redirects
> * move_type_validator.record.from_location => the from_location of the current move ??
> 
> Could you try moving these new location validations (is police, is hospital) to the location validator, passing in any extra params as required to the location validator to support that?

Hmm... I'm not convinced that's a good approach. Your understanding above is correct - we have some attributes that are currently on the move, and a new `to_location` to be set by the redirect event. We need to validate that the proposed new `to_location` is suitable for the move and that depends on the move type.

However, the `LocationValidator` currently sits outside any specific module and provides generic functionality to verify that a location id is present and a Location exists with that id. (As an aside, perhaps this should be renamed to `LocationIdValidator`). I think it's fine to leave this as is with responsibility only for verifying location id's. If instead this needed a move type attribute to verify a location then that becomes a move specific concern (and so better lives under `Moves` module) and needs to understand the difference between a from and to location - so in effect I think this ends up back where we started with the `Moves::MoveTypeValidator`.

I'm open to other suggestions for improving the clarity on this - perhaps further documentation, but this isn't an easy thing to solve. I'm keen that we don't repeat validation rules (particularly as the rules will soon grow further in complexity) in two places so hence the validator approach that can work with both a `Move` model and the `ParamsValidator` class - both of those expose the same `move_type`, `from_location` and `to_location` attributes so the validator can work effectively in both scenarios. This integrates nicely with the `Move` model but is a little less obvious within the `ParamsValidator` unfortunately.

spec/requests/api/moves_controller_create_spec.rb
```diff
@@ -281,6 +281,22 @@
         end
       end
 
+      context 'with explicit hospital `move_type`' do
+        let(:move_attributes) { attributes_for(:move, move_type: 'hospital') }
+        let(:to_location) { create :location, :hospital, suppliers: [supplier] }
+
+        it_behaves_like 'an endpoint that responds with success 201'
+
+        it 'creates a move', skip_before: true do
```
martyn-w, line 290: 🙀 would be nice to do this without the evil `skip_before: true` 🙀

smoothcontract:
> 👍 LG and much clearer; thanks for talking through my concerns about the params validator. One small gripe: the use of `skip_before: true`, but not blocking and happy to leave that for a rainy day.

Thanks @martyn-w - I'll rewrite the spec in the next PR, still working in this same area.

PR #718, opened 2020-07-15
P4-1791 GET /person_escort_records/{person_escort_record_id}

cwrw:
### Jira link

[P4-1791](https://dsdmoj.atlassian.net/browse/P4-1791)

### What?

I have added/removed/altered:

- [x] Added responded field to responses
- [x] Added meta section to person escort record serializer
- [x] Added person escort record GET endpoint

### Why?

To support getting an existing PER, add new endpoint which accepts an id and returns a PER. To help the FE render different section progress, a new `meta` section has been added. This is supported by the [JSON:API spec](https://jsonapi.org/format/#document-resource-objects) in the data section. This will summarise all sections and their progress according to the new `responded` attribute on responses. The `responded` field is by default set to false. If a response is supplied and valid, the `responded` value is updated to true.

### Have you? (optional)

### DIFF
--- a/app/models/framework_response.rb
+++ b/app/models/framework_response.rb
@@ -2,6 +2,7 @@
 
 class FrameworkResponse < VersionedModel
   validates :type, presence: true
+  validates :responded, inclusion: { in: [true, false] }
 
   belongs_to :framework_question
   belongs_to :person_escort_record
@@ -13,11 +14,19 @@ class FrameworkResponse < VersionedModel
     record.errors.add(:value, :blank) if requires_value?(value, record)
   end
 
+  after_validation :set_responded_value
+
   def self.requires_value?(value, record)
     return false if value.present? || !record.framework_question.required
 
     return true if record.parent.blank?
 
     record.parent.option_selected?(record.framework_question.dependent_value)
+  end
+
+private
+
+  def set_responded_value
+    self.responded = true if value_text.present? || value_json.present?
   end
 end
--- a/app/models/framework_response.rb
+++ b/app/models/framework_response.rb
@@ -14,7 +14,7 @@ class FrameworkResponse < VersionedModel
     record.errors.add(:value, :blank) if requires_value?(value, record)
   end
 
-  after_validation :set_responded_value
+  after_validation :set_responded_value, on: :update
 
   def self.requires_value?(value, record)
     return false if value.present? || !record.framework_question.required
@@ -27,6 +27,6 @@ class FrameworkResponse < VersionedModel
 private
 
   def set_responded_value
-    self.responded = true if value_text.present? || value_json.present?
+    self.responded = true
   end
 end
--- a/app/models/person_escort_record.rb
+++ b/app/models/person_escort_record.rb
@@ -47,8 +47,11 @@ class PersonEscortRecord < VersionedModel
   end
 
   def section_progress
-    sections_to_responded.each do |section, responses|
-      sections_to_responded[section] = set_progress(responses)
+    sections_to_responded.map do |section, responses|
+      {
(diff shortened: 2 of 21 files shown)


app/models/person_escort_record.rb
```diff
@@ -40,4 +45,28 @@ def build_responses!
...
+    @sections_to_responded ||= framework_responses.joins(:framework_question).select('DISTINCT responded, framework_questions.section').group_by(&:section)
+  end
+
+  def set_progress(responses)
+    responded = responses.pluck(:responded).uniq
+    if responded.include?(true)
+      responded.include?(false) ? PERSON_ESCORT_RECORD_IN_PROGRESS : PERSON_ESCORT_RECORD_COMPLETED
+    else
+      PERSON_ESCORT_RECORD_NOT_STARTED
+    end
```
alexdesi, line 70: Maybe we could simplify the logic here, the methods all and any could help:
```
responses.all?
responses.any?(true/false)
```

app/models/person_escort_record.rb
```diff
@@ -1,8 +1,13 @@
 class PersonEscortRecord < VersionedModel
+  PERSON_ESCORT_RECORD_NOT_STARTED = 'not_started'.freeze
+  PERSON_ESCORT_RECORD_IN_PROGRESS = 'in_progress'.freeze
+  PERSON_ESCORT_RECORD_COMPLETED = 'completed'.freeze
+  PERSON_ESCORT_RECORD_CONFIRMED = 'confirmed'.freeze
```
alexdesi, line 5: maybe we can avoid heading all this constant with `PERSON_ESCORT_RECORD_` since they are scoped into the `PersonEscortRecord` class?

app/models/framework_response.rb
```diff
@@ -13,11 +14,19 @@ class FrameworkResponse < VersionedModel
...
 
     return true if record.parent.blank?
 
     record.parent.option_selected?(record.framework_question.dependent_value)
   end
+
+private
+
+  def set_responded_value
+    self.responded = true if value_text.present? || value_json.present?
```
alexdesi, line 30: we can avoid the id
```
self.responded = value_text.present? || value_json.present?
```

app/models/person_escort_record.rb
cwrw: This seems to be the convention in all other models, also I think it's best to prepend the name of the model before the constant in this case as we might have other models with the similar constants in the future, so using the model name allows values to be explicit and scoped

app/models/framework_response.rb
cwrw: So this was intentional as this flag is set to mark if a question has been seen/answered and when set to true cannot be changed. That said I realised there is missing logic around if a question is *not* required and we do get empty values, we should allow this to be set to true. I'll amend

app/models/person_escort_record.rb
cwrw: updated 👍

spec/models/person_escort_record_spec.rb
```diff
@@ -216,4 +216,65 @@
...
+
+      expect(person_escort_record.section_progress).to be_empty
+    end
+
+    it 'returns all section values for framework questions' do
+      person_escort_record = create(:person_escort_record, :with_responses)
+      question_sections = person_escort_record.framework_questions.pluck(:section).uniq
+      progress_sections = person_escort_record.section_progress.map { |section| section[:key] }
+
+      expect(progress_sections).to contain_exactly(*question_sections)
```
smoothcontract, line 232: You could use `match_array(question_sections)` here instead of splatting the argument - I think

PR #719, opened 2020-07-15
[P4-1879] Filter moves by supplier

willfish:
### Jira link

P4-1879

### What?

- [x] Replaces existing/undocumented filter of supplier_id on the
from_location with supplier_id directly on the move
- [x] Simplifies shared examples around filtering
- [x] Extends swagger documentation to include supplier id filter
- [x] Duplicates filter tests for v1/v2 (removing v1 tests later)

### Why?

We're moving to supporting multiple suppliers on a location so our
authorisation model of move ownership through the location no longer
works.

### Have you? (optional)

- We may find that this breaks something for frontend since we technically supported - but didn't document - a version of supplier filter in the past based on the from location -> suppliers relationship.

### DIFF
--- a/spec/requests/api/moves_controller_filter_spec.rb
+++ b/spec/requests/api/moves_controller_filter_spec.rb
@@ -13,16 +13,12 @@ RSpec.describe Api::MovesController do
   let(:expected_moves) { nil }
   let(:unexpected_moves) { nil }
 
-  shared_examples 'it returns the correct moves' do
+  shared_examples 'an api that filters moves correctly' do
     let(:size) { response_json['data'].size }
     let(:ids) { response_json['data'].map { |move| move['id'] } }
 
-    it 'contains all the expected moves' do
+    it 'contains the correct moves' do
       expect(ids).to match_array(expected_moves.pluck(:id))
-    end
-
-    it 'does not contain any unexpected moves' do
-      # NB: this is a necessary test to catch a dodgy test which has unexpected moves in the expected moves
       expect(ids & unexpected_moves.pluck(:id)).to be_empty
     end
   end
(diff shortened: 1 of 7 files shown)


willfish:
Clarified with @teneightfive in slack that we aren't using the undocumented supplier_id filter in the frontend. Going to run the e2e tests against this branch to make sure existing filtering works. Otherwise, I think this is fine.

willfish:
```
 Running tests in:

 Cancel allocation
 ✓ Cancel allocation

 New PMU allocation

 ✓ With unfound person
 ✓ With existing person

 ✓ With a new person

 ✓ Reason - `Made in error`
 ✓ Reason - `Another reason`

 New move from Prison to Court
 ✓ With unfound person
 ✓ With existing person

 ✓ With existing person

 Download moves
 ✓ As Police user
 ✓ As Prison user
 ✓ As STC user
 ✓ As Supplier user

 Smoke tests
 ✓ As Police user
 ✓ As STC user
 ✓ As Prison user
 ✓ As Supplier user
 ✓ As OCA user


 21 passed (3m 38s)
```

spec/requests/api/moves_controller_filter_spec.rb
```diff
@@ -13,16 +13,12 @@
   let(:expected_moves) { nil }
   let(:unexpected_moves) { nil }
 
-  shared_examples 'it returns the correct moves' do
+  shared_examples 'an api that filters moves correctly' do
```
alexdesi, line 16: much more clear 👏

spec/requests/api/moves_controller_filter_v2_spec.rb
```diff
@@ -0,0 +1,232 @@
...
+      context 'when other' do
+        let(:cancellation_reason) { 'other' }
+        let(:expected_moves) { other_moves }
+        let(:unexpected_moves) { made_in_error_moves + supplier_declined_to_move_moves + rejected_moves }
+
+        it_behaves_like 'an api that filters moves correctly'
+      end
+    end
+  end
+end
```
alexdesi, line 232: Do we need to duplicate this spec? (from `moves_controller_filter_spec.rb`)

Maybe it makes sense to have a filter for `supplier_id` in the V1 too, it does no harm, right?

willfish: The `Move::Finder` class is shared between both versions of this api so we automatically get the same behaviour for filters/sorting.

I have already written a test for v1 (see line 37 of spec/requests/api/moves_controller_filter_spec.rb)

willfish: It's confusing because we already supported filtering by supplier id but against the from location suppliers which is what we're moving away from.

spec/requests/api/moves_controller_filter_spec.rb
```diff
@@ -13,16 +13,12 @@
...
     let(:size) { response_json['data'].size }
     let(:ids) { response_json['data'].map { |move| move['id'] } }
 
-    it 'contains all the expected moves' do
+    it 'contains the correct moves' do
       expect(ids).to match_array(expected_moves.pluck(:id))
-    end
-
-    it 'does not contain any unexpected moves' do
```
willfish, line 24: I've combined these two tests into one to half the amount of request/response cycles we do

PR #721, opened 2020-07-15
Clean ECR images weekly

mafonso:
### Jira link

[P4-1476](https://dsdmoj.atlassian.net/browse/P4-1476)

### What?

I have added/removed/altered:

- [ ] Automates the goal of ticket P4-1476 

### Why?

I am doing this because:

- Automate everything™ 
- Apply the process  outlined in [Confluence](https://dsdmoj.atlassian.net/l/c/b0u0ExVQ) automatically every Monday.

### Have you? (optional)

- This was tested on a number of conditions to prevent the current images to being deleted, but there is always a slight change that it could happen do to a bug.
- Only the newest 100 < images < 200 will be left behind.
- Images containing "latest" on its tag designation will be always left behind untouched.
- Up to 100 images will be deleted per run (max aws batch-delete size)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -147,7 +147,7 @@ executors:
           RACK_ENV: test
           DATABASE_URL: "postgres://postgres@localhost/hmpps-book-secure-move-api"
           EXTERNAL_URL: mocked_in_tests
-          ENCRYPTOR_SALT: "*\xDB\x9E\a\x1ES1\xB7\xE3\xB8{\xF9Fl\x8Fnn?\x83\xC6\xBBR\x9Eq\x0F\x01./U\x0E\xF1\xDE"
+          ENCRYPTOR_SALT: "2EREZ8Xub/vt0ya1ZM6YKUwIMN72MbmqeWMq7KS4BV8oSJJc27rDpZYmA6AQGYcS"
       - image: circleci/postgres:10-alpine-ram
         environment:
           POSTGRES_USER: postgres
@@ -378,6 +378,38 @@ jobs:
           env: "production"
           notify_slack: "true"
 
+  ecr-cleanup:
+    executor: cloud-platform-executor
+    steps:
+      - run:
+          name: Check for too many images
+          command: |
+            COUNT=$(aws ecr list-images --repository-name 'book-a-secure-move/hmpps-book-secure-move-api' | jq '.imageIds | unique_by(.imageDigest) | length')
+            if [ "$COUNT" -lt 200 ]; then circleci step halt; exit 0;  fi
+            echo "More than 200 images found. Proceeding with cleanup"
+      - run:
+          name: Build manifest
+          command: |
+            TO_DELETE=$(aws ecr describe-images --region=eu-west-2 \
+              --repository-name book-a-secure-move/hmpps-book-secure-move-api \
+              --query 'imageDetails[]' --output json |\
+              jq 'sort_by(.imagePushedAt)|map(select(any(.imageTags[]? ; contains("latest"))|not))|map({imageDigest})|.[:100]')
+            echo "Tags to be deleted:"
+            aws ecr describe-images --region=eu-west-2 \
+              --repository-name book-a-secure-move/hmpps-book-secure-move-api \
+              --image-ids "${TO_DELETE}" --output json  | jq '.[][].imageTags[]'
+            echo "Deleting:"
+            aws ecr batch-delete-image --region eu-west-2 \
+            --repository-name book-a-secure-move/hmpps-book-secure-move-api \
+            --image-ids ${TO_DELETE}
+      - run:
+          name: Confirm live latest images
+          command: |
+            echo "Images tagged latest left behind"
+            aws ecr describe-images --region=eu-west-2 \
+            --repository-name  book-a-secure-move/hmpps-book-secure-move-api \
+            --query 'imageDetails[]' --output json | \
+            jq 'map(select(any(.imageTags[]? ; contains("latest"))))|map(.imageTags)'
 workflows:
   version: 2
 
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -147,7 +147,7 @@ executors:
           RACK_ENV: test
           DATABASE_URL: "postgres://postgres@localhost/hmpps-book-secure-move-api"
           EXTERNAL_URL: mocked_in_tests
-          ENCRYPTOR_SALT: "*\xDB\x9E\a\x1ES1\xB7\xE3\xB8{\xF9Fl\x8Fnn?\x83\xC6\xBBR\x9Eq\x0F\x01./U\x0E\xF1\xDE"
+          ENCRYPTOR_SALT: "2EREZ8Xub/vt0ya1ZM6YKUwIMN72MbmqeWMq7KS4BV8oSJJc27rDpZYmA6AQGYcS"
```
mafonso, line 150: This binary string was causing YAML parsing issues.  (Same happens with Kubernetes YAML files.)
We could keep the binary and use the `!binary |`  operator but no need when a string will do.

.circleci/config.yml
```diff
@@ -378,6 +378,38 @@ jobs:
...
 
+  ecr-cleanup:
+    executor: cloud-platform-executor
+    steps:
+      - run:
+          name: Check for too many images
+          command: |
+            COUNT=$(aws ecr list-images --repository-name 'book-a-secure-move/hmpps-book-secure-move-api' | jq '.imageIds | unique_by(.imageDigest) | length')
+            if [ "$COUNT" -lt 200 ]; then circleci step halt; exit 0;  fi
+            echo "More than 200 images found. Proceeding with cleanup"
```
mafonso, line 389: Abort job gracefully if we have less than 200 images

.circleci/config.yml
```diff
@@ -378,6 +378,38 @@ jobs:
...
+      - run:
+          name: Build manifest
+          command: |
+            TO_DELETE=$(aws ecr describe-images --region=eu-west-2 \
+              --repository-name book-a-secure-move/hmpps-book-secure-move-api \
+              --query 'imageDetails[]' --output json |\
+              jq 'sort_by(.imagePushedAt)|map(select(any(.imageTags[]? ; contains("latest"))|not))|map({imageDigest})|.[:100]')
+            echo "Tags to be deleted:"
```
mafonso, line 397: Verbosely display the images that we plan to delete in next step

.circleci/config.yml
```diff
@@ -378,6 +378,38 @@ jobs:
...
+            aws ecr describe-images --region=eu-west-2 \
+              --repository-name book-a-secure-move/hmpps-book-secure-move-api \
+              --image-ids "${TO_DELETE}" --output json  | jq '.[][].imageTags[]'
+            echo "Deleting:"
+            aws ecr batch-delete-image --region eu-west-2 \
+            --repository-name book-a-secure-move/hmpps-book-secure-move-api \
+            --image-ids ${TO_DELETE}
+      - run:
+          name: Confirm live latest images
+          command: |
```
mafonso, line 407: Display the images whose tags contain "latest" that were left after the run. 
Just for peace of mind

.circleci/config.yml
```diff
@@ -434,3 +466,10 @@ workflows:
           <<: *only_deploy_tags
           requires:
             - hold_production
+  ecr-cleanup:
+    jobs: [ecr-cleanup]
+    triggers:
+      - schedule:
+          cron: "0 9 * * 1"
```
mafonso, line 473: 9am every monday.

.circleci/config.yml
```diff
@@ -378,6 +378,38 @@ jobs:
...
           notify_slack: "true"
 
+  ecr-cleanup:
+    executor: cloud-platform-executor
+    steps:
+      - run:
+          name: Check for too many images
+          command: |
+            COUNT=$(aws ecr list-images --repository-name 'book-a-secure-move/hmpps-book-secure-move-api' | jq '.imageIds | unique_by(.imageDigest) | length')
+            if [ "$COUNT" -lt 200 ]; then circleci step halt; exit 0;  fi
```
willfish, line 388: TIL `circleci step halt` 🎉

.circleci/config.yml
```diff
@@ -378,6 +378,38 @@ jobs:
...
+            TO_DELETE=$(aws ecr describe-images --region=eu-west-2 \
+              --repository-name book-a-secure-move/hmpps-book-secure-move-api \
+              --query 'imageDetails[]' --output json |\
+              jq 'sort_by(.imagePushedAt)|map(select(any(.imageTags[]? ; contains("latest"))|not))|map({imageDigest})|.[:100]')
+            echo "Tags to be deleted:"
+            aws ecr describe-images --region=eu-west-2 \
+              --repository-name book-a-secure-move/hmpps-book-secure-move-api \
+              --image-ids "${TO_DELETE}" --output json  | jq '.[][].imageTags[]'
+            echo "Deleting:"
+            aws ecr batch-delete-image --region eu-west-2 \
```
willfish, line 402: Is the region something better left to the environment?

mafonso: I'll check if we have `AWS_DEFAULT_REGION=eu-west-2` available and defined in CI.  
But all  AWS resources in live-1 are in that region, anyway.

mafonso:
> So this looks totally fine to me. I'd be interested in whether it works end-to-end by manually invoking it somehow and seeing it run twice (once where our images are > 200 in number and then once again.

Yes, you can run the commands one by one and tweak the values. We have about 40 images now, so it's quite low to change just one param. 

The reason for the 100 < 200, is that we'll delete up to a maximum batch of the oldest images, which as a limit of 100.
So I've set the threshold to be 200, so that when you delete a full batch, it still leaves 100 behind. 

But we can adjust this to be any interval.  The jq array will determine the batch size and shell condition the trigger.

.circleci/config.yml
mafonso: We do have AWS_DEFAULT_REGION=eu-west-2  so removing this.

PR #724, opened 2020-07-15
Expose PER in profile and update swagger docs

cwrw:
### What?

I have added/removed/altered:

- [x] Added PER include in profile
- [x] Added PER relationship in profile
- [x] Updated swagger docs for move includes and PER endpoints
- [x] Added tests for includes in PER serializers as they are keyed differently

### Why?

To allow the FE to access the PER from the moves detail dashboard, expose PER relationship in profile. I also updated swagger documentation as it was missing includes. I added some tests for serializers as we are keying questions/responses differently 

### Have you? (optional)

### DIFF
--- a/app/serializers/profile_serializer.rb
+++ b/app/serializers/profile_serializer.rb
@@ -7,6 +7,7 @@ class ProfileSerializer < ActiveModel::Serializer
 
   belongs_to :person, serializer: PersonSerializer
   has_many :documents, serializer: DocumentSerializer
+  has_one :person_escort_record, serializer: PersonEscortRecordSerializer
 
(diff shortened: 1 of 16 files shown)


app/serializers/profile_serializer.rb
```diff
@@ -7,6 +7,7 @@ class ProfileSerializer < ActiveModel::Serializer
 
   belongs_to :person, serializer: PersonSerializer
   has_many :documents, serializer: DocumentSerializer
+  has_one :person_escort_record, serializer: PersonEscortRecordSerializer
```
willfish, line 10: Do you want to add this to v2 profile serializer?

PR #725, opened 2020-07-15
Add support for prison_recall and court_other move types

smoothcontract:
### Jira link

P4-1873 P4-1860

### What?

- [x] Added new `court_other` and `prison_remand` move types
- [x] Added new helper methods to Location model to determine prison/STC/SCH types
- [x] Refactored existing create moves request spec (v1) to avoid use of `skip_before` tag

### Why?

- We want to support additional move types in the near future.

### Have you? (optional)

- Adds support for new types in move type param but these are not currently used, so no risk

### DIFF
--- a/spec/models/location_spec.rb
+++ b/spec/models/location_spec.rb
@@ -11,12 +11,14 @@ RSpec.describe Location do
   it { is_expected.to validate_presence_of(:location_type) }
   it { is_expected.to define_enum_for(:location_type).backed_by_column_of_type(:string) }
 
-  context 'when location is a prison' do
-    subject(:location) { build :location }
+  context 'when location is a court' do
+    subject(:location) { build :location, :court }
 
-    it { expect(location.prison?).to be true }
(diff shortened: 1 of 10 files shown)


spec/models/location_spec.rb
```diff
@@ -11,28 +11,54 @@
   it { is_expected.to validate_presence_of(:location_type) }
   it { is_expected.to define_enum_for(:location_type).backed_by_column_of_type(:string) }
 
+  context 'when location is a court' do
+    subject(:location) { build :location, :court }
+
+    it { expect(location.prison?).to be false }
```
cwrw, line 17: although this seems to have been an existing pattern in our test, I'm wondering if we should be testing enum types? maybe testing detained/not_detained is enough?

smoothcontract: Yes, tend to agree. Those were tests we had with the old implementation prior to adding the enum, so those were custom methods. They can be stripped out now.

PR #727, opened 2020-07-17
Make fields not required for Patch /move

alexdesi:
### Jira link

https://dsdmoj.atlassian.net/browse/P4-1929

### What?

Now, in the swagger docs, for patch move, it shows many fields that should not be possible to update.

Separate the example in Post and Patch, instead of using a common YAML to describe both.

### Why?
in the `patch /move` there should not be required attributes, if you don't pass them, the attributes will stay untouched.

### DIFF
--- a/swagger/v2/patch_move.yaml
+++ b/swagger/v2/patch_move.yaml
@@ -1,0 +1,142 @@
+Move:
+  type: object
+  required:
+    - type
+    - attributes
+  properties:
+    id:
+      type: string
+      format: uuid
+      example: f0a91e16-1b0e-4e1f-93fe-319dda9933e6
+      description: The unique identifier (UUID) of this object
+    type:
+      type: string
+      example: moves
+      enum:
+        - moves
+      description: The type of this object - always `moves`
+    attributes:
+      type: object
+      properties:
+        reference:
+          type: string
+          example: TM7B3A2S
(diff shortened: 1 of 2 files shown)


swagger/v2/patch_move.yaml
```diff
@@ -0,0 +1,142 @@
...
+        status:
+          type: string
+          enum:
+            - proposed
+            - requested
+            - booked
+            - in_transit
+            - cancelled
+            - completed
+          description: Indicates the stage in it's lifecycle that this move is at
```
matt-crowson-moj, line 35: `it's` should be `its`

swagger/v2/patch_move.yaml
```diff
@@ -0,0 +1,142 @@
...
+              format: date
+            - type: "null"
+          example: "2020-05-17"
+          description: Start date for potential move (mandatory if status is proposed)
+        date_to:
+          oneOf:
+            - type: string
+              format: date
+            - type: "null"
+          example: "2020-05-17"
```
matt-crowson-moj, line 94: I'd suggest changing this as it is the same as the `date_from` and the description says it should be after...

swagger/v2/patch_move.yaml
alexdesi: nice one, thanks!

PR #728, opened 2020-07-17
P4-1771 Extend PER validations

cwrw:
### Jira link

[P4-1771](https://dsdmoj.atlassian.net/browse/P4-1771)

### What?

I have added/removed/altered:

- [x] Added type validation across responses
- [x] Fixed error validations
- [x] Bubble up errors from nested detail objects and collections
- [x] Added extra testing for posterity

### Why?

Correct the following validations:
- Array responses inclusion: inclusion validation works on strings not arrays use custom validation instead
- Validate option inclusion if option present to allow empty responses to be valid if response optional
- Validate details presence only if question answered with dependent answer

Added the following validations:
- Type validation for each response type
- Uniqueness validation on collection options incase someone sends a checklist response more than once
- Move option presence to only be required if details sent, as presence is done in main object class

### DIFF
--- a/app/models/framework_response/array.rb
+++ b/app/models/framework_response/array.rb
@@ -1,7 +1,8 @@
 class FrameworkResponse
   class Array < FrameworkResponse
+    validate :validate_array_type
     validates :value_text, absence: true
-    validates :value_json, on: :update, inclusion: { in: :question_options }, if: :question_options
+    validate :validate_array_values, on: :update
 
     def value
       value_json.presence || []
@@ -17,8 +18,16 @@ class FrameworkResponse
 
   private
 
-    def question_options
-      @question_options ||= framework_question.options.presence
+    def validate_array_values
+      if (invalid_options = value - framework_question.options).any?
+        errors.add(:value, invalid_options.join(', ') + ' are not a valid option')
+      end
+    end
+
+    def validate_array_type
+      unless value.is_a?(::Array)
+        errors.add(:value, 'is incorrect type')
+      end
     end
   end
 end
--- a/app/models/framework_response/array.rb
+++ b/app/models/framework_response/array.rb
@@ -2,7 +2,7 @@ class FrameworkResponse
   class Array < FrameworkResponse
     validate :validate_array_type
     validates :value_text, absence: true
-    validate :validate_array_values, on: :update
+    validate :validate_value_inclusion, on: :update
 
     def value
       value_json.presence || []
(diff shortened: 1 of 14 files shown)


app/models/framework_response/array.rb
```diff
@@ -17,8 +18,16 @@ def option_selected?(option)
 
   private
 
-    def question_options
-      @question_options ||= framework_question.options.presence
+    def validate_array_values
+      if (invalid_options = value - framework_question.options).any?
+        errors.add(:value, invalid_options.join(', ') + ' are not a valid option')
```
smoothcontract, line 23:
```suggestion
        errors.add(:value, invalid_options.join(', ') + ' are not valid options')
```

app/models/framework_response/collection.rb
```diff
@@ -31,11 +32,26 @@ def details_collection(collection)
     end
 
     def validate_details_collection
-      errors.add(:value, :invalid) if details_collection(value).invalid?
+      validated_collection = details_collection(value)
+      if validated_collection.invalid?
+        validated_collection.errors.each do |field, message|
+          errors.add(field, message)
+        end
```
smoothcontract, line 39: Could you do this instead of looping through the errors maybe?

```ruby
errors.merge!(validated_collection.errors)
```

app/models/framework_response/details_collection.rb
```diff
@@ -25,7 +26,20 @@ def to_a
...
 
     def details_objects
-      errors.add(:collection, :invalid) if collection.any?(&:invalid?)
+      return unless collection.any?(&:invalid?)
+
+      collection.each do |detail_object|
+        detail_object.errors.each do |field, message|
+          errors.add(field, message)
+        end
+      end
```
smoothcontract, line 35: Use `errors.merge!(detail_object.errors)` ?

app/models/framework_response/details_object.rb
```diff
@@ -6,9 +6,9 @@ class DetailsObject
 
     attr_accessor :question_options, :details_options, :option, :details
 
-    validates :option, presence: true
-    validates :option, inclusion: { in: :question_options }, if: :question_options
-    validates :details, presence: true, if: :details_options
+    validates :option, presence: true, if: :details
+    validates :option, inclusion: { in: :question_options }, if: ->(object) { object.question_options && object.option.present? }
```
smoothcontract, line 10: What type is `object` here? I'm wondering if some of this logic expression could be moved to the relevant class and given a method name to make the intent more obvious.

app/models/framework_response/string.rb
```diff
@@ -1,7 +1,7 @@
 class FrameworkResponse
   class String < FrameworkResponse
     validates :value_json, absence: true
-    validates :value_text, on: :update, inclusion: { in: :question_options }, if: :question_options
+    validates :value, inclusion: { in: :question_options }, if: ->(response) { response.question_options && response.value.present? }
```
smoothcontract, line 4: Same as above (What type is object here? I'm wondering if some of this logic expression could be moved to the relevant class and given a method name to make the intent more obvious.)

PR #730, opened 2020-07-17
Enforce move_type attribute to be specified for v2 move creation

smoothcontract:
### Jira link

P4-1851

### What?

- [x] Added a new transient `version` attribute on `Move` model to avoid automatically setting a `move_type` for moves created via v2 `POST /moves` endpoint
- [x] Enforce specification of `move_type` when creating a move via v2 `POST /moves` endpoint (by disabling behaviour described above)
- [x] Added validation rules for original move types that were previously derived automatically - reflecting existing rules for these so that moves created under `v1` endpoint without specifying `move_type` are still valid

### Why?

- We need to support a richer set of move types (already implemented in previous PR's) so cannot rely on move type being derived automatically in the back end API. Instead clients (front end and suppliers) must specify the `move_type` explicitly when creating a move under v2 `POST /moves` endpoint. v1 endpoint will still allow `move_type` to be omitted.

### Have you? (optional)

- This is a breaking change for v2 - which is in use now, so needs collaboration with front end devs and communication with suppliers before this can be merged. Will also require end to end test suite to pass successfully (currently this will break as `move_type` is not specified by front end code).

### DIFF
--- a/app/models/move.rb
+++ b/app/models/move.rb
@@ -90,6 +90,13 @@ class Move < VersionedModel
 
   scope :not_cancelled, -> { where.not(status: MOVE_STATUS_CANCELLED) }
 
+  attr_accessor :version
+
+  # TODO: Temporary method to apply correct validation rules when creating v2 move
+  def v2?
+    version == 2
+  end
+
   def rebooked
     self.class.find_by(original_move_id: id)
   end
@@ -146,7 +153,7 @@ private
   end
 
   def set_move_type
-    return if move_type.present?
+    return if move_type.present? || v2?
 
     # TODO: The order is not important, here.
     #       Remove this from the model when we migrate to mandatory move_type under v2
--- a/spec/factories/move.rb
+++ b/spec/factories/move.rb
@@ -9,7 +9,6 @@ FactoryBot.define do
     time_due { Time.now }
     status { 'requested' }
     additional_information { 'some more info about the move that the supplier might need to know' }
-    move_type { 'court_appearance' }
     sequence(:created_at) { |n| Time.now - n.minutes }
     sequence(:date_from) { |n| Date.today - n.days }
 
@@ -41,6 +40,8 @@ FactoryBot.define do
 
     trait :police_transfer do
       move_type { 'police_transfer' }
+      association(:from_location, :police, factory: :location)
+      association(:to_location, :police, factory: :location)
     end
 
     trait :video_remand do
@@ -150,10 +151,10 @@ FactoryBot.define do
     end
   end
 
-  factory :from_court_to_prison, class: 'Move' do
+  factory :from_prison_to_court, class: 'Move' do
     association(:profile)
-    association(:from_location, :court, factory: :location)
-    association(:to_location, :prison, factory: :location)
+    association(:from_location, :prison, factory: :location)
+    association(:to_location, :court, factory: :location)
     date { Date.today }
     time_due { Time.now }
     status { 'requested' }
--- a/spec/models/move_spec.rb
+++ b/spec/models/move_spec.rb
@@ -229,11 +229,36 @@ RSpec.describe Move do
   end
 
   describe '#move_type' do
-    subject(:move) { build :move, from_location: from_location, to_location: to_location, move_type: nil }
+    subject(:move) { build :move, from_location: from_location, to_location: to_location, move_type: move_type, version: version }
 
     let(:from_location) { build :location, :police }
+    let(:move_type) { nil }
+    let(:version) { nil }
 
     before { move.valid? }
+
+    context 'when creating a v2 move' do
+      let(:version) { 2 }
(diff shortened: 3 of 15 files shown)


swagger/v1/move.yaml
```diff
@@ -19,7 +19,6 @@ Move:
       type: object
       required:
       - status
-      - move_type
```
smoothcontract, line 22: Under v1 `move_type` is not mandatory, so this documentation was incorrect.

app/models/move.rb
```diff
@@ -93,6 +93,13 @@ class Move < VersionedModel
 
   scope :not_cancelled, -> { where.not(status: MOVE_STATUS_CANCELLED) }
 
+  attr_accessor :version
+
+  # TODO: Temporary method to apply correct validation rules when creating v2 move
```
smoothcontract, line 98: This can be removed when we drop support for v1 - along with the `set_move_type` method below

spec/factories/move.rb
```diff
@@ -9,7 +9,6 @@
     time_due { Time.now }
     status { 'requested' }
     additional_information { 'some more info about the move that the supplier might need to know' }
-    move_type { 'court_appearance' }
```
smoothcontract, line 12: There was a balance here - many tests used this as the default move type without specifying the correct from and to location types. Removing this as the default - i.e. move_type is not set and instead derived using the existing `set_move_type` method within `Move` model is less invasive for now. When we move to removing v1 support completely we can revisit this - lots of tests will break again but at least then we only have to worry about half as many request specs.

A few request specs were relying on this assumption so those have been fixed to set an explicit move type and the correct location types (i.e. those won't break in future when `set_move_type` is removed).

spec/factories/move.rb
```diff
@@ -135,10 +136,10 @@
     end
   end
 
-  factory :from_court_to_prison, class: 'Move' do
+  factory :from_prison_to_court, class: 'Move' do
```
smoothcontract, line 139: This was back to front

spec/models/move_spec.rb
```diff
@@ -256,12 +256,37 @@
...
-    subject(:move) { build :move, from_location: from_location, to_location: to_location, move_type: nil }
+    subject(:move) { build :move, from_location: from_location, to_location: to_location, move_type: move_type, version: version }
 
     let(:from_location) { build :location, :police }
+    let(:move_type) { nil }
+    let(:version) { nil }
 
     before { move.valid? }
 
+    context 'when creating a v2 move' do
```
smoothcontract, line 267: At some point this will become the default behaviour (i.e. when v1 support is removed)

spec/requests/api/moves_controller_update_v2_spec.rb
```diff
@@ -40,7 +40,6 @@
           additional_information: 'some more info',
           cancellation_reason: nil,
           cancellation_reason_comment: nil,
-          move_type: 'court_appearance',
```
smoothcontract, line 43: The API endpoint doesn't support changing this attribute - it's not in the included list of permitted params so not really sure why it's present here.

app/models/move.rb
```diff
@@ -157,7 +164,7 @@ def set_reference
   end
 
   def set_move_type
```
cwrw, line 166: We could move this logic to the move_actions v1? this could help avoid having to define anything regarding v2

smoothcontract: Explored this approach and it didn't end well. The problem is we have many (200+) existing tests that currently use the move factory and then go on to set invalid combinations of move and location types. I was able to resolve most of those without making the PR too noisy by setting the default move_type in the factory to `nil`. That then means that `set_move_type` is called when the model is validated and move type is correctly determined based on the location types set for each test case. Removing the `before_validation` callback on this method and calling it explicitly from the `create_and_render` method in v1 works for the create move request specs, but reintroduces the issue with inconsistent test data in other specs and exposes the 200+ broken specs.

I think it's just too noisy to fix all of the broken specs as part of this PR so my preference is to keep the implementation as is for now (it's temporary anyway), and address the broken specs at a later stage; when we remove support for v1 there will be a lot of redundant code and so many of these broken specs will also be removed and there will be a lot less stuff to fix.

martyn-w:
my view is that we should not make breaking changes in V2, so I think we should continue to maintain the halfway house in V2 - with a view to move_type being mandatory in V3. I think this will be a sensitive issue for suppliers so best to get input from @jcavell and @jason-sims when he's back

martyn-w:
I've been thinking about this overnight: it turns out that only supplier `S` has actually created some move records, and they all look to be set with a move type, so _possibly_ this change might be merged under the radar. But best to discuss with @jcavell and @jason-sims as there is potential here to cause upset.

jason-sims:
After discussion confirmed both suppliers are already passing in `move_type ` so this is ok to merge.

PR #731, opened 2020-07-20
log number of moves exported

mafonso:
### Jira link

n/a

### What?

I have added/removed/altered:

- [ ] Outputting the number of exported moves per supplier 

### Why?

I am doing this because:
- So that we can keep track of the report evolution without having to open the actual reports. 

### Have you? (optional)

### DIFF
--- a/lib/tasks/journeys_report.rake
+++ b/lib/tasks/journeys_report.rake
@@ -78,6 +78,7 @@ namespace :journeys do
           journeys: journeys,
         }
       end
+      print " #{moves.length} moves exported. "
       puts 'done.'
 
       File.open(filename, 'w') do |file|


lib/tasks/journeys_report.rake
```diff
@@ -78,6 +78,7 @@ namespace :journeys do
           journeys: journeys,
         }
       end
+      print " #{moves.length} moves exported. "
```
martyn-w, line 81: `puts` not `print`

martyn-w: but I think you want `puts` here :-)

mafonso: I wanted to get this effect `Generating 2020-07-07-geoamey.json... 183 moves exported. done.`
Can I have puts not adding the newline, or should I not do that at all?    e.g potential issue with flushing the console buffer, or similar?

mafonso: I wanted that newline behaviour. Maybe I should add a ` $stdout.flush` as well?

martyn-w: thats fine as it is - no need for `$stdout.flush`

PR #732, opened 2020-07-20
P4-1778 Add FrameworkResponse patch endpoint

cwrw:
### Jira link

[P4-1778](https://dsdmoj.atlassian.net/browse/P4-1778)

### What?

I have added/removed/altered:

- [x] Added new framework response patch endpoint
- [x] Add join table for flags/framework responses
- [x] Add ability to attach/detach flags from a response according to it's answer
- [x] Add swagger docs for new endpoint and missing models

### Why?

To allow a client to update a response on a PER, add a new endpoint which accepts a value for the question. According to the answer a flag should be attached/detached from the response if the response matched the value on the flag. Flag should be renamed to `framework_flags` which will be done in a following PR

### Have you? (optional)

### DIFF
--- a/app/controllers/api/framework_responses_controller.rb
+++ b/app/controllers/api/framework_responses_controller.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+module Api
+  class FrameworkResponsesController < ApiController
+    PPERMITTED_PARAMS = [
+      :type,
+      attributes: [:value, { value: %i[option details] }, { value: [] }],
+    ].freeze
+
+    def update
+      framework_response.update!(update_attributes)
+
+      render json: framework_response, status: :ok, include: included_relationships
+    end
+
+  private
+
+    def update_params
+      params.require(:data).permit(PPERMITTED_PARAMS)
+    end
+
+    def update_attributes
+      update_params.to_h[:attributes]
+    end
+
+    def supported_relationships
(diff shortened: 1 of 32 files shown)


spec/swagger/swagger_doc_v1.yaml
```diff
@@ -448,6 +456,59 @@
             application/vnd.api+json:
               schema:
                 "$ref": "../v1/post_documents_responses.yaml#/422"
+  "/framework_responses/{framework_response_id}":
+    patch:
+      summary: Updates a framework response's value
+      tags:
+      - FrameworkResponse
```
teneightfive, line 463: Is this what maps to the heading? If so, could it be a plural to keep consistency with the other headings. And if possible, could it contain a space - `Framework responses` :D?

app/controllers/api/framework_responses_controller.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module Api
+  class FrameworkResponsesController < ApiController
+    PPERMITTED_PARAMS = [
+      :type,
+      attributes: [:value, { value: %i[option details] }, { value: [] }],
```
smoothcontract, line 7: Interesting syntax! Might be worth a comment here to explain as it's not immediately clear what shape the params have. Also I can't see any mention of `details` in the specs - there's one case for passing `options`.

PR #736, opened 2020-07-22
P4-1945 Clear dependent answers

cwrw:
### Jira link

[P4-1945](https://dsdmoj.atlassian.net/browse/P4-1945)

### What?

I have added/removed/altered:

- [x] Added ability to clear dependent responses, flags, and reset responded attribute

### Why?

If the parent answer is changed, we would like to clear all answered dependent values and flags. We should clear
the full tree of dependents and not only the first level down. To determine the correct dependent responses to clear, only
select the responses with dependent values used to match the answered value, this is found
in the framework question attached to the response under `dependent_value`.

I have added a recursive CTE to retrieve a full list of descendants for a list of response ids, rather than recursively attempting to retrieve them as its more efficient. Full explanation can be found [here](https://www.gmarik.info/blog/2012/recursive-data-structures-with-rails/).

To bulk update, I am using the active record import method to avoid any callbacks and to allow associations to be updated.

### DIFF
--- a/app/models/framework_response.rb
+++ b/app/models/framework_response.rb
@@ -57,9 +57,11 @@ private
     dependent_ids = dependents.includes(:framework_question).reject { |dependent| option_selected?(dependent.framework_question.dependent_value) }
     return unless dependent_ids.any?
 
-    FrameworkResponse
+    dependent_tree = FrameworkResponse
       .where("framework_responses.id IN (#{recursive_tree})", dependent_ids)
-      .update(value_json: nil, value_text: nil, flags: [])
+
+    dependent_tree.joins(:flags).update(flags: [])
+    dependent_tree.update_all(value_json: nil, value_text: nil, responded: false)
   end
 
   def recursive_tree
--- a/app/models/framework_response.rb
+++ b/app/models/framework_response.rb
@@ -52,19 +52,36 @@ private
   end
 
   def clear_dependent_values_and_flags!(old_value)
-    return unless (old_value != value) && dependents.any?
+    return unless old_value != value
 
     dependent_ids = dependents.includes(:framework_question).reject { |dependent| option_selected?(dependent.framework_question.dependent_value) }
+
+    update_dependent_responses!(dependent_ids)
+  end
+
+  def update_dependent_responses!(dependent_ids)
     return unless dependent_ids.any?
 
-    dependent_tree = FrameworkResponse
-      .where("framework_responses.id IN (#{recursive_tree})", dependent_ids)
+    descendants = descendants_tree(dependent_ids).includes(:framework_question, :flags).map do |descendant|
+      descendant.assign_attributes(
+        value: nil,
+        responded: false,
+        flags: [],
+      )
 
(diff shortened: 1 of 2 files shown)


app/models/framework_response.rb
```diff
@@ -45,4 +50,52 @@ def build_flags
...
+  end
+
+  def descendants_tree(ids)
+    FrameworkResponse
+      .where("framework_responses.id IN (#{recursive_tree})", ids)
+  end
+
+  def recursive_tree
+    # build full descendants tree
+    <<-SQL
```
martyn-w, line 85: This approach is pretty low-level sql and will be hard to maintain... is than an alternative more active-record friendly approach? could the acts_as_tree gem possibly help?

cwrw: I usually try to avoid building complex low level queries and default to using active record DSL provided, however the ability to build a recursive function wasn't possible unfortunately. I had a look at a few gems that could help, however `acts_as_tree` gem doe not support this kind of querying and only normally walks the tree. Another gem that does support this seems to not be actively maintained https://github.com/chrisroberts/acts_as_sane_tree. I did look into the ancestry gem initially but it did not play well with bulk creation of dependencies so I had to abandon it.

I kept this query quite simple in structure, where the only thing that is different is that it utilises the [recursive functionality](https://en.wikipedia.org/wiki/Hierarchical_and_recursive_queries_in_SQL#Common_table_expression) in [Postgres](https://www.postgresql.org/docs/12/queries-with.html).  It starts by selecting the root nodes required, and then the next section defines the recursive query to retrieve the next nodes. I think as long as we keep this query documented well and relatively simple in structure it can be the tradeoff.

martyn-w: Thanks for looking into alternatives 👍

spec/models/framework_response_spec.rb
```diff
@@ -227,5 +227,163 @@
...
+    context 'when dependent responses' do
+      it 'clears dependent responses if value changed' do
+        parent_response = create(:string_response)
+        child1_question = create(:framework_question, dependent_value: 'Yes', parent: parent_response.framework_question)
+        child2_question = create(:framework_question, dependent_value: 'Yes', parent: parent_response.framework_question)
+        create(:string_response, framework_question: child1_question, parent: parent_response)
+        create(:string_response, framework_question: child2_question, parent: parent_response)
+        parent_response.update_with_flags!('No')
+
+        parent_response.dependents.each do |child_response|
```
martyn-w, line 240: maybe add a test to very parent_response has dependents, as this test will skipped if it does not

martyn-w:
Could you add a test that exercises the `recursive_tree` functionality - e.g. a question tree with several levels of hierarchy?

cwrw:
@martyn-w I have the following test for testing multiple levels of hierarchy: https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/736/files#diff-82143bd465fa09ee8f8c351f3aa9b58cR320
Unless you meant something different?

martyn-w:
@cwrw sorry I missed that grand-child test, that's fine 👍

PR #737, opened 2020-07-22
tag api story rsepc

martyn-w:
Tags api story rspec to it can be easily excluded

api story tests can be excluded with:
```
rspec --tag ~@api_story
```


smoothcontract:
Would it be possible to add this to `.rspec` to make these *not* run by default? Would also need to override this for CI. What do people think - what's a sensible default?

```
--tag ~@api_story
```

martyn-w:
my view is that in an ideal world it _should_ run by default: when it fails it will help us catch problems. This test (and more ones to follow) are the operations suppliers will be doing as part of their assurance testing.
I appreciate though there is a pain in managing the java/docker dependency.

PR #741, opened 2020-07-24
Add Person Escort Record state machine

cwrw:
### Jira link

[P4-1591](https://dsdmoj.atlassian.net/browse/P4-1783)

### What?

I have added/removed/altered:

- [x] Added person escort record state machine
- [x] Added two new timestamp fields to person escort record for confirmed and printed state changes
- [x] Added ability to calculate state of person escort record on response update

### Why?

- The PersonEscortRecord state machine has been added to handle the different transitions of the PER, as well as add two new states: `not_started` and `printed`. Since we cannot have `not_started` in the enum values due to it overlapping with the negation generated methods `not_`, only expose that value in the serializer.

- To track when a PER is confirmed and printed, timestamps for each event have been added. After a state is changed to either set the timestamps respectively in the after callback.

- To calculate the state changes on a person escort record, utilise the same query for sections, but do not group them to the specified section to calculate the general state. If the person escort record is confirmed/printed, this method will
throw an error to stop the ability to calculate the state, as this is no longer editable and can only be done by the patch endpoint.

- When a response is updated, the PER state should be recalculated. If the PER is either confirmed/printed, it should not be updated, and returns a forbidden error instead.

### Have you? (optional)

### DIFF
--- a/app/models/framework_response.rb
+++ b/app/models/framework_response.rb
@@ -32,7 +32,11 @@ class FrameworkResponse < VersionedModel
 
       update!(flags: build_flags)
       clear_dependent_values_and_flags!(old_value)
-      person_escort_record.update_state!
+
+      # lock the state update to avoid race condition on multiple response patches
+      person_escort_record.with_lock do
+        person_escort_record.update_state!
+      end
     end
   rescue FiniteMachine::InvalidStateError
     raise ActiveRecord::ReadOnlyRecord, "Can't update framework_responses because person_escort_record is #{person_escort_record.state}"
--- a/app/models/framework_response.rb
+++ b/app/models/framework_response.rb
@@ -33,13 +33,13 @@ class FrameworkResponse < VersionedModel
       update!(flags: build_flags)
       clear_dependent_values_and_flags!(old_value)
 
-      # lock the state update to avoid race condition on multiple response patches
+      # lock the status update to avoid race condition on multiple response patches
       person_escort_record.with_lock do
-        person_escort_record.update_state!
+        person_escort_record.update_status!
       end
     end
   rescue FiniteMachine::InvalidStateError
-    raise ActiveRecord::ReadOnlyRecord, "Can't update framework_responses because person_escort_record is #{person_escort_record.state}"
+    raise ActiveRecord::ReadOnlyRecord, "Can't update framework_responses because person_escort_record is #{person_escort_record.status}"
   end
 
 private
--- a/app/models/framework_response.rb
+++ b/app/models/framework_response.rb
@@ -32,7 +32,10 @@ class FrameworkResponse < VersionedModel
 
       update!(flags: build_flags)
       clear_dependent_values_and_flags!(old_value)
+      person_escort_record.update_state!
     end
+  rescue FiniteMachine::InvalidStateError
+    raise ActiveRecord::ReadOnlyRecord, "Can't update framework_responses because person_escort_record is #{person_escort_record.state}"
   end
 
 private
--- a/app/serializers/person_escort_record_serializer.rb
+++ b/app/serializers/person_escort_record_serializer.rb
@@ -4,7 +4,7 @@ class PersonEscortRecordSerializer < ActiveModel::Serializer
   has_many :framework_responses, serializer: FrameworkResponseSerializer, key: :responses
   has_many :flags
 
-  attributes :version, :status, :confirmed_at, :printed_at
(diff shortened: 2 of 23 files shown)


app/serializers/person_escort_record_serializer.rb
```diff
@@ -19,6 +18,14 @@ def framework_responses
     object.framework_responses.includes(:flags, framework_question: :framework)
   end
 
+  def status
+    if object.state == 'unstarted'
+      'not_started'
```
cwrw, line 23: we cannot use `not_started` in the enum as negative scopes and enum elements prefixed with "not_" will cause a warning, so only surface here

app/models/framework_response.rb
```diff
@@ -32,7 +32,10 @@ def update_with_flags!(value)
 
       update!(flags: build_flags)
       clear_dependent_values_and_flags!(old_value)
+      person_escort_record.update_state!
     end
+  rescue FiniteMachine::InvalidStateError
+    raise ActiveRecord::ReadOnlyRecord, "Can't update framework_responses because person_escort_record is #{person_escort_record.state}"
```
cwrw, line 38: If an attempt to update a response when the PER is confirmed/printed, a forbidden error is raised for the record being read only now

app/state_machines/person_escort_record_state_machine.rb
```diff
@@ -0,0 +1,22 @@
+class PersonEscortRecordStateMachine < FiniteMachine::Definition
+  initial :unstarted
+
+  event :complete, %i[unstarted in_progress completed] => :completed
```
cwrw, line 4: we have to allow completed to be run on completed if a response is updated, and the same for in_progress state as well to allow further update to responses

app/models/framework_response.rb
```diff
@@ -32,7 +32,14 @@ def update_with_flags!(value)
 
       update!(flags: build_flags)
       clear_dependent_values_and_flags!(old_value)
+
+      # lock the state update to avoid race condition on multiple response patches
+      person_escort_record.with_lock do
```
cwrw, line 37: when attempting to calculate the state, a race condition is occurring due to multiple response submission per page. For instance if two responses are sent at the same time, the state is calculated as in progress although it is now completed due to each response being seen as missing at the current time of calculation. Adding a lock prevents this from happening and waits for the first status to be calculated, then the second making sure all response values have been updated.

app/state_machines/person_escort_record_state_machine.rb
```diff
@@ -0,0 +1,22 @@
...
+  event :uncomplete, %i[unstarted in_progress completed] => :in_progress
+  event :confirm, completed: :confirmed
+  event :to_print, confirmed: :printed
```
cwrw, line 7: `print` is a reserved word so had to use something else

app/serializers/person_escort_record_serializer.rb
smoothcontract: Presumably the state is never sent back to the back end in a request, so one way conversion suffices?

app/state_machines/person_escort_record_state_machine.rb
smoothcontract: Even so, I don't think you need to map that state transition. Does calling `complete` when the state is already completed raise an exception or is it ignored instead? If it raises an exception then I'd agree you need this.

smoothcontract: Maybe this is the wrong approach to take with the state machine. It might be cleaner/preferable to add guard clauses on the transition and add a simple `update` event in the state machine. You can always call that from the controller, and within the state machine check whether progress is complete or not and make the correct transition. Something like this:

```ruby
event :update, %i[unstarted in_progress] => :completed, if: -> { target.progress == PERSON_ESCORT_RECORD_IN_PROGRESS }
event :update, %i[completed] => :in_progress, if -> { target.progress == PERSON_ESCORT_RECORD_IN_PROGRESS }
```

Then in the controller

```ruby
def update_status!
  self.progress = set_progress(sections_to_responded)
  state_machine.update!
  save!
end
```

Something to think about, but I believe this is a more idiomatic approach for the state machine.

app/serializers/person_escort_record_serializer.rb
smoothcontract: Discussed offline, front end never sends this attribute back.

PR #742, opened 2020-07-24
P4-1591 Add Person escort record PATCH endpoint

cwrw:
### Jira link

[P4-1783](https://dsdmoj.atlassian.net/browse/P4-1783)

### What?

- [x] Added PATCH endpoint for person escort record
- [x] Added param validator for endpoint
- [x] Added state updater in person escort record

### Why?

Add new PATCH endpoint for person escort record, which accepts a status, which could be `confirmed` or `printed`, and updates the person escort record state. It also sets the timestamps for each respectively.

If either the start state is invalid (`completed -> confirmed` or `confirmed -> printed`), or any other params are passed in, it returns an error.

### Have you? (optional)

### DIFF
--- a/app/controllers/api/person_escort_records_controller.rb
+++ b/app/controllers/api/person_escort_records_controller.rb
@@ -23,16 +23,13 @@ module Api
     end
 
     def update
-      PersonEscortRecords::ParamsValidator.new(update_person_escort_record_attributes).validate!
-      person_escort_record = PersonEscortRecord.find(params[:id])
-      person_escort_record.set_status!(update_person_escort_record_attributes)
+      PersonEscortRecords::ParamsValidator.new(update_person_escort_record_status).validate!
+      person_escort_record.confirm!(update_person_escort_record_status)
 
       render_person_escort_record(person_escort_record, :ok)
     end
 
     def show
-      person_escort_record = PersonEscortRecord.find(params[:id])
-
       render_person_escort_record(person_escort_record, :ok)
     end
 
@@ -46,12 +43,16 @@ module Api
       params.require(:data).permit(UPDATE_PERMITTED_PER_PARAMS)
     end
 
-    def update_person_escort_record_attributes
+    def update_person_escort_record_status
       update_person_escort_record_params.to_h.dig(:attributes, :status)
     end
 
     def supported_relationships
       PersonEscortRecordSerializer::SUPPORTED_RELATIONSHIPS
+    end
+
+    def person_escort_record
+      @person_escort_record ||= PersonEscortRecord.find(params[:id])
     end
 
     def render_person_escort_record(person_escort_record, status)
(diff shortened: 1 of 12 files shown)


app/controllers/api/person_escort_records_controller.rb
```diff
@@ -17,6 +22,14 @@ def create
       render_person_escort_record(person_escort_record, :created)
     end
 
+    def update
+      PersonEscortRecords::ParamsValidator.new(update_person_escort_record_attributes).validate!
+      person_escort_record = PersonEscortRecord.find(params[:id])
```
smoothcontract, line 27: Might be nice to add a `person_escort_record` method as this is now used in a couple of places.

```ruby
def person_escort_record
  @person_escort_record ||= PersonEscortRecord.find(params[:id])
end
```

app/models/person_escort_record.rb
```diff
@@ -77,6 +77,17 @@ def update_status!
     save!
   end
 
+  def set_status!(new_status)
```
smoothcontract, line 80: This feels a little strange... it only supports changing status to one value but saves the PER regardless. Maybe renaming the method would make the intent clearer.. or perhaps from the controller call `confirm!` directly.

app/controllers/api/person_escort_records_controller.rb
```diff
@@ -8,6 +8,11 @@ class PersonEscortRecordsController < ApiController
       relationships: [profile: {}],
     ].freeze
 
+    UPDATE_PERMITTED_PER_PARAMS = [
+      :type,
+      attributes: [:status],
+    ].freeze
+
     def create
       person_escort_record = PersonEscortRecord.save_with_responses!(
```
martyn-w, line 17: should the `PersonEscortRecords::ParamsValidator` also be called on create?

app/controllers/api/person_escort_records_controller.rb
```diff
@@ -17,9 +22,14 @@ def create
       render_person_escort_record(person_escort_record, :created)
     end
 
-    def show
-      person_escort_record = PersonEscortRecord.find(params[:id])
+    def update
+      PersonEscortRecords::ParamsValidator.new(update_person_escort_record_attributes).validate!
+      person_escort_record.confirm!(update_person_escort_record_attributes)
```
martyn-w, line 27: is an update _always_ a confirmation ?

cwrw: yes, a client can only confirm a PER, and can't change anything else on the status/attributes on the record

app/controllers/api/person_escort_records_controller.rb
cwrw: the validator is only specific to the update action, and not to the create values which are validated differently in the model as they are all relationships instead of attributes.

PR #743, opened 2020-07-27
[P4-1967] Fixes case sensitivity for all person identifiers

willfish:
### Jira link

P4-1967

### What?

The v2 finder doesn't use the same mechanism as the Nomis Importer to pull out the prison numbers and we had no tests that covered us actually filtering by prison number.

- [x] Adds a test for v2 filtering by prison number
- [x] Fixes bug in filtering prison numbers that are downcased with migration to case insensitive citext.
- [x] Indicate case insensitivity in swagger documentation

V1 doesn't also need to be fixed as it is currently unused.

### Why?

Users have noticed that they can't search people when the prison number is downcased.

### DIFF
--- a/app/services/v2/people/finder.rb
+++ b/app/services/v2/people/finder.rb
@@ -19,6 +19,14 @@ module V2
         filter_params[name]&.split(',')
       end
 
+      def split_and_upcase_params(name)
+        params = split_params(name)
+
+        return params&.map(&:upcase) if name == :prison_number
+
+        params
+      end
+
       def apply_filters(scope)
         scope = scope.includes(:profiles, :ethnicity, :gender)
         %i[police_national_computer criminal_records_office prison_number].each do |param|
@@ -29,7 +37,7 @@ module V2
       end
 
       def apply_filter(param, scope)
-        scope = scope.where(param => split_params(param)) if filter_params.key?(param)
+        scope = scope.where(param => split_and_upcase_params(param)) if filter_params.key?(param)
         scope
       end
     end
--- a/app/services/v2/people/finder.rb
+++ b/app/services/v2/people/finder.rb
@@ -19,14 +19,6 @@ module V2
         filter_params[name]&.split(',')
       end
 
-      def split_and_upcase_params(name)
-        params = split_params(name)
-
-        return params&.map(&:upcase) if name == :prison_number
-
-        params
-      end
-
       def apply_filters(scope)
         scope = scope.includes(:profiles, :ethnicity, :gender)
         %i[police_national_computer criminal_records_office prison_number].each do |param|
@@ -37,7 +29,7 @@ module V2
       end
 
       def apply_filter(param, scope)
-        scope = scope.where(param => split_and_upcase_params(param)) if filter_params.key?(param)
+        scope = scope.where(param => split_params(param)) if filter_params.key?(param)
         scope
       end
     end
--- a/db/migrate/20200727143634_change_people_identifiers_case_insensitive.rb
+++ b/db/migrate/20200727143634_change_people_identifiers_case_insensitive.rb
@@ -1,0 +1,15 @@
+class ChangePeopleIdentifiersCaseInsensitive < ActiveRecord::Migration[6.0]
+  def up
+    change_column :people, :nomis_prison_number, :citext
+    change_column :people, :prison_number, :citext
+    change_column :people, :criminal_records_office, :citext
+    change_column :people, :police_national_computer, :citext
+  end
+
+  def down
+    change_column :people, :nomis_prison_number, :text
+    change_column :people, :prison_number, :text
+    change_column :people, :criminal_records_office, :text
+    change_column :people, :police_national_computer, :text
+  end
+end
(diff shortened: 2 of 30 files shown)


app/services/v2/people/finder.rb
```diff
@@ -19,6 +19,14 @@ def split_params(name)
         filter_params[name]&.split(',')
       end
 
+      def split_and_upcase_params(name)
+        params = split_params(name)
+
+        return params&.map(&:upcase) if name == :prison_number
```
alexdesi, line 25: It sounds strange to upcase the params only for `prison_number`,
also instead of upcase the filter, is it not better to perform a case insensitive search?

spec/requests/api/people_controller_index_v2_spec.rb
```diff
@@ -47,6 +47,14 @@
       let(:query) { '?filter[prison_number]=G3239GV,GV345VG' }
       let(:import_from_nomis) { instance_double('People::ImportFromNomis', call: nil) }
 
+      let!(:people) do
+        [
+          create(:person, prison_number: 'G3239GV'),
+          create(:person, prison_number: 'GV345VG'),
+          create(:person, prison_number: 'flibble'),
```
willfish, line 54:
```suggestion
          create(:person, prison_number: 'FLIBBLE'),
```

app/services/v2/people/finder.rb
willfish: Chatted with Alessandro and Dom about this. Prison numbers can never be downcased so this is fine. We're going to review the other identifiers separately to potentially simplify this method - stop special casing prison_number

cwrw: I suggest allowing a case insensitive search instead, by utilising the [CITEXT](https://www.postgresql.org/docs/9.1/citext.html) extension on Postgres instead?

spec/requests/api/people_controller_index_v2_spec.rb
```diff
@@ -47,6 +47,14 @@
       let(:query) { '?filter[prison_number]=G3239GV,GV345VG' }
       let(:import_from_nomis) { instance_double('People::ImportFromNomis', call: nil) }
 
+      let!(:people) do
+        [
+          create(:person, prison_number: 'G3239GV'),
```
teneightfive, line 52: Would it be worth creating some test cases for other identifiers too?

app/services/v2/people/finder.rb
willfish: I don't really understand the advantage?

willfish: Okay, I've been convinced. I'll make all the identifiers case insensitive. Thanks @alexdesi and @cwrw :)

spec/requests/api/people_controller_index_v2_spec.rb
willfish: Done.

db/schema.rb
```diff
@@ -382,10 +382,10 @@
   create_table "people", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.string "nomis_prison_number"
-    t.string "prison_number"
-    t.string "criminal_records_office"
-    t.string "police_national_computer"
+    t.citext "nomis_prison_number"
```
willfish, line 385: I've run migrations locally and the values are casted correctly as far as I could see. Rolling back works cleanly, too.

db/migrate/20200727143634_change_people_identifiers_case_insensitive.rb
```diff
@@ -0,0 +1,15 @@
+class ChangePeopleIdentifiersCaseInsensitive < ActiveRecord::Migration[6.0]
+  def up
+    change_column :people, :nomis_prison_number, :citext
+    change_column :people, :prison_number, :citext
+    change_column :people, :criminal_records_office, :citext
+    change_column :people, :police_national_computer, :citext
+  end
+
+  def down
+    change_column :people, :nomis_prison_number, :text
```
martyn-w, line 10: these are strings, not text

db/migrate/20200727143634_change_people_identifiers_case_insensitive.rb
```diff
@@ -0,0 +1,15 @@
+class ChangePeopleIdentifiersCaseInsensitive < ActiveRecord::Migration[6.0]
+  def up
+    change_column :people, :nomis_prison_number, :citext
```
martyn-w, line 3: I don't `citext` think this is the right datatype for these fields

willfish:
So we aren't specifying a length for the varchar so I can't see any benefit of using it over text, anyway.

willfish:
Apparently, they use the same data structure under the hood and there's no performance benefit from using one over the other.

willfish:
I think CITEXT is fine and clean and is faster than using LOWER from what I've read.

willfish:
ILIKE is also slow and not generally recommended

willfish:
I've fixed the rollback datatype :)

PR #744, opened 2020-07-27
Allow prison to prison transfers (allocations) to be created in Nomis

smoothcontract:
### Jira link

P4-1949

### What?

- [x] Add new `create_in_nomis` parameter to existing `POST /move_events/approve' endpoint
- [x] Add new `Allocations::CreateInNomis` service
- [x] Add new `NomisClient::Allocations` client service

### Why?

- This is to support creating prison to prison transfers (aka allocations) in Nomis automatically when the singleton move request is approved in the front end. Further PR's will do the same when updating existing allocation profiles.

- Note that the front end always receives a success response regardless of the Nomis response. We do not wish to block front end in the event of Nomis failure, so must return a success response. Unfortunately we do not have a mechanism to return a success response with associated warnings (e.g. the move was approved but Nomis was not updated).

### Have you? (optional)

- Integrates with a new (or at least not used yet) Nomis endpoint, so will require end to end testing
- Introduces a new optional attribute to existing API endpoint, so does not affect production API

### DIFF
--- a/app/controllers/api/move_events_controller.rb
+++ b/app/controllers/api/move_events_controller.rb
@@ -8,7 +8,7 @@ module Api
     before_action :validate_idempotency_key
     around_action :idempotent_action
 
-    APPROVE_PARAMS = [:type, attributes: %i[timestamp date]].freeze
+    APPROVE_PARAMS = [:type, attributes: %i[timestamp date create_in_nomis]].freeze
     CANCEL_PARAMS = [:type, attributes: %i[timestamp cancellation_reason cancellation_reason_comment notes]].freeze
     LOCKOUT_PARAMS = [:type, attributes: %i[timestamp notes], relationships: { from_location: {} }].freeze
     REDIRECT_PARAMS = [:type, attributes: %i[timestamp notes], relationships: { to_location: {} }].freeze
--- a/app/lib/nomis_client/allocations.rb
+++ b/app/lib/nomis_client/allocations.rb
@@ -8,18 +8,7 @@ module NomisClient
 
         NomisClient::Base.post(allocations_path, body: body_params.to_json)
       rescue OAuth2::Error => e
-        Raven.capture_message(
-          'Allocations::CreateInNomis Error!',
-          extra: {
-            allocations_route: allocations_path,
-            body_params: body_params,
-            nomis_response: {
-              status: e.response.status,
-              body: e.response.body,
-            },
-          },
-          level: 'error',
-        )
+        log_exception('Allocations::CreateInNomis Error!', allocations_path, body_params, e)
 
         e.response
       end
--- a/app/lib/nomis_client/allocations.rb
+++ b/app/lib/nomis_client/allocations.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+module NomisClient
+  class Allocations < NomisClient::Base
+    class << self
+      def post(booking_id:, body_params: {})
+        allocations_path = "/bookings/#{booking_id}/prison-to-prison"
+
+        NomisClient::Base.post(allocations_path, body: body_params.to_json)
+      rescue OAuth2::Error => e
+        Raven.capture_message(
+          'Allocations::CreateInNomis Error!',
+          extra: {
+            allocations_route: allocations_path,
+            body_params: body_params,
+            nomis_response: {
+              status: e.response.status,
+              body: e.response.body,
+            },
+          },
+          level: 'error',
+        )
+
+        e.response
+      end
+    end
+  end
+end
--- a/app/models/move_event.rb
+++ b/app/models/move_event.rb
@@ -1,8 +1,12 @@
 class MoveEvent < Event
   # NB: this class exposes a few methods specific to moves to the Event model
 
+  def create_in_nomis?
+    @create_in_nomis ||= option_selected?(:create_in_nomis)
+  end
+
   def rebook?
-    @rebook ||= event_params.dig(:attributes, :rebook).to_s == 'true'
+    @rebook ||= option_selected?(:rebook)
   end
 
   def date
@@ -19,5 +23,11 @@ class MoveEvent < Event
 
   def cancellation_reason_comment
     @cancellation_reason_comment ||= event_params.dig(:attributes, :cancellation_reason_comment)
+  end
+
+private
+
+  def option_selected?(attribute_name)
+    event_params.dig(:attributes, attribute_name).to_s == 'true'
   end
 end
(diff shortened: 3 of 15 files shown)


app/controllers/api/move_events_controller.rb
```diff
@@ -8,7 +8,7 @@ class MoveEventsController < ApiController
     before_action :validate_idempotency_key
     around_action :idempotent_action
 
-    APPROVE_PARAMS = [:type, attributes: %i[timestamp date]].freeze
+    APPROVE_PARAMS = [:type, attributes: %i[timestamp date create_in_nomis]].freeze
```
willfish, line 11: This feels like a weird place for a special snowflake optional parameter. How do you feel about a query param or making it part of meta? The resource attributes usually map directly onto the entity attributes in the database, etc?

app/lib/nomis_client/allocations.rb
```diff
@@ -0,0 +1,28 @@
...
+    class << self
+      def post(booking_id:, body_params: {})
+        allocations_path = "/bookings/#{booking_id}/prison-to-prison"
+
+        NomisClient::Base.post(allocations_path, body: body_params.to_json)
+      rescue OAuth2::Error => e
+        Raven.capture_message(
+          'Allocations::CreateInNomis Error!',
+          extra: {
+            allocations_route: allocations_path,
```
willfish, line 14: Could we generalise this to path and have this exception method in the base client or something that makes other similarly propagated exceptions conform to the same schema?

app/models/move.rb
```diff
@@ -119,6 +119,10 @@ def rejected?
     cancellation_reason == CANCELLATION_REASON_REJECTED
   end
 
+  def nomis_event_id
```
willfish, line 122: What's the reasoning behind last, here? I'm wondering whether we could flag the meaning better with a rename of this method: `most_recent_nomis_event_id`?

app/services/event_log/move_runner.rb
```diff
@@ -16,6 +16,7 @@ def call
         when Event::APPROVE
           move.status = Move::MOVE_STATUS_REQUESTED
           move.date = event.date
+          Allocations::CreateInNomis.call(move) if event.create_in_nomis?
```
willfish, line 19:
```suggestion
          Allocations::CreateInNomis.call(move) if event.create_in_nomis? && move.latest_nomis_booking_id
```

willfish, line 19:
```suggestion
          Allocations::CreateInNomis.call(move) if event.create_in_nomis? && move&.person&.latest_nomis_booking_id.present?
```

spec/models/move_spec.rb
```diff
@@ -219,8 +219,28 @@
...
+
+      it 'returns the most recent nomis_event_id' do
+        expect(move.nomis_event_id).to eq(654_321)
+      end
+    end
+  end
+
   describe '#nomis_event_id=' do
-    subject(:move) { create :move }
+    subject(:move) { build :move }
```
willfish, line 243: 🎉

app/models/move_event.rb
```diff
@@ -1,8 +1,12 @@
 class MoveEvent < Event
   # NB: this class exposes a few methods specific to moves to the Event model
 
+  def create_in_nomis?
```
cwrw, line 4: can you add a test for this? (although rest of those aren't tested previously 🤔 )

app/models/move.rb
cwrw: this is slightly confusing as we have a `nomis_event_id` column on the model as well, so not sure if that is the most recent?

app/services/allocations/create_in_nomis.rb
```diff
@@ -0,0 +1,28 @@
+module Allocations
+  class CreateInNomis
+    def self.call(move)
+      booking_id = move.person.latest_nomis_booking_id
+
+      body = {
+        booking_id: booking_id,
+        body_params: {
+          fromPrisonLocation: move.from_location.nomis_agency_id,
```
cwrw, line 9: could we ever have an empty nomis agency id?

app/services/allocations/create_in_nomis.rb
```diff
@@ -0,0 +1,28 @@
...
+          toPrisonLocation: move.to_location.nomis_agency_id,
```
cwrw, line 10: if we don't have a to location, will that be ok with nomis (I know we have a type of move that allows this to be nil) but wondering what the implications are on nomis?

app/controllers/api/move_events_controller.rb
smoothcontract: Not sure I agree. This will be a checkbox on the form that the user ticks if they want the event created in Nomis, or leave unticked if they don't. So it seems reasonable to treat this in the same way as the date input on the same form.

app/models/move.rb
smoothcontract: Agree that this is confusing - it's a long story but this PR explains things. Once merged I can rebase and remove this custom method. https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/747

app/models/move_event.rb
smoothcontract: Discussed offline - has coverage via the move runner specs for true/false scenarios.

app/services/allocations/create_in_nomis.rb
smoothcontract: No, the location data is all imported from Nomis and is returned from the Nomis API with an id (primary key) for each location.

app/services/allocations/create_in_nomis.rb
smoothcontract: Good catch. Will add coverage and a nil guard for this

app/services/event_log/move_runner.rb
smoothcontract: I'd prefer to add that check inside the service. Will update.

spec/models/move_spec.rb
smoothcontract: Every optimisation helps... but this is soon to be removed completely 💣

spec/models/move_spec.rb
```diff
@@ -219,8 +219,28 @@
     it { is_expected.to validate_presence_of(:reference) }
   end
 
+  describe '#nomis_event_id' do
```
smoothcontract, line 222: This will be removed after this gets merged: https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/747

app/controllers/api/move_events_controller.rb
willfish: Beyond a subjective sense of how things should be I don't really have much more to say 🎉 .

spec/models/move_spec.rb
willfish: 👍

PR #745, opened 2020-07-28
P4-1783 Add webhook on confirming a person escort record

cwrw:
jira-ticket: https://dsdmoj.atlassian.net/browse/P4-1783

If a person escort record is confirmed, send an update move webhook and email. We have written up a proposal for suppliers,  if using the same `update_move` notification is not preferable we can create a specific action name instead (we will wait for their feedback).

### DIFF
--- a/app/jobs/prepare_person_escort_record_notifications_job.rb
+++ b/app/jobs/prepare_person_escort_record_notifications_job.rb
@@ -4,7 +4,8 @@ class PreparePersonEscortRecordNotificationsJob < ApplicationJob
   queue_as :notifications
 
   def perform(topic_id:, action_name:)
-    move = PersonEscortRecord.find(topic_id).profile.move
+    return unless (move = PersonEscortRecord.find(topic_id).profile.move)
+
     PrepareMoveNotificationsJob.perform_now(topic_id: move.id, action_name: action_name)
   end
 end
--- a/app/jobs/prepare_person_escort_record_notifications_job.rb
+++ b/app/jobs/prepare_person_escort_record_notifications_job.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class PreparePersonEscortRecordNotificationsJob < ApplicationJob
+  queue_as :notifications
+
+  def perform(topic_id:, action_name:)
+    move = PersonEscortRecord.find(topic_id).profile.move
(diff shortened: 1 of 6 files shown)


app/jobs/prepare_person_escort_record_notifications_job.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class PreparePersonEscortRecordNotificationsJob < ApplicationJob
+  queue_as :notifications
+
+  def perform(topic_id:, action_name:)
+    move = PersonEscortRecord.find(topic_id).profile.move
```
martyn-w, line 7: profiles can exist without moves; is there any danger here of the move being null?

spec/requests/api/person_escort_records_controller_update_spec.rb
```diff
@@ -6,6 +6,12 @@
   describe 'PATCH /person_escort_records/:person_escort_record_id' do
     include_context 'with supplier with spoofed access token'
 
+    subject(:patch_person_escort_record) do
+      patch "/api/v1/person_escort_records/#{person_escort_record_id}", params: person_escort_record_params, headers: headers, as: :json
```
martyn-w, line 10: we're moving away from the versioned urls, use `/api/person_escort_records/#{person_escort_record_id}"` instead (it will go to the same place)

spec/requests/api/person_escort_records_controller_update_spec.rb
```diff
@@ -6,6 +6,12 @@
   describe 'PATCH /person_escort_records/:person_escort_record_id' do
     include_context 'with supplier with spoofed access token'
 
+    subject(:patch_person_escort_record) do
```
willfish, line 9: This is a new style for making the request. Currently we have:

1. Directly in before block
2. Directly in before block with skip before
3. Not in before block (and just in examples)
4. Directly in the before block with method
5. Directly in the before block with method and skip before
6. Directly in the before block with subject let

I'm not saying this is a bad way to do it. Just that we have so many permutations of making a request in our request specs 🤔

martyn-w: I don't think it's a new style?  see e.g. `/api/v1/person_escort_records/#{person_escort_record_id}"` ?

willfish: Ah, in PRs I've reviewed.

PR #747, opened 2020-07-28
Remove dead nomis_event_ids attribute from moves

smoothcontract:
### Jira link

P4-1949 P4-1499

### What?
- [x] Removed `nomis_event_ids` from Move model
- [x] Removed overridden assignment of `nomis_event_id` attribute on same model
- [x] Removed `from_nomis?` method and associated checks for read only moves

### Why?

It seems that in the distant past, Nomis events were imported from Nomis and used to create associated moves in PECS. When first implemented, the original Nomis event id was stored in the move `nomis_event_id` attribute as these have a 1:1 mapping with moves. Also these imported moves were treated as read only so that they could not be modified via the API (which would presumably be a bad thing).

Time passed...

Subsequently it was discovered that court hearings break the original 1:1 mapping approach with the same prisoner potentially having multiple court hearings (for different offences) on the same date and location, which means a move cannot be created for each event. Nomis stores each individual court hearing as a separate event; so at that point `nomis_event_ids` was added to store an array of all associated Nomis events for the move. A rake task was added to back fill this array attribute on existing data, and the `nomis_event_ids` array was then used to determine if a move originated from Nomis.

Time passed...

At some point it was further decided not to import events from Nomis to create moves any longer. This meant that in production a number of unwanted moves were left that needed to be cancelled. Another rake task was built for this. There is no remaining evidence of the Nomis event importer in the code base so presumably this was removed, although the legacy `nomis_event_id` and `nomis_event_ids` attributes plus checks for enforcing read only access to moves from Nomis were left behind. This looks to be an oversight as there is no reference to these attributes in the code which means they are not exposed via the API.

In an attempt to clean up the residual mess I have removed the `nomis_event_ids` attribute. There will be ancient cancelled moves in production that were from Nomis originally that will hence lose the association with the original Nomis event - however this data is unwanted and probably should have been deleted completely instead. The original `nomis_event_id` attribute however does have some future use, as it will be repurposed to store the created prison to prison transfer event id. This means that the custom accessor code for this attribute can be removed and it can revert to a simple integer type.

- Drops a redundant attribute of production data (which was never exposed in the API). The only potential knock on impact from this would be that it may now be possible to modify legacy cancelled Nomis moves (but perhaps the front end prevents amendments to cancelled moves - unsure on that).

### DIFF
--- a/app/controllers/api/v2/moves_actions.rb
+++ b/app/controllers/api/v2/moves_actions.rb
@@ -24,16 +24,14 @@ module Api::V2
     end
 
     def update_and_render
-      raise ActiveRecord::ReadOnlyRecord, "Can't change moves coming from Nomis" if move.from_nomis?
+      move.assign_attributes(common_move_attributes)
+      action_name = move.status_changed? ? 'update_status' : 'update'
+      move.save!
+      move.allocation&.refresh_status_and_moves_count!
 
-      @move.assign_attributes(common_move_attributes)
-      action_name = @move.status_changed? ? 'update_status' : 'update'
-      @move.save!
-      @move.allocation&.refresh_status_and_moves_count!
+      Notifier.prepare_notifications(topic: move, action_name: action_name)
 
-      Notifier.prepare_notifications(topic: @move, action_name: action_name)
-
-      render_move(@move, :ok)
+      render_move(move, :ok)
     end
 
   private
--- a/db/migrate/20200728090010_remove_nomis_event_ids_from_moves.rb
+++ b/db/migrate/20200728090010_remove_nomis_event_ids_from_moves.rb
@@ -1,0 +1,9 @@
+class RemoveNomisEventIdsFromMoves < ActiveRecord::Migration[6.0]
+  def up
+    remove_column :moves, :nomis_event_ids
+  end
+
+  def down
+    add_column :moves, :nomis_event_ids, :integer, array: true, null: false, default: []
+  end
+end
--- a/app/controllers/api/v1/moves_actions.rb
+++ b/app/controllers/api/v1/moves_actions.rb
@@ -25,8 +25,6 @@ module Api::V1
     end
 
     def update_and_render
-      raise ActiveRecord::ReadOnlyRecord, 'Can\'t change moves coming from Nomis' if move.from_nomis?
-
       updater.call
 
       Notifier.prepare_notifications(topic: updater.move, action_name: updater.status_changed ? 'update_status' : 'update')
--- a/app/models/move.rb
+++ b/app/models/move.rb
@@ -87,7 +87,6 @@ class Move < VersionedModel
 
   before_validation :set_reference
   before_validation :set_move_type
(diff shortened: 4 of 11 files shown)


app/controllers/api/v2/moves_actions.rb
```diff
@@ -24,16 +24,14 @@ def create_and_render
...
 
     def update_and_render
-      raise ActiveRecord::ReadOnlyRecord, "Can't change moves coming from Nomis" if move.from_nomis?
-
-      @move.assign_attributes(common_move_attributes)
-      action_name = @move.status_changed? ? 'update_status' : 'update'
-      @move.save!
-      @move.allocation&.refresh_status_and_moves_count!
+      move.assign_attributes(common_move_attributes)
+      action_name = move.status_changed? ? 'update_status' : 'update'
```
smoothcontract, line 28: `@move` is already memoized in the `move` method so this is just as fast as before, but cleaner (and resolves some strangeness where this method initially refers to `move` and then `@move`, which is what broke the specs originally).

app/controllers/api/v2/moves_actions.rb
```diff
@@ -24,16 +24,14 @@ def create_and_render
     end
 
     def update_and_render
-      raise ActiveRecord::ReadOnlyRecord, "Can't change moves coming from Nomis" if move.from_nomis?
```
willfish, line 27: Do we need to update the swagger docs to remove this error?

willfish:
Absolutely love the context in the PR description :heart:

db/migrate/20200728090010_remove_nomis_event_ids_from_moves.rb
```diff
@@ -0,0 +1,9 @@
+class RemoveNomisEventIdsFromMoves < ActiveRecord::Migration[6.0]
+  def up
+    remove_column :moves, :nomis_event_ids
+  end
+
+  def down
+    add_column :moves, :nomis_event_ids, :integer, array: true, null: false, default: [], index: true
```
willfish, line 7: Should we be removing an index in the up method?

willfish:
```suggestion
    add_column :moves, :nomis_event_ids, :integer, array: true, null: false, default: []
```

smoothcontract: Good catch, updated.

smoothcontract:
> Thanks for the cleanup and thorough investigation. The pr looks good, however I was wondering around a future need to move back to having multiple `nomis_event_ids` again. Maybe it's safer to keep the array and remove the 1:1 relationship instead, as the one to many concept exists generally in NOMIS just incase, to avoid a rework again?

The only scenario where more than one event occurs on a given date/time/prisoner in NOMIS is a court hearing. We have those events in a dedicated model with each created NOMIS event held in the `nomis_case_id` and `nomis_hearing_id` attributes. It's confusing having two attributes in the Move model with very similar names; this PR aims to fix that confusion raised on my other PR.

PR #748, opened 2020-07-29
P4-1793 Add `ready_for_transit` filter on moves

cwrw:
### Jira link

[P4-1793](https://dsdmoj.atlassian.net/browse/P4-1793)

### What?

I have added/removed/altered:

- [x] Added a new `ready_for_transit` filter on moves endpoint

### Why?

Add new filter to return all moves ready to be in transit if set to true. Moves ready for transit are moves where the person escort record is confirmed.

A move is not ready to be in transit if there is no person escort record attached, or if the status is not confirmed. This filter will be used on the FE dashboard for moves.

### Have you? (optional)

- Adds a filter to an api that is used in production

### DIFF
--- a/spec/factories/move.rb
+++ b/spec/factories/move.rb
@@ -120,6 +120,21 @@ FactoryBot.define do
       end
     end
 
+    trait :with_person_escort_record do
+      transient do
+        person_escort_record_status { 'unstarted' }
+      end
+
+      after(:create) do |move, evaluator|
+        create(
+          :person_escort_record,
+          profile: move.profile,
+          status: evaluator.person_escort_record_status,
+          confirmed_at: evaluator.person_escort_record_status == 'confirmed' ? Time.zone.now : nil,
+        )
+      end
+    end
+
     trait :with_date_to do
       date_to { date + 3.days }
     end
--- a/app/controllers/api/moves_controller.rb
+++ b/app/controllers/api/moves_controller.rb
@@ -25,7 +25,7 @@ module Api
     end
 
     PERMITTED_FILTER_PARAMS = %i[
-      date_from date_to created_at_from created_at_to location_type status from_location_id to_location_id supplier_id move_type cancellation_reason has_relationship_to_allocation
+      date_from date_to created_at_from created_at_to location_type status from_location_id to_location_id supplier_id move_type cancellation_reason has_relationship_to_allocation ready_for_transit
     ].freeze
 
     def filter_params
--- a/app/services/moves/finder.rb
+++ b/app/services/moves/finder.rb
@@ -46,6 +46,7 @@ module Moves
       scope = apply_date_range_filters(scope)
       scope = apply_location_type_filters(scope)
       scope = apply_allocation_relationship_filters(scope)
+      scope = apply_ready_for_transit_filters(scope)
       scope = apply_filter(scope, :supplier_id)
       scope = apply_filter(scope, :from_location_id)
       scope = apply_filter(scope, :to_location_id)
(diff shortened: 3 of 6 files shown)


spec/factories/move.rb
```diff
@@ -120,6 +120,21 @@
       end
     end
 
+    trait :with_person_escort_record do
+      transient do
+        person_escort_record_status { 'unstarted' }
```
cwrw, line 125: adding a transient value will allow optionally passing in the `person_escort_record_status` when creating the factory

martyn-w:
I think the filter name `ready_for_transit` could easily be confused with the move status `in_transit`.

The current move statuses are: Proposed -> Requested -> Booked -> In Transit -> Completed / Cancelled.
This filter seems to ignore the move status and rely solely on the PER status? I think the two need to be combined somehow,

cwrw:
@martyn-w The naming of this filter was indeed tricky but the reasoning behind the name: `ready_for_transit` was chosen to avoid confusing it with a `status` on a **move**, on a **person escort record** as well as avoid tying the filter name to a person escort record, since we will in the future have further assessments for different roles which might not be called person escort record but something different. We also wanted to allow moves without a person escort record associated to be included in the false variation of the filter.

Also since we have a `status` filter as well on moves, I think this would conflict with the `ready_for_transit` if we were to combine both filters causing unintended behaviour. We can allow the status where a move is `ready_for_transit` to be decided by the client instead, where they could query for a `ready_for_transit` moves which have been cancelled or active moves which are `ready_for_transit`. 

I'm still open to suggestions though for a better name, as understand that it might not be very clear. Let me know your thoughts. 
cc @teneightfive

martyn-w:
Hi @cwrw,

Thanks for the clarifications. It doesn't seem logical that a cancelled move can be returned in a list of `ready_for_transit` moves. I think a different name would certainly help. How about `assessments_completed`? `checks_completed` ?or `evaluations_completed` 

That way, it is clear the filter relates to some processes have been done, and not that the move can be transitioned to some state.

cwrw:
@martyn-w We can't use anything appended with `completed` as it is a status on the PER which is not viable for transit unfortunately. Also this set of moves can include moves without an actual evaluation/PER as well. Im thinking we keep as is for now and revisit later when things evolve further

teneightfive:
Thanks for explaining the thought process we went through @cwrw. @martyn-w I'm not sure I agree that a cancelled move shouldn't show in this list. This filter is only to give the ability for a client to return any moves that have passed the checks we have deemed necessary in this instance for a move to have the `in_transit` status. 

I suppose the slightly conflicting thing is that we don't actually enforce these rules when a move becomes `in_transit`. However, there is nothing stopping us doing that in future.

The key thing for this filter is to communicate to the client that any moves returned in that list are either ready to be handed over to the supplier for the next state (_ready for transit_) or not. 

We landed on `ready_for_transit`, because it does in fact show that link to the transit move state, but without conflicting with status filters. 

I think if you introduce move statuses into this filter it gets a little messier so it felt nicer to separate the two and give the client the control whether to include or exclude cancelled moves in that list.

teneightfive:
For a bit more context, here is the filter we essentially need to be able to build (ignore ready to print - this has been removed):

<kbd></kbd>

- `Incomplete` = moves without a Person Escort Record **OR** moves with a Person Escort Record with a status of `not_started`, `in_progress` or `completed`
- `Ready to move` = moves with a Person Escort Record with a status of `confirmed`

In the future "Ready to move" will also include the same but for Youth Risk Assessment. A single filter of `ready_for_transit` gives us the ability to use this for both.

martyn-w:
Hi @cwrw, @teneightfive,

thanks for the further clarifications, and sorry to be so picky about such a small thing as a name. But I'm afraid that it still doesn't sit well: a move included by this filter `ready_for_transit` is *not* necessarily ready for `in_transit`.

(To be clear, only moves which are in a status of `booked` may progress to a status of `in_transit`.)

I agree with your reasons that adding the move status into this filter is not a good idea, and allowing the status to be filtered independently will give you maximum flexibility. I think if you can come up with another name, which does not imply the move can be moved to a status of `in_transit`, then all will be well.

In appreciate and accept your concerns about using the word `completed` in the filter name. So how about `assessments_done`? `assessments_finished`? `almost_ready`?! or indeed anything you can suggest which doesn't have the same connotations as `ready_for_transit`.

Keep in mind that although the filter will be used by the FE for now, it will be published in our API for all to see; so the name must have congruence with other filters and states.

martyn-w:
@teneightfive  Re:
> I suppose the slightly conflicting thing is that we don't actually enforce these rules when a move becomes in_transit. However, there is nothing stopping us doing that in future.

I think we should be adding a state machine to the move model (in the same way as we have on allocations and journeys) which will prevent a move transitioning from invalid states. I'd recommend that API clients/FE should assume that there _is_ a state machine, so that when we do add, nothing breaks.

martyn-w:
@cwrw re:
> I suggest we keep as is for now and revisit later when things evolve further?
unfortunately, once published that (theoretically) locks it into this version - it would be better to settle on the right name now

teneightfive:
@martyn-w what about `ready_for_handover`?

martyn-w:
Discussed offline with @cwrw and @teneightfive - we've agreed to stick with `ready_for_transit` on the basis that the documentation is very clear that additional filters on status are required in order to find moves which can be progressed to the `in_transit` status

PR #755, opened 2020-07-30
[P4-1498] Removes person_id from moves table

willfish:
### Jira link

P4-1498

### What?

- [x] Removes ignored_columns which we used to safely (i.e. non-destructively) remove person id attribute previously on 2020-07-03 09:56:39.
- [x] Removed person id from moves table and foreign key

### Why?

We no longer need the person id on the moves table since we now have a relationship on profile.

- Includes destructive/migratory actions which may effect critical offender data.

The above risk was significantly mitigated by use proving that the person id hasn't been in use for the last month nearly.


db/migrate/20200729191640_drop_person_id_from_moves.rb
```diff
@@ -0,0 +1,6 @@
+class DropPersonIdFromMoves < ActiveRecord::Migration[6.0]
+  def change
+    remove_foreign_key :moves, column: :person_id, name: 'fk_rails_moves_person_id', to_table: :people
```
willfish, line 3: Unfortunately the `remove_reference` wrapper for `remove_foreign_key`, `remove_index` and `remove_column` doesn't let you pass options to the remove_column method meaning we can't have rollback compatibility/set the type to `:uuid`.

PR #756, opened 2020-07-30
Adds postgresql-client to enable bin/rails dbconsole on all pods

willfish:
### Jira link

P4-<TODO>

### What?

Adds postgresql-client to all pods

### Why?

So that I can bring up raw dbconsole


Dockerfile
```diff
@@ -40,7 +40,12 @@ EXPOSE $PUMA_PORT
 RUN addgroup -g $APPUID -S appgroup && \
     adduser -u $APPUID -S appuser -G appgroup -h /app
 
-RUN apk add --update --no-cache tzdata postgresql-dev
+RUN apk add \
+      --update \
+      --no-cache \
+      tzdata \
+      postgresql-dev \
+      postgresql-client
```
mafonso, line 48: As a matter of principle I like to keep the image footprint to a minimum needed for the runtime environment.
But I also appreciate the convenience this brings.

On that trade-off I think it's OK to have this for now. We can always remove it at a later time when this need is no longer relevant.

willfish:
Thanks @mafonso . I think your instincts not to add clutter/surface area for attacks is a good one. We can achieve this with a small amount of inconvenience.

PR #757, opened 2020-07-30
P4-1943 Pull framework versions from repository

cwrw:
### Jira link

[P4-1943](https://dsdmoj.atlassian.net/browse/P4-1943)

### What?

- [x] Added git in Docker image as now we need it for the app 
- [x] Amended populating framework rake task to pull from Github tags if no filename/version provided

### Why?

If no version and filename supplied pull in all published tags from github to populate the framework. The repository is pulled into the temp directly and destroyed after the task runs. If a filename and version are supplied use instead otherwise throw an error.

tested on Dev environment:
```
~ $ bundle exec rake frameworks:populate_data
{"severity":"INFO","timestamp":"2020-07-30T15:24:35.440+00:00","msg":"Raven 2.11.0 ready to catch errors"}
Populating Frameworks from Github tags
Successfully persisted Framework: 'person-escort-record' with version: '0.3.0'
Successfully persisted Framework: 'person-escort-record' with version: '0.1.0'
```


lib/tasks/populate_framework_data.rake
```diff
@@ -1,13 +1,36 @@
...
   desc 'Populate versioned questions from frameworks repository'
   task :populate_data, %i[filepath version] => :environment do |_, args|
-    unless (filepath = args[:filepath]).present? && (version = args[:version]).present?
-      abort('No filepath or version provided')
-    end
+    if (filepath = args[:filepath]).present? && (version = args[:version]).present?
+      print("Populating Framework from filepath: #{filepath} and version: #{version}\n")
+
+      import_framework(filepath, version)
+    elsif (filepath = args[:filepath]).blank? && (version = args[:version]).blank?
```
smoothcontract, line 12: You don't need to assign these variables as they're not used.

```suggestion
    elsif args[:filepath].blank? && args[:version].blank?
```

PR #760, opened 2020-07-30
P4-1990 fix over-notify on profile updates

martyn-w:
Also fixes profile - move incorrect 1-to-1 relationship

### Jira link

P4-1990
### What?

I have added/removed/altered:

- [x] ensure profile updates only notify other moves related by profile, not by person
- [x] Fix 1-to-1 relationship between profile and move, now 1-to-many

### Why?

I am doing this because:

- a profile can be re-used across multiple moves
- notifications based on person were over-notifying
-

### Have you? (optional)

- [ ] Updated API docs if necessary N/A

- changes relationship between profile and moves


app/jobs/prepare_person_escort_record_notifications_job.rb
```diff
@@ -4,8 +4,8 @@ class PreparePersonEscortRecordNotificationsJob < ApplicationJob
   queue_as :notifications
 
   def perform(topic_id:, action_name:)
-    return unless (move = PersonEscortRecord.find(topic_id).profile.move)
-
-    PrepareMoveNotificationsJob.perform_now(topic_id: move.id, action_name: action_name)
+    PersonEscortRecord.find(topic_id).profile.moves.each do |move|
```
cwrw, line 7: could you use `find_each` here as well?

app/models/profile.rb
```diff
@@ -5,7 +5,7 @@ class Profile < VersionedModel
 
   belongs_to :person
 
-  has_one :move, dependent: :nullify
+  has_many :moves, dependent: :nullify
```
cwrw, line 8: could you add a test for this?
```
it { is_expected.to have_many(:moves) }
```

app/jobs/prepare_person_escort_record_notifications_job.rb
martyn-w: good spot, yes of course 👍

app/models/profile.rb
martyn-w: 👍

PR #767, opened 2020-08-03
Allow filter for /people to be empty

alexdesi:
### Jira link

https://dsdmoj.atlassian.net/browse/P4-1252

### What?

If you query the GET /people endpoint with no filters it returns a `400` saying a filter is required.

The expected behaviour is:
Without the filter, or without a valid filter, the index people endpoint should return all the entries.

### Why?
Other endpoints (see /moves)  do not expect the filters to be mandatory.

### Have you? (optional)

- [ ] Updated API docs if necessary # WIP

### DIFF
--- a/app/controllers/api/v1/people_actions.rb
+++ b/app/controllers/api/v1/people_actions.rb
@@ -62,7 +62,7 @@ module Api::V1
     end
 
     def filter_params
-      params.require(:filter).permit(PERMITTED_FILTER_PARAMS).to_h
+      params.fetch(:filter, {}).permit(PERMITTED_FILTER_PARAMS).to_h
     end
 
     def person_params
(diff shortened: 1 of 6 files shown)


app/controllers/api/v1/people_actions.rb
```diff
@@ -62,7 +62,7 @@ def updater
     end
 
     def filter_params
-      params.require(:filter).permit(PERMITTED_FILTER_PARAMS).to_h
+      params.fetch(:filter, {}).permit(PERMITTED_FILTER_PARAMS).to_h
```
martyn-w, line 65: do you also need to update v2 people_actions ?

martyn-w: (and also `spec/requests/api/people_controller_index_v2_spec.rb`)?

alexdesi: > do you also need to update v2 people_actions ?
That one is correct, but now there is a bit of duplication that can be removed.

PR #768, opened 2020-08-03
Automatically cancel Nomis prison transfer events when cancelling move

smoothcontract:
### Jira link

P4-1950

### What?
- [x] Extended Nomis base client to support `put` method and refactored to DRY things up.
- [x] Added new `Allocations::RemoveFromNomis` service to complement previous `CreateInNomis` service.

### Why?

- When moves are cancelled any corresponding transfer in event should be removed to keep Nomis in sync. The new service uses the persisted `nomis_event_id` attribute to determine if there is an existing Nomis transfer event (this is assigned when calling `Allocations::CreateInNomis` service). On success the `nomis_event_id` is cleared so that calling the service again has no effect. This cancellation behaviour is automatically applied for all move cancellations but only calls Nomis when an event id is stored on the move.

- Changes behaviour for an existing API (cancelling move), but has no effect unless event was previously created

### DIFF
--- a/app/lib/nomis_client/allocations.rb
+++ b/app/lib/nomis_client/allocations.rb
@@ -12,6 +12,16 @@ module NomisClient
 
         e.response
       end
+
+      def put(booking_id:, event_id:, body_params: {})
+        allocations_path = "/bookings/#{booking_id}/prison-to-prison/#{event_id}/cancel"
+
+        NomisClient::Base.put(allocations_path, body: body_params.to_json)
+      rescue OAuth2::Error => e
+        log_exception('Allocations::RemoveFromNomis Error!', allocations_path, body_params, e)
+
+        e.response
+      end
     end
   end
 end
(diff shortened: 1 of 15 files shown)


martyn-w:
Discussed  offline with @smoothcontract  the need to handle repeated calls to the event runner without creating/deleting duplicates in Nomis

app/lib/nomis_client/allocations.rb
```diff
@@ -12,6 +12,16 @@ def post(booking_id:, body_params: {})
 
         e.response
       end
+
+      def put(booking_id:, event_id:, body_params: {})
```
martyn-w, line 16: I see the Nomis routes and names are crazy; nevertheless `put` doesn't seem like a good name here if this method will _always_ cancel an event? if so, how about `cancel`?

smoothcontract: This matches the convention with the other nomis client services - i.e. the HTTP method that will be executed. I really don't want to break that convention - it delegates to `put` on the oauth token under the covers too.

martyn-w: 👍  fair enough

PR #774, opened 2020-08-05
Add support for filtering moves by person date_of_birth attribute

smoothcontract:
### Jira link

P4-1995

### What?

I have added/removed/altered:

- [x] Added support for additional `date_of_birth_from` and `date_of_birth_to` filters for `GET /moves` endpoint
- [x] Updated `Moves::Finder` and `Moves::ParamsValidator` services to permit/validate additional params

### Why?

- We wish to support showing moves only for people with specific age ranges. Rather than pass in an age (which could include a number of years and/or months), this requires the front end to determine a specific date based on the current date and pass corresponding date range. Agreed this approach during refinement as it is more flexible and avoids having to invent a new mechanism for passing age in year and month parameters.

### Have you? (optional)

- [x] Updated API docs if necessary - new filter is supported in both v1 and v2 API

- Includes support for additional filters, but these do not affect returned moves if not specified

### DIFF
--- a/app/services/moves/finder.rb
+++ b/app/services/moves/finder.rb
@@ -46,6 +46,7 @@ module Moves
       scope = scope.accessible_by(ability)
       scope = scope.includes(MOVE_INCLUDES)
       scope = apply_date_range_filters(scope)
+      scope = apply_date_of_birth_filters(scope)
       scope = apply_location_type_filters(scope)
       scope = apply_allocation_relationship_filters(scope)
       scope = apply_ready_for_transit_filters(scope)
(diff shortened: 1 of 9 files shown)


app/services/moves/finder.rb
```diff
@@ -69,12 +70,22 @@ def apply_filter(scope, param_name)
...
       scope = scope.where('moves.created_at < ?', Date.parse(filter_params[:created_at_to]) + 1) if filter_params.key?(:created_at_to)
       scope
     end
 
+    def apply_date_of_birth_filters(scope)
+      return scope unless filter_params.key?(:date_of_birth_from) || filter_params.key?(:date_of_birth_to)
+
+      # only join on person if necessary, otherwise moves without people are not included
+      scope = scope.joins(:person)
+      scope = scope.where('people.date_of_birth >= ?', filter_params[:date_of_birth_from]) if filter_params.key?(:date_of_birth_from)
```
cwrw, line 84: what do you think about indexing `date_of_birth`?

spec/services/moves/finder_spec.rb
```diff
@@ -121,6 +121,53 @@
...
+
+      context 'with mis-matching date of birth range in past' do
+        let(:filter_params) { { date_of_birth_from: (date_of_birth - 5.days).to_s, date_of_birth_to: (date_of_birth - 3.days).to_s } }
+
+        it 'returns empty results set' do
+          expect(results).to be_empty
+        end
+      end
+
+      context 'with mis-matching date of birth range in future' do
```
cwrw, line 162: can you add a test to check the filter is inclusive as well?

db/schema.rb
```diff
@@ -51,7 +51,6 @@
     t.string "prisoner_category"
     t.string "sentence_length"
     t.jsonb "complex_cases"
-    t.integer "moves_count", null: false
```
cwrw, line 54: we shouldn't commit those, just did a drop/create/migrate and those shouldn't move around

smoothcontract: Good spot - reverted 👍

app/services/moves/finder.rb
smoothcontract: Index added.

PR #778, opened 2020-08-06
Update PER documentation

cwrw:
* Add specific PATCH framework response body
* Add section to framework question attributes
* Amend framework responses value types to include:
  - string
  - object
  - array of strings
  - array of objects

### DIFF
--- a/swagger/v1/framework.yaml
+++ b/swagger/v1/framework.yaml
@@ -37,6 +37,6 @@ Framework:
       required:
       - questions
       properties:
-        question:
+        questions:
           $ref: framework_question_reference.yaml#/FrameworkQuestionReference
           description: The questions for the framework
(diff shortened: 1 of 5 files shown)


swagger/v1/framework.yaml
```diff
@@ -37,6 +37,6 @@ Framework:
       required:
       - questions
       properties:
-        question:
+        questions:
```
mafonso, line 40: 👍

swagger/v1/framework_response_reference.yaml
```diff
@@ -10,6 +10,9 @@ FrameworkResponseReference:
       - type: 'null'
       - type: array
         description:  Multiple objects returned if resource is associated to multiple framework_responses
+      example:
+        - id: ea5ace8e-e9ad-4ca3-9977-9bf69e3b6154
+          type: framework_responses
       required:
```
mafonso, line 16: Is indent right?

cwrw: Unfortunately if I move it into array it won't pull this example out when rendering the docs, and will either pull the single example or default to it's own (with removing the specific examples specified below)

PR #799, opened 2020-08-10
[P4-2012] Moves feed

willfish:
### Jira link

P4-2012

### What?

I have added/removed/altered:

- [x] Adds for_feed formatters for each model in the move feed
- [x] Added feed generator for moves
- [x] Added rake task for moves feed

### Why?

- We're moving to a flexible data feed


app/models/move.rb
```diff
@@ -138,6 +161,23 @@ def current?
       (date.nil? && date_to.nil? && date_from.present? && date_from >= Time.zone.today)
   end
 
+  # rake feed:moves
+  # rake feed:journeys
```
willfish, line 165: Remove this nonsense

app/services/feeds/move.rb
```diff
@@ -0,0 +1,18 @@
+module Feeds
+  class Move
+    def initialize(updated_at_from = nil, updated_at_to = nil)
+      @updated_at_from = updated_at_from || Time.zone.now.beginning_of_day - 1.day
+      @updated_at_to = updated_at_to || Time.zone.now.end_of_day - 1.day
```
alexdesi, line 5:
```suggestion
      @updated_at_from = updated_at_from || Time.zone.yesterday.beginning_of_day
      @updated_at_to = updated_at_to || Time.zone.yesterday.end_of_day
```

app/models/move.rb
```diff
@@ -90,6 +113,11 @@ class Move < VersionedModel
 
   scope :not_cancelled, -> { where.not(status: MOVE_STATUS_CANCELLED) }
 
+  scope :updated_at_from_and_to, lambda { |from, to|
```
alexdesi, line 116: maybe `updated_range` is easier to remember/to work with?

lib/tasks/feeds/all.rake
```diff
@@ -0,0 +1,12 @@
+namespace :feeds do
+  desc 'Exports a JSON feed of all possible feed to s3'
+  task all: :environment do
+    feeds = %w[
+      move
+    ]
+
+    feeds.each do |feed_name|
+      Rake::Task["feeds:#{feed_name}"].invoke
```
alexdesi, line 9:
```suggestion
      puts "Generating #{feed_name} feed ..."
      Rake::Task["feeds:#{feed_name}"].invoke
```
some output is usefull

lib/tasks/feeds/move.rake
```diff
@@ -0,0 +1,22 @@
+class CloudDataFeed
```
alexdesi, line 1: my branch should be ready, we can merge that and rebase this from master

app/models/move.rb
willfish: Feexed

lib/tasks/feeds/all.rake
```diff
@@ -0,0 +1,13 @@
+namespace :feeds do
+  desc 'Exports a JSON feed of all possible feed to s3'
```
willfish, line 2:
```suggestion
  desc 'Exports a JSON feed of all possible feeds to s3'
```

spec/services/feeds/move_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    let!(:on_end_move) { create(:move, updated_at: updated_at_to) }
+
+    let(:expected_json) do
+      [on_start_move, on_end_move].sort_by(&:id).map { |move| JSON.parse(move.for_feed.to_json) }
+    end
+
+    it 'returns correctly formatted feed' do
+      result = feed.call
+      result = result.split("\n")
+      result = result.map { |move| JSON.parse(move) }
```
alexdesi, line 18: - maybe we can get rid off all this re-assignment
- it does not look great having to split the result to apply the parse, but I suppose it does not on the whole result
- I think we do not need to specify `\n` in the split

PR #803, opened 2020-08-11
Add CloudDataFeed to make simple to update report on S3

alexdesi:
### Jira link

https://dsdmoj.atlassian.net/browse/P4-2018

### What?
Create a module to upload reporting to S3

### Why?
To make uploading report on S3 easy.

### DIFF
--- a/app/lib/cloud_data_feed.rb
+++ b/app/lib/cloud_data_feed.rb
@@ -1,13 +1,16 @@
 class CloudDataFeed
   def initialize(bucket_name = ENV['S3_REPORTING_BUCKET_NAME'])
-    s3 = Aws::S3::Resource.new
-    @bucket = s3.bucket(bucket_name)
+    @s3 = Aws::S3::Resource.new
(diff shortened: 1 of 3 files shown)


alexdesi:
The specs I wrote, partially test the `CloudDataFeed`, for a `comprehensive` tests we'd need to create a stubbed S3 obj and inject it in the class. 

Anyway, since the method is quite simple I think the tests I wrote are enough.

app/lib/cloud_data_feed.rb
```diff
@@ -0,0 +1,16 @@
+class CloudDataFeed
+  def initialize(bucket_name = ENV['S3_REPORTING_BUCKET_NAME'])
+    @s3 = Aws::S3::Resource.new
+    @bucket = @s3.bucket(bucket_name)
+  end
+
+  def write(content, obj_name)
```
willfish, line 7: I've run this code in dev and it works 🎉

PR #804, opened 2020-08-11
Moves feed deploydev

willfish:
### Jira link

P4-<TODO>

### What?

I have added/removed/altered:

- [ ] Added foo in bar
- [ ] Removed baz

### Why?

I am doing this because:

-
-
-

### Have you? (optional)


lib/tasks/feeds/all.rake
```diff
@@ -0,0 +1,12 @@
+namespace :feeds do
+  desc 'Exports a JSON feed of all possible feed to s3'
+  task all: :environment do
+    feeds = %w[
+      move
+    ]
+
+    feeds.each do |feed_name|
+      Rake::Task["feeds:#{feed_name}"].invoke
```
alexdesi, line 9:
```suggestion
      puts "Generating '#{feed_name}' ..."
      Rake::Task["feeds:#{feed_name}"].invoke
```
a little output would be great :)

lib/tasks/feeds/move.rake
```diff
@@ -0,0 +1,22 @@
+class CloudDataFeed
```
alexdesi, line 1: this PR should be fine, so we can merge that, and the `CloudDataFeed` will be included
https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/803

it took me a bit more than expected to write the tests 😭

PR #809, opened 2020-08-12
[P4-2019] Adds a feed for Profile records.

willfish:
### Jira link

P4-2019

### What?

I have added/removed/altered:

- [x] Adds feed tasks
- [x] Adds feed unit tests

### Why?

See epic


app/models/profile.rb
```diff
@@ -1,6 +1,14 @@
 # frozen_string_literal: true
 
 class Profile < VersionedModel
+  FEED_ATTRIBUTES = %w[
+    id
+    person_id
```
willfish, line 6: This isn't being denormalized and will be joined in the hub's reports.

app/models/profile.rb
```diff
@@ -1,6 +1,14 @@
...
+    created_at
+    updated_at
+    assessment_answers
```
willfish, line 9: We might end up removing this.

willfish: Or just pulling out pregnant/special vehicle ones

PR #810, opened 2020-08-12
Add report feed for Person

mafonso:
### Jira link

P4-2014

### What?

I have added/removed/altered:

- [x] Adds for_feed formatters for each model in the Person feed
- [x] Added feed generator for Person
- [x] Added rake task for Person feed

### Why?

I am doing this because:
-  We're moving to a flexible data feed


app/models/ethnicity.rb
```diff
@@ -3,4 +3,12 @@
 class Ethnicity < ApplicationRecord
   validates :key, presence: true
   validates :title, presence: true
+
+  def for_feed()
+    {
+        "ethnicity" => key,
```
willfish, line 9: ✅

app/models/gender.rb
```diff
@@ -3,4 +3,12 @@
 class Gender < ApplicationRecord
   validates :title, presence: true
   validates :key, presence: true
+
+
+  def for_feed()
+    {
+        "gender" => key,
```
willfish, line 10: ✅

app/models/person.rb
```diff
@@ -1,6 +1,21 @@
 # frozen_string_literal: true
-
 class Person < VersionedModel
+  FEED_ATTRIBUTES = %w[
+    id
```
willfish, line 4: I'm sure they'll come back to us with something about anonymising these but that can be a separate PR :).

lib/tasks/feeds/all.rake
```diff
@@ -3,6 +3,7 @@ namespace :feeds do
   task all: :environment do
     feeds = %w[
       move
+      person
```
willfish, line 6: ✅

willfish:
I think we need need a few more tests for the gender, ethnicity and person for_feed methods: https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/spec/models/move_spec.rb#L509-L539.

Other than that I think this is done 🎉

app/models/gender.rb
```diff
@@ -3,4 +3,12 @@
 class Gender < ApplicationRecord
   validates :title, presence: true
   validates :key, presence: true
+
+
+  def for_feed()
```
willfish, line 8:
```suggestion
  def for_feed
```

app/models/ethnicity.rb
```diff
@@ -3,4 +3,12 @@
 class Ethnicity < ApplicationRecord
   validates :key, presence: true
   validates :title, presence: true
+
+  def for_feed()
```
willfish, line 7:
```suggestion
  def for_feed
```

app/models/person.rb
```diff
@@ -1,6 +1,21 @@
...
+    created_at
+    updated_at
+    first_names
```
willfish, line 7: Remove me

app/models/person.rb
```diff
@@ -1,6 +1,21 @@
...
+    last_name
```
willfish, line 8: Remove me

app/models/person.rb
```diff
@@ -1,6 +1,21 @@
...
+    date_of_birth
```
willfish, line 9: Replace me with age at time of updated_at

app/models/person.rb
```diff
@@ -1,6 +1,21 @@
...
+    gender_additional_information
```
willfish, line 10: Remove me

app/models/person.rb
```diff
@@ -1,6 +1,21 @@
...
+    criminal_records_office
+    nomis_prison_number
+    police_national_computer
+    prison_number
+    last_synced_with_nomis
```
willfish, line 15: Remove me

app/models/person.rb
```diff
@@ -32,4 +50,14 @@ def attach_image(image_blob)
       end
     end
   end
+
+  def for_feed
+    feed_attributes = attributes.slice(*FEED_ATTRIBUTES)
+
+    feed_attributes.merge!(gender.for_feed) if gender
+    feed_attributes.merge!(ethnicity.for_feed) if ethnicity
```
willfish, line 58: Remove me

willfish: Might add back later.

spec/services/feeds/person_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+  describe '#call' do
+    let!(:on_start_person) { create(:person, updated_at: updated_at_from) }
+    let!(:on_end_person) { create(:person, updated_at: updated_at_to) }
+
+    let(:expected_json) do
+      [on_start_person, on_end_person].sort_by(&:id).map { |person| JSON.parse(person.for_feed.to_json) }
+    end
+
+    it 'returns correctly formatted feed' do
+      on_start_person.update(updated_at: updated_at_from)
```
willfish, line 16: We're doing this because the associations update the updated_at as part of the factory so our initial value gets overwritten

PR #811, opened 2020-08-12
Generate feed for journeys

alexdesi:
### Jira link

https://dsdmoj.atlassian.net/browse/P4-2013

### What?

Generate feed for Journey

### Why?

To generate reports

### DIFF
--- a/app/models/journey.rb
+++ b/app/models/journey.rb
@@ -59,4 +59,22 @@ class Journey < ApplicationRecord
            :cancelled?,
            :rejected?,
            to: :state_machine
+
+  def for_feed
+    {
+      'id' => id,
+      'move_id' => move.id,
+      'supplier' => supplier.key,
+      'from_location' => from_location.nomis_agency_id,
+      'from_location_type' => from_location.location_type,
+      'to_location' => to_location.nomis_agency_id,
+      'to_location_type' => to_location.location_type,
+      'billable' => billable,
+      'state' => state,
+      'vehicle_registration' => vehicle['registration'],
+      'client_timestamp': client_timestamp,
+      'created_at': created_at,
+      'updated_at': updated_at,
+    }
+  end
 end
--- a/app/models/journey.rb
+++ b/app/models/journey.rb
@@ -45,6 +45,11 @@ class Journey < ApplicationRecord
 
   scope :default_order, -> { order(client_timestamp: :asc) }
 
+  scope :updated_at_range, lambda { |from, to|
+    includes(:supplier, :from_location, :to_location).where(updated_at: from..to)
+  }
+
+
   has_state_machine JourneyStateMachine
 
   delegate :start,
--- a/app/models/journey.rb
+++ b/app/models/journey.rb
@@ -18,6 +18,17 @@
 # The supplier is paid for only the first journey A-->B but not B-->C.
 #
 class Journey < ApplicationRecord
+  FEED_ATTRIBUTES = %w[
+    id
+    move_id
+    billable
+    state
+    vehicle_registration
+    client_timestamp
+    created_at
+    updated_at
+  ].freeze
+
   include StateMachineable
 
   belongs_to :move, touch: true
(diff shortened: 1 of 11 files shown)


app/models/journey.rb
```diff
@@ -59,4 +63,22 @@ class Journey < ApplicationRecord
            :cancelled?,
            :rejected?,
            to: :state_machine
+
+  def for_feed
+    {
```
willfish, line 68: We should keep consistent with the other feeds so that we can update the relationships/attributes in one way/place:

https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/models/move.rb#L172-L174

app/models/journey.rb
```diff
@@ -45,6 +45,10 @@ class Journey < ApplicationRecord
 
   scope :default_order, -> { order(client_timestamp: :asc) }
 
+  scope :updated_at_range, lambda { |from, to|
```
willfish, line 48: We should add a spec for this https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/spec/models/move_spec.rb#L561-L577

app/models/journey.rb
alexdesi: I agree 👍

app/models/journey.rb
```diff
@@ -59,4 +74,16 @@ class Journey < ApplicationRecord
...
+  def for_feed
+    attributes.slice(*FEED_ATTRIBUTES)
+      .merge(from_location.for_feed(prefix: :from))
+      .merge(to_location.for_feed(prefix: :to))
+      .merge(supplier.for_feed)
+      .merge(vehicle_registration: vehicle_registration)
+  end
+
+  def vehicle_registration
+    vehicle['registration']
```
willfish, line 87: Vehicle can be nil so we'll need to handle that :)

spec/models/move_spec.rb
```diff
@@ -558,20 +558,18 @@
...
+  describe '.updated_at_range scope' do
+    let(:updated_at_from) { Time.zone.yesterday.beginning_of_day }
+    let(:updated_at_to) { Time.zone.yesterday.end_of_day }
 
     it 'returns the expected moves' do
-      actual_moves = described_class.updated_at_range(
-        updated_at_from,
-        updated_at_to,
-      )
+      create(:move, updated_at: updated_at_from - 1.second)
```
willfish, line 566: Can you make this same refactor to people/profiles? I agree it's cleaner/clearer :)

app/models/journey.rb
alexdesi: I'll handle it this way:
```
  def vehicle_registration
    vehicle['registration'] if vehicle
  end
```

(another way could be not adding the vehicle_registration key)

spec/models/move_spec.rb
alexdesi: sure, even though, we are using the same scope there, I'd suggest  (not now) to clean this up, we got 4 identical scopes in 4 different models, and we test it in all the places, it sounds we should use a generic scope, include it in the models, and test just once

willfish: I was thinking the same thing earlier today :)

app/models/journey.rb
```diff
@@ -59,4 +74,16 @@ class Journey < ApplicationRecord
...
+  def for_feed
+    attributes.slice(*FEED_ATTRIBUTES)
+      .merge(from_location.for_feed(prefix: :from))
+      .merge(to_location.for_feed(prefix: :to))
+      .merge(supplier.for_feed)
+      .merge(vehicle_registration: vehicle_registration)
+  end
+
+  def vehicle_registration
+    vehicle['registration'] if vehicle
```
willfish, line 87: 👍

spec/models/person_spec.rb
```diff
@@ -76,27 +76,18 @@
...
-    let!(:before_start_person) { create(:person) }
-    let!(:on_start_person) { create(:person) }
-    let!(:on_end_person) {  create(:person) }
-    let!(:after_end_person) { create(:person) }
+  describe '.updated_at_range scope' do
+    let(:updated_at_from) { Time.zone.yesterday.beginning_of_day }
+    let(:updated_at_to) { Time.zone.yesterday.end_of_day }
 
     it 'returns the expected persons' do
-      # NB: Associations touch parents in factories so updated_at needs to be patched
```
willfish, line 89: Nice 👍

PR #816, opened 2020-08-13
Feeds: Fix wrong task name

mafonso:
### Jira link

N/A 

### What?

I have added/removed/altered:

- [ ] Fixes a wrong task name on the feeds task
- [ ] Make feeds filenames consistent using  plural form.

### Why?

I am doing this because:

- Needs to be addressed.

### DIFF
--- a/lib/tasks/feeds/person.rake
+++ b/lib/tasks/feeds/person.rake
@@ -3,6 +3,6 @@ namespace :feeds do
   task person: :environment do
     feed = Feeds::Person.new.call
 
-    CloudDataFeed.new.write(feed, 'person.json')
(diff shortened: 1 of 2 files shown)


lib/tasks/feeds/person.rake
```diff
@@ -3,6 +3,6 @@ namespace :feeds do
   task person: :environment do
     feed = Feeds::Person.new.call
 
-    CloudDataFeed.new.write(feed, 'person.json')
+    CloudDataFeed.new.write(feed, 'persons.json')
```
willfish, line 6:
```suggestion
    CloudDataFeed.new.write(feed, 'people.json')
```

PR #821, opened 2020-08-13
[P4-2015] Adds feed for Notification model

willfish:
### Jira link

P4-2015

### What?

I have added/removed/altered:

- [x] Adds feed tasks
- [x] Adds feed unit tests
- [x] Adds updated_at index
- [x] Refactors notifications factory to add flexibility for creating multiple notifications of the same type

### Why?

See epic

### DIFF
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -3,6 +3,21 @@
 class Notification < ApplicationRecord
   include Discard::Model
 
+  FEED_ATTRIBUTES = %w[
+    id
+    event_type
+    topic_id
+    topic_type
+    delivery_attempts
+    delivery_attempted_at
+    delivered_at
+    discarded_at
+    created_at
+    updated_at
+    response_id
+    notification_type_id
+  ].freeze
+
   belongs_to :subscription
   belongs_to :notification_type # NB: should be either NotificationType::WEBHOOK or NotificationType::EMAIL
   belongs_to :topic, polymorphic: true, touch: true # NB: polymorphic association because topic could be associated with a Move or a Profile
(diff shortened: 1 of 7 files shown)


spec/factories/notification.rb
```diff
@@ -1,7 +1,11 @@
 FactoryBot.define do
   factory :notification do
     association :subscription
-    association :notification_type, :webhook # defaults to webhook
+
+    notification_type do
+      NotificationType.find_or_create_by(id: 'webhook', title: 'Webhook')
```
willfish, line 6: NotificationType is a table with two entries in it. The primary keys chosen are one of: 'email' or 'webhook'. This is so far out of the standard approach to primary keys in rails that I'm having to do some special snowflake wrapper around the association as this approach isn't compatible with default factory bot rails approaches to handling relationships.

app/models/notification.rb
```diff
@@ -3,6 +3,21 @@
...
+    delivery_attempts
+    delivery_attempted_at
+    delivered_at
+    discarded_at
+    created_at
+    updated_at
+    response_id
+    notification_type_id
+  ].freeze
+
```
mafonso, line 20: 👍

db/migrate/20200813104335_adds_index_on_updated_at_to_notification.rb
```diff
@@ -0,0 +1,7 @@
+class AddsIndexOnUpdatedAtToNotification < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction!
+
+  def change
+    add_index :notifications, :updated_at, algorithm: :concurrently
+  end
```
mafonso, line 6: 👍

lib/tasks/feeds/notification.rake
```diff
@@ -0,0 +1,8 @@
+namespace :feeds do
+  desc 'Exports a JSON feed of all profiles to s3'
+  task notification: :environment do
+    feed = Feeds::Notification.new.call
+
+    CloudDataFeed.new.write(feed, 'notifications.json')
+  end
+end
```
mafonso, line 8: 👍

spec/models/notification_spec.rb
```diff
@@ -58,4 +58,47 @@
...
+      on_start_notification = create(:notification, updated_at: updated_at_from)
+      on_end_notification = create(:notification, updated_at: updated_at_to)
+
+      actual_notifications = described_class.updated_at_range(
+        updated_at_from,
+        updated_at_to,
+      )
+      expect(actual_notifications).to eq([on_start_notification, on_end_notification])
+    end
+  end
```
mafonso, line 103: 👍

PR #823, opened 2020-08-13
Tidies

willfish:
### What?

- [x] Added missing indexes for feed
- [x] Added notification to all.rake

### Why?

Indexes are needed for optimal reporting
Notification feed needs to be reported on

### DIFF
--- a/db/migrate/20200813125352_adds_index_updated_at.rb
+++ b/db/migrate/20200813125352_adds_index_updated_at.rb
@@ -1,0 +1,10 @@
+class AddsIndexUpdatedAt < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction!
+
+  def change
+    add_index :people, :updated_at, algorithm: :concurrently
+    add_index :profiles, :updated_at, algorithm: :concurrently
+    add_index :events, :updated_at, algorithm: :concurrently
+    add_index :journeys, :updated_at, algorithm: :concurrently
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_08_13_104335) do
+ActiveRecord::Schema.define(version: 2020_08_13_125352) do
 
   # These are extensions that must be enabled in order to support this database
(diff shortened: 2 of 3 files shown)


db/migrate/20200813125352_adds_index_updated_at.rb
```diff
@@ -0,0 +1,10 @@
+class AddsIndexUpdatedAt < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction!
+
+  def change
+    add_index :people, :updated_at, algorithm: :concurrently
+    add_index :profiles, :updated_at, algorithm: :concurrently
+    add_index :events, :updated_at, algorithm: :concurrently
+    add_index :journeys, :updated_at, algorithm: :concurrently
+  end
```
mafonso, line 9: 👍

db/schema.rb
```diff
@@ -123,6 +123,7 @@
     t.index ["client_timestamp"], name: "index_events_on_client_timestamp"
     t.index ["eventable_id", "eventable_type", "event_name"], name: "index_events_on_eventable_id_and_eventable_type_and_event_name"
     t.index ["eventable_id", "eventable_type"], name: "index_events_on_eventable_id_and_eventable_type"
+    t.index ["updated_at"], name: "index_events_on_updated_at"
   end
```
mafonso, line 127: It does not seem to apply the `concurrently` modifier.

db/schema.rb
```diff
@@ -123,6 +123,7 @@
     t.index ["client_timestamp"], name: "index_events_on_client_timestamp"
     t.index ["eventable_id", "eventable_type", "event_name"], name: "index_events_on_eventable_id_and_eventable_type_and_event_name"
     t.index ["eventable_id", "eventable_type"], name: "index_events_on_eventable_id_and_eventable_type"
+    t.index ["updated_at"], name: "index_events_on_updated_at"
```
mafonso, line 126: It does not seem to apply the concurrently modifier.

lib/tasks/feeds/all.rake
```diff
@@ -6,6 +6,7 @@ namespace :feeds do
       profile
       person
       journey
+      notification
```
mafonso, line 9: 👍

PR #824, opened 2020-08-13
Add support for returning GET /moves endpoint in CSV format

smoothcontract:
### Jira link

P4-2028

### What?

- [x] Add support for specifying 'text/csv' or 'application/csv' type in 'Accept' header on 'GET /moves' v2 endpoint
- [x] Add new `Moves::Exporter` service to render found moves in CSV file format (matching the existing front end format)

### Why?

- The existing approach for downloading CSV files in the front end uses the API to retrieve all of the relevant moves (normally 1 weeks worth for a given location) in JSONAPI format, and then reformats to a second JSON representation, then serializes that to a CSV format. This is far from efficient, so this new approach generates the CSV in the backend without involving any unnecessary JSON conversion - it simply uses ActiveRecord relationships and writes to a temporary file.
- Adding an additional content type to the existing v2 endpoint means that we can leverage all of the existing filters on the `GET /moves` endpoint - these are already well tested, so clients can obtain exactly the required moves.
- The simple synchronous approach adopted means we only require a `Tempfile` on the API server serving the request; nothing needs to be shared across servers and the temporary files are automatically removed when garbage collection is performed. The server load and response time should be the same or lower than returning the same data in JSONAPI format too. Local testing using a snapshot of the entire staging database (11,470 moves) returns the file in around 15 seconds. A more representative example of 1,000 moves (1 weeks worth across all locations) takes around 1 - 1.5 seconds. If this is scoped to a single location then this should be sub second.

### Have you? (optional)

- Extends the existing `GET /moves` endpoint to support new `Accept` header value - no change to behaviour if this is not specified

### DIFF
--- a/app/controllers/api/v2/moves_actions.rb
+++ b/app/controllers/api/v2/moves_actions.rb
@@ -3,10 +3,14 @@ module Api::V2
     def index_and_render
       moves = Moves::Finder.new(filter_params, current_ability, params[:sort] || {}).call
 
-      paginate moves,
-               each_serializer: ::V2::MoveSerializer,
-               include: included_relationships,
-               fields: ::V2::MoveSerializer::INCLUDED_FIELDS
+      if csv?
+        send_file(Moves::Exporter.new(moves).call, type: 'text/csv', disposition: :inline)
+      else
+        paginate moves,
+                 each_serializer: ::V2::MoveSerializer,
+                 include: included_relationships,
+                 fields: ::V2::MoveSerializer::INCLUDED_FIELDS
+      end
     end
 
     def show_and_render
--- a/app/models/move.rb
+++ b/app/models/move.rb
@@ -83,7 +83,6 @@ class Move < VersionedModel
   has_many :journeys, -> { default_order }, dependent: :restrict_with_exception, inverse_of: :move
   has_many :court_hearings, dependent: :restrict_with_exception
   has_many :move_events, as: :eventable, dependent: :destroy # NB: polymorphic association
-  has_many :documents
 
   validates :from_location, presence: true
   validates :to_location, presence: true, unless: -> { prison_recall? || video_remand? }
--- a/app/models/move.rb
+++ b/app/models/move.rb
@@ -83,6 +83,7 @@ class Move < VersionedModel
   has_many :journeys, -> { default_order }, dependent: :restrict_with_exception, inverse_of: :move
   has_many :court_hearings, dependent: :restrict_with_exception
   has_many :move_events, as: :eventable, dependent: :destroy # NB: polymorphic association
+  has_many :documents
 
   validates :from_location, presence: true
   validates :to_location, presence: true, unless: -> { prison_recall? || video_remand? }
--- a/app/services/moves/exporter.rb
+++ b/app/services/moves/exporter.rb
@@ -63,7 +63,7 @@ module Moves
       'Requires special vehicle',
       'Requires special vehicle details',
       'Uploaded documents',
-    ]
+    ].freeze
 
     MOVE_INCLUDES = [:from_location, :to_location, :profile, :documents, person: %i[gender ethnicity]].freeze
 
@@ -131,7 +131,7 @@ module Moves
       ].flatten
     end
 
-   def answer_details(answers, key)
+    def answer_details(answers, key)
       selected_answers = answers&.select { |answer| answer.key == key }
       comments = selected_answers&.collect do |answer|
         description = answer.nomis_alert_description
@@ -139,6 +139,6 @@ module Moves
       end
 
       [comments.present?.to_s.upcase, comments&.join("\n\n")]
-    end
+     end
   end
 end
--- a/app/services/moves/exporter.rb
+++ b/app/services/moves/exporter.rb
@@ -65,8 +65,6 @@ module Moves
       'Uploaded documents',
     ].freeze
 
-    MOVE_INCLUDES = [:from_location, :to_location, :profile, :documents, person: %i[gender ethnicity]].freeze
-
     def initialize(moves)
       @moves = moves
     end
@@ -75,10 +73,8 @@ module Moves
       Tempfile.new('export', Rails.root.join('tmp')).tap do |file|
         csv = CSV.new(file)
         csv << HEADINGS
-        moves.includes(MOVE_INCLUDES).find_in_batches do |batched_moves|
-          batched_moves.each do |move|
-            csv << attributes_row(move)
-          end
+        moves.find_each do |move|
+          csv << attributes_row(move)
         end
       end
     end
@@ -87,7 +83,8 @@ module Moves
 
     def attributes_row(move)
       person = move.person
-      answers = move.profile&.assessment_answers
+      profile = move.profile
+      answers = profile&.assessment_answers
 
       [
         move.status, # Status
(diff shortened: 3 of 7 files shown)


app/models/move.rb
```diff
@@ -83,6 +83,7 @@ class Move < VersionedModel
   has_many :journeys, -> { default_order }, dependent: :restrict_with_exception, inverse_of: :move
   has_many :court_hearings, dependent: :restrict_with_exception
   has_many :move_events, as: :eventable, dependent: :destroy # NB: polymorphic association
+  has_many :documents
```
smoothcontract, line 86: This relationship was missing - documents can belong to moves optionally, and the front end download code looks to be counting move documents so I matched that behaviour.

app/services/moves/exporter.rb
```diff
@@ -0,0 +1,144 @@
...
+
+    def initialize(moves)
+      @moves = moves
+    end
+
+    def call
+      Tempfile.new('export', Rails.root.join('tmp')).tap do |file|
+        csv = CSV.new(file)
+        csv << HEADINGS
+        moves.includes(MOVE_INCLUDES).find_in_batches do |batched_moves|
```
smoothcontract, line 78: Defaults to batches of 1000 - it's doubtful we would ever go higher than this but I've tested successfully with 11k moves and this holds up fine. Without using batches the eager fetch SQL queries and memory usage would be huge.

app/services/moves/exporter.rb
```diff
@@ -0,0 +1,144 @@
...
+    ].freeze
+
+    MOVE_INCLUDES = [:from_location, :to_location, :profile, :documents, person: %i[gender ethnicity]].freeze
+
+    def initialize(moves)
+      @moves = moves
+    end
+
+    def call
+      Tempfile.new('export', Rails.root.join('tmp')).tap do |file|
```
smoothcontract, line 75: I am intentionally not cleaning this file up explicitly - as we need to access it in the controller actions. Tempfiles do get deleted during garbage collection though so shouldn't be an issue. See https://ruby-doc.org/stdlib-2.7.1/libdoc/tempfile/rdoc/Tempfile.html#class-Tempfile-label-Explicit+close

app/controllers/api/v2/moves_actions.rb
```diff
@@ -3,10 +3,14 @@ module MovesActions
     def index_and_render
       moves = Moves::Finder.new(filter_params, current_ability, params[:sort] || {}).call
 
-      paginate moves,
-               each_serializer: ::V2::MoveSerializer,
-               include: included_relationships,
-               fields: ::V2::MoveSerializer::INCLUDED_FIELDS
+      if csv?
```
alexdesi, line 6: I was thinking if we could use the standard rails way:
```
respond_to do |format|
...
end
```
and using format.csv and format.json
The advantage would be removing the conditional and a more standard rails way.

spec/services/moves/exporter_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+    expect(row).to include(person.gender.title)
+  end
+
+  it 'includes ethnicity details' do
+    expect(row).to include(person.ethnicity.title, person.ethnicity.key)
+  end
+
+  %w[violent escape hold_separately self_harm concealed_items other_risks special_diet_or_allergy health_issue medication wheelchair pregnant other_health solicitor interpreter other_court not_for_release not_to_be_released special_vehicle].each do |alert_type|
+    it "includes TRUE flag when #{alert_type} is present" do
+      question.update(key: alert_type)
```
alexdesi, line 65: we might wrap this into a context here to setup the question with a correct alert_type, since this is done in all these tests.

spec/services/moves/exporter_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+      move.profile.update(assessment_answers: [{ assessment_question_id: question.id, comments: 'Yikes!' }])
```
alexdesi, line 66: the comment about could be valid for this line too

app/services/moves/exporter.rb
```diff
@@ -0,0 +1,144 @@
...
+        answer_details(answers, 'other_health'), # Any other details
+        answer_details(answers, 'solicitor'), # Solicitor or other legal representation details
+        answer_details(answers, 'interpreter'), # Sign or other language interpreter details
+        answer_details(answers, 'other_court'), # Any other information details
+        answer_details(answers, 'not_for_release'), # Not for release details
+        answer_details(answers, 'not_to_be_released'), # Not to be released details
+        answer_details(answers, 'special_vehicle'), # Requires special vehicle details
+        move.documents.size, # 'Uploaded documents',
+      ].flatten
+    end
```
alexdesi, line 132: This list is quite big, 
I was wondering if we could reuse the serializers, or maybe the `for_feed` method (that generate the JSON used for reporting)

Is this the only model we need the CSV we need? if so maybe it's ok

app/controllers/api/v2/moves_actions.rb
willfish: We don't use plain json format? Curious about how that works for json api mime type

app/models/move.rb
willfish: Documents are now on a profile

app/services/moves/exporter.rb
willfish: If you do the following:

```suggestion
        moves.includes(MOVE_INCLUDES).find_each do |move|
```

It will only instantiate 1000 moves at a time and yield at the level (move) that you actually want.

app/services/moves/exporter.rb
```diff
@@ -0,0 +1,144 @@
...
+          batched_moves.each do |move|
+            csv << attributes_row(move)
+          end
+        end
+      end
+    end
+
+  private
+
+    def attributes_row(move)
```
willfish, line 88: Wondering whether we can have a `to_csv` implementation on the move model?

willfish: Not super precious about this.

app/models/move.rb
smoothcontract: Will revert this 👍

app/services/moves/exporter.rb
smoothcontract: I'm not sure I like that idea - as it would mean splitting the responsibility for the correct columns into two places i.e. the column headers are defined here but the order of the array values returned in the move model has to match. Also the move model is pretty bloated already so on balance I prefer to keep everything in this service.

app/services/moves/exporter.rb
smoothcontract: The existing approach to download via the front end uses JSON serialisation which is one of the main reasons for the poor performance. A key goal with this work is to get down to the metal to improve the export speed - I've spent a fair bit of time tuning and optimising this to make it performant and doubt that introducing a JSON wrapper would help in that goal.

app/controllers/api/v2/moves_actions.rb
smoothcontract: I could be wrong but I thought that `respond_to` / `respond_with` relies on routes defining a format (i.e. a file extension) - so that wouldn't work with our approach of using `Accept` header to determine the required type.

app/services/moves/exporter.rb
smoothcontract: I'll investigate that suggestion, thanks

spec/services/moves/exporter_spec.rb
smoothcontract: I might rework this as it's probably overkill to test all the behaviour permutations for every assessment key.

app/services/moves/exporter.rb
```diff
@@ -127,8 +124,8 @@ def attributes_row(move)
         answer_details(answers, 'not_for_release'), # Not for release details
         answer_details(answers, 'not_to_be_released'), # Not to be released details
         answer_details(answers, 'special_vehicle'), # Requires special vehicle details
-        move.documents.size, # 'Uploaded documents',
-      ].flatten
+        profile&.documents&.size, # 'Uploaded documents',
+      ].flatten # Expand answer_details column pairs into individual columns
```
willfish, line 128: 👍

PR #826, opened 2020-08-14
[P4-2011] Adds event feed

willfish:
### Jira link

P4-2011

### What?

I have added/removed/altered:

- [x] Adds feed tasks
- [x] Adds feed unit tests
- [x] Adds elapsed time for each feed to job log output

### Why?

See epic

### DIFF
--- a/spec/models/event_spec.rb
+++ b/spec/models/event_spec.rb
@@ -60,4 +60,30 @@ RSpec.describe Event, type: :model do
       expect { create(:event, eventable: eventable) }.to change { eventable.reload.updated_at }
     end
   end
+
+  describe '#for_feed' do
+    subject(:event) { create(:event) }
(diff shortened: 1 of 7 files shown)


spec/models/event_spec.rb
```diff
@@ -60,4 +60,30 @@
       expect { create(:event, eventable: eventable) }.to change { eventable.reload.updated_at }
     end
   end
+
+  describe '#for_feed' do
+    subject(:event) { create(:event) }
```
alexdesi, line 65:
```suggestion
```

willfish: We can remove it but discussed that it doesn't have a performance cost and makes the test more readable/more consistent with the other specs for the `#for_feed` tests

PR #833, opened 2020-08-16
1 - P4-2004 Add support for importing `add_multiple_items` questions

cwrw:
### Jira link

### What?

- [x] Add new type to framework questions
- [x] Link child questions on a add_multiple_items question without dependent values

### Why?

To support importing questions which require multiple items, add new type of question which allows multiple questions to exist grouped by this question. The child questions are created without a dependent value.

### DIFF
--- a/app/models/framework_question.rb
+++ b/app/models/framework_question.rb
@@ -6,6 +6,7 @@ class FrameworkQuestion < VersionedModel
     checkbox: 'checkbox',
     text: 'text',
     textarea: 'textarea',
+    add_multiple_items: 'add_multiple_items',
   }
 
   validates :key, presence: true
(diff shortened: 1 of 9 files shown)


app/models/framework_question.rb
```diff
@@ -6,6 +6,7 @@ class FrameworkQuestion < VersionedModel
     checkbox: 'checkbox',
     text: 'text',
     textarea: 'textarea',
+    add_multiple_items: 'add_multiple_items',
```
martyn-w, line 9: the `add_multiple_items` seems a bit of an odd name: would `multiple_items` work better?

cwrw: Indeed naming this question was tough. Our reasoning for prepending it with add is to ensure that it is communicated as allowing a user to `add` or amend the items number, and felt more explicit. `multiple_items` as a name seemed a bit vague for a type of question. We initially started with `add_another` based on https://moj-design-system.herokuapp.com/components/add-another but we didn't feel that was clear either so went with this as a middle ground.

martyn-w: 👍

PR #834, opened 2020-08-16
2 - P4-2004 Add support for creating `add_multiple_items` responses

cwrw:
### Jira link

### What?

- [x] Added new type of collection response for multiple item questions
- [x] Added support for creating multiple item responses when creating the PER

### Why?

Add a new collection response type: multiple items. Each multiple items list contains a multiple item object which encapsulates the item and the corresponding responses. It applies the validations to that object and bubbles them up to the collection appropriately.

This type of question has nested questions which do not have corresponding responses, but rather those responses are encapsulated in the parent question, mapped to the item they are associated to. 

Since the responses provided for the item mirror framework responses, validate those on update (if normal validation added it allows them to be empty which we don't want). Scope validations to an index of the responses and the items to avoid overwriting validation messages. This is similar to `index_errors` on rails associations: https://blog.bigbinary.com/2016/07/07/errors-can-be-indexed-with-nested-attrbutes-in-rails-5.html

When a question is an add multiple items question, create a response of type collection. Do not create dependent responses for multiple items dependent questions, as those will be nested in the parent response.

### DIFF
--- a/app/models/framework_response/collection.rb
+++ b/app/models/framework_response/collection.rb
@@ -3,6 +3,7 @@ class FrameworkResponse
     validate :validate_collection_type
     validates :value_text, absence: true
     validate :validate_details_collection, on: :update, if: -> { response_details }
+    validate :validate_multiple_items_collection, on: :update, if: -> { multiple_items? }
 
     def value
       value_json.delete_if(&:empty?) if collection_type_valid?(value_json)
@@ -13,6 +14,8 @@ class FrameworkResponse
       self.value_json =
         if response_details && collection_type_valid?(raw_value)
           details_collection(raw_value).to_a
+        elsif multiple_items? && collection_type_valid?(raw_value)
+          multiple_items_collection(raw_value).to_a
         else
           raw_value.presence || []
         end
@@ -32,6 +35,14 @@ class FrameworkResponse
       )
     end
 
+    def multiple_items_collection(collection)
+      MultipleItemsCollection.new(
+        collection: collection,
+        questions: framework_question.dependents,
+        person_escort_record: person_escort_record
+      )
+    end
+
     def validate_details_collection
       return if errors.present?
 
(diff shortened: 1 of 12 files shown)


app/models/framework_response/multiple_item_object.rb
```diff
@@ -0,0 +1,86 @@
...
+    end
+
+  private
+
+    def multiple_response_objects
+      return unless responses_type_valid? && response_objects.any? { |object| object.invalid?(:update) }
+
+      response_objects.each_with_index do |object, index|
+        object.errors.each do |key, value|
+          errors.add("responses[#{index}].#{key}", value)
```
cwrw, line 41: This is based on https://blog.bigbinary.com/2016/07/07/errors-can-be-indexed-with-nested-attrbutes-in-rails-5.html

app/models/framework_response/multiple_items_collection.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+  private
+
+    def multiple_item_objects
+      return unless collection.any?(&:invalid?)
+
+      collection.each_with_index do |object, index|
+        object.errors.each do |key, value|
+          errors.add("items[#{index}].#{key}", value)
```
cwrw, line 32: This is based on https://blog.bigbinary.com/2016/07/07/errors-can-be-indexed-with-nested-attrbutes-in-rails-5.html

app/models/framework_response/multiple_item_object.rb
```diff
@@ -0,0 +1,86 @@
...
+      {
+        item: item,
+        responses: response_objects.map { |response| response_as_json(response) },
+      }
+    end
+
+  private
+
+    def multiple_response_objects
+      return unless responses_type_valid? && response_objects.any? { |object| object.invalid?(:update) }
```
cwrw, line 37: validation rules are different creation vs update on responses. Ensure we are validating on update which includes required values

app/models/framework_response/collection.rb
```diff
@@ -13,6 +14,8 @@ def value=(raw_value)
       self.value_json =
         if response_details && collection_type_valid?(raw_value)
           details_collection(raw_value).to_a
+        elsif multiple_items? && collection_type_valid?(raw_value)
```
cwrw, line 17: I intend to resolve the type validation and raise an error before going into this for all response types, as this is not ideal to check the type at every point. Will open a pr after all this is merged

PR #835, opened 2020-08-16
3 - P4-2004 Add support for PATCHing `add_multiple_items` responses

cwrw:
### Jira link

### What?

- [x] Added support in framework responses controller to permit new values for add multiple items
- [x] Expose dependent questions in framework question serializer
- [x] Expose response type in framework question serializer

### Why?

To allow multiple item responses to be updated, permit nested `item` and `responses` attributes for framework responses.
Since value supports multiple types we have to mirror that in the permitted attributes by supporting both a nested string and object.

Since we are allowing the client to build the nested responses, they require the response type to be exposed. We can do this on the framework questions, which now have both `question_type` as well as `response_type`.

Expose a different response type for add multiple items, since this will indicate that this type of response is structured differently. Since migrating existing types is more involved, we are deferring it for later.

### Have you? (optional)

### DIFF
--- a/app/serializers/framework_question_serializer.rb
+++ b/app/serializers/framework_question_serializer.rb
@@ -1,9 +1,24 @@
 class FrameworkQuestionSerializer < ActiveModel::Serializer
   belongs_to :framework
 
-  attributes :key, :section, :question_type, :options
+  attributes :key, :section, :question_type, :options, :response_type
+  has_many :dependents, key: :descendants
+
+  def response_type
+    case object.question_type
+    when 'radio'
+      object.followup_comment ? 'object' : 'string'
+    when 'checkbox'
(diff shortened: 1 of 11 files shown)


app/serializers/framework_question_serializer.rb
```diff
@@ -1,9 +1,24 @@
...
+  has_many :dependents, key: :descendants
+
+  def response_type
+    case object.question_type
+    when 'radio'
+      object.followup_comment ? 'object' : 'string'
+    when 'checkbox'
+      object.followup_comment ? 'collection' : 'array'
+    when 'add_multiple_items'
+      'collection::add_multiple_items'
```
cwrw, line 14: We'll be changing other types to be more explicit, but that requires a timed deploy, so only amending the new type till then

PR #837, opened 2020-08-17
Improve error message for webhook notifications

cwrw:
Parse response and extract error message from body, to help debugging of webhook errors. The raw response with an error coming from the server has the following structure:

```
{"resource"=>"/suite/webapi/getMOJEvents", "error"=>"APNX-1-4187-005", "message"=>"User password has expired. Please sign in using a web browser to change your password.", "title"=>"An Error Has Occurred"}
```
If the structure is different we should get empty values

### DIFF
--- a/app/jobs/notify_webhook_job.rb
+++ b/app/jobs/notify_webhook_job.rb
@@ -18,7 +18,9 @@ class NotifyWebhookJob < ApplicationJob
       client = get_client(subscription)
       response = client.post(subscription.callback_url, data, 'PECS-SIGNATURE': hmac, 'PECS-NOTIFICATION-ID': notification_id)
 
-      raise "non-success status received from #{subscription.callback_url} (#{response.status} #{response.reason_phrase})" unless response.success?
+      unless response.success?
+        raise "non-success status received from #{subscription.callback_url}. Status: #{response.status}, Reason: #{response.reason_phrase}, Response body: '#{response.body}'"
+      end
 
       notification.update(
         delivered_at: Time.zone.now,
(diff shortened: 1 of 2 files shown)


app/jobs/notify_webhook_job.rb
```diff
@@ -18,7 +18,10 @@ def perform(notification_id)
       client = get_client(subscription)
       response = client.post(subscription.callback_url, data, 'PECS-SIGNATURE': hmac, 'PECS-NOTIFICATION-ID': notification_id)
 
-      raise "non-success status received from #{subscription.callback_url} (#{response.status} #{response.reason_phrase})" unless response.success?
+      unless response.success?
+        parsed_response = JSON.parse(response.body)
```
martyn-w, line 22: I don't think we can assume that every response is guaranteed to be JSON

martyn-w: maybe check the response content type and parse as json if it is json.  Actually, thinking about this more, maybe it is safer just to send through the un-parsed response as a string in the error message.

Keep in mind that suppliers are running different code / different architectures and the responses from each are likely to be different.

martyn-w: Or, even better: if there is a `JSON.parse(response.body)['message']` use that, otherwise just the unparsed `response.body`

cwrw: good point, I think maybe we should just include the `response.body` as you suggest without parsing, as we'll have to handle JSON parsing errors otherwise.

martyn-w: 👍

PR #839, opened 2020-08-17
fix updated_at, add supplier

martyn-w:
fixes updated_at time for diagnostics; adds supplier if available

### DIFF
--- a/app/services/diagnostics/move_inspector.rb
+++ b/app/services/diagnostics/move_inspector.rb
@@ -24,8 +24,9 @@ module Diagnostics
         move type:\t#{move.move_type}
         from location:\t#{move.from_location&.title}
         to location:\t#{move.to_location&.title}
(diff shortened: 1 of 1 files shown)


app/services/diagnostics/move_inspector.rb
```diff
@@ -24,8 +24,9 @@ def generate
         move type:\t#{move.move_type}
         from location:\t#{move.from_location&.title}
         to location:\t#{move.to_location&.title}
+        supplier:\t#{move.supplier&.name}
```
cwrw, line 27: do we always have suppliers on moves? checking validations seems optional

martyn-w: there are historic moves which don't have a supplier_id defined; and in the case of multiple suppliers to the same location I think there was a case when a move can be initialised (proposed) without a supplier and it updated later?

PR #841, opened 2020-08-18
Expose dependents on move endpoint for PER questions

cwrw:
To expose the descendant questions resource on a Move, add to include list.
Update since descendants are nested and hierarchical we need to use `**` to allow includes for more than one level: https://github.com/rails-api/active_model_serializers/blob/v0.10.6/docs/general/adapters.md#include-option 

I kept the response at a single level as we will never need to support multiple nested resources at that level, this is only on the PER or Move

### DIFF
--- a/app/serializers/move_serializer.rb
+++ b/app/serializers/move_serializer.rb
@@ -41,6 +41,7 @@ class MoveSerializer < ActiveModel::Serializer
     profile.person_escort_record.framework
     profile.person_escort_record.responses
     profile.person_escort_record.responses.question
+    profile.person_escort_record.responses.question.descendants.**
     from_location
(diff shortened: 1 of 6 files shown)


app/serializers/move_serializer.rb
```diff
@@ -41,6 +41,7 @@ class MoveSerializer < ActiveModel::Serializer
     profile.person_escort_record.framework
     profile.person_escort_record.responses
     profile.person_escort_record.responses.question
+    profile.person_escort_record.responses.question.descendants.**
```
cwrw, line 44: from https://github.com/rails-api/active_model_serializers/blob/v0.10.6/docs/general/adapters.md#include-option

PR #842, opened 2020-08-18
Add new endpoint /suppliers/:supplier_id/locations 

alexdesi:
### Jira link

https://dsdmoj.atlassian.net/browse/P4-2063

### What?

One of the tasks, before going live, is

to provide an endpoint to get a list of locations for which the suppliers own the moves (from_location)
The locations will be a unique list of the relevant move from_location locations.

I have added a new controller SuppliersController and the action locations.

### Why?

read the https://dsdmoj.atlassian.net/browse/P4-2063 for an extensive explanation

### Have you? (optional)

Not necessary, we do not want to make it public for now

No, it does not change the existing endpoints.

### DIFF
--- a/app/controllers/api/suppliers_controller.rb
+++ b/app/controllers/api/suppliers_controller.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+module Api
+  class SuppliersController < ApiController
+    def locations
+      paginate locations_from_supplier_moves
+    end
+
+  private
+
+    def locations_from_supplier_moves
+      supplier = Supplier.find(params[:supplier_id])
+
+      Location.joins(:moves_from).where(moves: { supplier_id: supplier.id }).distinct
+    end
+  end
+end
--- a/app/controllers/api/suppliers_controller.rb
+++ b/app/controllers/api/suppliers_controller.rb
@@ -12,6 +12,7 @@ module Api
       supplier = Supplier.find(params[:supplier_id])
 
       Location.joins(:moves_from)
+              .includes(:suppliers)
               .where(moves: { supplier_id: supplier.id })
               .order(key: :asc)
               .distinct
--- a/app/controllers/api/suppliers_controller.rb
+++ b/app/controllers/api/suppliers_controller.rb
@@ -3,7 +3,7 @@
 module Api
   class SuppliersController < ApiController
     def locations
-      paginate locations_from_supplier_moves
+      paginate locations_from_supplier_moves, include: included_relationships
     end
 
   private
@@ -11,7 +11,14 @@ module Api
     def locations_from_supplier_moves
       supplier = Supplier.find(params[:supplier_id])
 
-      Location.joins(:moves_from).where(moves: { supplier_id: supplier.id }).distinct
+      Location.joins(:moves_from)
+              .where(moves: { supplier_id: supplier.id })
+              .order(key: :asc)
+              .distinct
+    end
+
+    def supported_relationships
+      LocationSerializer::SUPPORTED_RELATIONSHIPS
     end
   end
 end
--- a/spec/requests/api/suppliers_controller_locations_spec.rb
+++ b/spec/requests/api/suppliers_controller_locations_spec.rb
@@ -3,28 +3,57 @@
 require 'rails_helper'
 
 RSpec.describe Api::SuppliersController do
+  subject(:get_supplier_locations) { get "/api/v1/suppliers/#{supplier_id}/locations#{query_params}", headers: headers, params: {} }
+
   let(:access_token) { 'spoofed-token' }
-  let(:headers) { { 'CONTENT_TYPE': content_type, 'Authorization': "Bearer #{access_token}" } }
-  let(:content_type) { ApiController::CONTENT_TYPE }
+  let(:headers) { { 'Authorization': "Bearer #{access_token}" } }
   let(:response_json) { JSON.parse(response.body) }
+  let(:query_params) { '' }
 
   describe 'GET /suppliers/:supplier_id/locations' do
+    let(:schema) { load_yaml_schema('get_locations_responses.yaml') }
+
     let(:supplier) { create(:supplier) }
+    let(:supplier_id) { supplier.id }
+    let(:location_a) { create(:location, suppliers: [supplier], key: 'key_aaa') }
+    let(:location_b) { create(:location, suppliers: [supplier], key: 'key_bbb') }
+
+    it_behaves_like 'an endpoint that responds with success 200' do
+      before do
+        create(:move, :requested, supplier: supplier)
+
+        get_supplier_locations
+      end
+    end
 
     it 'returns the locations' do
-      location1 = create(:location, suppliers: [supplier])
-      location2 = create(:location, suppliers: [supplier])
-      create_list(:move, 2, :requested, supplier: supplier, from_location: location1) # to tests that duplicates locations are not shown
-      create(:move, :requested, supplier: supplier, from_location: location2)
+      create_list(:move, 2, :requested, supplier: supplier, from_location: location_a) # to tests that duplicates locations are not shown
+      create(:move, :requested, supplier: supplier, from_location: location_b)
 
-      get "/api/v1/suppliers/#{supplier.id}/locations", headers: headers, params: {}
+      get_supplier_locations
 
       response_location_keys = response_json['data']
                                  .select { |e| e['type'] == 'locations' }
                                  .map { |l| l['attributes']['key'] }
 
-      expect(response).to have_http_status(:success)
-      expect(response_location_keys).to match_array([location1.key, location2.key])
+      expect(response_location_keys).to eq(%w[key_aaa key_bbb])
+    end
+
+    context 'when other locations are present' do
+      it 'does not include those into the response' do
+        another_supplier = create(:supplier)
+        another_location = create(:location, suppliers: [another_supplier])
+        create(:move, :requested, supplier: another_supplier, from_location: another_location)
+        create(:move, :requested, supplier: supplier, from_location: location_a)
+
+        get_supplier_locations
+
+        response_location_keys = response_json['data']
+                                     .select { |e| e['type'] == 'locations' }
+                                     .map { |l| l['attributes']['key'] }
+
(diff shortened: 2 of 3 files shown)


app/controllers/api/suppliers_controller.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Api
+  class SuppliersController < ApiController
+    def locations
+      paginate locations_from_supplier_moves
```
smoothcontract, line 6: Would be good to add support for includes here, these can be the same as the ones on the locations reference data endpoint I think. Something like:

```ruby
        paginate locations_from_supplier_moves, include: included_relationships
```

You'll also need to add this private method

```ruby
      def supported_relationships
        LocationSerializer::SUPPORTED_RELATIONSHIPS
      end
```

app/controllers/api/suppliers_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+    end
+
+  private
+
+    def locations_from_supplier_moves
+      supplier = Supplier.find(params[:supplier_id])
+
+      Location.joins(:moves_from).where(moves: { supplier_id: supplier.id }).distinct
```
smoothcontract, line 14: I like the simplicity of this, but there's a potential issue with pagination as it doesn't specify any ordering - so requesting a few pages in sequence doesn't guarantee the client would get the correct locations in each page. Maybe add an order by clause on something which is unique, perhaps the `nomis_agency_id` on locations?

Also this looks like it has an N+1 query for suppliers; check the development logs when hitting the endpoint via Postman. Would be good to resolve that if possible.

spec/requests/api/suppliers_controller_locations_spec.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Api::SuppliersController do
+  let(:access_token) { 'spoofed-token' }
+  let(:headers) { { 'CONTENT_TYPE': content_type, 'Authorization': "Bearer #{access_token}" } }
+  let(:content_type) { ApiController::CONTENT_TYPE }
```
smoothcontract, line 8: You can leave out `CONTENT_TYPE` in the headers and the spec will still work - we have coverage for that elsewhere (it's optional) so is just adding noise here. Alternatively we have a shared context you could include that sets all of this up:

```ruby
```

spec/requests/api/suppliers_controller_locations_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  let(:response_json) { JSON.parse(response.body) }
+
+  describe 'GET /suppliers/:supplier_id/locations' do
+    let(:supplier) { create(:supplier) }
+
+    it 'returns the locations' do
+      location1 = create(:location, suppliers: [supplier])
+      location2 = create(:location, suppliers: [supplier])
+      create_list(:move, 2, :requested, supplier: supplier, from_location: location1) # to tests that duplicates locations are not shown
```
smoothcontract, line 17: 👍

spec/requests/api/suppliers_controller_locations_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      create(:move, :requested, supplier: supplier, from_location: location2)
+
+      get "/api/v1/suppliers/#{supplier.id}/locations", headers: headers, params: {}
+
+      response_location_keys = response_json['data']
+                                 .select { |e| e['type'] == 'locations' }
+                                 .map { |l| l['attributes']['key'] }
+
+      expect(response).to have_http_status(:success)
+      expect(response_location_keys).to match_array([location1.key, location2.key])
```
smoothcontract, line 27: Would prefer to split this into two separate tests ideally - also could this use the standard approach:

```ruby
      it_behaves_like 'an endpoint that responds with success 200'
```

Would be nice to move the request out into a subject line so it can be called as needed too.

e.g.

```ruby
subject(:get_supplier_locations) { get "/api/v1/suppliers/#{supplier_id}/locations", headers: headers, params: {} }

it_behaves_like 'an endpoint that responds with success 200' do
  get_supplier_locations
end

File: spec/requests/api/suppliers_controller_locations_spec.rb [hunk 6]
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  context 'when the supplier_id is not found' do
+    let(:schema) { load_yaml_schema('error_responses.yaml') }
+
+    let(:supplier_id) { 'invalid-supplier-id' }
+    let(:detail_404) { "Couldn't find Supplier with 'id'=invalid-supplier-id" }
+
+    before do
+      get "/api/v1/suppliers/#{supplier_id}/locations", headers: headers, params: {}
```
smoothcontract, line 38: Could use `get_supplier_locations` here too

spec/requests/api/suppliers_controller_locations_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+require 'rails_helper'
+
+RSpec.describe Api::SuppliersController do
+  let(:access_token) { 'spoofed-token' }
+  let(:headers) { { 'CONTENT_TYPE': content_type, 'Authorization': "Bearer #{access_token}" } }
+  let(:content_type) { ApiController::CONTENT_TYPE }
+  let(:response_json) { JSON.parse(response.body) }
+
+  describe 'GET /suppliers/:supplier_id/locations' do
+    let(:supplier) { create(:supplier) }
```
smoothcontract, line 12: Add

```ruby
let(:supplier_id) { supplier.id }
```

spec/requests/api/suppliers_controller_locations_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    it 'returns the locations' do
+      location1 = create(:location, suppliers: [supplier])
+      location2 = create(:location, suppliers: [supplier])
```
smoothcontract, line 16: Maybe add a another location that's not linked to a move - or maybe linked only via `to_location` to demonstrate that the move `from_location` is the only thing that associates a move to a supplier. Perhaps add a dedicated spec for this just to make the behaviour crystal clear.

smoothcontract (commented):
Great progress, left a few suggestions / ideas for you 👍

app/controllers/api/suppliers_controller.rb
alexdesi: Good point, it might be useful to FE.

app/controllers/api/suppliers_controller.rb
alexdesi: We validate the presence of `key`, so we could order by key.
I'll check about the N+1 queries issues, it might be enough just to change joins to includes 🤞

alexdesi: I've checked for the N+1 error, and it looks like the ONLY query generated is:
```SELECT DISTINCT "locations".* FROM "locations" INNER JOIN "moves" ON "moves"."from_location_id" = "locations"."id" WHERE "moves"."supplier_id" = $1 ORDER BY "locations"."key" ASC LIMIT $2  [["supplier_id", "xxxxxxx"], ["LIMIT", 11]]```

I've also created more locations and suppliers in my tests, to highlight the issue, but that's the only query I got in the test log.

File: spec/requests/api/suppliers_controller_locations_spec.rb (hunk 3 above)
Comment (alexdesi) [reply spec/requests/api/suppliers_controller_locations_spec.rb:8]: 👍 I didn't know that.
Removed!

File: app/controllers/api/suppliers_controller.rb (hunk 2 above)
Comment (smoothcontract) [reply app/controllers/api/suppliers_controller.rb:14]: Hmm.. strange. I pulled this branch and ran locally to test the endpoint via Postman and saw extra queries for each supplier. I suspect it's the LocationSerializer that's causing the extra supplier fetches. Have you tried via postman with a bit more data rather than just the test data? I used an old dump from staging. You can resolve it with a simple `includes(:supplier)` anyway.

File: spec/requests/api/suppliers_controller_locations_spec.rb (hunk 5 above)
Comment (alexdesi) [reply spec/requests/api/suppliers_controller_locations_spec.rb:27]: nice one, yes I'agree, let's make this test more "standard"

File: spec/requests/api/suppliers_controller_locations_spec.rb [hunk 9]
```diff
@@ -0,0 +1,88 @@
...
+    let(:schema) { load_yaml_schema('get_locations_responses.yaml') }
+
+    let(:supplier) { create(:supplier) }
+    let(:supplier_id) { supplier.id }
+    let(:location_a) { create(:location, suppliers: [supplier], key: 'key_aaa') }
+    let(:location_b) { create(:location, suppliers: [supplier], key: 'key_bbb') }
+
+    it_behaves_like 'an endpoint that responds with success 200' do
+      before do
+        create(:move, :requested, supplier: supplier)
```
alexdesi, line 23: I added this because I wanted to test the schema with a proper response (and not empty array)

We could achieve the same with 'let!', but we try to avoid it.

spec/requests/api/suppliers_controller_locations_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+        another_supplier = create(:supplier)
+        another_location = create(:location, suppliers: [another_supplier])
+        create(:move, :requested, supplier: another_supplier, from_location: another_location)
+        create(:move, :requested, supplier: supplier, from_location: location_a)
+
+        get_supplier_locations
+
+        response_location_keys = response_json['data']
+                                     .select { |e| e['type'] == 'locations' }
+                                     .map { |l| l['attributes']['key'] }
```
alexdesi, line 53: we use this in only 2 places, so I think it's fine not to DRY this for now

alexdesi:
@smoothcontract thanks for the comments, I've done many improvements on it, 
possibly we can merge it,
have a second look, thanks

smoothcontract (approved):
Looks good, nice job 👍

app/controllers/api/suppliers_controller.rb
alexdesi: yes, you are right,  I was adding the includes with the moves, and it did not make any difference,
of course we should do it with the suppliers (`.includes(:suppliers)`) 
since you can add them with the `include` params

Fixed!

PR #843, opened 2020-08-18
P4 2038 generate future move notifications

martyn-w:
### Jira link

P4-2038

### What?

I have added/removed/altered:

- [x] Added rake task `notifications:send` for seeding suppliers with webhooks, for go-live

### Why?

I am doing this because:

- so that on go live suppliers can synchronise their system with ours

- pretty minimal

### DIFF
--- a/app/jobs/prepare_move_notifications_job.rb
+++ b/app/jobs/prepare_move_notifications_job.rb
@@ -4,22 +4,25 @@
 class PrepareMoveNotificationsJob < ApplicationJob
   queue_as :notifications
 
-  def perform(topic_id:, action_name:)
+  def perform(topic_id:, action_name:, send_webhooks: true, send_emails: true, only_supplier_id: nil)
     move = Move.find(topic_id)
 
-    move.suppliers.each do |supplier|
+    # if the move has a specified supplier, use it; otherwise use the move.suppliers delegate based on from_location
+    [move.supplier || move.suppliers].flatten.each do |supplier|
+      next unless only_supplier_id.nil? || only_supplier_id == supplier.id
+
       supplier.subscriptions.kept.each do |subscription|
         next unless subscription.enabled?
 
         # NB: always notify the webhook (if defined) on any change, even for back-dated historic moves
-        if subscription.callback_url.present?
+        if send_webhooks && subscription.callback_url.present?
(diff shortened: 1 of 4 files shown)


app/jobs/prepare_move_notifications_job.rb
```diff
@@ -4,22 +4,25 @@
 class PrepareMoveNotificationsJob < ApplicationJob
   queue_as :notifications
 
-  def perform(topic_id:, action_name:)
+  def perform(topic_id:, action_name:, send_webhooks: true, send_emails: true, only_supplier_id: nil)
     move = Move.find(topic_id)
 
-    move.suppliers.each do |supplier|
+    # if the move has a specified supplier, use it; otherwise use the move.suppliers delegate based on from_location
+    [move.supplier || move.suppliers].flatten.each do |supplier|
```
willfish, line 11: As far as I know, every move has a supplier attached to it, currently. So this could be simplified I think?

martyn-w: I'm afraid not: there are currently 21 moves in production which do not have a `supplier_id`

cwrw, line 11: is there any chance we'll ever have one of each at the same time? Maybe we should merge if that's the case, otherwise this is fine

martyn-w: I think the logic should be: use the `move.supplier_id` if it exists, otherwise infer the supplier from the move's `to_location`: that is how this is currently written. I don't think we should be using _both_ the supplier_id and the inferred suppliers

willfish: So that's interesting. I think those moves were created before the supplier FELTHAM A (HMPYOI) had the location/supplier yaml link applied.

cwrw (approved):

app/jobs/prepare_move_notifications_job.rb
martyn-w: If we had a validation on moves to prevent a nil supplier_id in place then this could be simplified... but there might be other unexpected consequences, so best to tackle that problem separately

willfish: Sounds good

PR #844, opened 2020-08-19
V2.3.2 Update CHANGELOG.md

cwrw:


smoothcontract (approved):
Things are still moving around in here, which is weird but otherwise looks good 👍

willfish (approved):

PR #846, opened 2020-08-19
[P4-1975] Adds type to events table

willfish:
### Jira link

P4-1975

### What?

I have added/removed/altered:

- [x] Added type column to the events table

### Why?

I am doing this because:

- This is required for STI to work in the events table

### DIFF
--- a/db/migrate/20200819130412_add_type_to_event.rb
+++ b/db/migrate/20200819130412_add_type_to_event.rb
@@ -1,0 +1,7 @@
+class AddTypeToEvent < ActiveRecord::Migration[6.0]
+  def change
+    # This column is going to be null for some events (specifically,
+    # those ones that have been created through the legacy api).
+    add_column :events, :type, :string
+  end
+end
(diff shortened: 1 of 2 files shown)


db/migrate/20200819130412_add_type_to_event.rb
```diff
@@ -0,0 +1,7 @@
+class AddTypeToEvent < ActiveRecord::Migration[6.0]
+  def change
+    # This column is going to be null for some events (specifically,
+    # those ones that have been created through the legacy api).
+    add_column :events, :type, :string
```
mafonso, line 5: 👍

db/schema.rb
```diff
@@ -120,6 +120,7 @@
     t.datetime "updated_at", null: false
     t.uuid "eventable_id", null: false
     t.string "eventable_type", null: false
+    t.string "type"
```
mafonso, line 123: 👍

mafonso (approved):

PR #848, opened 2020-08-19
Revert notes

willfish:
### Jira link

P4-<TODO>

### What?

Revert accidentally committed change to db to master

### Why?

I am doing this because:

- We have a process of review and this was committed by accident


mafonso (approved):

willfish:
Spoken to Miguel about this. We actually want the change and rollback is going to be manual.

PR #854, opened 2020-08-20
Move Nomis file fixtures to a dedicated folder

smoothcontract:
### Jira link

P4-<TODO>

### What?

- [x] Moved and renamed Nomis test fixtures to new `nomis` folder

### Why?

- I was working on adding a fixture for another story and the existing arrangement struck me as slightly messy. So this is tidier.


martyn-w (approved):
👍

PR #859, opened 2020-08-20
Add support for importing supplier location mapping with effective dates

smoothcontract:
### Jira link

P4-2064

### What?

- [x]  Updated format of existing `supplier_locations.yaml` file to include extra keys with `effective_from` and `effective_to` dates.
- [x] Modified existing importer (using HABTM join table "locations_suppliers) to work with new file format (but ignores dates)
- [x] Add new model `SupplierLocation` to represent the join between `Supplier` and `Location`.
- [x] Add new service `SupplierLocations::Importer` to import yaml file with effective date ranges.
- [x] Add new rake task to remove and recreate all `SupplierLocation` relations for all effective date ranges (i.e. process all yaml files)
- [x] Add another yaml file to be used for go live - this has all of the supplier to location mappings needed for new contract effective from 29th.

### Why?

- This allows all the needed data to be prepared for go live with the new supplier contract. The existing approach to link suppliers with locations uses a HABTM relationship with an inferred `locations_suppliers` join table. We need to introduce extra attributes for `effective_from` and `effective_to` dates on the join, which means moving to a has many through relation with an explicit `SupplierLocation` model.
- Normally this situation is a pain to migrate the data but in this case it's easy as the data is entirely recreated from the YAML files. The approach in this PR is to build out a parallel model so that the data can be populated on all environments, then in a future story use this data within the `SupplierChooser` to determine move ownership. For now this data is ignored so can be merged and deployed safely. We can even continue to run the legacy `link_suppliers` task if needed. After go live the legacy task and `locations_suppliers` table can be removed.

### Have you? (optional)

- [x] Updated README to include new task

- Adds new table and model, but is not yet used so minimal risk


spec/models/supplier_location_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+RSpec.describe SupplierLocation do
+  it { is_expected.to belong_to(:supplier) }
+  it { is_expected.to belong_to(:location) }
+
+  it 'prevents effective_from > effective_to' do
+    expect(build(:supplier_location, effective_from: '2020-03-04', effective_to: '2020-03-03')).not_to be_valid
+  end
+
+  it 'allows effective_from == effective_to' do
```
alexdesi, line 13: `allows effective_from to be greater than effective_to`
and 
`allows effective_from to be equal to effective_to`
are a bit clearer, 
not very opinionated, up to you

spec/services/supplier_locations/importer_spec.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe SupplierLocations::Importer do
+  subject(:call_importer) { importer.call }
+
+  let(:importer) { described_class.new(filename) }
```
alexdesi, line 8: we use importer only in the subject,
maybe we can simplify it with just single line: `subject(:call_importer) { described_class.new(filename).call }`

spec/services/supplier_locations/importer_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+    let(:effective_to) { Date.parse('2020-12-31') }
+
+    let!(:supplier1) { create(:supplier, key: 'supplier1') }
+    let!(:supplier2) { create(:supplier, key: 'supplier2') }
+    let!(:location1) { create(:location, nomis_agency_id: 'LOC1') }
+    let!(:location2) { create(:location, nomis_agency_id: 'LOC2') }
+    let!(:location3) { create(:location, nomis_agency_id: 'LOC3') }
+
+    it 'creates expected supplier locations' do
+      expect { call_importer }.to change(SupplierLocation, :count).by(4)
```
alexdesi, line 22: I am totally sure why 4 is the expected supplier locations? 
is there a way to make it clear  from the test, or maybe just add a comment 😄

app/services/supplier_locations/importer.rb
```diff
@@ -0,0 +1,57 @@
...
+    def yaml
+      @yaml ||= ActiveSupport::HashWithIndifferentAccess.new(YAML.safe_load(File.read(filename)))
+    end
+
+    def suppliers
+      @suppliers ||= yaml[:suppliers]
+    end
+
+    def effective_from
+      @effective_from ||= yaml[:effective_from].presence && Date.strptime(yaml[:effective_from], DATE_FORMAT)
```
willfish, line 41: I don't think we need to get fiddly with dates, here? Can't we just do `Date.parse`?

spec/services/supplier_locations/importer_spec.rb
willfish: See spec/fixtures/files/supplier_locations/valid.yaml @alexdesi

PR #863, opened 2020-08-21
Rake to update suppliers for all moves

alexdesi:
### Jira link

https://dsdmoj.atlassian.net/browse/P4-2036

### What?

This is a small rake task to update all the Supplier,
This is not yet completed since the new SupplierChooser is WIP.

### DIFF
--- a/lib/tasks/update_supplier_for_moves.rake
+++ b/lib/tasks/update_supplier_for_moves.rake
@@ -5,19 +5,34 @@ namespace :moves do
   task update_suppliers: :environment do
     total = Move.count
 
-    i = 0
-
     puts 'Updating suppliers for moves...'
-    Move.find_each(batch_size: 200) do |move|
-      # TODO: need to update this whith the new SupplierChooser
-      new_supplier = SupplierChooser.new(nil, nil).call
-      move.update(supplier: new_supplier)
+    Move.find_each(batch_size: 200).with_index do |move, i|
+      effective_date = move.date.presence || move.date_from
+      supplier = SupplierChooser.new(effective_date, move.from_location).call
 
-      if (total % 200).zero?
-        puts "Updated #{i += 200} out of #{total} moves..."
+      if supplier.nil?
+        puts_error(move)
+      else
+        move.supplier = supplier
+        move.save(validate: false)
+      end
+
+      if (i % 200).zero?
+        puts "Processed #{i} / #{total} moves..."
       end
     end
 
     puts 'Task completed.'
+    puts 'To confirm you want to save the move, please specify the VAR "ENV[confirm]=true""' unless ENV['confirm_save'] == 'true'
   end
+end
+
+private
+
+def puts_error(move)
+  supplier_location = SupplierLocation.location(move.from_location.id) if move.from_location
+
+  puts "ERROR: #{move.id} has supplier = nil!"
+  puts "  Debug info: move.date: move.date:#{move.date}, move.date_from: #{move.date_from}, " \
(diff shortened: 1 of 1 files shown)


alexdesi:
@Andrew Retmanski I've made some changes on the rake tasks: 
now, we can run it just to simulate the execution, without changing the moves.

To confirm we want to update the moves, run the rake with the confirm_save ENV var, this way:
`confirm_save=true bundle exec rake moves:update_suppliers`

lib/tasks/update_supplier_for_moves.rake
```diff
@@ -0,0 +1,28 @@
...
+namespace :moves do
+  desc 'Update the suppliers for all the moves based on the effective dates of the Suppliers'
+  task update_suppliers: :environment do
+    total = Move.count
+
+    i = 0
+
+    puts 'Updating suppliers for moves...'
+    Move.find_each(batch_size: 200) do |move|
+      move.determine_supplier # TODO: this might need to change
```
smoothcontract, line 12: This would have worked with the `force` parameter, but I removed that so instead this will need to replicate some code from that method. Something like:

```ruby
effective_date = move.date.presence || move.date_from
move.supplier = SupplierChooser.new(effective_date, move.from_location).call
```

smoothcontract: It would be really good to flag up any moves that we can't find a suitable supplier for - this would indicate some missing data in `SupplierLocation`, so perhaps log those to the console with the relevant move reference, effective date and from location? Then we can investigate any of those we see when running the rake task. The theory is that this should run and if everything works, every move will have a supplier id assigned, but if there is missing data we need to know.

lib/tasks/update_supplier_for_moves.rake
```diff
@@ -0,0 +1,28 @@
...
+
+      if ENV['confirm_save'] == 'true'
```
smoothcontract, line 14: Not really keen on using environment variables for this - or that we need this extra complexity to be honest. If you want to keep it then would be better to use a parameter passed into the rake task instead - and document it in the rake task description.

lib/tasks/update_supplier_for_moves.rake
```diff
@@ -0,0 +1,28 @@
...
+        move.save!
+      else # Simulation mode
+        puts("Move: #{move.id} is not valid!") unless move.valid?
+      end
+
+      if (total % 200).zero?
```
smoothcontract, line 20: I don't follow the logic of this - `total` value doesn't change, so why evaluate this expression for every move?

lib/tasks/update_supplier_for_moves.rake
alexdesi: I agree! we should log the moves in which is not possible to get the supplier id, and figuring out why.

lib/tasks/update_supplier_for_moves.rake
alexdesi: Yeah I'll remove it and I'll add more useful log

lib/tasks/update_supplier_for_moves.rake
alexdesi: good catch,
that's should be  `if (i % 200).zero?`
thanks

smoothcontract: I don't think that quite works either, as `i` is only incremented inside this check

smoothcontract: Try something like this

```ruby
i += 1
if (i % 200).zero?
  puts "Updated #{i} out of #{total} moves..."
end
```

smoothcontract (approved):
Looks good 👍

PR #864, opened 2020-08-21
Use move date/date_from to determine correct supplier for move

smoothcontract:
### Jira link

P4-2041 P4-2062

### What?

- [x] Update SupplierChooser service to consider move date/date_from to determine correct supplier
- [x] Ensure that SupplierChooser is invoked whenever the move date or date_from is changed to ensure that the supplier choice remains correct.
- [x] Replace existing HABTM relationship between Supplier and Locations with SupplierLocation join
- [x] Remove legacy rake task `reference_data:link_suppliers` and associated helper methods
- [x] Update Locations reference data service to provide locations linked to supplier for current date (i.e. these change before/after go live date)
- [x] Remove redundant `supplier` method on Location model - this used to be used prior to introduction of SupplierChooser service.
- [x] Updated specs that were demonstrating supplier assignment based solely on API token - this behaviour is now removed in favour of using move date.

### Why?

- New SupplierLocation model supports effective dates for supplier contracts. We need to ensure that moves are assigned the correct supplier based on the move date - the location mapping changes either side of the go live date.
- Whenever the move date changes we need to determine the supplier location again as it may potentially have changed. If the date doesn't change then we can skip the update and avoid an extra database query.
- The legacy table `locations_suppliers` was used only for the HABTM relationship and is now ignored (will be removed in a future PR). The legacy rake task to populate that is hence no longer needed.

### Have you? (optional)

- [x] Updated README

- This affects a few API's but should not be a breaking change providing the SupplierLocation data is correctly populated before deploying this PR. A previous PR introduced a new rake task `create_supplier_locations` which needs to be run first on each environment before deploying this work, otherwise moves will potentially have a `nil` supplier assigned which means suppliers cannot see their moves.

### DIFF
--- a/app/controllers/api/v1/moves_actions.rb
+++ b/app/controllers/api/v1/moves_actions.rb
@@ -12,6 +12,7 @@ module Api::V1
 
     def create_and_render
       move = Move.new(new_move_attributes)
+      move.determine_supplier
       move.profile.documents = profile_documents
 
       authorize!(:create, move)
@@ -83,7 +84,6 @@ module Api::V1
           to_location: Location.find_by(id: new_move_params.dig(:relationships, :to_location, :data, :id)),
           court_hearings: CourtHearing.where(id: (new_move_params.dig(:relationships, :court_hearings, :data) || []).map { |court_hearing| court_hearing[:id] }),
           prison_transfer_reason: PrisonTransferReason.find_by(id: new_move_params.dig(:relationships, :prison_transfer_reason, :data, :id)),
-          supplier: SupplierChooser.new(doorkeeper_application_owner, from_location).call,
         )
       end
     end
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -46,7 +46,8 @@ class Location < ApplicationRecord
 
   NOMIS_TYPES_WITH_DOCUMENTS = %w[STC SCH].freeze
 
-  has_and_belongs_to_many :suppliers
+  has_many :supplier_locations
+  has_many :suppliers, through: :supplier_locations
   has_and_belongs_to_many :regions
   # Deleting locations isn't really a thing in practice - so dependent: :destroy is a pragmatic choice
   has_many :moves_from, class_name: 'Move', foreign_key: :from_location_id, inverse_of: :from_location, dependent: :destroy
@@ -56,7 +57,6 @@ class Location < ApplicationRecord
   validates :title, presence: true
   validates :location_type, presence: true, inclusion: { in: location_types }
 
-  scope :supplier, ->(supplier_id) { joins(:suppliers).where(locations_suppliers: { supplier_id: supplier_id }) }
   scope :ordered_by_title, ->(direction) { order('locations.title' => direction) }
   scope :search_by_title, ->(search) { select(:id).where('title ILIKE :search', search: "%#{search}%") }
 
--- a/app/models/move.rb
+++ b/app/models/move.rb
@@ -174,7 +174,7 @@ class Move < VersionedModel
   def determine_supplier
     # Supplier choice is based on from_location and effective date. As the from location cannot change after the
     # move is created we only need to look for date changes - if no change then skip the supplier update for speed.
-    return supplier unless date_changed? || date_from_changed?
+    return supplier unless new_record? || date_changed? || date_from_changed?
 
     effective_date = date.presence || date_from
     self.supplier = SupplierChooser.new(effective_date, from_location).call
--- a/app/models/move.rb
+++ b/app/models/move.rb
@@ -171,6 +171,15 @@ class Move < VersionedModel
     feed_attributes
   end
 
+  def determine_supplier
+    # Supplier choice is based on from_location and effective date. As the from location cannot change after the
+    # move is created we only need to look for date changes - if no change then skip the supplier update for speed.
+    return supplier unless date_changed? || date_from_changed?
+
+    effective_date = date.presence || date_from
+    self.supplier = SupplierChooser.new(effective_date, from_location).call
+  end
+
 private
 
   def date_to_after_date_from
--- a/app/models/supplier_location.rb
+++ b/app/models/supplier_location.rb
@@ -6,6 +6,11 @@ class SupplierLocation < ApplicationRecord
 
   validate :effective_to_after_effective_from
 
+  scope :location, ->(location_id) { where(location_id: location_id) }
+  scope :effective_from, ->(date) { where(effective_from: nil).or(where('effective_from <= ?', date)) }
+  scope :effective_to, ->(date) { where(effective_to: nil).or(where('effective_to >= ?', date)) }
+  scope :effective_on, ->(date) { effective_from(date).effective_to(date) }
+
   def self.link_locations(effective_from: nil, effective_to: nil, supplier:, locations:)
     locations.each do |location|
       create!(effective_from: effective_from, effective_to: effective_to, supplier: supplier, location: location)
--- a/app/services/allocations/creator.rb
+++ b/app/services/allocations/creator.rb
@@ -2,10 +2,9 @@
 
 module Allocations
   class Creator
-    attr_accessor :doorkeeper_application_owner, :allocation_params, :complex_case_params, :allocation
+    attr_accessor :allocation_params, :complex_case_params, :allocation
 
-    def initialize(doorkeeper_application_owner:, allocation_params:, complex_case_params:)
-      self.doorkeeper_application_owner = doorkeeper_application_owner
+    def initialize(allocation_params:, complex_case_params:)
       self.allocation_params = allocation_params
       self.complex_case_params = complex_case_params
     end
@@ -28,7 +27,7 @@ module Allocations
     end
 
     def moves
-      supplier = SupplierChooser.new(doorkeeper_application_owner, allocation.from_location).call
+      supplier = SupplierChooser.new(allocation.date, allocation.from_location).call
 
       Array.new(allocation.moves_count) do
         Move.new(
--- a/app/services/event_log/move_runner.rb
+++ b/app/services/event_log/move_runner.rb
@@ -16,6 +16,7 @@ module EventLog
         when Event::APPROVE
           move.status = Move::MOVE_STATUS_REQUESTED
           move.date = event.date
+          move.determine_supplier
           Allocations::CreateInNomis.call(move) if event.create_in_nomis?
         when Event::CANCEL
           move.status = Move::MOVE_STATUS_CANCELLED
--- a/app/services/locations/finder.rb
+++ b/app/services/locations/finder.rb
@@ -24,7 +24,8 @@ module Locations
     def apply_supplier_filters(scope)
       return scope unless filter_params.key?(:supplier_id)
 
-      scope.joins(:locations_suppliers).where(suppliers: { id: filter_params[:supplier_id] })
+      scope = scope.where(suppliers: { id: filter_params[:supplier_id] })
+      scope.merge(SupplierLocation.effective_on(Time.zone.today))
     end
   end
 end
--- a/app/services/moves/updater.rb
+++ b/app/services/moves/updater.rb
@@ -11,6 +11,7 @@ module Moves
 
     def call
       move.assign_attributes(attributes)
+      move.determine_supplier
       # NB: rather than update directly, we need to detect whether the move status has changed before saving the record
       self.status_changed = move.status_changed?
 
--- a/app/services/supplier_chooser.rb
+++ b/app/services/supplier_chooser.rb
@@ -9,6 +9,7 @@ class SupplierChooser
   def call
     return unless effective_date.present? && location.present?
 
+    # NB: business rules _should_ preclude multiple suppliers for visibility
     supplier_location = SupplierLocation.location(location.id).effective_on(effective_date).first
     supplier_location&.supplier
   end
--- a/app/services/supplier_chooser.rb
+++ b/app/services/supplier_chooser.rb
@@ -1,14 +1,15 @@
 class SupplierChooser
-  def initialize(doorkeeper_application_owner, from_location)
-    @doorkeeper_application_owner = doorkeeper_application_owner
-    @from_location = from_location
+  attr_accessor :effective_date, :location
+
+  def initialize(effective_date, location)
+    self.effective_date = effective_date
+    self.location = location
   end
 
   def call
-    return @doorkeeper_application_owner if @doorkeeper_application_owner
+    return unless effective_date.present? && location.present?
 
-    # TODO: Replace with the frontend supplied supplier
-    #       There can only ever be one supplier per location, currently
-    @from_location.suppliers.first
+    supplier_location = SupplierLocation.location(location.id).effective_on(effective_date).first
+    supplier_location&.supplier
   end
 end
--- a/spec/models/move_spec.rb
+++ b/spec/models/move_spec.rb
@@ -543,6 +543,63 @@ RSpec.describe Move do
     end
   end
 
+  describe '#determine_supplier' do
+    subject(:determined_supplier) { move.determine_supplier }
+
+    let(:location) { create(:location, :police) }
+    let(:old_supplier) { create(:supplier) }
+    let(:new_supplier) { build(:supplier) }
+    let(:date) { Date.today }
+    let(:date_from) { Date.today }
+    let(:move) { create(:move, :cancelled, from_location: location, to_location: nil, supplier: old_supplier, date: date, date_from: date_from) }
+
+    before do
+      supplier_chooser = instance_double('SupplierChooser', call: new_supplier)
+      allow(SupplierChooser).to receive(:new).and_return(supplier_chooser)
+    end
+
+    context 'when date and date_from have not changed' do
+      it 'returns existing supplier' do
+        expect(determined_supplier).to eq old_supplier
+      end
+
+      it 'does not call SupplierChooser service' do
+        determined_supplier
+        expect(SupplierChooser).not_to have_received(:new)
+      end
+    end
+
+    context 'when date is present and has changed' do
+      let(:date_from) { nil }
+
+      before { move.date = Date.tomorrow }
+
+      it 'returns updated supplier' do
+        expect(determined_supplier).to eq new_supplier
+      end
+
+      it 'calls SupplierChooser service to update supplier' do
+        determined_supplier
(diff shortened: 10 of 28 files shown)


app/controllers/api/v1/moves_actions.rb
```diff
@@ -14,6 +14,7 @@ def show_and_render
 
     def create_and_render
       move = Move.new(new_move_attributes)
+      move.determine_supplier
```
alexdesi, line 17: I think now we can remove this extra step and add the `supplier` in the `new_move_attributes`.

app/controllers/api/v1/moves_actions.rb
```diff
@@ -83,7 +84,6 @@ def new_move_attributes
           to_location: Location.find_by(id: new_move_params.dig(:relationships, :to_location, :data, :id)),
           court_hearings: CourtHearing.where(id: (new_move_params.dig(:relationships, :court_hearings, :data) || []).map { |court_hearing| court_hearing[:id] }),
           prison_transfer_reason: PrisonTransferReason.find_by(id: new_move_params.dig(:relationships, :prison_transfer_reason, :data, :id)),
-          supplier: SupplierChooser.new(doorkeeper_application_owner, from_location).call,
```
alexdesi, line 86: are there any vantages removing this?

this could be 
```
  supplier: determine_supplier
```
with some changes on the `determine_supplier`, that should just return the supplier, no assignment in it.
If this is possible, we could keep some controller's actions cleaner

app/models/supplier_location.rb
```diff
@@ -6,6 +6,11 @@ class SupplierLocation < ApplicationRecord
 
   validate :effective_to_after_effective_from
 
+  scope :location, ->(location_id) { where(location_id: location_id) }
```
alexdesi, line 9: Do we need to make the location a scope? that's so simple

app/models/supplier_location.rb
```diff
@@ -6,6 +6,11 @@ class SupplierLocation < ApplicationRecord
...
+  scope :effective_from, ->(date) { where(effective_from: nil).or(where('effective_from <= ?', date)) }
+  scope :effective_to, ->(date) { where(effective_to: nil).or(where('effective_to >= ?', date)) }
+  scope :effective_on, ->(date) { effective_from(date).effective_to(date) }
+
```
alexdesi, line 13: nice and clean! 👌

app/services/allocations/creator.rb
```diff
@@ -2,10 +2,9 @@
...
 module Allocations
   class Creator
-    attr_accessor :doorkeeper_application_owner, :allocation_params, :complex_case_params, :allocation
+    attr_accessor :allocation_params, :complex_case_params, :allocation
 
-    def initialize(doorkeeper_application_owner:, allocation_params:, complex_case_params:)
-      self.doorkeeper_application_owner = doorkeeper_application_owner
+    def initialize(allocation_params:, complex_case_params:)
       self.allocation_params = allocation_params
       self.complex_case_params = complex_case_params
```
alexdesi, line 9: meanwhile could you change these to:
```suggestion
      @allocation_params = allocation_params
      @complex_case_params = complex_case_params
```
it just sounds right with instance variables

app/services/event_log/move_runner.rb
```diff
@@ -16,6 +16,7 @@ def call
         when Event::APPROVE
           move.status = Move::MOVE_STATUS_REQUESTED
           move.date = event.date
+          move.determine_supplier
```
alexdesi, line 19: so once it is approved we need to determine again the supplier  ... yeah it sounds right, there might be other cases like this.
Also:
in case you change `determine_supplier`
this would be something like:
`move.supplier = determine_supplier`

app/services/locations/finder.rb
```diff
@@ -24,7 +24,8 @@ def apply_filters(scope)
     def apply_supplier_filters(scope)
       return scope unless filter_params.key?(:supplier_id)
 
-      scope.joins(:locations_suppliers).where(suppliers: { id: filter_params[:supplier_id] })
+      scope = scope.where(suppliers: { id: filter_params[:supplier_id] })
+      scope.merge(SupplierLocation.effective_on(Time.zone.today))
```
alexdesi, line 28: Not sure if this should be here 🤔 
maybe just rename the method to `apply_effective_supplier_filters`

app/services/moves/updater.rb
```diff
@@ -11,6 +11,7 @@ def initialize(move, move_params)
 
     def call
       move.assign_attributes(attributes)
+      move.determine_supplier
```
alexdesi, line 14: clean up this if possiblele

app/services/supplier_chooser.rb
```diff
@@ -1,14 +1,15 @@
 class SupplierChooser
-  def initialize(doorkeeper_application_owner, from_location)
-    @doorkeeper_application_owner = doorkeeper_application_owner
-    @from_location = from_location
+  attr_accessor :effective_date, :location
+
+  def initialize(effective_date, location)
+    self.effective_date = effective_date
+    self.location = location
```
alexdesi, line 6: I think generally in this project we prefer using instance variables `@`, 
also, it is less ambiguous, self.effective_date could be a variable or a method, with `@effective_date` it's very clear what it is.
https://stackoverflow.com/questions/1693243/instance-variable-self-vs

spec/models/move_spec.rb
```diff
@@ -543,6 +543,63 @@
...
+      end
+
+      it 'does not call SupplierChooser service' do
+        determined_supplier
+        expect(SupplierChooser).not_to have_received(:new)
+      end
+    end
+
+    context 'when date is present and has changed' do
+      let(:date_from) { nil }
```
alexdesi, line 573: we need to set date_from to nil?
this makes the test a bit confusing: "when date is present and has changed"

spec/models/move_spec.rb
```diff
@@ -543,6 +543,63 @@
...
+      end
+
+      it 'calls SupplierChooser service to update supplier' do
+        determined_supplier
+        expect(SupplierChooser).to have_received(:new).with(Date.tomorrow, location)
+      end
+    end
+
+    context 'when date_from is present and has changed' do
+      let(:date) { nil }
```
alexdesi, line 588: similar comment, 
why if the test states "when date_from is present and has changed"
does date need to be set to nil?

spec/models/supplier_location_spec.rb
```diff
@@ -43,4 +43,58 @@
...
+    let(:location) { create(:location) }
+    let(:supplier) { create(:supplier) }
+    let(:date) { Date.today }
+
+    context 'with a nil date parameter' do
+      let!(:supplier_location1) { create(:supplier_location, supplier: supplier, location: location) }
+      let!(:supplier_location2) { create(:supplier_location, supplier: supplier, location: location, effective_from: Date.today, effective_to: Date.today) }
+      let(:date) { nil }
+
+      it 'returns matching supplier locations' do
```
alexdesi, line 59: I did not fully understand this test:

with a nil date parameter
  returns matching supplier locations (only supplier_location1)

why? what's happen when date is nil?
it might be enough only to amend the test description, thanks

spec/models/supplier_location_spec.rb
```diff
@@ -43,4 +43,58 @@
...
+        expect(effective_on).to contain_exactly(supplier_location1)
+      end
+    end
+
+    context 'without an effective from or to date' do
+      let!(:supplier_location) { create(:supplier_location, supplier: supplier, location: location) }
+
+      it 'returns matching supplier locations' do
```
alexdesi, line 67: same comments, I think "matching supplier locations" should clarify what it is returning

spec/models/supplier_location_spec.rb
```diff
@@ -43,4 +43,58 @@
...
+        expect(effective_on).to contain_exactly(supplier_location1)
+      end
+    end
+
+    context 'with only an effective to date' do
+      let!(:supplier_location1) { create(:supplier_location, supplier: supplier, location: location, effective_to: date) }
+      let!(:supplier_location2) { create(:supplier_location, supplier: supplier, location: location, effective_to: date.yesterday) }
+
+      it 'returns matching supplier locations' do
```
alexdesi, line 85: same

spec/models/supplier_location_spec.rb
```diff
@@ -43,4 +43,58 @@
...
+        expect(effective_on).to contain_exactly(supplier_location1)
+      end
+    end
+
+    context 'with effective from and to date' do
+      let!(:supplier_location1) { create(:supplier_location, supplier: supplier, location: location, effective_from: date, effective_to: date) }
+      let!(:supplier_location2) { create(:supplier_location, supplier: supplier, location: location, effective_to: date.yesterday) }
+      let!(:supplier_location3) { create(:supplier_location, supplier: supplier, location: location, effective_from: date.tomorrow) }
+
+      it 'returns matching supplier locations' do
```
alexdesi, line 95: same

spec/services/locations/finder_spec.rb
```diff
@@ -6,18 +6,23 @@
...
   describe 'filtering' do
     context 'with a supplier' do
       let(:filter_params) { { supplier_id: supplier.id } }
 
       before do
-        location.suppliers << supplier
+        create(:location) # Not linked to supplier
+        create(:supplier_location, supplier: supplier, location: other_location, effective_to: Date.yesterday) # Not effective today
+        create(:supplier_location, supplier: supplier, location: location1) # Linked and effective
+        create(:supplier_location, supplier: supplier, location: location2) # Linked and effective
```
alexdesi, line 21: 👍  for the comments, very helpful! 😄

spec/services/supplier_chooser_spec.rb
```diff
@@ -1,23 +1,82 @@
...
 
-  context 'when doorkeeper_application_owner is nil' do
-    let(:doorkeeper_application_owner) { nil }
+  context 'without an effective from or to date' do
+    let!(:supplier_location) { create(:supplier_location, supplier: supplier1, location: location) }
 
-    it 'returns the first supplier from move.from_location.suppliers' do
-      expect(service.call).to eq(supplier_one)
+    it 'returns matching supplier' do
+      expect(service.call).to eq(supplier1)
```
alexdesi, line 13: sorry it's not clear why it is returning only supplier1 to me

alexdesi (commented):
Great job here, I added few comments about cleaning up things a bit,
and I struggled to understand many tests only described with:

`it 'returns matching ...' do`

it took me like 5 min to understand some of them.
A better description could really help.

app/controllers/api/v1/moves_actions.rb
smoothcontract: err.. no I don't think so. We need the move instance before we can call methods on it.

app/controllers/api/v1/moves_actions.rb
smoothcontract: see previous comment. This is encapsulated in the move instance method `determine_supplier` - and needs to be passed either the move or move_date - better to keep that logic within the move model. So that means we need to create the move first, and then call `determine_supplier` on the move.

app/models/supplier_location.rb
smoothcontract: why not?

app/services/allocations/creator.rb
smoothcontract: I replied to this suggestion before. We have the same established pattern in almost every service so I don't think this should be changed.

app/services/event_log/move_runner.rb
smoothcontract: This has test coverage to ensure the supplier gets updated, so that would catch any implementation changes. I've covered all the cases where the move date can change too.

app/services/locations/finder.rb
smoothcontract: The locations for a given supplier depend on the effective date - as this is a reference data endpoint it should return the current set of locations for the supplier. Then when suppliers call this before/after the go live date they receive the correct set of locations according to their contract.

app/services/moves/updater.rb
smoothcontract: I don't follow what you mean?

app/services/supplier_chooser.rb
smoothcontract: This seems to be a personal preference thing - and we use this style extensively:

Screenshot 2020-08-25 at 13 31 10

spec/models/move_spec.rb
smoothcontract: Yes because the implementation will use the `date` in preference if it's present.

spec/models/move_spec.rb
smoothcontract: See above

app/controllers/api/v1/moves_actions.rb
alexdesi: ok, it makes sense, I am just a bit worried every time we add new lines in the controllers

app/models/supplier_location.rb
alexdesi: Scope does make sense only with a meaningful name (like :cancelled, :active, published, ...)
The name "location" does not say anything about "the scope", the name is confusing, it sounds like a belongs_to relationship, 
if you want to keep it I'd suggest to rename it  with `with_location` or `having_location`, ... up to you :) 

`SupplierLocation.location(location.id)....`

app/models/supplier_location.rb
```diff
@@ -6,6 +6,11 @@ class SupplierLocation < ApplicationRecord
 
   validate :effective_to_after_effective_from
 
+  scope :location, ->(location_id) { where(location_id: location_id) }
+  scope :effective_from, ->(date) { where(effective_from: nil).or(where('effective_from <= ?', date)) }
+  scope :effective_to, ->(date) { where(effective_to: nil).or(where('effective_to >= ?', date)) }
+  scope :effective_on, ->(date) { effective_from(date).effective_to(date) }
```
alexdesi, line 12: 👍 these are perfect names for a scope

app/services/allocations/creator.rb
alexdesi: really? I thought it was a small minority, ok no problem

app/services/supplier_chooser.rb
alexdesi: 👍

alexdesi (approved):
Hey  @smoothcontract , thanks for the improvements and to make the specs clearer, 
if you can rename the scope called `location` would be great (see new comment in there).

spec/services/supplier_chooser_spec.rb
smoothcontract: I updated the description in the spec, clearer now?

alexdesi:
@smoothcontract:  SyntaxError!
```
 SyntaxError:
  /home/circleci/project/spec/models/supplier_location_spec.rb:100: syntax error, unexpected `end', expecting end-of-input
```

app/models/location.rb
```diff
@@ -56,7 +57,6 @@ class Location < ApplicationRecord
   validates :title, presence: true
   validates :location_type, presence: true, inclusion: { in: location_types }
 
-  scope :supplier, ->(supplier_id) { joins(:suppliers).where(locations_suppliers: { supplier_id: supplier_id }) }
```
cwrw, line 59: I think you missed: https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/lib/tasks/notifications.rake#L23

app/services/supplier_chooser.rb
```diff
@@ -1,14 +1,15 @@
...
   end
 
   def call
-    return @doorkeeper_application_owner if @doorkeeper_application_owner
+    return unless effective_date.present? && location.present?
 
-    # TODO: Replace with the frontend supplied supplier
-    #       There can only ever be one supplier per location, currently
-    @from_location.suppliers.first
+    supplier_location = SupplierLocation.location(location.id).effective_on(effective_date).first
```
cwrw, line 12: is there a chance we might have multiple results returned?

cwrw:
We discussed this offline but for posterity: the possibility of moving the determine supplier method to a callback on move. Even though we are being very explicit here, my only concern is someone adds a change that updates a date and misses determining the supplier, so I think it might be safer to move that to the model. Let me know your thoughts.

cwrw:
Other than the few comments all good 👍 
This is quite an involved and intricate change, well done. I've tested briefly on https://booksecuremove-br-master.herokuapp.com/ and seems ok regarding determining the supplier. I also checked the reference location endpoint for posterity. Hopefully didn't miss anything

app/models/location.rb
smoothcontract: Great catch - that rake task is new and wasn't in master when I removed this scope. The scope doesn't work as intended now anyway, so the rake task will need some tweaking. I'll discuss that with Martyn.

app/services/supplier_chooser.rb
smoothcontract: Potentially, although not intentionally. So this picks one arbitrarily if there are multiple results - which would indeed indicate overlapping effective dates and possibly multiple suppliers that can service the location. The business rule however is that only one supplier serves a location during the contract period.

cwrw: doesn't have to be this pr, but something to think about regarding enforcing this rule to have one either in validations or the rake task

app/models/location.rb
smoothcontract: Discussed and the rake task will be updated to remove use of this method (the method no longer works as intended anyway) prior to running it.

martyn-w: 👍

app/models/move.rb
```diff
@@ -171,6 +171,15 @@ def for_feed
     feed_attributes
   end
 
+  def determine_supplier
+    # Supplier choice is based on from_location and effective date. As the from location cannot change after the
+    # move is created we only need to look for date changes - if no change then skip the supplier update for speed.
+    return supplier unless date_changed? || date_from_changed?
```
martyn-w, line 177: I think this is working for new moves because they must have either a date or a from_date, and so `date_changed? || date_from_changed?` evaluates to `true`. But, it might be a little more readable to have instead:
```
return supplier unless new_record? || date_changed? || date_from_changed?
```

app/services/supplier_chooser.rb
martyn-w: yep, good point: worth adding a comment about this line, e.g. `# NB: business rules _should_ preclude multiple suppliers` for visibility

cwrw (approved):

PR #868, opened 2020-08-24
Fix ecr cleanup

mafonso:
### Jira link

n/a

### What?

I have added/removed/altered:

- [ ] Deletes all UNTAGGED images first while cleaning up ECR

### Why?

I am doing this because:
- If there are images without any tags, it will cause an issue later when we try to filter out images based on tags.
  If they don't have a tag, we don't want to keep it anyway.


martyn-w (approved):
👍

PR #869, opened 2020-08-24
Different endpoint for returning moves in CSV format

smoothcontract:
### Jira link

P4-2028

### What?

- [x] Moved the Moves CSV endpoint from `GET /moves` to `POST /moves/csv`
- [x] Refactored some moves controller/actions stuff to DRY this up across v1 and v2

### Why?

- The original endpoint chosen for this aligned with Rails idioms with a GET verb - primarily chosen as it's returning a list of moves. Whilst this works fine the use of GET imposes a limitation in the length of the passed URL query string - tested to be around 8k characters. This limit is set within the web hosting infrastructure so is not easy to get around.

- Some of the users have a large list of locations they can access, so building a request that specifies each of the required location id filters results in a 10k request - this was blowing up with a 413 error because of the request size. This could be better architected by performing location authorisation in the API rather than the front end (passing some user info to the API instead of locations), but that's a longer term goal and not something we can achieve now.

- To work around the immediate problem, this PR swaps to a POST request which doesn't have the same size limit, but does mean that the filter params must be passed in JSON format rather than as part of the query string. The new endpoint supports the same filter parameters as `GET /moves`. We can't use `POST /moves` for this as that is used to create a new move, so we have chosen `POST /moves/csv` as the new endpoint. This is intentionally undocumented as it's not intended for supplier use (only by the front end). All references to"csv" have been removed in the swagger documentation.

### Have you? (optional)

- [x] Updated API docs if necessary (to undocument the feature!)

- Not yet used in production

### DIFF
--- a/app/controllers/api/v2/moves_actions.rb
+++ b/app/controllers/api/v2/moves_actions.rb
@@ -1,16 +1,10 @@
 module Api::V2
   module MovesActions
     def index_and_render
-      moves = Moves::Finder.new(filter_params, current_ability, params[:sort] || {}).call
-
-      if csv?
-        send_file(Moves::Exporter.new(moves).call, type: 'text/csv', disposition: :inline)
-      else
-        paginate moves,
-                 each_serializer: ::V2::MoveSerializer,
-                 include: included_relationships,
-                 fields: ::V2::MoveSerializer::INCLUDED_FIELDS
-      end
+      paginate moves,
+               each_serializer: ::V2::MoveSerializer,
+               include: included_relationships,
+               fields: ::V2::MoveSerializer::INCLUDED_FIELDS
     end
 
     def show_and_render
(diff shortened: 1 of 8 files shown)


app/controllers/api/v2/moves_actions.rb
```diff
@@ -1,16 +1,10 @@
...
-      else
-        paginate moves,
-                 each_serializer: ::V2::MoveSerializer,
-                 include: included_relationships,
-                 fields: ::V2::MoveSerializer::INCLUDED_FIELDS
-      end
+      paginate moves,
+               each_serializer: ::V2::MoveSerializer,
+               include: included_relationships,
+               fields: ::V2::MoveSerializer::INCLUDED_FIELDS
```
alexdesi, line 7: I like this cleanup 😏

alexdesi (approved):
It looks good to me! 🚀

PR #870, opened 2020-08-24
P4 2081 bulk load moves

martyn-w:
### Jira link

P4-2081

### What?

I have added/removed/altered:

- [x] Added a rake task to bulk load N identical moves, for load testing

### Why?

I am doing this because:

- so we can load test

- none


smoothcontract (commented):
This seems like a strange approach to adopt just for some test data - creating these via the API is going to be a bit slow, and seems to be replicating logic that we already have in the `fake_data` rake tasks.. rather than adding this new code would you consider adapting the existing fake data tasks, or building something along similar lines that leverages FactoryBot fixtures and Faker, creating records directly rather than via the API?

martyn-w:
@smoothcontract thanks for the review. The point here was to load directly via the API (rather than at a lower level) in order to load-test the system, and also to load-test the resulting webhooks sent to suppliers.

smoothcontract:
> @smoothcontract thanks for the review. The point here was to load directly via the API (rather than at a lower level) in order to load-test the system, and also to load-test the resulting webhooks sent to suppliers.

Fair enough, I misunderstood the intent of this PR then. I thought this was simply setting up a large volume of data to then perform load testing against, but this is actually performing the load testing too? If that's the case then that's also a slightly odd approach as there are plenty of great tools to do this rather than hand cranking stuff in Ruby.

martyn-w:
Closing for now, will revisit later.

PR #873, opened 2020-08-25
P4-2061 Fix type validation in Framework Responses

cwrw:
### Jira link

[P4-2061](https://dsdmoj.atlassian.net/browse/P4-2061)

### What?

I have added/removed/altered:

- [x] Moved type validation to run before setting the value in framework responses
- [x] Moved type validation in multiple item objects for framework responses to initializer

### Why?

Check the response type on setting the value in the base response class to avoid having to check errors in validations
later on, which has caused us to duplicate type validation across multiple points when creating/updating records to avoid random errors.

Now we first call super on value setting which checks the type, and fails if it is invalid.
The same is done in the multiple item object, where now the validation error is in the initialisation to avoid duplicating checking the type everywhere.

Can be tested on https://booksecuremove-br-master.herokuapp.com/

### DIFF
--- a/app/models/framework_response.rb
+++ b/app/models/framework_response.rb
@@ -42,6 +42,13 @@ class FrameworkResponse < VersionedModel
     raise ActiveRecord::ReadOnlyRecord, "Can't update framework_responses because person_escort_record is #{person_escort_record.status}"
   end
 
+  def value=(raw_value)
+    unless raw_value.blank? || value_type_valid?(raw_value)
+      errors.add(:value, 'is incorrect type')
+      raise ActiveModel::ValidationError, self
+    end
+  end
+
 private
 
   def set_responded_value
--- a/app/models/framework_response.rb
+++ b/app/models/framework_response.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class FrameworkResponse < VersionedModel
+  ValueTypeError = Class.new(StandardError)
+
   validates :type, presence: true
   validates :responded, inclusion: { in: [true, false] }
 
@@ -44,8 +46,7 @@ class FrameworkResponse < VersionedModel
 
   def value=(raw_value)
     unless raw_value.blank? || value_type_valid?(raw_value)
-      errors.add(:value, 'is incorrect type')
-      raise ActiveModel::ValidationError, self
+      raise FrameworkResponse::ValueTypeError, raw_value
     end
   end
 
--- a/app/models/framework_response/multiple_item_object.rb
+++ b/app/models/framework_response/multiple_item_object.rb
@@ -58,8 +58,7 @@ class FrameworkResponse
 
     def validate_responses_type
       unless responses.blank? || responses_type_valid?
-        errors.add(:responses, 'is incorrect type')
-        raise ActiveModel::ValidationError, self
+        raise FrameworkResponse::ValueTypeError, responses
       end
     end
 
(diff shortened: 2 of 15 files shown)


app/models/framework_response.rb
```diff
@@ -42,6 +42,13 @@ def update_with_flags!(value)
     raise ActiveRecord::ReadOnlyRecord, "Can't update framework_responses because person_escort_record is #{person_escort_record.status}"
   end
 
+  def value=(raw_value)
+    unless raw_value.blank? || value_type_valid?(raw_value)
```
willfish, line 46: Feels weird to reach up and then reach down the inheritance hierarchy to do common then specific validations. I don't think it's unforgivable but certainly it's unexpected.

Would it be more obvious to just duplicate the common blank value check and error adding, here?

willfish (approved):

app/models/framework_response.rb
cwrw: I agree this isn't the best approach, I tried going through the before validation callback initially, but the problem is that we need this to fail before trying to set the value, due to the nested structures and the way they are built. Happy to discuss other ways of failing early.
Not sure I understand what you mean with duplicating it, as in move this to each STI class? I think it would be nice to utilise the base class to avoid duplication for type validations

willfish: Yeah, duplicate in the STI class was my thinking. It's okay

app/models/framework_response.rb
```diff
@@ -42,6 +42,13 @@ def update_with_flags!(value)
...
+      errors.add(:value, 'is incorrect type')
+      raise ActiveModel::ValidationError, self
```
smoothcontract, line 48: Discussed offline - liking the simplicity this provides further down the STI tree, but I find it a little strange to be throwing a `ValidationError` without calling `save` or `valid?`. If this was changed to a different exception rather than just adding to the model errors array this would perhaps feel more explicit, and the controller code could catch the exception and convert to a suitable error message for the response.

Ruby isn't typed, and this is essentially a type error - so maybe use `ArgumentError` as the closest approximation to that, or add a custom exception class - and pass in whatever arguments are useful to help build the eventual error to be served to the client.

app/models/framework_response/multiple_item_object.rb
```diff
@@ -56,8 +57,9 @@ def response_objects
     end
 
     def validate_responses_type
-      unless responses_type_valid?
+      unless responses.blank? || responses_type_valid?
         errors.add(:responses, 'is incorrect type')
+        raise ActiveModel::ValidationError, self
```
smoothcontract, line 62: Same comment applies here I think too - as this exception is potentially raised via the constructor, perhaps consider throwing the custom exception instead. Also no need to append to `errors`.

app/models/framework_response.rb
```diff
@@ -42,6 +44,12 @@ def update_with_flags!(value)
     raise ActiveRecord::ReadOnlyRecord, "Can't update framework_responses because person_escort_record is #{person_escort_record.status}"
   end
 
+  def value=(raw_value)
+    unless raw_value.blank? || value_type_valid?(raw_value)
+      raise FrameworkResponse::ValueTypeError, raw_value
```
smoothcontract, line 49: Yay 👍

smoothcontract (approved):
The behaviour is really clear with the custom exception type now, thank you for adding that. Nice job 👍

PR #874, opened 2020-08-25
[P4-1975] Generic event model

willfish:
### Jira link

P4-1975

### What?

I have added/removed/altered:

- [x] Added generic_events table
- [x] Added GenericEvent model

### Why?

We've agreed that committing to using the preexisting implementation/tables is creating a bunch of special cases which will make it harder to unpick version 1 (really 2) of the events implementation later. This is simpler, cleaner and requires fewer special cases.


app/models/events/move_cancel.rb
```diff
@@ -0,0 +1,24 @@
+class Events::MoveCancel < GenericEvent
+  attr_writer :cancellation_reason
+
+  EVENTABLE_TYPES = %w[Move].freeze
```
willfish, line 4: Plan is that each event can support one or more different types of eventable. We validate this and give feedback to the client.

app/models/events/move_cancel.rb
```diff
@@ -0,0 +1,24 @@
+class Events::MoveCancel < GenericEvent
```
cwrw, line 1: what do you think about the hierarchy:
```

class GenericEvent
  class MoveCancel < GenericEvent
```
and `generic_event/move_cancel.rb `
this way there is no confusion around previous events and the STI for generic events?

app/models/events/move_cancel.rb
```diff
@@ -0,0 +1,24 @@
...
+  attr_writer :cancellation_reason
+
+  EVENTABLE_TYPES = %w[Move].freeze
+
+  validates :cancellation_reason, inclusion: { in: Move::CANCELLATION_REASONS }
+  validates :eventable_type, inclusion: { in: EVENTABLE_TYPES }
```
cwrw, line 7: since we will be doing this across multiple types, does it make sense to move this to the parent class?

db/migrate/20200825080245_create_generic_events.rb
```diff
@@ -0,0 +1,18 @@
+class CreateGenericEvents < ActiveRecord::Migration[6.0]
+  def change
+    create_table :generic_events, id: :uuid do |t|
+      t.uuid :eventable_id, null: false     # polymorphic eventable/subject of the event
```
cwrw, line 4: are we always sure we will have an associated subjects per event, can an event not be associated to anything for more general events?

app/models/events/move_cancel.rb
cwrw: Our table currently can hold one type of eventable, how will that look when we add multiple eventables?

app/models/generic_event.rb
```diff
@@ -0,0 +1,18 @@
+class GenericEvent < ApplicationRecord
+  CREATED_BY_OPTIONS = %w[serco geoamey unknown].freeze
+
+  belongs_to :eventable, polymorphic: true, touch: true
```
cwrw, line 4: do we not need the reverse `has_many` on moves?

app/models/events/move_cancel.rb
willfish: Each event will define it's own list of valid eventable types

db/migrate/20200825080245_create_generic_events.rb
willfish: Not currently, no. It'll be about a Person if it's more generic.

app/models/generic_event.rb
willfish: Not currently needed. Probably will be needed in a later PR around running through triggers on each event. Laterbase

app/models/events/move_cancel.rb
willfish: Not got strong feelings. Does feel weird using a class as a namespace _and_ for inheritance, though.

app/models/events/move_cancel.rb
cwrw: confirmed there is only one per Event 👍

app/models/events/move_cancel.rb
cwrw: I did think so too, but found this pattern common with STIs, especially if you already defined `events` as a module somewhere else (I used it in framework response as I had a services module in that name). Not a blocker, however if we go with the current structure of `events` maybe rename to `generic_events` to avoid confusion with the old model.

app/models/events/move_cancel.rb
cwrw: sorry meant the validation itself, if this validation is duplicated across event types, might be better to move to the parent

cwrw (approved):
some minor comments otherwise 👍

PR #878, opened 2020-08-26
Making supplier_id not optional.

alexdesi:
### Jira link

https://dsdmoj.atlassian.net/browse/P4-2101

### What?
Make the relation **Move - Supplier** not optional.

### Why?
When a move is created or updated, a supplier should always be set.

### DIFF
--- a/app/models/move.rb
+++ b/app/models/move.rb
@@ -128,7 +128,7 @@ class Move < VersionedModel
   def rebook
     return rebooked if rebooked.present?
 
-    new_move = self.class.new(
+    Move.create!(
       original_move_id: id,
       from_location_id: from_location_id,
       to_location_id: to_location_id,
@@ -138,10 +138,8 @@ class Move < VersionedModel
       date: date && date + 7.days,
       date_from: date_from && date_from + 7.days,
       date_to: date_to && date_to + 7.days,
+      supplier: supplier
     )
-    new_move.determine_supplier
-    new_move.save!
-    new_move
   end
 
   def self.unfilled?
--- a/app/models/move.rb
+++ b/app/models/move.rb
@@ -70,7 +70,7 @@ class Move < VersionedModel
     REJECTION_REASON_NO_TRANSPORT = 'no_transport_available',
   ].freeze
 
-  belongs_to :supplier, optional: true
+  belongs_to :supplier
   belongs_to :from_location, class_name: 'Location'
   belongs_to :to_location, class_name: 'Location', optional: true
   belongs_to :profile, optional: true, touch: true
@@ -126,7 +126,9 @@ class Move < VersionedModel
   end
 
   def rebook
-    rebooked || self.class.create(
+    return rebooked if rebooked.present?
+
+    new_booking = self.class.new(
       original_move_id: id,
       from_location_id: from_location_id,
       to_location_id: to_location_id,
@@ -134,9 +136,12 @@ class Move < VersionedModel
       profile_id: profile_id,
       status: MOVE_STATUS_PROPOSED,
       date: date && date + 7.days,
-      date_from: date_from && date_from + 7.days,
+      date_from: (date_from || date) + 7.days,
       date_to: date_to && date_to + 7.days,
     )
+    new_booking.determine_supplier
+    new_booking.save!
+    new_booking
   end
 
   def self.unfilled?
--- a/app/models/move.rb
+++ b/app/models/move.rb
@@ -138,7 +138,7 @@ class Move < VersionedModel
       date: date && date + 7.days,
       date_from: date_from && date_from + 7.days,
       date_to: date_to && date_to + 7.days,
-      supplier: supplier
+      supplier: supplier,
     )
   end
 
--- a/app/models/move.rb
+++ b/app/models/move.rb
@@ -128,7 +128,7 @@ class Move < VersionedModel
   def rebook
     return rebooked if rebooked.present?
 
-    new_booking = self.class.new(
+    new_move = self.class.new(
       original_move_id: id,
       from_location_id: from_location_id,
       to_location_id: to_location_id,
@@ -136,12 +136,12 @@ class Move < VersionedModel
       profile_id: profile_id,
       status: MOVE_STATUS_PROPOSED,
       date: date && date + 7.days,
-      date_from: (date_from || date) + 7.days,
+      date_from: date_from && date_from + 7.days,
       date_to: date_to && date_to + 7.days,
     )
-    new_booking.determine_supplier
-    new_booking.save!
-    new_booking
+    new_move.determine_supplier
+    new_move.save!
+    new_move
   end
 
   def self.unfilled?
--- a/spec/factories/move.rb
+++ b/spec/factories/move.rb
@@ -12,9 +12,8 @@ FactoryBot.define do
     sequence(:created_at) { |n| Time.now - n.minutes }
     sequence(:date_from) { |n| Date.today - n.days }
 
-    trait :with_supplier do
-      association(:supplier)
-    end
+    association(:supplier)
+
     # Move types
     trait :court_appearance do
       # NB: Police / Prison / STC / SCH --> Court
@@ -169,6 +168,8 @@ FactoryBot.define do
     association(:profile)
     association(:from_location, :prison, factory: :location)
     association(:to_location, :court, factory: :location)
+    association(:supplier)
+
     date { Date.today }
     time_due { Time.now }
     status { 'requested' }
--- a/spec/jobs/prepare_move_notifications_job_spec.rb
+++ b/spec/jobs/prepare_move_notifications_job_spec.rb
@@ -8,7 +8,7 @@ RSpec.describe PrepareMoveNotificationsJob, type: :job do
   let(:subscription) { create :subscription }
   let(:supplier) { create :supplier, name: 'test', subscriptions: [subscription] }
   let(:location) { create :location, suppliers: [supplier] }
-  let(:move) { create :move, from_location: location }
+  let(:move) { create :move, from_location: location, supplier: supplier }
   let(:action_name) { 'create' }
   let(:send_webhooks) { true }
   let(:send_emails) { true }
@@ -17,6 +17,7 @@ RSpec.describe PrepareMoveNotificationsJob, type: :job do
   before do
     create(:notification_type, :webhook)
     create(:notification_type, :email)
+
     allow(NotifyWebhookJob).to receive(:perform_later)
     allow(NotifyEmailJob).to receive(:perform_later)
   end
@@ -97,7 +98,7 @@ RSpec.describe PrepareMoveNotificationsJob, type: :job do
   end
 
   context 'when creating a back-dated move' do
-    let(:move) { create :move, from_location: location, date: 2.days.ago }
+    let(:move) { create :move, from_location: location, date: 2.days.ago, supplier: supplier }
 
     it_behaves_like 'it creates a webhook notification record'
     it_behaves_like 'it does not create an email notification record'
@@ -106,7 +107,7 @@ RSpec.describe PrepareMoveNotificationsJob, type: :job do
   end
 
   context 'when creating a move for today' do
-    let(:move) { create :move, from_location: location, date: Time.zone.today }
+    let(:move) { create :move, from_location: location, date: Time.zone.today, supplier: supplier }
 
     it_behaves_like 'it creates a webhook notification record'
     it_behaves_like 'it creates an email notification record'
@@ -115,7 +116,7 @@ RSpec.describe PrepareMoveNotificationsJob, type: :job do
   end
 
   context 'when creating a forward-dated move' do
-    let(:move) { create :move, from_location: location, date: 2.days.from_now }
+    let(:move) { create :move, from_location: location, date: 2.days.from_now, supplier: supplier }
 
     it_behaves_like 'it creates a webhook notification record'
     it_behaves_like 'it creates an email notification record'
--- a/spec/models/move_spec.rb
+++ b/spec/models/move_spec.rb
@@ -446,9 +446,12 @@ RSpec.describe Move do
         expect(original_move.rebook.date_from).to eq(original_move.date_from + 7.days)
       end
 
-      it 'sets the move from_date to 7 days after date if from_date is NOT present' do
+      it 'raises an error is date_from is nil' do
         original_move.date_from = nil
-        expect(original_move.rebook.date_from).to eq(original_move.date + 7.days)
+
+        expect {
+          original_move.rebook.date_from
+        }.to raise_exception ActiveRecord::RecordInvalid
       end
 
       it 'sets the move to date to 7 days in the future if present' do
--- a/spec/models/move_spec.rb
+++ b/spec/models/move_spec.rb
@@ -446,7 +446,7 @@ RSpec.describe Move do
         expect(original_move.rebook.date_from).to eq(original_move.date_from + 7.days)
       end
 
-      it 'raises an error is date_from is nil' do
+      it 'raises an error if date_from is nil' do
         original_move.date_from = nil
 
         expect {
--- a/spec/models/move_spec.rb
+++ b/spec/models/move_spec.rb
@@ -3,7 +3,7 @@
 require 'rails_helper'
 
 RSpec.describe Move do
-  it { is_expected.to belong_to(:supplier).optional }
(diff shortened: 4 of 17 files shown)


app/models/move.rb
```diff
@@ -126,17 +126,22 @@ def rebooked
   end
 
   def rebook
-    rebooked || self.class.create(
+    return rebooked if rebooked.present?
+
+    new_booking = self.class.new(
```
willfish, line 131:
```suggestion
    move = self.class.new(
```

app/models/move.rb
```diff
@@ -126,17 +126,22 @@ def rebooked
...
       to_location_id: to_location_id,
       allocation_id: allocation_id,
       profile_id: profile_id,
       status: MOVE_STATUS_PROPOSED,
       date: date && date + 7.days,
-      date_from: date_from && date_from + 7.days,
+      date_from: (date_from || date) + 7.days,
       date_to: date_to && date_to + 7.days,
     )
+    new_booking.determine_supplier
```
willfish, line 142:
```suggestion
    new_move.determine_supplier
```

app/models/move.rb
```diff
@@ -126,17 +126,22 @@ def rebooked
...
+    new_booking.save!
```
willfish, line 143:
```suggestion
    rebooked_move.save!
```

app/models/move.rb
```diff
@@ -126,17 +126,22 @@ def rebooked
...
+    new_booking
```
willfish, line 144:
```suggestion
    new_move
```

app/services/event_log/move_runner.rb
```diff
@@ -16,7 +16,6 @@ def call
         when Event::APPROVE
           move.status = Move::MOVE_STATUS_REQUESTED
           move.date = event.date
-          move.determine_supplier
```
willfish, line 19: Not sure I understand why we need to stop determining the supplier again if the effective date comparison might yield a different result in the supplier chooser. @smoothcontract I don't suppose you know why this has changed, do you?

app/models/move.rb
```diff
@@ -126,17 +126,22 @@ def rebooked
...
+    new_booking = self.class.new(
       original_move_id: id,
       from_location_id: from_location_id,
       to_location_id: to_location_id,
       allocation_id: allocation_id,
       profile_id: profile_id,
       status: MOVE_STATUS_PROPOSED,
       date: date && date + 7.days,
-      date_from: date_from && date_from + 7.days,
+      date_from: (date_from || date) + 7.days,
```
willfish, line 139:
```suggestion
      date_from: (date || date_from) + 7.days,
```

We should use the move date first because that's the confirmed date

martyn-w: are you certain that changing the behaviour of rebooking a move to be based on `date` rather than `date_from` is sensible? I think that might be wiser to tackle that in its own issue rather than as part of this.

app/models/move.rb
martyn-w: I think the `determine_supplier` method is about to disappear, but check with Andrew

martyn-w:
I'm a bit wary this is tagged with `breaking_change` - what is it likely to break? and what should we tell supplies to mitigate any problems?

app/models/move.rb
willfish: I think we should revert to the original, yeah.

app/services/event_log/move_runner.rb
alexdesi, line 19: now that we are live, we determine the supplier on move-creation and then it never changes on move-update

spec/jobs/prepare_move_notifications_job_spec.rb
```diff
@@ -8,7 +8,7 @@
   let(:subscription) { create :subscription }
   let(:supplier) { create :supplier, name: 'test', subscriptions: [subscription] }
   let(:location) { create :location, suppliers: [supplier] }
-  let(:move) { create :move, from_location: location }
+  let(:move) { create :move, from_location: location, supplier: supplier }
```
willfish, line 11: We can checkout the master version of this file and the tests will still pass :)

spec/models/move_spec.rb
```diff
@@ -442,9 +446,12 @@
         expect(original_move.rebook.date_from).to eq(original_move.date_from + 7.days)
       end
 
-      it 'sets the move date from to nil if not present' do
+      it 'raises an error is date_from is nil' do
```
willfish, line 449:
```suggestion
      it 'raises an error if date_from is nil' do
```

willfish (approved):

willfish:
Let's hold off until @smoothcontract has finished his bits with `Move#determine_supplier`

app/models/move.rb
```diff
@@ -137,6 +139,9 @@ def rebook
       date_from: date_from && date_from + 7.days,
       date_to: date_to && date_to + 7.days,
     )
+    new_move.determine_supplier
```
smoothcontract, line 142: You can remove this line now - isn't needed and the method will be removed. See https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/905

app/services/event_log/move_runner.rb
smoothcontract: See https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/905 - explained there

spec/factories/move.rb
```diff
@@ -155,6 +154,8 @@
     association(:profile)
     association(:from_location, :prison, factory: :location)
     association(:to_location, :court, factory: :location)
+    association(:supplier)
```
smoothcontract, line 157: Isn't this duplicated from line 15? Looks like every one of these factories now includes a supplier (which also might be kind of expensive and affect the test suite performance).

spec/requests/api/moves_controller_update_spec.rb
```diff
@@ -400,7 +400,7 @@
           end
 
           context 'when the supplier has a webhook subscription', :skip_before do
-            let!(:subscription) { create(:subscription, :no_email_address, supplier: supplier) }
+            let!(:subscription) { create(:subscription, :no_email_address, supplier: move.supplier) }
```
smoothcontract, line 403: Presumably this spec file defines a `supplier` let somewhere - better instead of this change to use that supplier when creating the move instead; otherwise this ends up creating two suppliers which is slower.

smoothcontract: Ditto other amended specs in this file.

spec/requests/api_stories/move_workflows/1_police_to_police/1_1_person_with_police_national_computer_spec.rb
```diff
@@ -150,6 +150,9 @@
     police1
     police2
 
+    # Assign police1 location to Serco supplier
+    create :supplier_location, location: police1, supplier: serco_supplier, effective_from: Date.new(1900, 1, 1)
```
smoothcontract, line 154: It's cleaner to just leave `effective_from` as `nil`. The behaviour of all the supplier chooser and supplier location models is well covered elsewhere and it's perfectly fine to create a supplier location with nil effective dates - makes it much easier to test.

spec/requests/api_stories/move_workflows/2_police_to_unspecified/2_1_person_with_police_national_computer_to_unspecified_prison_spec.rb
```diff
@@ -168,6 +168,9 @@
     police
     prison
 
+    # Assign the police location to the supplier
+    create :supplier_location, location: police, supplier: serco_supplier, effective_from: Date.new(1900, 1, 1)
```
smoothcontract, line 172: As above

spec/requests/api_stories/move_workflows/5_court_to_prison/5_1_person_with_nomis_prison_number_spec.rb
```diff
@@ -143,6 +143,9 @@
     header 'Content-Type', 'application/vnd.api+json'
     header 'Accept', 'application/vnd.api+json; version=2'
     header 'Authorization', 'Bearer spoofed-token'
+
+    # Assign location court to Serco supplier
+    create :supplier_location, location: court, supplier: serco_supplier, effective_from: Date.new(1900, 1, 1)
```
smoothcontract, line 148: and here

spec/requests/api_stories/move_workflows/6_prison_to_prison_ipt/6_1_singleton_known_person_spec.rb
```diff
@@ -164,6 +164,9 @@
     header 'Accept', 'application/vnd.api+json; version=2'
     header 'Authorization', 'Bearer spoofed-token'
 
+    # Assign prison1 location to Serco supplier
+    create :supplier_location, location: prison1, supplier: serco_supplier, effective_from: Date.new(1900, 1, 1)
```
smoothcontract, line 168: and here

spec/services/moves/updater_spec.rb
```diff
@@ -31,6 +31,8 @@
     }
   end
 
+  before { create :supplier_location, supplier: move.supplier, location: move.from_location } # Used by SupplierChooser
```
smoothcontract, line 34: This will soon be redundant as supplier chooser is not called when updating a move. Maybe add a TODO here as a reminder to clean this up if this gets merged before the other PR.

spec/support/with_nomis_alerts_seed_data.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 RSpec.shared_context 'with Nomis alerts reference data' do
+  let(:serco_supplier) { create :supplier, key: 'serco', name: 'Serco' }
```
smoothcontract, line 4: Doesn't look like this gets called anywhere so is probably redundant.

smoothcontract (commented):
A few things that could be tidied/improved, but mainly I think it would be better to wait for https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/905 to be merged first, then revise - this should simplify quite a few scenarios.

app/models/move.rb
```diff
@@ -126,7 +126,9 @@ def rebooked
   end
 
   def rebook
-    rebooked || self.class.create(
+    return rebooked if rebooked.present?
+
+    new_move = self.class.new(
```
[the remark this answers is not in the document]
willfish: Don't need it any more since we're not going to be determining the supplier and are just going to hand the new  move the original move's supplier.

app/services/event_log/move_runner.rb
willfish: Okay, not sure what happens when you've both removed stuff.

willfish: We'll get yours merged and then we can resolve conflicts

spec/requests/api/moves_controller_update_v2_spec.rb
```diff
@@ -24,7 +24,10 @@
   end
 
   describe 'PATCH /moves' do
-    let!(:move) { create :move, :proposed, :prison_recall, from_location: from_location, profile: profile }
+    # <<<<<<< HEAD
```
willfish, line 27: I think this comment can be removed?

spec/requests/api/moves_controller_update_v2_spec.rb
```diff
@@ -24,7 +24,10 @@
...
+    let!(:move) { create :move, :proposed, :prison_recall, from_location: from_location, profile: profile, supplier: supplier }
+
+    # let!(:move) { create :move, :proposed, :prison_recall, from_location: from_location, profile: profile, date_from: 2.days.ago, supplier: supplier }
```
willfish, line 30: And this one?

spec/services/moves/updater_spec.rb
```diff
@@ -30,6 +31,8 @@
...
```
willfish, line 34: Remove me

spec/factories/move.rb
willfish: Doesn't use the other factory as a parent so I guess it's needed.

alexdesi: yes, I tried to remove it, but test didn't pass

spec/models/move_spec.rb
willfish, line 449:
```suggestion
      it 'raises an error if date_from is nil' do
```

spec/requests/api/moves_controller_update_spec.rb
alexdesi: specifying the supplier, the subscription factory does not create its own supplier, so it reduces the test of -1 supplier

spec/requests/api_stories/move_workflows/1_police_to_police/1_1_person_with_police_national_computer_spec.rb
alexdesi: 👍

spec/requests/api_stories/move_workflows/2_police_to_unspecified/2_1_person_with_police_national_computer_to_unspecified_prison_spec.rb
alexdesi: 👍

spec/requests/api/moves_controller_update_spec.rb
willfish: Fixed.

spec/services/moves/updater_spec.rb
alexdesi: good spot!

spec/support/with_nomis_alerts_seed_data.rb
willfish: We're using it in the api stories

alexdesi: this is called in 4 places.

app/models/move.rb
```diff
@@ -126,7 +126,9 @@ def rebooked
   end
 
   def rebook
-    rebooked || self.class.create(
+    return rebooked if rebooked.present?
+
+    Move.create!(
```
smoothcontract, line 131: just a minor point, but surprised rubocop hasn't flagged a preference for `self.class.create!`

spec/requests/api_stories/move_workflows/1_police_to_police/1_1_person_with_police_national_computer_spec.rb
```diff
@@ -150,6 +150,9 @@
     police1
     police2
 
+    # Assign police1 location to Serco supplier
+    create :supplier_location, location: police1
```
willfish, line 154:
```suggestion
    create :supplier_location, location: police1, supplier: serco_supplier
```

spec/factories/move.rb
smoothcontract: Okay, thanks for the explanation.

spec/requests/api_stories/move_workflows/1_police_to_police/1_1_person_with_police_national_computer_spec.rb
smoothcontract: Shouldn't this specify a supplier too, otherwise it will create a new supplier association. Either the comment is wrong or this should pass a supplier parameter I think - the examples in the other workflows look correct.

spec/support/with_nomis_alerts_seed_data.rb
smoothcontract: Sorry, I didn't spot it was a shared context - my mistake.

smoothcontract (approved):
Looks really good - and this got a lot simpler with the revised supplier chooser rules, so that's a bonus! Just a couple of very minor points if you want to address, but otherwise 👍

spec/requests/api_stories/move_workflows/1_police_to_police/1_1_person_with_police_national_computer_spec.rb
willfish: See https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/878/commits/3557320883b4e6f9e534335cb84f434dfb3bfa94

willfish (approved):

app/models/move.rb
smoothcontract: Feel free to ignore this comment if you like 😀

PR #880, opened 2020-08-26
Update notifications rake task to avoid assuming ownership of moves

smoothcontract:
### Jira link

P4-2038

### What?

- [x] Avoid use of removed `Location.supplier` scope in notification rake task

### Why?

- We can't determine move owner solely by location any more - this instead needs to consider the effective date (implemented in SupplierChooser service). However it's also a little beyond the scope of this rake task to be updating the supplier - we have another rake task to do that, so I've updated this to just report if there any moves found without a supplier assigned, but otherwise ignore those when sending out notifications.

### DIFF
--- a/lib/tasks/notifications.rake
+++ b/lib/tasks/notifications.rake
@@ -17,16 +17,13 @@ namespace :notifications do
       .where('date >= ?', from_date)
       .where(status: statuses)
 
-    # moves assumed to belong to the supplier (based on from_location)
-    moves_assumed = Move
+    # moves that don't belong to any supplier
+    moves_unassigned = Move
       .where(supplier_id: nil)
-      .where(from_location: Location.supplier(supplier))
-      .where('date >= ?', from_date)
-      .where(status: statuses)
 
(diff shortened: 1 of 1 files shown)


lib/tasks/notifications.rake
```diff
@@ -39,13 +36,10 @@ namespace :notifications do
 
     puts 'Processing assigned moves...'
     moves_assigned.find_each do |move|
-      PrepareMoveNotificationsJob.perform_now(topic_id: move.id, action_name: 'update', send_webhooks: send_webhooks, send_emails: send_emails, only_supplier_id: supplier.id)
+      puts move
+      # PrepareMoveNotificationsJob.perform_now(topic_id: move.id, action_name: 'update', send_webhooks: send_webhooks, send_emails: send_emails, only_supplier_id: supplier.id)
```
martyn-w, line 40: I don't think this should be commented out

smoothcontract: bollocks! oops, didn't mean to commit that!

martyn-w (approved):
👍 LGTM

PR #887, opened 2020-08-27
P4-2080 Support additional framework response types

cwrw:
### Jira link

[P4-2080](https://dsdmoj.atlassian.net/browse/P4-2080)

### What?

I have added/removed/altered:

- [x] Added new types for object and collection for framework responses
- [x] Moved the response type method to framework question and remove from serializers

### Why?

We currently support `collection::add_multiple_items` for a response that is a collection and of type multiple items. To
extend this pattern to other exact collections and objects, append `::followup_comment` to object and collection if it has followup comments. If another type of collection or object is added, we will add that type to those respectively. This will help the FE render the correct type of response.

This is a breaking change and has to be coordinated with the FE.

### Have you? (optional)

### DIFF
--- a/app/models/framework_question.rb
+++ b/app/models/framework_question.rb
@@ -38,9 +38,9 @@ class FrameworkQuestion < VersionedModel
   def response_type
     case question_type
     when 'radio'
-      followup_comment ? 'object::details' : 'string'
+      followup_comment ? 'object::followup_comment' : 'string'
     when 'checkbox'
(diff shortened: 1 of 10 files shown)


cwrw:
testing can be done on: https://booksecuremove-br-master.herokuapp.com/

smoothcontract (approved):
Looks good and is a nice cleanup - liking the move of `response_type` from serializers to the model. 👍

martyn-w (approved):
👍  LGTM

PR #888, opened 2020-08-27
Remove "Not for release" columns from CSV export

smoothcontract:
### Jira link

P4-2102

### What?

- [x] Remove "Not for release" columns from CSV export

### Why?

- Suppliers don't want this column as apparently it's redundant, and the information needed can be obtained from the "Not to be released" column instead.

- Resolves a production API issue


martyn-w (approved):
👍 LGTM

PR #889, opened 2020-08-27
Update CHANGELOG.md

martyn-w:
Changelog v2.3.6


cwrw (approved):
I've stopped committing the weird changes to the changelog as a workaround for now, as still not sure how those are happening

PR #890, opened 2020-08-27
Add new move cancellation reason

smoothcontract:
### Jira link

P4-2055

### What?

- [x] Add new move cancellation reason for "Cancelled by PMU"

### Why?

- We need to support this in the front end.

### Have you? (optional)

- Changes an api that is used in production (but is not a breaking change)


martyn-w (approved):
👍 LGTM

PR #891, opened 2020-08-27
P4 2109 ensure supplier created moves owned by supplier

martyn-w:
### Jira link

P4-2109

### What?

I have added/removed/altered:

- [x] changed SupplierChooser logic to consider doorkeeper_app_owner, existing owner and whether its a new record or not
- [x] updated tests

### Why?

I am doing this because:

- suppliers must be able to create moves for any location: even one which they do not ordinarily service
- the previous logic would prevent suppliers creating e.g. court to prison moves

### Have you? (optional)

- wide ranging and needs careful but urgent review

### DIFF
--- a/app/controllers/api/v2/moves_actions.rb
+++ b/app/controllers/api/v2/moves_actions.rb
@@ -13,18 +13,16 @@ module Api::V2
 
     def create_and_render
       move = Move.new(move_attributes)
-      move.determine_supplier
-      authorize!(:create, move)
-      move.save!
-
+      move.determine_supplier(doorkeeper_application_owner)
+      move.save! # NB: the move should be saved and notified even if the current user is no longer authorised to view it
       Notifier.prepare_notifications(topic: move, action_name: 'create')
-
+      authorize!(:create, move)
       render_move(move, :created)
     end
 
     def update_and_render
       move.assign_attributes(common_move_attributes)
-      move.determine_supplier
+      move.determine_supplier(doorkeeper_application_owner)
       action_name = move.status_changed? ? 'update_status' : 'update'
       move.save!
       move.allocation&.refresh_status_and_moves_count!
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -30,6 +30,10 @@ class Event < ApplicationRecord
     @supplier_id ||= details[:supplier_id]
   end
 
+  def supplier
+    Supplier.find_by(id: supplier_id) # NB: not a conventional table field, could be nil
+  end
+
   def event_params
     @event_params ||= details[:event_params]
   end
--- a/app/services/supplier_chooser.rb
+++ b/app/services/supplier_chooser.rb
@@ -1,16 +1,85 @@
 class SupplierChooser
-  attr_accessor :effective_date, :location
+  attr_accessor :effective_date, :location, :new_record, :doorkeeper_application_owner, :existing_owner
 
-  def initialize(effective_date, location)
-    self.effective_date = effective_date
-    self.location = location
+  def initialize(effective_date:, location:, new_record:, doorkeeper_application_owner: nil, existing_owner: nil)
+    @effective_date = effective_date
+    @location = location
+    @new_record = new_record
+    @doorkeeper_application_owner = doorkeeper_application_owner
+    @existing_owner = existing_owner
   end
 
-  def call
-    return unless effective_date.present? && location.present?
+  # CREATING NEW RECORDS
+  # | Doorkeeper App Owner | Location+Date Owner | Resulting Owner |
+  # |----------------------|---------------------|-----------------|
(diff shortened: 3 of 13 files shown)


app/services/supplier_chooser.rb
```diff
@@ -1,16 +1,86 @@
...
+  # | Supplier2           | nil            | Supplier1            | Supplier2       |
+  # | Supplier2           | nil            | Supplier2            | Supplier2       |
+  #
+  # | Supplier2           | Supplier1      | nil                  | Supplier2       |
+  # | Supplier2           | Supplier1      | Supplier1            | Supplier2       |
+  # | Supplier2           | Supplier1      | Supplier2            | Supplier2       |
+  #
+  # | Supplier2           | Supplier2      | nil                  | Supplier2       |
+  # | Supplier2           | Supplier2      | Supplier1            | Supplier2       |
+  # | Supplier2           | Supplier2      | Supplier2            | Supplier2       |
```
smoothcontract, line 64: @martyn-w it would be great if you could talk people through your thinking with this logic, I'm not fully following the approach. Eventually it would be good to capture this truth table in the form of specs too - appreciate that wasn't the focus whilst burning the midnight oil but maybe mark with a TODO for the future.

app/services/supplier_chooser.rb
```diff
@@ -1,16 +1,86 @@
...
+
   def call
+    if new_record
+      doorkeeper_application_owner || effective_supplier_for_location
+    else
+      effective_supplier_for_location || existing_owner || doorkeeper_application_owner
```
smoothcontract, line 70: I don't follow why this would need to consider `doorkeeper_application_owner` - as `existing_owner` should always be present? Could this be simplified to:

```ruby
effective_supplier_for_location || existing_owner
```

If so then that means we only need to worry about the doorkeeper token when creating the move, which happens in fewer places so there's less need to pass the doorkeeper token around which should simplify the PR.

smoothcontract (commented):
It would be good to get a walkthrough of this to understand some of the chosen logic. The approach seems very complicated and it's introduced a lot of additional wiring to pass doorkeeper through everywhere; I'm not convinced we need quite so much of that. Also it would be good to add coverage for the revised logic as it looks like there are a few scenarios that are modelled only in the truth table documentation rather than as specs, so we wouldn't know if we broke anything inadvertently.

I'm also thinking we could look at storing the original supplier id on the move when it gets created - and that would never change throughout the lifetime of the move. That also helps when we need to send notifications when the move changes - even if the current supplier/owner of the move changes we still know who the original owner was so can notify them too. That too should help simplify some of the logic here. Happy to talk through these ideas on a call.

app/controllers/api/v2/moves_actions.rb
```diff
@@ -13,18 +13,16 @@ def show_and_render
 
     def create_and_render
       move = Move.new(move_attributes)
-      move.determine_supplier
-      authorize!(:create, move)
-      move.save!
-
+      move.determine_supplier(doorkeeper_application_owner)
+      move.save! # NB: the move should be saved and notified even if the current user is no longer authorised to view it
```
smoothcontract, line 17: Hmm... I'm not sure this is the correct rule to apply. If we're now assuming on creation that the doorkeeper supplier token has priority then when a supplier creates the move we would always have a token. So this doesn't need to be changed from the current implementation.

My understanding is that moves created in the front end will always be mapped to a supplier (via SupplierLocation), and moves created by suppliers will always have a supplier API token. So the existing logic should work. Also there is another story in flight to ensure that a move have a supplier before being created - that I think is a better fix to ensuring data integrity. See: https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/878

The potential behaviour here of creating a move but returning a 403 means that clients are likely to retry the same operation repeatedly meaning we could end up with garbage data.

app/controllers/api/v2/moves_actions.rb
```diff
@@ -13,18 +13,16 @@ def show_and_render
...
       Notifier.prepare_notifications(topic: move, action_name: 'create')
-
+      authorize!(:create, move)
```
willfish, line 19: Seems very confusing for a consumer of this api to get a failure message and actually succeed in creating a move 😕 .

app/models/event.rb
```diff
@@ -30,6 +30,10 @@ def supplier_id
     @supplier_id ||= details[:supplier_id]
   end
 
+  def supplier
+    Supplier.find_by(id: supplier_id) # NB: not a conventional table field, could be nil
```
willfish, line 34: Could be nil == not for long

willfish, line 34:
```suggestion
    Supplier.find(supplier_id)
```

martyn-w:
Closing this PR as @smoothcontract is working on a new approach

PR #892, opened 2020-08-28
Quick fix for go live with Serco

smoothcontract:
### Jira link

P4-<TODO>

### What?

- [x] Assume that if we can't find a supplier location mapping for a move then it was created via the API by Serco, so assign them as the move owner.

### Why?

- Moves created by suppliers include locations that are not mapped in the supplier_location.yml file. We need to come up with a nice solution for this, potentially reverting to the previous approach based on the doorkeeper_application_owner. As a short term fix we are going live with just one supplier, Serco, so can make the assumption that any moves created without an appropriate SupplierLocation mapping will have been created by Serco.

- This removes the immediate blocker to going live and gives us time to work on a longer term fix to support further suppliers, and is also a simple change to understand/review.

### DIFF
--- a/app/services/supplier_chooser.rb
+++ b/app/services/supplier_chooser.rb
@@ -11,6 +11,12 @@ class SupplierChooser
 
     # NB: business rules _should_ preclude multiple suppliers for visibility
     supplier_location = SupplierLocation.location(location.id).effective_on(effective_date).first
-    supplier_location&.supplier
+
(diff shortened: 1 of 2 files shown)


martyn-w:
This will work for Serco, but will break for Geo. Whether that is acceptable or not is a question for big cheeses...

martyn-w (approved):
😬

PR #894, opened 2020-08-28
Disable Prometheus by default

cwrw:
To avoid `Prometheus Exporter, failed to send message Connection refused` in ENVs that don't have prometheus hooked up, make this optional according to ENV variable. This mirrors the setup in prometheus exporter initializer

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -35,13 +35,15 @@ plugin :tmp_restart
 
 preload_app!
 # run a new process instrumenter after work boot
-after_worker_boot do
-  require 'prometheus_exporter/instrumentation'
+if ENV['PROMETHEUS_METRICS'].in? ['on', 'true']
+  after_worker_boot do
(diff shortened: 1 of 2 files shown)


config/puma.rb
```diff
@@ -35,13 +35,15 @@
 
 preload_app!
 # run a new process instrumenter after work boot
-after_worker_boot do
-  require 'prometheus_exporter/instrumentation'
+if ENV['PROMETHEUS_METRICS'].in? ['on', 'true']
```
cwrw, line 38: same as: https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/config/initializers/prometheus_exporter.rb

mafonso (approved):

smoothcontract (approved):
LGTM 👍

martyn-w (approved):
👍 YES PLEASE 👍

PR #895, opened 2020-08-28
Change in? to include? on checking ENV

cwrw:
It seems Rails methods aren't being loaded on loading Puma config, use native ruby include? instead

### DIFF
--- a/config/initializers/prometheus_exporter.rb
+++ b/config/initializers/prometheus_exporter.rb
@@ -2,7 +2,7 @@
 
 # check for environment variable so that we can run production rake task in Dockerfile
 # otherwise we try to load the metrics outside of the actual app itself
-if ENV['PROMETHEUS_METRICS'].in? ['on', 'true']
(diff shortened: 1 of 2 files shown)


mafonso (approved):

config/initializers/prometheus_exporter.rb
```diff
@@ -2,7 +2,7 @@
 
 # check for environment variable so that we can run production rake task in Dockerfile
 # otherwise we try to load the metrics outside of the actual app itself
-if ENV['PROMETHEUS_METRICS'].in? ['on', 'true']
+if ['on', 'true'].include? ENV['PROMETHEUS_METRICS']
```
alexdesi, line 5: sure these downcase? :) 
does rubocop like this or it prefers the `w%[on true]`

alexdesi: rubocop seems happy 😄

PR #896, opened 2020-08-28
Add a new GET /moves filter for rejection reason

smoothcontract:
### Jira link

P4-2059

### What?

- [x] Add support for a new filter on rejection reason to `Moves::Finder` service; this is exposed on v1 and v2 `GET /moves` endpoint.

### Why?

- This supports front end requirements to get a list of moves that are not rejected (By specifying a value of `nil` for the new filter). Follows existing patterns for other filters, also improved spec coverage slightly and refactored filter specs to avoid creating lots of unnecessary fixtures (now slightly faster).

### Have you? (optional)

- Extends production API, but is backwards compatible. New filter is optional.

### DIFF
--- a/spec/requests/api/moves_controller_filter_spec.rb
+++ b/spec/requests/api/moves_controller_filter_spec.rb
@@ -32,8 +32,8 @@ RSpec.describe Api::MovesController do
 
     describe 'by supplier_id' do
       let(:filter_params) { { filter: { supplier_id: supplier.id } } }
-      let(:expected_moves) { create_list :move, 2, supplier: supplier }
-      let(:unexpected_moves) { create_list :move, 2, supplier: alternative_supplier }
+      let(:expected_moves) { create_list :move, 1, supplier: supplier }
+      let(:unexpected_moves) { create_list :move, 1, supplier: alternative_supplier }
 
       it_behaves_like 'an api that filters moves correctly'
     end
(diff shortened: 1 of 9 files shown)


spec/requests/api/moves_controller_filter_spec.rb
```diff
@@ -32,17 +32,17 @@
 
     describe 'by supplier_id' do
       let(:filter_params) { { filter: { supplier_id: supplier.id } } }
-      let(:expected_moves) { create_list :move, 2, supplier: supplier }
-      let(:unexpected_moves) { create_list :move, 2, supplier: alternative_supplier }
+      let(:expected_moves) { create_list :move, 1, supplier: supplier }
+      let(:unexpected_moves) { create_list :move, 1, supplier: alternative_supplier }
```
alexdesi, line 36: I like these kinds of changes, the only things is that it
sounds a bit weird using `create_list` with 1 element,
I suppose it is because it returns an array,
the alternative would be to wrap it in an array 🤔 
`[ create(:move, supplier: supplier ) ]`

Not very opinionated though

alexdesi (approved):
It looks fine to me.

spec/requests/api/moves_controller_filter_spec.rb
martyn-w: I guess creating a list with one element is twice as fast as creating it with two elements 🤔

martyn-w (approved):
👍 LGTM

spec/requests/api/moves_controller_filter_spec.rb
alexdesi: my "dilemma" was more about
`let(:expected_moves) { create_list :move, 1, supplier: supplier }`
versus the equivalent form:
`let(:expected_moves) { [ create(:move, supplier: supplier) ] }`  
but that's fine, not very important.

martyn-w: potatohs vs potatoes :-)

PR #897, opened 2020-08-28
Remove redundant `locations_suppliers` table

alexdesi:
### Jira link

https://dsdmoj.atlassian.net/browse/P4-2100

### What?
The table `locations_suppliers` has been replaced by `supplier_locations` (which now keep the relationship suppliers-locations within effective dates)

So `locations_suppliers` should be removed.

### Why?

`locations_suppliers` is not used anymore.


smoothcontract (approved):
LGTM 👍

PR #899, opened 2020-09-01
Guard against empty move dates when calculating age in Journey reports

cwrw:
Make sure a person, person DOB and move date exist before attempting to calculate person's age in journey report

### DIFF
--- a/lib/tasks/journeys_report.rake
+++ b/lib/tasks/journeys_report.rake
@@ -67,7 +67,7 @@ namespace :journeys do
           prison_id: person&.nomis_prison_number,
           gender: person&.gender&.key,
           dob: person&.date_of_birth,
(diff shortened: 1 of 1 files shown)


lib/tasks/journeys_report.rake
```diff
@@ -67,7 +67,7 @@ namespace :journeys do
           prison_id: person&.nomis_prison_number,
           gender: person&.gender&.key,
           dob: person&.date_of_birth,
-          age: person.present? && person.date_of_birth.present? ? ((move.date.to_date - person.date_of_birth&.to_date) / 365.25).to_i : nil,
+          age: age_from(person, move),
```
martyn-w, line 70: maybe `age_at` is a clearer name?

martyn-w (approved):
👍

PR #902, opened 2020-09-01
P4 2095 prioritise notifications queues

martyn-w:
### Jira link

P4-2095

### What?

I have added/removed/altered:

- [x] prioritising notifications based on move date
- [x] non-move queues default to medium priority

* Notifications about moves for today are processed in the high priority queue
* Notifications about moves for tomorrow are processed in the medium priority queue
* Notifications about moves for any other future/past date are processed in the low priority queue
* Notifications about other types (person/profile/per) are processed in the medium priority queue

### Why?

I am doing this because:

- to ensure bulk loads prioritise moves for today/tomorrow over back-dated or future-dated moves

### Have you? (optional)

- [x] ~TODO: review sidekiq queues and set priorities~

- ~Verify sidekiq queue settings in deploy project before merging~


db/schema.rb
```diff
@@ -274,15 +274,15 @@
     t.string "reference", null: false
     t.string "move_type"
     t.string "additional_information"
-    t.integer "nomis_event_id"
```
willfish, line 277: Weird but these haven't changed?

martyn-w: @willfish sorry, I shouldn't have committed that: just annoying jumping fields because they are ordered a bit differently in different local databases. Now fixed.

willfish:
Wondering whether this is extra complication that we don't need inside of the application? Could we just increase the number of sidekiq pods pulling jobs off of the queue?

db/schema.rb
willfish: No worries :)

martyn-w:
Hi @willfish see the ticket https://dsdmoj.atlassian.net/browse/P4-2095 for some of the thinking behind prioritisation.

There are two situations in particular: we will load a large volume of back-dated moves for Geoamey, and we want these to be processed with low priority so that they do not block urgent moves. And secondly, in the steady state as the system gets busier, moves for today most take precedence over moves for next week.

smoothcontract (approved):
Very clean implementation, good stuff 👍

PR #905, opened 2020-09-02
Revised business rules for determining move supplier

smoothcontract:
### Jira link

P4-2109

### What?

- [x] Only determine supplier for a move when the move is created. After it's been created, the supplier does not change.
- [x] If a supplier creates a move with an owned API key, then the move is always assigned to that supplier
- [x] If the move is created by the front end (no owner on the API key) then the move is assigned to the relevant supplier according to the move date and from_location (using the supplier_location.yaml mapping files)
- [x] Removed redundant `moves:update_suppliers` rake task

### Why?

- The new rules introduced just before go live did not provide the intended behaviour. The SupplierLocation (aka yaml file) mapping only supports a subset of locations. For court moves, suppliers create moves directly but the court locations are not present in the yaml file. For these we must instead assume the supplier creating the move is the owner of the move, irrespective of any other conditions.

- This PR largely reverts previous work to introduce supplier determination when the move is updated; instead the supplier for a move is only determined when the move is initially created. Once created the assigned supplier owner never changes. Also this reverts the previous reliance solely on the SupplierLocation mapping and reinstates checks for supplier API keys. If a supplier calls the API then the move is assigned to the supplier irrespective of SupplierLocation mapping. If the front end calls the supplier then SupplierLocation mapping is used to determine the correct supplier.

- Removed rake task was needed for go live, but is actually now dangerous if we were to run this again as it does not consider supplier API keys and will in fact revert any existing court moves to have a nil supplier (which we don't want!).


martyn-w:
👍 This looks good to me. Given the importance of this PR, it would be good to get reviews from other developers (cc @willfish @alexdesi @cwrw), focusing on the business logic:

* the supplier is assigned on creation
* the supplier doesn't change on update

martyn-w (approved):

willfish (approved):
Looks good to me.

willfish:
I think we may in the future need to allow the frontend to specify the supplier directly but this is totally fine until they've agreed/we've worked out what that looks like.

jcavell (approved):

jason-sims (approved):
👍 LGTM

jcavell:
Yes, the logic of setting the supplier only when creating, based on the supplier if Serco/Geo are calling the API, or inferring if it's the front-end is correct. As you've done, the supplier should not change when updating.

PR #906, opened 2020-09-02
P4 2114 improve webhook rake task

martyn-w:
### Jira link

P4-2114

### What?

I have added/removed/altered:

- [x] Added ability for webhook script to re-run and continue where it left off
- [x] Added random delay to improve stability and load

### Why?

I am doing this because:

- the initial version without a delay overwhelmed one of the pods a little

- minimal

### DIFF
--- a/lib/tasks/notifications.rake
+++ b/lib/tasks/notifications.rake
@@ -35,8 +35,6 @@ namespace :notifications do
       moves_assigned = moves_assigned.where("id NOT IN (#{already_sent_notifications.to_sql})")
     end
 
-    puts moves_assigned.to_sql
-
     # moves that don't belong to any supplier
     moves_unassigned = Move
       .where(supplier_id: nil)
--- a/lib/tasks/notifications.rake
+++ b/lib/tasks/notifications.rake
@@ -1,6 +1,6 @@
 namespace :notifications do
   desc 'Send webhook and email notifications to the specified supplier for future moves'
-  task :send, %i[supplier from_date send_webhooks send_emails] => :environment do |_, args|
+  task :send, %i[supplier from_date send_webhooks send_emails resend action_name] => :environment do |_, args|
     abort "Please specify a supplier, e.g. $ rake 'notifications:test_webhook[serco]'" if args[:supplier].blank?
     supplier = Supplier.find_by(key: args[:supplier]) || Supplier.find_by(id: args[:supplier]) || Supplier.find_by(name: args[:supplier])
     abort "Unknown supplier: #{args[:supplier]}" if supplier.blank?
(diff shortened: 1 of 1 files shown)


lib/tasks/notifications.rake
```diff
@@ -1,21 +1,35 @@
...
+                                         .where(notification_type: notification_types)
+                                         .select(:topic_id)
+                                         .to_sql
+
     # moves directly assigned to the supplier
     moves_assigned = Move
       .where(supplier: supplier)
       .where('date >= ?', from_date)
       .where(status: statuses)
+      .where(resend ? '' : "id NOT IN (#{already_sent_notifications_sql})")
```
smoothcontract, line 32: Not seen this technique used before `where('')` - surprised it works actually. Just a minor thing but maybe worth splitting out separately for readability. Also maybe it's simpler to just fetch a list of the move id's that have already been processed and pass that as a parameter to a more idiomatic Active Record chain. Something like this...

```ruby
if resend
  previously_sent_move_ids = Notification
                                         .where.not(delivered_at: nil)
                                         .where(topic_type: 'Move')
                                         .where(notification_type: notification_types)
                                         .select(:topic_id)
  moves_assigned = moves_assigned.where.not(id: previously_sent_move_ids)
end
```

smoothcontract (commented):
Looks good, just some minor improvement suggestions if you feel inclined.

lib/tasks/notifications.rake
```diff
@@ -1,21 +1,35 @@
 namespace :notifications do
   desc 'Send webhook and email notifications to the specified supplier for future moves'
-  task :send, %i[supplier from_date send_webhooks send_emails] => :environment do |_, args|
+  task :send, %i[supplier from_date send_webhooks send_emails resend action_name] => :environment do |_, args|
     abort "Please specify a supplier, e.g. $ rake 'notifications:test_webhook[serco]'" if args[:supplier].blank?
```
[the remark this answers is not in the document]
martyn-w: oops, good spot, thats a left over from its old name - will fix

lib/tasks/notifications.rake
martyn-w: so, I don't want to pass the `id`s directly to active record as there could be tens of thousands - the sql approach, although a bit clunky is definitely more efficient. However, I've re-written it along the lines you suggest to improve readability.

BTW the ActiveRecord trick:  `where(nil)`, `where({})` and `where("")` works and can be handy sometimes for conditional querying.

smoothcontract: Fair enough, better to avoid passing a huge list of ID's to Postgres and good to learn something new 😄

smoothcontract (approved):
Thanks for updating, it's much clearer now 👍

PR #908, opened 2020-09-03
[P4-2068] Separate auth and prison api urls and update prefix

willfish:
### Jira link

P4-2068

### What?

I have added/removed/altered:

- [x] Updates prison api client implementation to use a separate token url and prefix
- [x] Updates specs to reflect change of prefix

### Why?

I am doing this because:

- The prison api is now separate from the auth api

### DIFF
--- a/app/lib/nomis_client/base.rb
+++ b/app/lib/nomis_client/base.rb
@@ -25,7 +25,7 @@ module NomisClient
       REFRESH_TOKEN_TIMEFRAME_IN_SECONDS = 5
 
       def token_request(method, path, params)
-        token.send(method, "#{ENV['NOMIS_API_PATH_PREFIX']}#{path}", params)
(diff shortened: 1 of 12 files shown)


app/lib/nomis_client/base.rb
```diff
@@ -25,7 +25,7 @@ def put(path, params = {})
       REFRESH_TOKEN_TIMEFRAME_IN_SECONDS = 5
 
       def token_request(method, path, params)
-        token.send(method, "#{ENV['NOMIS_API_PATH_PREFIX']}#{path}", params)
+        token.send(method, "#{ENV['NOMIS_PRISON_API_PATH_PREFIX']}#{path}", params)
```
willfish, line 28: We've gone with completely new environment variables for prefixes and site urls so that we can very easily rollback our configuration and make sure devs don't suddenly find their ci/local development doesn't work anymore.

alexdesi (approved):
It looks good to me!

PR #910, opened 2020-09-03
Update .env.test to include new environment variables

willfish:
### Jira link

P4-<TODO>

### What?

I have added/removed/altered:

- [x] Updates .env.test values to match what wiremock expects and to stop calling nomis in ci

### Why?

I am doing this because:

- We use wiremock so we don't call nomis in test envs. We should match this behaviour like before.


martyn-w (approved):
👍

PR #914, opened 2020-09-07
[P4-1978] Journey events

willfish:
### Jira link

P4-1978

### What?

I have added/removed/altered:

- [x] Added Journey event STI classes and specs

### Why?

I am doing this because:

- We need these to dual write events to the v1/v2 event models
- Because we want to support these in the generic events api


app/models/generic_event/journey_cancel.rb
```diff
@@ -0,0 +1,11 @@
+class GenericEvent
+  class JourneyCancel < GenericEvent
+    EVENTABLE_TYPES = %w[Journey].freeze
+
+    validates :eventable_type, inclusion: { in: EVENTABLE_TYPES }
```
willfish, line 5: I've looked into drying these validations up and my options are:

1. Use `Tracepoint` to implement behaviour that will add validations via the parent after the `EVENTABLE_TYPES` constant has been defined in the child class (I don't like this - it's non-idiomatic/too complex/unusual and is worse than duplicating in my opinion)
2. Use another level of inheritance for the eventable type (e.g. one for JourneyEvent, PeopleEvent, MoveEvent) which adds eventable-specific validations. Inheritance hierarchies can be very hard to reason about and this implementation actually lacks the flexibility that we might want going forward. I genuinely think this is a can of worms that would end up opening us up to a pattern of very complex STI classes.
3. Duplicate the validation rules in the STI classes for each eventable. This smells a bit, gives us maximum flexibility and is really clear when you read the code.

I've opted for 3 for the reasons noted above. Happy to have a discussion about alternatives, here.

spec/models/generic_event_spec.rb
```diff
@@ -26,4 +26,33 @@
 
     expect { create(:event_move_cancel, eventable: eventable) }.to change { eventable.reload.updated_at }
   end
+
+  describe '#for_feed' do
```
willfish, line 30:
```suggestion
  describe '#trigger' do
```

app/models/generic_event/journey_cancel.rb
martyn-w: I'd personally go with (2) and introduce another layer: `JourneyXXXEvent < JourneyGenericEvent < GenericEvent`, with eventable_type validations at the JourneyGenericEvent level. However, I appreciate hierarchies of classes are not everyone's cup of tea.

Alternatively, how about including the validations as a concern?

app/models/generic_event.rb
```diff
@@ -1,7 +1,29 @@
 class GenericEvent < ApplicationRecord
   CREATED_BY_OPTIONS = %w[serco geoamey unknown].freeze
```
martyn-w, line 2: is it necessary to hard-code the supplier's names? could we use supplier_id instead?

willfish: Was just what we agreed over calls. I think this is an iteration on the final thing, tbh. Happy to talk about it

willfish: We could use the supplier_id, yeah. I'd be happy to do that and if the hub want the nomis agency id, etc, in the feed that can have it.

app/models/generic_event/journey_cancel.rb
willfish: I'd be happy to use concerns to dry it up. That way we don't get the pitfalls/inflexibility of inheritance.

app/models/generic_event.rb
willfish: Speaking with Martyn, we're going to change this to be more consistent with our approach elsewhere. I'll flag up in the reporting doc that we're changing this to be consistent and talk to @jcavell

willfish: In a separate PR, too

martyn-w: 👍

app/models/generic_event/journey_cancel.rb
martyn-w: 👍

martyn-w (approved):
👍 LGTM

app/models/generic_event/journey_cancel.rb
```diff
@@ -0,0 +1,9 @@
+class GenericEvent
+  class JourneyCancel < GenericEvent
+    include JourneyEventValidations
+
+    def trigger
+      eventable.cancel
+    end
+  end
+end
```
alexdesi, line 9:
```
class GenericEvent
  class subclass < GenericEvent
```
this pattern drives me crazy :D
My brain goes into loop,
do you have a article that explain it?

alexdesi: what's happen if we just create something like
`JourneyCancel < GenericEvent`
but not as a subclass of GenericEvent?
We cound add a module too, to add a scope

PR #921, opened 2020-09-10
Add new endpoint for bulk update of PER framework responses

smoothcontract:
### Jira link

P4-1799

### What?

- [x] Added a new endpoint to `PATCH /person_transfer_records/:id/framework_responses` that accepts an array of framework responses to update in bulk.
- [x] Added a new `FrameworkResponses::BulkUpdater` service to wrap the implementation details of the new endpoint
- [x] Refactored existing `FrameworkResponse` method to allow updates to multiple instances at once

### Why?

- Suppliers cannot use the existing endpoint to `PATCH /framework_responses/:id` as they will be using mobile devices with potentially patchy communication. It's more convenient and performant for them to submit the entire set of framework responses for a PER in a single request, rather than send ~60 individual requests for each response.

- JSONAPI typically only supports updating a singular resource, but there is a draft specification for a bulk update extension which this follows. Essentially this passes an array of resources rather than just one, but each resource has the usual structure (see request specs for details).

- Implementation makes extensive use of the `activerecord-import` gem; this is able to provide bulk data upserts in a single query which means that 60 framework response values can be updated in a single hit. The code is optimised to avoid updating responses where the value doesn't change, which gives a best to worst case response time (depending how many of the 60 framework responses were modified) in the range 180-900ms. This should be acceptable and avoids the need to introduce further complexity with asynchronous updating.

### Have you? (optional)

- Introduces a new endpoint, so minimal risk as not yet used

### DIFF
--- a/app/models/framework_response.rb
+++ b/app/models/framework_response.rb
@@ -101,6 +101,8 @@ class FrameworkResponse < VersionedModel
     SQL
   end
 
+  private_class_method :update_dependent_responses!, :descendants_tree, :recursive_tree
+
 private
 
   def set_responded_value
--- a/app/models/framework_response.rb
+++ b/app/models/framework_response.rb
@@ -5,6 +5,9 @@ class FrameworkResponse < VersionedModel
 
   validates :type, presence: true
   validates :responded, inclusion: { in: [true, false] }
+  validates_each :value, on: :update do |record, _attr, value|
+    record.errors.add(:value, :blank) if requires_value?(value, record)
+  end
 
   belongs_to :framework_question
   belongs_to :person_escort_record
@@ -13,27 +16,17 @@ class FrameworkResponse < VersionedModel
 
   belongs_to :parent, class_name: 'FrameworkResponse', optional: true
   has_and_belongs_to_many :framework_flags, autosave: true
-  validates_each :value, on: :update do |record, _attr, value|
-    record.errors.add(:value, :blank) if requires_value?(value, record)
-  end
 
   after_validation :set_responded_value, on: :update
 
-  def self.requires_value?(value, record)
-    return false if value.present? || !record.framework_question.required
-
-    return true if record.parent.blank?
-
-    record.parent.option_selected?(record.framework_question.dependent_value)
-  end
-
   def update_with_flags!(value)
     ActiveRecord::Base.transaction do
-      old_value = self.value
       self.value = value
+      return unless value_text_changed? || value_json_changed?
 
-      update!(framework_flags: build_flags)
-      clear_dependent_values_and_flags!(old_value)
+      rebuild_flags
+      save!
+      self.class.clear_dependent_values_and_flags!([self])
 
       # lock the status update to avoid race condition on multiple response patches
       person_escort_record.with_lock do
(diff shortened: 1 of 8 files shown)


app/models/framework_response.rb
```diff
@@ -44,12 +52,53 @@ def update_with_flags!(value)
...
+    clear_dependent_values_and_flags!
+  end
+
   def value=(raw_value)
     unless raw_value.blank? || value_type_valid?(raw_value)
       raise FrameworkResponse::ValueTypeError, raw_value
     end
   end
 
+  def self.bulk_update(person_escort_record, response_values_hash)
```
smoothcontract, line 66: This method is getting a bit bloaty, so considering moving this to a service instead. Thoughts?

app/models/framework_response.rb
```diff
@@ -13,27 +16,17 @@ class FrameworkResponse < VersionedModel
...
-    return true if record.parent.blank?
-
-    record.parent.option_selected?(record.framework_question.dependent_value)
-  end
-
   def update_with_flags!(value)
     ActiveRecord::Base.transaction do
-      old_value = self.value
       self.value = value
+      return unless value_text_changed? || value_json_changed?
```
cwrw, line 25: so I double checked why I needed the `old_value = self.value` instead of `value_text_changed? || value_json_changed?` and it seems that this is required for jsonb responses, as it always picks this up as changed although it hasn't so I think we will need to revert to the old implementation of using `old_value`

app/services/framework_responses/bulk_updater.rb
```diff
@@ -0,0 +1,63 @@
...
+  private
+
+    def build_value_updates_list
+      [].tap do |updated_responses|
+        validator = ActiveRecord::Import::Validator.new(FrameworkResponse)
+
+        FrameworkResponse.includes(framework_question: %i[framework_flags]).where(person_escort_record: person_escort_record).find(response_values_hash.keys).each do |response|
+          response.value = response_values_hash[response.id]
+          if validator.valid_model?(response)
+            updated_responses << response if response.changed?
```
cwrw, line 34: same thing here re `changed?`, this is picking up any `jsonb` columns as changed although they haven't we'll need to use old/new value instead to make sure we skip responses that haven't changed

app/services/framework_responses/bulk_updater.rb
```diff
@@ -0,0 +1,63 @@
...
+            errors[response.id] = response.errors.full_messages.first
+          end
+        rescue FrameworkResponse::ValueTypeError => e
+          errors[response.id] = "Value: #{e.message} is incorrect type"
+        end
+      end
+    end
+
+    def apply_bulk_response_changes(updated_responses)
+      # Bulk update all modified response values
```
cwrw, line 45: We should return early here is `updated_responses` is empty to avoid doing any updates on unchanged responses otherwise

app/models/framework_response.rb
```diff
@@ -50,31 +43,27 @@ def value=(raw_value)
...
+    return true if record.parent.blank?
 
-  def clear_dependent_values_and_flags!(old_value)
-    return unless old_value != value
+    record.parent.option_selected?(record.framework_question.dependent_value)
+  end
 
-    dependent_ids = dependents.includes(:framework_question).reject { |dependent| option_selected?(dependent.framework_question.dependent_value) }
+  def self.clear_dependent_values_and_flags!(responses_to_update)
+    all_dependent_ids = responses_to_update.map do |r|
```
cwrw, line 59: we can change this to:
```
all_dependent_ids = FrameworkResponse.where(id: responses_to_update.map(&:id)).includes(dependents: :framework_question).map do |response|
      response.dependents.reject { |dependent| response.option_selected?(dependent.framework_question.dependent_value) }
    end
```
This will avoid the n+1 queries

app/services/framework_responses/bulk_updater.rb
```diff
@@ -0,0 +1,63 @@
...
+      FrameworkResponse.import(updated_responses, validate: false, on_duplicate_key_update: { conflict_target: [:id], columns: %i[value_text value_json responded] })
+
+      # Update associated flags for all modified response values
+      updated_flags = updated_responses.each(&:rebuild_flags)
+      FrameworkResponse.import(updated_flags, validate: false, recursive: true, all_or_none: true, on_duplicate_key_update: { conflict_target: [:id] })
```
cwrw, line 50: We don't need the import here as `rebuild_flags` will be setting the flags directly as we're using `=` to set an association. Maybe it should be renamed to `rebuild_flags!`

cwrw (commented):
A few comments otherwise looking good

app/models/framework_response.rb
smoothcontract: Nice 👍

app/services/framework_responses/bulk_updater.rb
smoothcontract: Good point, also the PER update can be short circuited too (as nothing has changed), so I'll check for this scenario in the `call` method.

cwrw (approved):
👏

PR #923, opened 2020-09-11
Disable dual writing temporarily

willfish:
### Jira link

P4-<TODO>

### What?

I'm adding a reference from event -> generic_event in my rake task that enables use to avoid creating duplicate generic_events/means we can run the rake task over and over and get the same effect.

Dual writing events currently breaks this contract. In this PR I'm disabling the dual write functionality and will deploy/clear out generic events in all environments in a console. We then need to merge the rake task/migration which enables the reference and then switch on dual writing events.

I have added/removed/altered:

- [x] Disables dual writing events in journey eventable module.

### Why?

I am doing this because:

- We need to make sure we can safely rollback copied events/avoid duplicates via a reference. Currently written generic events are missing this reference and need removing.

### DIFF
--- a/app/controllers/concerns/journeys/eventable.rb
+++ b/app/controllers/concerns/journeys/eventable.rb
@@ -27,7 +27,7 @@ module Journeys
     end
 
     def create_generic_event(event)
-      GenericEvent.from_event(event).save!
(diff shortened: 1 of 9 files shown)


alexdesi (approved):
🌔 🚶

PR #926, opened 2020-09-11
Remove any repeat nested includes to avoid circular loading

cwrw:
Having a nested include is causing some queries to be very slow,
due to the circular association of the dependency. This now loads quite
quickly for the filter `ready_in_transit`.

Keeping the other includes loads the moves generally quicker, but
might be slightly slower with filters.


martyn-w (approved):
👍 LGTM

willfish (approved):

PR #930, opened 2020-09-14
Add some console log to event copier

alexdesi:
### What?
This PR add some output to the event copier

### Why?

To have a clue on how the process is progressing (if stuck, if progressing and how fast it is)

### DIFF
--- a/app/services/event_copier.rb
+++ b/app/services/event_copier.rb
@@ -1,3 +1,4 @@
+# rubocop:disable Rails/Output
 class EventCopier
   def initialize(dry_run:)
     @dry_run = dry_run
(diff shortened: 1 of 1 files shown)


app/services/event_copier.rb
```diff
@@ -12,7 +13,14 @@ def initialize(dry_run:)
   end
 
   def call
-    Event.not_copied.find_each do |event|
+    event_not_copied = Event.not_copied
+    event_not_copied_count = not_copied.count
```
martyn-w, line 17: should this be `event_not_copied.count` not `not_copied.count` ?

martyn-w (approved):
👍 LGTM

PR #931, opened 2020-09-14
P4-2165 Add some new prometheus counters

martyn-w:
### Jira link

P4-2165

### What?

I have added/removed/altered:

- [x] Adds a range of move counters to Prometheus
- [x] Adds a range of PER counters to Prometheus

### Why?

I am doing this because:

- so that move and PER counters can be consumed in dashboards

### DIFF
--- a/lib/prometheus/move_collector.rb
+++ b/lib/prometheus/move_collector.rb
@@ -47,7 +47,7 @@ private
       moves = moves.where(date: Time.zone.now + date_from_offset.days)
     else
       moves = moves.where('date >= ?', Time.zone.now + date_from_offset.days) if date_from_offset.present?
-      moves = moves.where('date =< ?', Time.zone.now + date_to_offset.days) if date_to_offset.present?
+      moves = moves.where('date <= ?', Time.zone.now + date_to_offset.days) if date_to_offset.present?
     end
     metric.observe(moves.count, status: status, supplier: supplier, date_from_offset: date_from_offset, date_to_offset: date_to_offset)
   end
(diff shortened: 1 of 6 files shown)


lib/prometheus/move_collector.rb
```diff
@@ -10,8 +10,46 @@ def type
...
+    observe_moves_count(metric, status: status, supplier: supplier, date_from_offset: -1, date_to_offset: -1) # yesterday
+    observe_moves_count(metric, status: status, supplier: supplier, date_from_offset: 0, date_to_offset: 0) # today
+    observe_moves_count(metric, status: status, supplier: supplier, date_from_offset: 1, date_to_offset: 1) # tomorrow
+    observe_moves_count(metric, status: status, supplier: supplier, date_from_offset: 0, date_to_offset: 6) # the next 7 days
+    observe_moves_count(metric, status: status, supplier: supplier, date_from_offset: 0, date_to_offset: 29) # the next 30 days
+    observe_moves_count(metric, status: status, supplier: supplier, date_from_offset: 0, date_to_offset: nil) # the future
+  end
+
+  def observe_moves_count(metric, status: nil, supplier: nil, date_from_offset: nil, date_to_offset: nil)
+    # TODO: investigate how to move these queries to a read-only replica database
```
cwrw, line 43: This is interesting: https://guides.rubyonrails.org/active_record_multiple_databases.html#using-manual-connection-switching

martyn-w: thanks, very interesting reading 👍

We do have read-replicas set up in most environments (not UAT), but more work will be required to make them accessible in the environments. And also need to see how well Rails behaves with multiple databases. I'll do that as a separate PR in due course.

smoothcontract (approved):
LGTM 👍

PR #933, opened 2020-09-15
Explicitly define the FrameworkResponse class in class methods

cwrw:
Since we are utilising STI in framework responses, missing the class name is causing issues when loading certain methods, as they are defined differently per response type. For instance the parent response type can be different to dependent responses, however when attempting to retrieve descendants, the parent class is being loaded, which means we are not looking at ALL framework responses.


teneightfive (approved):

smoothcontract (approved):
Pesky STI! Great catch 👍

PR #935, opened 2020-09-16
Add comment to explain enum usage for unstarted and fix test

cwrw:
* Due to enum value name restrictions around prepending `not_` unstarted is used as the main enum name, but the value `not_started` is surfaced through the API.

* Fix test to use values from the set options.


martyn-w (approved):
👍 LGTM

PR #937, opened 2020-09-17
Use correct key in setting expiry for keys in cache

cwrw:
The `expires_in` setting will set the expiration time on the cache. It accepts seconds. 
https://guides.rubyonrails.org/caching_with_rails.html#activesupport-cache-store

`expiry` does not set the `expires_in` value:
```
irb(main):019:0> Rails.cache.write('test', 'test', expiry: 3)
=> "OK"
irb(main):021:0> Rails.cache.send(:read_entry, 'test')
=> #<ActiveSupport::Cache::Entry:0x000055708b71e9a8 @value="test", @version=nil, @created_at=1600324732.0032308, @expires_in=nil>
```
`expires_in` sets the `expires_in` value:
```
irb(main):023:0> Rails.cache.write('test', 'test', expires_in: 300)
=> "OK"
irb(main):024:0> Rails.cache.send(:read_entry, 'test')
=> #<ActiveSupport::Cache::Entry:0x000055708a3475d8 @value="test", @version=nil, @created_at=1600324902.830221, @expires_in=300.0>
```

We now have around ~40k conflict_keys and cache_response_keys in production. After this is deployed, we will need to cleanup the keys. To make sure we do not delete any keys being used, we can delete the keys after 12 hours, the expiry time of conflict keys.
```
Rails.cache.keys.each do |key|
  Rails.cache.delete(key) if Rails.cache.send(:read_entry, key).expires_at.nil? && (key =~ /conf/ || key =~ /resp/)
end
```

### DIFF
--- a/app/services/idempotency/store.rb
+++ b/app/services/idempotency/store.rb
@@ -36,10 +36,10 @@ module Idempotency
       return if @idempotency_key.blank?
 
       # cache the response for a short time
-      Rails.cache.write(cache_response_key, response, expiry: CACHE_RESPONSE_TTL)
+      Rails.cache.write(cache_response_key, response, expires_in: CACHE_RESPONSE_TTL)
 
       # set conflict flag for a longer time
-      Rails.cache.write(conflict_key, 1, expiry: CONFLICT_TTL)
+      Rails.cache.write(conflict_key, 1, expires_in: CONFLICT_TTL)
     end
   end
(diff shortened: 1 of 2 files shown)


martin-ballhatchet-moj (approved):
LGTM

martyn-w (approved):
👍 LGTM

PR #938, opened 2020-09-17
Add support for filtering moves by `location_id`

smoothcontract:
### Jira link

P4-2206

### What?

- [x] Added support for filtering moves (via `GET /moves` endpoint) using `location_id`

### Why?

- This returns moves matching on **either** `from_location_id` **or** `to_location_id` - which allows the list of transfers in and out of a specific location to be returned in a single API call. This is to support initial work in the front end for PMU dashboard screens.

### Have you? (optional)

- Extends existing API but is backwards compatible; behaviour is unchanged if new filter is not specified.

### DIFF
--- a/app/controllers/api/moves_controller.rb
+++ b/app/controllers/api/moves_controller.rb
@@ -35,7 +35,7 @@ module Api
     end
 
     PERMITTED_FILTER_PARAMS = %i[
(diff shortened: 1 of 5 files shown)


martyn-w (approved):
👍  LGTM

PR #942, opened 2020-09-21
P4-2239 relax validation of profile in booked moves

martyn-w:
### Jira link

P4-2239

### What?

I have added/removed/altered:

- [x] Relaxed profile validation for booked moves

### Why?

I am doing this because:

- users need to be able to un-assign profiles from booked moves


spec/models/move_spec.rb
```diff
@@ -77,7 +77,7 @@
     )
   end
 
-  it 'validates presence of `profile` if `status` is NOT requested or cancelled' do
+  it 'validates presence of `profile` if `status` is proposed' do
```
willfish, line 80: 🎉

willfish (approved):

cwrw (approved):

PR #943, opened 2020-09-21
[P4-2244] Updates event feed to use generic events

willfish:
### Jira link

P4-2244

### What?

I have added/removed/altered:

- [x] Updates events feed to use generic events

### Why?

I am doing this because:

- This will produce the agreed format for the hub.


alexdesi (approved):
this sounds good!

PR #944, opened 2020-09-21
Dynamically validate new event STI classes

willfish:
### What?

I have added/removed/altered:

- [x] Move to dynamically adding event classes
- [x] Corrects missing use of mixin idiom for STI validations

### Why?

I am doing this because:

- We missed adding the corresponding STI classes when migrating the Move events in the specific events controller.

### DIFF
--- a/app/models/generic_event.rb
+++ b/app/models/generic_event.rb
@@ -13,6 +13,7 @@ class GenericEvent < ApplicationRecord
     eventable_type
     details
   ].freeze
+
   STI_CLASSES = %w[
     MoveAccept
     MoveApprove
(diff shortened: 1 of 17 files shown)


app/models/generic_event.rb
```diff
@@ -52,4 +60,12 @@ def self.from_event(event)
 
     type.constantize.from_event(event)
   end
+
+  def self.inherited(subclass)
```
willfish, line 64: We want to make sure that all STI classes are included in the static STI_CLASSES list. Alternative to this is to work out the constants by filename/grepping the constants that the GenericEvent model defines. The second approach isn't great because it breaks lazy loading in dev mode and we have to anticipate what to grep for. The first approach might break if a developer decides to put something else in the same generic events directory. This approach gives the richest feedback and is easier to test.

app/models/generic_event.rb
```diff
@@ -13,17 +13,13 @@ class GenericEvent < ApplicationRecord
...
-    JourneyUncancel
-    JourneyUncomplete
-  ].freeze
+
+  STI_CLASSES = Dir['app/models/generic_event/*'].map { |file|
+    file
+      .gsub('app/models/generic_event/', '')
+      .sub('.rb', '')
+      .camelcase
+  }.freeze
```
alexdesi, line 22: IMO this makes things hard to understand, and less explicit.
but I like the code, I'd move it into the test.

willfish: Have updated 👍

alexdesi (approved):
🚀

PR #945, opened 2020-09-21
[P4-2163] Use new executors in existing event apis

willfish:
### Jira link

P4-2163

### What?

I have added/removed/altered:

- [x] Use new executors in existing event apis

### Why?

I am doing this because:

- So that we can remove the existing runners and references to the old event model.


alexdesi (approved):
👍

PR #946, opened 2020-09-21
Remove journeys report as is no longer used

willfish:
### Jira link

P4-2255

### What?

I have added/removed/altered:

- [x] Removes unused report

### Why?

I am doing this because:

- Suppliers don't integrate with this report


mafonso (approved):
LGTM

PR #950, opened 2020-09-22
P4-2112 Add Reasonable Adjustments NOMIS query

cwrw:
jira-ticket: https://dsdmoj.atlassian.net/browse/P4-2112

To get a reasonable adjustment for an offender, the booking id and the list of treatment codes are passed into a GET query to return a list of reasonable adjustments.

### DIFF
--- a/app/lib/nomis_client/reasonable_adjustments.rb
+++ b/app/lib/nomis_client/reasonable_adjustments.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+module NomisClient
+  class ReasonableAdjustments < NomisClient::Base
+    class << self
+      def get(booking_id:, reasonable_adjustment_types:)
+        return [] unless booking_id.present? && reasonable_adjustment_types.present?
+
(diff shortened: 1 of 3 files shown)


app/lib/nomis_client/reasonable_adjustments.rb
```diff
@@ -0,0 +1,33 @@
...
+        reasonable_adjustments_response['reasonableAdjustments'].map do |reasonable_adjustment_attributes|
+          attributes_for(reasonable_adjustment_attributes)
+        end
+      end
+
+      def get_response(booking_id:, reasonable_adjustment_types:)
+        NomisClient::Base.get(
+          "/bookings/#{booking_id}/personal-care-needs?type=#{reasonable_adjustment_types}",
+        ).parsed
+      end
```
cwrw, line 20: I'm intentionally not handling any errors at this level, as I'd like them to bubble up to be handled in the PER NOMIS importer

martyn-w (approved):
👍 LGTM

PR #951, opened 2020-09-22
[P4-2164] Adds missing JourneyCreate and JourneyUpdate events

willfish:
### Jira link

P4-2164

### What?

I have added/removed/altered:

- [x] Added missing STI event classes `JourneyCreate` and `JourneyUpdate`

These events aren't documented, don't follow the standard process_event
behaviour and aren't affected by the event log.

We discovered these by auditing a migration outcome in production just
as we were going to make a breaking change.

These events are created via a controller callback for the update and
create actions: https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/5d1064f63a557f2b69eef16878ca8eb39e714e71/app/controllers/api/journeys_controller.rb#L10

### Why?

I am doing this because:

- Make sure that all events are copied by the event copier

### DIFF
--- a/app/models/generic_event.rb
+++ b/app/models/generic_event.rb
@@ -17,12 +17,14 @@ class GenericEvent < ApplicationRecord
   STI_CLASSES = %w[
     JourneyCancel
     JourneyComplete
+    JourneyCreate
     JourneyLockout
     JourneyLodging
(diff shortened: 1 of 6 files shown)


alexdesi (approved):
👍

PR #953, opened 2020-09-22
[P4-2164] Dual write generic events in after_action callback

willfish:
### Jira link

P4-2164

### What?

I have added/removed/altered:

- [x] Dual write generic events in after_action callback

We've added support for the JourneyUpdate and JourneyCreate
generic events https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/951.

We need to dual write these new events where the original Event model
gets created in the JourneysController: https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/5d1064f63a557f2b69eef16878ca8eb39e714e71/app/controllers/api/journeys_controller.rb#L10.

This change adds support for dual writing these events ultimately so
that we can drop support the Event model from being used.

### Why?

I am doing this because:

- We need this so that we can stop using two separate event models.

### DIFF
--- a/app/controllers/api/journeys_controller.rb
+++ b/app/controllers/api/journeys_controller.rb
@@ -146,8 +146,6 @@ module Api
         },
       )
       create_generic_event(event)
-
-      event
     end
 
     def create_generic_event(event)
--- a/app/controllers/api/journeys_controller.rb
+++ b/app/controllers/api/journeys_controller.rb
@@ -137,7 +137,7 @@ module Api
 
     def create_event
       # Logs the event for posterity and the immutable event log
-      journey.events.create!(
+      event = journey.events.create!(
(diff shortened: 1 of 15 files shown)


app/controllers/api/journeys_controller.rb
```diff
@@ -137,14 +137,23 @@ def update_journey_attributes
...
         event_name: action_name,
         client_timestamp: Time.zone.parse(data_params.dig(:attributes, :timestamp)),
         details: {
           data_params: data_params,
           supplier_id: supplier.id,
         },
       )
+      create_generic_event(event)
+
+      event
```
willfish, line 150: Remove me

alexdesi (approved):
🕺

PR #957, opened 2020-09-23
Allow personal care need types to be passed into the NOMIS client

cwrw:
To allow flexibility around the NOMIS personal care needs query, allow types to be passed in, or default to already defined personal care need types. This will allow the PER to define its own personal care need types, which can be different to the PTR.


martyn-w (approved):
👍 LGTM

willfish (approved):

PR #959, opened 2020-09-23
Removes broken fake data generation

willfish:
### Jira link

P4-<TODO>

### What?

I have added/removed/altered:

- [x] Removes broken cases for journey fake data that don't map onto real ways of redirecting journeys (an impossible event - we only redirect moves)

### Why?

I am doing this because:

- This is an impossible state to get into

### DIFF
--- a/lib/tasks/fake_data/journeys.rb
+++ b/lib/tasks/fake_data/journeys.rb
@@ -35,43 +35,6 @@ module Tasks
 
           # randomly create some events with intermediate journeys
           case random_event # add some random events to the journeys
-          when :lockout_journey_unbillable # unbillable lockout at current_location, redirect journey (not move) to intermediate_location, e.g. "delayed by flat tyre"
-            # lockout event
-            create_lockout_journey_event(
-              journey,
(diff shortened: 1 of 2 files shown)


lib/tasks/fake_data/journeys.rb
```diff
@@ -177,12 +118,8 @@ def create_redirect_move_event(timestamp, notes, location)
       end
 
       def random_event
-        case rand(1..6)
+        case rand(1..2)
```
martyn-w, line 121: how about `if rand(0..1) == 1`

martyn-w (approved):
👍

lib/tasks/fake_data/journeys.rb
martyn-w: actually -just delete this `random_event()` method and move the if up into the main method ^

willfish: We mock this method in the test so requires updating. Leaving as is painful/not my main priority atm. Will likely remove this/refactor later.

PR #960, opened 2020-09-23
Adds supplier relationship to generic_event

willfish:
### Jira link

P4-<TODO>

### What?

I have added/removed/altered:

- [x] Adds supplier relationship to generic event

### Why?

I am doing this because:

- We want to make sure we're including the supplier that is currently present in the v1 apis on every generic event.

### DIFF
--- a/app/models/generic_event.rb
+++ b/app/models/generic_event.rb
@@ -36,6 +36,7 @@ class GenericEvent < ApplicationRecord
   ].freeze
 
   belongs_to :eventable, polymorphic: true, touch: true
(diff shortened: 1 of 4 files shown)


martyn-w:
will you need a task to update existing generic events copied over from v1 to import the supplier?

willfish:
I'll be doing that separately/rolling back/updating the copier/updating newly dual-written generic events to include this

alexdesi (approved):
💯

martyn-w (approved):
👍

PR #963, opened 2020-09-23
[P4-2265] Adds supplier to generic_event feed

willfish:
### Jira link

P4-2265

### What?

I have added/removed/altered:

- [x] Update event feed to include supplier

### Why?

I am doing this because:

- We need the supplier in the generic event feed for when we backfill the generic events for the JPC.


alexdesi (approved):
🤖

PR #966, opened 2020-09-24
P4-2111 Import NOMIS mappings and fallbacks into Frameworks

cwrw:
jira-ticket: https://dsdmoj.atlassian.net/browse/P4-2111

* Create new table for NOMIS codes and fallbacks. A question can be associated to multiple NOMIS codes, or can be a fallback question if no other questions link to a certain type of NOMIS resource. Those properties are not mutually exclusive, and a question could be a fallback question and also be linked to other codes.
* Import NOMIS codes for a question when pulling in new Frameworks, also set questions as fallback questions if the NOMIS fallback option is available.

### DIFF
--- a/app/models/framework_nomis_code.rb
+++ b/app/models/framework_nomis_code.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class FrameworkNomisCode < VersionedModel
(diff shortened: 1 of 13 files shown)


willfish (approved):
LGTM

PR #969, opened 2020-09-24
P4-2198 Fix unique profile when creating the PER error

cwrw:
jira-ticket: https://dsdmoj.atlassian.net/browse/P4-2198

When a user is double clicking create the PER button, they are seeing a 500 error as opposed to be gracefully redirected to the already created PER. Ideally the API should be catching this uniqueness validation error, and returning a 422 with the relevant message, however due to the way our app is setup in production, where we have multiple instances attempting the same thing,  this isn't being caught as a validation error on uniqueness, and is falling back to the database level of
errors, and raises an exception.

I found that this article explains it pretty well: https://thoughtbot.com/blog/the-perils-of-uniqueness-validations

We have to manually catch this error if it happens. I've added tests for both instances, so now we can handle the uniqueness validation as well as the DB error itself.

### DIFF
--- a/spec/requests/api/person_escort_records_controller_create_spec.rb
+++ b/spec/requests/api/person_escort_records_controller_create_spec.rb
@@ -105,6 +105,47 @@ RSpec.describe Api::PersonEscortRecordsController do
         it_behaves_like 'an endpoint that responds with error 404'
       end
 
+      context 'when a person escort record already exists on a profile' do
+        let(:errors_422) do
+          [
+            {
+              'title' => 'Unprocessable entity',
+              'detail' => 'Profile has already been taken',
+              'source' => { 'pointer' => '/data/attributes/profile' },
+              'code' => 'taken',
+            },
+          ]
+        end
+
+        before do
+          post '/api/v1/person_escort_records', params: person_escort_record_params, headers: headers, as: :json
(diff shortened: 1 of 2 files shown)


spec/requests/api/person_escort_records_controller_create_spec.rb
```diff
@@ -105,6 +105,47 @@
...
+          [
+            {
+              'title' => 'Invalid profile',
+              'detail' => 'Validation failed: Profile Profile has already been take',
+            },
+          ]
+        end
+
+        before do
+          person_escort_record = PersonEscortRecord.new
```
cwrw, line 139: attempted to use an `instance_double` but that didn't work as we will be calling methods regarding errors on this instance

spec/requests/api/person_escort_records_controller_create_spec.rb
```diff
@@ -105,6 +105,47 @@
...
+
+        it_behaves_like 'an endpoint that responds with error 422'
+      end
+
+      context 'when a person escort record already exists on a profile and unique index error thrown' do
+        let(:errors_422) do
+          [
+            {
+              'title' => 'Invalid profile',
+              'detail' => 'Validation failed: Profile Profile has already been take',
```
cwrw, line 133: remove duplicate message

martyn-w (approved):
👍  LGTM

PR #970, opened 2020-09-24
Rescue multiple errors on unique index error

cwrw:
This isn't being caught, so add all possible errors to rescue on


martyn-w (approved):
👍

willfish (approved):

PR #974, opened 2020-09-24
Add new tasks to backfill generic event starting from a certain date

alexdesi:
### Jira link

P4-2263

What

Write a script to backfill all generic events from updated_at from last 3 days of August to now.

Why

We need this for the JPC

### DIFF
--- a/lib/tasks/feeds/event.rake
+++ b/lib/tasks/feeds/event.rake
@@ -5,4 +5,24 @@ namespace :feeds do
 
     CloudDataFeed.new.write(feed, 'events.jsonl')
   end
+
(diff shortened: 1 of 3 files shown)


lib/tasks/feeds/event.rake
```diff
@@ -5,4 +5,24 @@ namespace :feeds do
 
     CloudDataFeed.new.write(feed, 'events.jsonl')
   end
+
+  desc 'Exports the JSON starting from a certain date'
```
willfish, line 9:
```suggestion
  desc 'Exports the JSON starting from a certain date (inclusive)'
```

willfish (approved):
Looks good 🎉

PR #977, opened 2020-09-25
Fix ambiguous date column error in Allocations::Finder service

smoothcontract:
### Jira link

IM-21

### What?

- [x] Explicitly spec table name for `date` column in Allocations::Finder service

### Why?

- This was only surfaced on production, so also added a failing spec to cover the relevant scenario that was missing from unit test suite.

### DIFF
--- a/app/services/allocations/finder.rb
+++ b/app/services/allocations/finder.rb
@@ -68,8 +68,8 @@ module Allocations
     end
 
     def apply_date_range_filters(scope)
-      scope = scope.where('date >= ?', filter_params[:date_from]) if filter_params.key?(:date_from)
-      scope = scope.where('date <= ?', filter_params[:date_to]) if filter_params.key?(:date_to)
+      scope = scope.where('allocations.date >= ?', filter_params[:date_from]) if filter_params.key?(:date_from)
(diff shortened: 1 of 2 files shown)


app/services/allocations/finder.rb
```diff
@@ -68,8 +68,8 @@ def apply_filters(scope)
     end
 
     def apply_date_range_filters(scope)
-      scope = scope.where('date >= ?', filter_params[:date_from]) if filter_params.key?(:date_from)
-      scope = scope.where('date <= ?', filter_params[:date_to]) if filter_params.key?(:date_to)
+      scope = scope.where('allocations.date >= ?', filter_params[:date_from]) if filter_params.key?(:date_from)
```
martyn-w, line 71: do you also need to order by `allocations.date` rather than `date` on line 12?

smoothcontract: Rails seems to handle that case (line 12) correctly and automatically applies the explicit scoping. I've added another spec to exercise filtering and ordering by date which passes fine, so I'm confident this is the only place where the scope needs to be modified. It's needed here because the `where` clause is passed as a string rather than a symbol.

martyn-w: 👍 jolly good

martyn-w (approved):
👍

PR #979, opened 2020-09-25
[P4-2264] Remove GenericEvent:: prefix in type in feed view

willfish:
### Jira link

P4-2264

### What?

I have added/removed/altered:

- [x] Removes the prefix GenericEvent:: on type in the for_feed view for the hub
- [x] Updates corresponding tests

### Why?

I am doing this because:

- This is ugly and namespacing implementation detail that isn't relevant to the hub


alexdesi (approved):
✨  nice cleaning up!

PR #981, opened 2020-09-28
[P4-2264] Handle event-specific relationships in generic events controller

willfish:
### Jira link

P4-2264

### What?

- [x] Support additional event-specific relationships in the generic api
- [x] Add tests to validate that we can send these relationships either in the json:api relationships section

### Why?

I am doing this because:

We need to support relationships in the json api style but store them in the event-specific details jsonb field.


cwrw:
We're gonna catchup in person as I don't have the full context regarding this change, but to capture my initial thoughts: I think it would be best to only support JSON:API complaint specifications, where we only set relationships that are set in the `relationships` field as the current implementation supports, to avoid breaking the spec and also to avoid confusion, specifically if there are no special and needed circumstances to do so.

spec/requests/api/generic_events_controller_create_spec.rb
```diff
@@ -47,8 +42,42 @@
       before { do_post }
     end
 
+    context 'when event-specific relationships are in the details attribute' do
```
cwrw, line 45: so I understand that we want to merge relationships into details for reporting purposes, which is ok. But I don't think this should pass, we shouldn't allow users to set relationships through details attribute by mistake, due to the previously mentioned note around JSON:API compliance.

app/controllers/api/generic_events_controller.rb
```diff
@@ -56,5 +60,15 @@ def event_type
     def created_by
       current_user&.owner&.name || 'unknown'
     end
+
+    def event_specific_relationships
```
cwrw, line 64: Not a blocker: this method might be better suited living outside the controller where we can unit test it

spec/requests/api/generic_events_controller_create_spec.rb
cwrw, line 45: spoke in person: We shouldn't allow accidentally setting relationships through the details attribute, we agreed that this test will be marked as pending, and will be tackled in future PRs to remove the creation of location through details, and only allow it through relationships

cwrw (approved):
Spoke in person, approving for now but issues will be tackled in later prs

PR #983, opened 2020-09-28
[P4-2264] Adds success case event controller specs

willfish:
### Jira link

P4-2264

### What?

I have added/removed/altered:

- [x] For each event, adds a controller spec which validates the event is successfully created via the controller action happy path.

### Why?

I am doing this because:

- Straightforward test of the event plumbing


martyn-w (approved):
👍

PR #984, opened 2020-09-28
Feed files are generated yesterday

willfish:
### Jira link

P4-<TODO>

### What?

I have added/removed/altered:

- [x] Updated CloudDataFeed writer to default to write a file with folder/file names of yesterday.

### Why?

I am doing this because:

- All reports feed range queries apply to yesterday


mafonso (approved):
LGTM

PR #987, opened 2020-09-29
P4-2067 Allow passing in a Move to create a Person Escort Record

cwrw:
### Jira link

[P4-2067](https://dsdmoj.atlassian.net/browse/P4-22067)

### What?

- [x] Added new relationship on Person escort record to Move
- [x] Allow Move optionally to be passed in on Person escort record creation
- [x] Surface Move relationship on the Person Escort Record endpoint

### Why?

To allow the Person escort record access to a location's type, to determine if a location is a Police or Prison location for importing NOMIS resources, we need access to the Move associated to the PER. We cannot do that through the Profile, as a Profile can belong to multiple Moves. Instead we need to pass in the Move to create the Person Escort Record, and associate the Move's profile instead.

Since transitioning to passing in a Move to create a PER is a breaking change, support both Profile and Move until the FE completes the transition to only use Move. For now the Move relationship is optional.

### Have you? (optional)

- [x] I will update API docs when we officially support only Move relationship creation when creating the PER

### DIFF
--- a/app/controllers/api/person_escort_records_controller.rb
+++ b/app/controllers/api/person_escort_records_controller.rb
@@ -4,10 +4,11 @@ module Api
   class PersonEscortRecordsController < ApiController
     after_action :send_notification, only: :update
 
+    # TODO: Remove profile id and transition to Move id, for now accept both
     NEW_PERMITTED_PER_PARAMS = [
       :type,
       attributes: [:version],
-      relationships: [profile: {}],
+      relationships: [profile: {}, move: {}],
     ].freeze
 
     UPDATE_PERMITTED_PER_PARAMS = [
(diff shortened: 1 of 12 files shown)


willfish:
From a PER perspective I guess the profile assessment answers (now literally the only thing on a profile) are pretty useless to you. The characteristic that was interesting about a profile was it enabled reuse of the assessment answers (and other things), and the capturing of the answers at a point in time - but so would making Move#assessment_answers and Move#person_escort_record via a table. I think this is the death knell of the profile concept and the profile table should just be renamed to `ptr` or equivalent. Given your specific goals, though, I don't think this is relevant to you now: I think you've just discovered it's pretty useless.

cwrw:
@willfish I think we still do need the profile and will still keep the relationship on the PER. However we cannot solely rely on it to create the PER due to the multiple moves on the profile issue. When suppliers wish to reuse a PER across multiple moves, they will do that through reusing the same profile across the other Move, so I do think there is definitely still use for a profile.

willfish (approved):
Seems to do what you say it does :).

willfish:
Okay @cwrw

PR #990, opened 2020-09-29
[P4-2264] Enable json:api only relationships in generic event api

willfish:
### Jira link

P4-2264

### What?

I have added/removed/altered:

- [x] Constrain acceptable attributes in the details json blob for all generic events (especially ignoring relationship fields).
- [x] Fix specs which were sending relationships in the details json blob to use json:api standard

### Why?

- After a conversation with @cwrw we agreed that restricting the way relationships can be set in the api is preferable/more secure/less confusing.

### Have you? (optional)

- [x] This reflects what is documented so no updates necessary

### DIFF
--- a/app/models/generic_event/journey_arrive_at_outer_gate.rb
+++ b/app/models/generic_event/journey_arrive_at_outer_gate.rb
@@ -1,5 +1,7 @@
 class GenericEvent
   class JourneyArriveAtOuterGate < GenericEvent
+    DETAILS_ATTRIBUTES = %w[].freeze
+
     include JourneyEventValidations
   end
 end
--- a/app/models/generic_event/journey_arrive_at_outer_gate.rb
+++ b/app/models/generic_event/journey_arrive_at_outer_gate.rb
@@ -1,7 +1,5 @@
 class GenericEvent
   class JourneyArriveAtOuterGate < GenericEvent
-    DETAILS_ATTRIBUTES = %w[].freeze
-
     include JourneyEventValidations
   end
 end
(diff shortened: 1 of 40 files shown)


app/models/generic_event/journey_arrive_at_outer_gate.rb
```diff
@@ -1,5 +1,7 @@
 class GenericEvent
   class JourneyArriveAtOuterGate < GenericEvent
+    DETAILS_ATTRIBUTES = %w[].freeze
```
cwrw, line 3: To avoid having empty constants in classes that don't support details, is it possible to default to no detail attributes unless they exist in a class, or maybe move the empty `DETAILS_ATTRIBUTES` to the generic event base class?

spec/support/a_generic_event_endpoint.rb
```diff
@@ -13,17 +13,37 @@
...
+  let(:event_specific_relationships) do
+    attributes = attributes_for(factory)
+    relationship_attributes = attributes[:details].slice(*attributes[:details].keys.grep(/_id/))
+
+    relationship_attributes.each_with_object({}) do |(key, value), acc|
+      named_relationship_key = key.to_s.sub('_id', '')
+      acc[named_relationship_key] = {
+        'data' => {
+          'id' => value,
+          'type' => 'locations',
```
cwrw, line 39: is the type of those relationships always location?

cwrw (commented):
Thanks for addressing the previous comments! left a few comments but otherwise 👍

spec/support/a_generic_event_endpoint.rb
willfish: Currently, yeah. In future, if there are other types I can just override the invocation block for the shared example with a `let(:event_specific_relationships)` and they'll get merged in.

app/models/generic_event/journey_arrive_at_outer_gate.rb
willfish: So I'm not sure. It requires a small amount of maintenance for more visibility of the possible details when you look at the individual event. I could/did swing either way on this. If you feel strongly I can DRY it up a bit.

cwrw (approved):

PR #993, opened 2020-09-30
[P4-2264] Adds GenericEvents::Runner and consumes it when creating GenericEvents

willfish:
### Jira link

P4-2264

### What?

- [x] Invoke a `GenericEvents::Runner` everytime we create an event (for allocations, move events, journey events and journey CRUD lifecycle audit events).
- [x] Generalise by pushing the behaviour around handling the applied order trigger result on the eventable into the eventable model and move unit tests there
- [x] Move away from special snowflake `move_events` model which was responsible for defining getters when triggering a move event in the old world. This means that we can treat event creation in the same way and consolidate the two Journey::Eventable Move::Eventable modules.

### Why?

I am doing this because:

- We need to make sure that generic events get triggered in the applied order as we can expect suppliers to send events in the wrong order but with the correct occurred_at timestamp
- This behaviour was missing from the generic event controller
- Duplicating the runner between the v1/v2 approaches didn't make sense

- This is a transparent/same behaviour change to the currently used event creation process.


app/models/move.rb
```diff
@@ -84,8 +84,8 @@ class Move < VersionedModel
   has_many :notifications, as: :topic, dependent: :destroy # NB: polymorphic association
   has_many :journeys, -> { default_order }, dependent: :restrict_with_exception, inverse_of: :move
   has_many :court_hearings, dependent: :restrict_with_exception
-  has_many :move_events, as: :eventable, dependent: :destroy # NB: polymorphic association
```
willfish, line 87: This model exposed getter methods for the previous implementation of the runner. These getters are no longer used and this change makes the relational api between the two currently-implemented eventables (Journey, Move) the same. It meant that in the new `GenericEvents::Runner` we can now call `Move#events#create` and `Journey#events#create` in a consistent way and consolidate the two (tidy up coming in separate PR once this has been removed).

app/models/move.rb
```diff
@@ -84,8 +84,9 @@ class Move < VersionedModel
...
 
+  has_many :move_events, as: :eventable, dependent: :destroy # NB: polymorphic association
```
martyn-w, line 88: what's the difference between `move_events` and `events` ? if they are the same thing (a v1 event), do we need both here?

app/controllers/concerns/eventable.rb
```diff
@@ -0,0 +1,38 @@
...
+      client_timestamp: Time.zone.parse(event_params.dig(:attributes, :timestamp)),
+      details: {
+        event_params: event_params,
+        supplier_id: supplier_id,
+      },
+    )
+  end
+
+  def create_generic_event(event)
+    generic_event = GenericEvent.from_event(event)
```
martyn-w, line 29: creating the generic event from the v1-event is a bit convoluted and means the event needs to be updated after its been saved. How about instead creating the generic event first, and then creating the v1-event referencing the generic event second?

app/controllers/concerns/eventable.rb
```diff
@@ -0,0 +1,38 @@
+module Eventable
+  extend ActiveSupport::Concern
+
+  def supplier_id
+    # NB: not all events will have a supplier_id so this could well be nil
+    @supplier_id ||= current_user.owner&.id
+  end
+
+  def process_event(eventables, event_name, event_params)
```
martyn-w, line 9: where is `process_event` called from?

spec/models/move_spec.rb
```diff
@@ -572,4 +572,70 @@
...
+      it 'does not trigger a move notification' do
+        move.handle_event_run
+
+        expect(Notifier).not_to have_received(:prepare_notifications)
+      end
+    end
+
+    context 'when the move status has changed' do
+      before do
+        move.status = 'in_transit'
```
martyn-w, line 597: what is being tested here? the test looks like the move goes from `in_transit` to `in_transit`?

app/models/move.rb
willfish: We need move_events because they're used in some outdated (to be removed in a separate PR) tests and in the diagnostics endpoint.

app/controllers/concerns/eventable.rb
willfish: This is an old pattern we're following and that I plan to refactor later when I come to remove dual writing.

app/controllers/concerns/eventable.rb
willfish: It's the same as the previous eventable implementations that I'll be removing in a separate PR. It get's included in the relevant controllers (AllocationEventsController, MoveEventsController, JourneyEventsController and invoked as part of each action)

spec/models/move_spec.rb
willfish: I'm simply testing the plumbing around when the eventable changes it should save the change and hit the has changed branch. I'm not actually interested in the transition itself. The responsibility of this method is to validate when the move has changed (initial state is `requested` (see the [factory](https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/852c78c2f48f3e7ca4a87834ca3fcee06bd372bc/spec/factories/move.rb#L10)). Would it be

willfish: For moves, the run handler needs to behave the same as the following https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/852c78c2f48f3e7ca4a87834ca3fcee06bd372bc/app/services/event_log/move_runner.rb#L46-L51. So we validate that a change leads to a save and a notification when it succeeds/is valid and a different message to the notifier when the status versus other attributes have changed.

app/controllers/concerns/eventable.rb
willfish: Explained this offline and we agreed that dual-writing is technically not needed anymore, but that because updating the `#from_event` builders is going to create quite a lot of PR noise this should be done separately. I'm going to prioritise that next.
 
1. I wanted a consistent approach to running event logs between the old and the new apis in this project. This led to

martyn-w (approved):
👍  approving on the basis that this is a stepping-stone and further PRs will simplify the methodology for creating generic events

PR #994, opened 2020-09-30
Add includes to allocation to avoid n+1 queries when including resources

cwrw:
Locally to query over 300 allocations this comes down from 16s to 4s, there are some circular dependencies for locations but because they are not pulling in much it seems ok

### DIFF
--- a/app/services/allocations/finder.rb
+++ b/app/services/allocations/finder.rb
@@ -4,7 +4,7 @@ module Allocations
   class Finder
     attr_reader :filter_params, :search_params
 
-    ALLOCATION_INCLUDES = [:from_location, :to_location, moves: %i[profile]].freeze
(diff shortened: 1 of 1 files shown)


app/services/allocations/finder.rb
```diff
@@ -4,7 +4,7 @@ module Allocations
   class Finder
     attr_reader :filter_params, :search_params
 
-    ALLOCATION_INCLUDES = [:from_location, :to_location, moves: %i[profile]].freeze
+    ALLOCATION_INCLUDES = [from_location: :suppliers, to_location: :suppliers, moves: %i[from_location to_location court_hearings profile person]].freeze
```
cwrw, line 7: gonna test this on staging again for posterity, tested locally with a lot of data

willfish (approved):
LGTM pending staging test

smoothcontract (approved):
Discussed offline, looks good and performs better than the existing implementation. Hoping that this improves further when moving away from AMS 🤞

PR #997, opened 2020-10-01
[P4-2264] Remove unused eventable, move_events and runner code

willfish:
### Jira link

P4-2264

### What?

I have added/removed/altered:

- [x] Removes eventable modules that have now been [consolidated](https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/controllers/concerns/eventable.rb)
- [x] Removes references to the MoveEvent model which defined getters on the details attributes for the v1 event runners. These have been replaced with getters that expose these details in the STI classes themselves (e.g. https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/096162f54db0777289760ffa98408da9b9e48621/app/models/generic_event/move_cancel.rb#L23-L27)
- [x] Removes now-superseded runner service with one which delegates to the individual STI model https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/096162f54db0777289760ffa98408da9b9e48621/app/services/generic_events/runner.rb

### Why?

All of this code is no longer used


app/services/diagnostics/move_inspector.rb
```diff
@@ -36,9 +36,9 @@ def generate
         -----------
       ENDMOVEEVENTS
 
-      if move.move_events.any?
+      if move.events.any?
```
willfish, line 39: When I come to remove dual writing I will replace these with the generic events relationship

lib/tasks/fake_data/journeys.rb
```diff
@@ -79,40 +79,15 @@ def create_journey(timestamp, journey_from, journey_to, billable)
...
-              relationships: {
-                from_location: { data: { id: location.id } },
-              },
-            },
-          },
-        )
-      end
-
       def create_redirect_move_event(timestamp, notes, location)
-        move.move_events.create!(
```
willfish, line 102: TODO: If we're not moving to v2 events in the diagnostics endpoint we should just dereference move_events here and preserve creation of v1 events until we migrate away from dual writing.

spec/services/event_log/journey_executor_spec.rb
```diff
@@ -1,65 +0,0 @@
-# frozen_string_literal: true
```
martyn-w, line 1: shouldn't these tests be reproduced in some form in the new generic event world?

willfish: Short answer, no. Long answer... So from the perspective of the runner we only know about whether an eventable is valid/invalid by an integration. We test the handler in the eventable model spec. We test the trigger in the STI model spec. Should we test the integration here at the unit level? I don't think so. Should we test these specific transitions in a workflow somewhere: if there's business value in doing so, yeah. But still not here. Happy to add a task to add some workflow/acceptance tests that cover these integrated behaviours. I'm not convinced we'd be learning/avoiding a breaking change in the behaviour by adding them back in, though. We also do a full integration at the request level for all events that are currently possible to create and the only thing that varies in different transitions is whether the eventable is valid.

willfish: Presumably whether the transitions are valid themselves are tested inside of the state machine specs, too.

martyn-w: I guess you are saying these are integration tests not unit tests, so shouldn't be here?

my concern is that simply deleting these tests and relying purely on unit tests does not offer guarantees that all the expected behaviours will continue to work. 

For example, where there are multiple events for an `eventable`, these should be ordered by timestamp and applied in that order: this may or may not result in a change to the `eventable`. 

I don't think simply testing each event in isolation does not replicate the real-life sceanario with multiple events?

willfish: So I guess I'm saying is if you could clarify a specific behaviour that is missed without an integration test then I'll add it separately.

willfish: Calling the trigger method in the applied order is tested, yeah.

martyn-w: E.g.:
System has a redirect to A with a timestamp at 15:00
Some time later a new event arrives: redirect to B with timestamp at 14:00
=> net result is log the event but don't change the destination

willfish: Okay, PR and story incoming :)

martyn-w: that is just one example ^ 
... ideally, I think all the existing tests should persist

willfish: https://dsdmoj.atlassian.net/browse/P4-2322

martyn-w: I am concerned that the generic events runner (already merged) doesn't have tests of complex cases (such as 2 unordered redirects) and I'd like to see one or more test(s) for complex cases...

martyn-w (changes requested):
I'm afraid I'm unhappy with deleting important complex tests without re-creating them for the v2-generic event world

willfish:
Tests reinstated here https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/1004

martyn-w (approved):
👍 LGTM

PR #998, opened 2020-10-01
[P4-2164] Removes dual writing of v1 and v2 events

willfish:
### Jira link

P4-2164

### What?

I have added/removed/altered:

- [x] Removes dual writing of events

### Why?

I am doing this because:

- The old event model is no longer referenced so this is no longer necessary.

### DIFF
--- a/app/controllers/concerns/eventable.rb
+++ b/app/controllers/concerns/eventable.rb
@@ -8,31 +8,41 @@ module Eventable
 
   def process_event(eventables, event_name, event_params)
     [eventables].flatten.each do |eventable|
-      event = create_event(eventable, event_name, event_params)
-      create_generic_event(event)
+      create_generic_event(eventable, event_name, event_params)
       run_event_logs(eventable)
     end
   end
 
-  def create_event(eventable, event_name, event_params)
-    eventable.events.create!(
-      event_name: event_name,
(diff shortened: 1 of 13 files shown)


app/controllers/concerns/eventable.rb
```diff
@@ -8,31 +8,41 @@ def supplier_id
...
     GenericEvents::Runner.new(eventable).call
   end
+
+private
+
+  def assign_common_attributes!(attributes, eventable, event_params)
+    timestamp = Time.zone.parse(event_params.dig(:attributes, :timestamp))
+
+    attributes[:eventable]   = eventable
+    attributes[:occurred_at] = timestamp
```
martyn-w, line 37: how about using attributes `occured_at` and `recorded_at` and fall back to `timestamp` if they are not provided?

willfish: This module is only used in an api where timestamp is provided so I'm not sure the flexibility adds value? I think our intention is probably to leisurely migrate all calls to be via `POST /api/events` rather than these existing endpoints and ultimately deprecate and then remove this code.

martyn-w: ah of course, sorry for not realising that 👍

martyn-w (approved):
👍 LGTM

PR #999, opened 2020-10-02
P4-2067 Add NOMIS Mappings and Importer

cwrw:
### Jira link

[P4-2067](https://dsdmoj.atlassian.net/browse/P4-2067)

### What?
- [x] Add Framework NOMIS mapping table and model
- [x] Add a mapping importer for each type of NOMIS mapping
- [x] Add NOMIS Mapping Importer

### Why?

NOMIS alerts, personal care needs as well as reasonable adjustments for a person will help users fill out the Person escort record more accurately. Each question in the PER is associated to the relevant mappings of NOMIS resource codes, or set as a fallback question if a person has NOMIS resources that are not associated to any question.

To allow users to view those associations, add an Importer which pulls in all the resources for a person, and maps them appropriately to the NOMIS response.

NOMIS codes are not unique to a question, so we need to make sure we associate mappings to each and every question a NOMIS code appears in. If this NOMIS code is not in out framework, we will group it under the fallback response with the correct type.

Failures of importing and validation/persisting of NOMIS mappings will be addressed in further tickets.
### Have you? (optional)

- Updated API docs if necessary: will happen in subsequent PRs

- The latest framework has not been added to production ENVs so we will not be using this importer yet

### DIFF
--- a/app/services/framework_nomis_mappings/importer.rb
+++ b/app/services/framework_nomis_mappings/importer.rb
@@ -65,7 +65,7 @@ module FrameworkNomisMappings
 
           if mapping_nomis_codes.any?
             mapping_nomis_codes.each do |nomis_code|
-              hash[nomis_code.id] = mapping
+              hash[nomis_code.id] = hash[nomis_code.id].to_a + [mapping]
             end
           elsif mapping_nomis_fallback
             hash[mapping_nomis_fallback.id] = hash[mapping_nomis_fallback.id].to_a + [mapping]
--- a/app/services/framework_nomis_mappings/importer.rb
+++ b/app/services/framework_nomis_mappings/importer.rb
@@ -3,6 +3,7 @@
 module FrameworkNomisMappings
   class Importer
     attr_reader :person, :framework_responses, :framework_nomis_codes
+
     def initialize(person:, framework_responses:, framework_nomis_codes:)
       @person = person
       @framework_responses = framework_responses
--- a/app/services/framework_nomis_mappings/importer.rb
+++ b/app/services/framework_nomis_mappings/importer.rb
@@ -1,0 +1,81 @@
+# frozen_string_literal: true
+
+module FrameworkNomisMappings
+  class Importer
+    attr_reader :person, :framework_responses, :framework_nomis_codes
+    def initialize(person:, framework_responses:, framework_nomis_codes:)
+      @person = person
+      @framework_responses = framework_responses
+      @framework_nomis_codes = framework_nomis_codes
+    end
+
+    def call
+      return unless person && framework_responses.any? && framework_nomis_codes.any?
+
+      ActiveRecord::Base.transaction do
+        return unless persist_framework_nomis_mappings.any?
+
+        framework_responses.includes(:framework_nomis_mappings).each do |response|
+          nomis_code_ids = responses_to_codes[response.id]&.pluck(:nomis_code_id)
+          response.framework_nomis_mappings = nomis_code_ids_to_mappings.slice(*nomis_code_ids).values.flatten
+        end
+      end
+    end
+
+  private
+
+    def persist_framework_nomis_mappings
+      @persist_framework_nomis_mappings ||= begin
+        mappings = alert_mappings + personal_care_need_mappings + reasonable_adjust_mappings
+        # TODO: log any validation failures
+        import = FrameworkNomisMapping.import(mappings, all_or_none: true)
+
+        FrameworkNomisMapping.where(id: import.ids)
+      end
+    end
+
+    def alert_mappings
+      FrameworkNomisMappings::Alerts.new(prison_number: person.prison_number).call
+    end
+
+    def personal_care_need_mappings
+      FrameworkNomisMappings::PersonalCareNeeds.new(prison_number: person.prison_number).call
+    end
+
+    def reasonable_adjust_mappings
+      FrameworkNomisMappings::ReasonableAdjustments.new(
+        booking_id: person.latest_nomis_booking_id,
+        nomis_codes: grouped_framework_nomis_codes['reasonable_adjustment'],
+      ).call
+    end
+
+    def grouped_framework_nomis_codes
+      framework_nomis_codes.group_by(&:code_type)
+    end
+
+    def fallback_nomis_codes
+      @fallback_nomis_codes ||= framework_nomis_codes.select(&:fallback?)
+    end
+
+    def nomis_code_ids_to_mappings
+      @nomis_code_ids_to_mappings ||= begin
+        persist_framework_nomis_mappings.each_with_object({}) do |mapping, hash|
+          mapping_nomis_codes = framework_nomis_codes.select { |nomis_code| nomis_code.code == mapping.code && nomis_code.code_type == mapping.code_type }
+          mapping_nomis_fallback = fallback_nomis_codes.find { |fallback| fallback.code_type == mapping.code_type }
+
+          if mapping_nomis_codes.any?
+            mapping_nomis_codes.each do |nomis_code|
+              hash[nomis_code.id] = mapping
+            end
+          elsif mapping_nomis_fallback
+            hash[mapping_nomis_fallback.id] = hash[mapping_nomis_fallback.id].to_a + [mapping]
+          end
+        end
+      end
+    end
+
+    def responses_to_codes
+      @responses_to_codes ||= framework_responses.joins(:framework_nomis_codes).select('framework_responses.id as id, framework_nomis_codes.id as nomis_code_id').group_by(&:id)
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_09_29_055929) do
+ActiveRecord::Schema.define(version: 2020_10_01_043655) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "citext"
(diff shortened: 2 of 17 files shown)


app/services/framework_nomis_mappings/importer.rb
```diff
@@ -0,0 +1,81 @@
...
+    def grouped_framework_nomis_codes
+      framework_nomis_codes.group_by(&:code_type)
+    end
+
+    def fallback_nomis_codes
+      @fallback_nomis_codes ||= framework_nomis_codes.select(&:fallback?)
+    end
+
+    def nomis_code_ids_to_mappings
+      @nomis_code_ids_to_mappings ||= begin
```
cwrw, line 61: add test and account for duplicate NOMIS alerts

app/services/framework_nomis_mappings/importer.rb
```diff
@@ -0,0 +1,81 @@
...
+          nomis_code_ids = responses_to_codes[response.id]&.pluck(:nomis_code_id)
+          response.framework_nomis_mappings = nomis_code_ids_to_mappings.slice(*nomis_code_ids).values.flatten
+        end
+      end
+    end
+
+  private
+
+    def persist_framework_nomis_mappings
+      @persist_framework_nomis_mappings ||= begin
```
smoothcontract, line 28: I can't see this being called more than once, so no need to memoize?

smoothcontract: Ignore that comment, I found another reference

app/services/framework_nomis_mappings/importer.rb
```diff
@@ -0,0 +1,81 @@
+# frozen_string_literal: true
+
+module FrameworkNomisMappings
+  class Importer
+    attr_reader :person, :framework_responses, :framework_nomis_codes
```
smoothcontract, line 5: Blank line here pls

app/services/framework_nomis_mappings/importer.rb
```diff
@@ -0,0 +1,81 @@
...
+            end
+          elsif mapping_nomis_fallback
+            hash[mapping_nomis_fallback.id] = hash[mapping_nomis_fallback.id].to_a + [mapping]
+          end
+        end
+      end
+    end
+
+    def responses_to_codes
+      @responses_to_codes ||= framework_responses.joins(:framework_nomis_codes).select('framework_responses.id as id, framework_nomis_codes.id as nomis_code_id').group_by(&:id)
```
smoothcontract, line 78: Not sure why this needs a `group_by` clause?

db/schema.rb
```diff
@@ -162,6 +162,27 @@
...
 
+  create_table "framework_nomis_mappings", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.jsonb "raw_nomis_mapping", null: false
+    t.string "code", null: false
+    t.string "code_type", null: false
+    t.text "code_description"
+    t.text "comments"
+    t.date "start_date"
+    t.date "end_date"
+    t.date "creation_date"
```
smoothcontract, line 173: Is `creation_date` different from `created_at` ?

spec/services/framework_nomis_mappings/importer_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+    allow(NomisClient::Alerts).to receive(:get).and_return([nomis_alert])
+    allow(NomisClient::PersonalCareNeeds).to receive(:get).and_return(nomis_personal_care_needs)
+    allow(NomisClient::ReasonableAdjustments).to receive(:get).and_return(nomis_reasonable_adjustments)
+  end
+
+  it 'persists alerts, reasonable adjustments and personal care needs NOMIS mappings' do
+    framework_responses = FrameworkResponse.where(id: [framework_response1.id, framework_response2.id])
+    framework_nomis_codes = framework_responses.flat_map(&:framework_nomis_codes)
+
+    expect { described_class.new(person: person, framework_responses: framework_responses, framework_nomis_codes: framework_nomis_codes).call }.to change(FrameworkNomisMapping, :count).by(5)
```
smoothcontract, line 90: Could you add a comment to explain why this adds 5 mappings please?

spec/services/framework_nomis_mappings/importer_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+    allow(NomisClient::Alerts).to receive(:get).and_return([])
+    allow(NomisClient::PersonalCareNeeds).to receive(:get).and_return([])
+    allow(NomisClient::ReasonableAdjustments).to receive(:get).and_return([])
+    framework_responses = FrameworkResponse.where(id: [framework_response1.id, framework_response2.id])
+    framework_nomis_codes = framework_responses.flat_map(&:framework_nomis_codes)
+
+    expect { described_class.new(person: person, framework_responses: framework_responses, framework_nomis_codes: framework_nomis_codes).call }.not_to change(FrameworkNomisMapping, :count)
+  end
+
+  it 'does nothing is no person present' do
```
smoothcontract, line 157: `is` => `if`

spec/services/framework_nomis_mappings/importer_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+  end
+
+  it 'does nothing if no framework responses present' do
+    framework_responses = FrameworkResponse.where(id: [framework_response1.id, framework_response2.id])
+    framework_nomis_codes = framework_responses.flat_map(&:framework_nomis_codes)
+
+    expect { described_class.new(person: person, framework_responses: [], framework_nomis_codes: framework_nomis_codes).call }.not_to change(FrameworkNomisMapping, :count)
+  end
+
+  it 'does nothing is no framework NOMIS codes present' do
```
smoothcontract, line 171: `is` => `if`

spec/services/framework_nomis_mappings/personal_care_needs_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+  end
+
+  it 'ignores personal care needs with status "I"' do
+    allow(NomisClient::PersonalCareNeeds).to receive(:get).and_return([nomis_personal_care_need(problem_status: 'I')])
+    mappings = described_class.new(prison_number: 'A9127EK').call
+
+    expect(mappings).to be_empty
+  end
+
+  it 'ignores personal care needs with status "EBS"' do
```
smoothcontract, line 48: Might be nice to have a comment somewhere indicating what `I` and `EBS` magic values represent

smoothcontract (approved):
A lot going on here so is a big PR. I think I follow it though and have left a few trivial comments for consideration.

app/services/framework_nomis_mappings/importer.rb
cwrw: We need to group the NOMIS codes by the response to make it easier and more efficient to lookup the NOMIS codes later on by response id, where now its through a hash key rather than having to select those every time. I added this to avoid doing more than one query to get those values, and instead have this in memory.

db/schema.rb
cwrw: we wanted to differentiate the `created_at` coming from NOMIS for some resources from the `created_at` datetime table value to avoid confusion, so we went with `creation_date` instead

cwrw: It also ties nicely with `expiry_date`

spec/services/framework_nomis_mappings/importer_spec.rb
cwrw: Not sure I understand, this test checks that all the resources defined earlier are persisted, could you elaborate on where the uncertainty is and will add a comment

app/services/framework_nomis_mappings/importer.rb
smoothcontract: Awesome, thanks for the explanation 👍

db/schema.rb
smoothcontract: Fair enough 👍

spec/services/framework_nomis_mappings/importer_spec.rb
```diff
@@ -0,0 +1,189 @@
...
+    allow(NomisClient::Alerts).to receive(:get).and_return(nomis_alerts)
+    allow(NomisClient::PersonalCareNeeds).to receive(:get).and_return(nomis_personal_care_needs)
+    allow(NomisClient::ReasonableAdjustments).to receive(:get).and_return(nomis_reasonable_adjustments)
+  end
+
+  it 'persists alerts, reasonable adjustments and personal care needs NOMIS mappings' do
+    framework_responses = FrameworkResponse.where(id: [framework_response1.id, framework_response2.id])
+    framework_nomis_codes = framework_responses.flat_map(&:framework_nomis_codes)
+
+    expect { described_class.new(person: person, framework_responses: framework_responses, framework_nomis_codes: framework_nomis_codes).call }.to change(FrameworkNomisMapping, :count).by(6)
```
smoothcontract, line 103: My question before was around why this expects 6 mappings to be created? I'm guessing, perhaps it's because there are two nomis alerts, two reasonable adjustments and two personal care needs? Or maybe something else.. but a little explanation here would be helpful for future readers.

spec/services/framework_nomis_mappings/personal_care_needs_spec.rb
smoothcontract: 👍

spec/services/framework_nomis_mappings/importer_spec.rb
smoothcontract: Updated spec description makes sense, thanks for addressing 👍

smoothcontract (approved):
LGTM 👍

PR #1000, opened 2020-10-02
P4-2067 Import NOMIS mappings when creating the PER

cwrw:
### Jira link

[P4-2067](https://dsdmoj.atlassian.net/browse/P4-2067)

### What?

I have added/removed/altered:

- [x] Add NOMIS mapping serializer
- [x] Add NOMIS mappings includes and associations to relevant serializers
- [x] When creating the PER, import NOMIS mappings and associate them to the responses of the PER

### Why?

After responses are created, import NOMIS mappings. We can only import mappings if a move is associated to the PER and the from location is a prison.

Importing is not part of the initial transaction as it is a standalone action, which is not performant can be moved to a sidekiq job in the future. 

### Have you? (optional)

### DIFF
--- a/app/models/person_escort_record.rb
+++ b/app/models/person_escort_record.rb
@@ -52,6 +52,8 @@ class PersonEscortRecord < VersionedModel
     record = new(profile: profile, move: move, framework: framework)
     record.build_responses!
     record.import_nomis_mappings!
+
+    record
   rescue PG::UniqueViolation, ActiveRecord::RecordNotUnique
     record.errors.add(:profile, :taken)
     raise ActiveRecord::RecordInvalid, record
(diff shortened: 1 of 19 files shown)


app/models/person_escort_record.rb
```diff
@@ -74,6 +75,20 @@ def build_responses!
     self
   end
 
+  def import_nomis_mappings!
```
cwrw, line 78: If this does not perform very well, we can send to a sidekiq job, but this needs to be coordinated with the FE to show the correct messaging and polling

app/models/person_escort_record.rb
```diff
@@ -51,6 +51,7 @@ def self.save_with_responses!(profile_id: nil, version: nil, move_id: nil)
 
     record = new(profile: profile, move: move, framework: framework)
     record.build_responses!
+    record.import_nomis_mappings!
```
martyn-w, line 54: might be more readable as
```
record.import_nomis_mappings! if move&.from_location&.prison?
record
...
```

martyn-w (approved):
👍

app/models/person_escort_record.rb
cwrw: good suggestion I tidied it up slightly,  however l left the check in the method itself as wanted to encapsulate the import logic into the `import_nomis_mappings` method

martyn-w: 👍

PR #1001, opened 2020-10-02
Remove pager_api gem and replace with equivalent code

smoothcontract:
### Jira link

P4-2300

### What?

- [x] Remove `pager_api` gem and replace with simplified custom implementation

### Why?

- This gem doesn't play nicely with the proposed new faster jsonapi gem (it only supports AMS) and is no longer maintained, so reimplementing this within the PECS repo allows further customisation plus test coverage to be added. In addition, we can remove unused features of the gem such as the pagination header links and metadata links (links within the top level JSON payload are rendered via AMS so are not affected).

- `link` section under `pagination` `meta` key is removed as it is not used and does not adhere to JSONAPI specifications. Links to other pages are provided under the top level `links` key which is unaffected by this PR (although had no test coverage previously). See example responses:

Before: https://gist.github.com/smoothcontract/0339d0657ccafe2178f8cfa9322bc964
After: https://gist.github.com/smoothcontract/19e018675116da104a6c1328dcdd9a91

### Have you? (optional)

- Changes an api that is used in production. Have checked with Front end implementation and suppliers that the `meta.pagination.links` keys are not used.


app/controllers/concerns/paginatable.rb
```diff
@@ -0,0 +1,35 @@
+module Paginatable
```
smoothcontract, line 1: I'll likely be changing the name of this as part of the other work to remove AMS

martyn-w:
Geo have confirmed they are OK with this change as they are not using the `links`. Awaiting confirmation from the other supplier.

martyn-w:
Removing `breaking change` label as both suppliers have confirmed they are not using the links

app/controllers/concerns/paginatable.rb
```diff
@@ -0,0 +1,38 @@
...
+    options[:json] = paginated_collection
+    options[:meta] = meta(paginated_collection, options)
+
+    render options
+  end
+
+private
+
+  def paginate_collection(collection, options)
+    options[:page] = params[:page] || 1
```
cwrw, line 22: The links generated for `page` example: `http://www.example.com/api/v1/moves?page%5Bnumber%5D=1&page%5Bsize%5D=5` include: `page[number]=1&page[size]=5` which causes an error, instead it should be `page=1&per_page=5`.

The error I get is:
```
undefined method `to_i' for #<ActionController::Parameters:0x00007fdcb5932a30>
Did you mean?  to_s
               to_h excluded from capture: DSN not set

NoMethodError (undefined method `to_i' for #<ActionController::Parameters:0x00007fdcb5932a30>
Did you mean?  to_s
               to_h):

app/controllers/concerns/paginatable.rb:25:in `paginate_collection'
app/controllers/concerns/paginatable.rb:11:in `paginate'
app/controllers/api/v1/moves_actions.rb:4:in `index_and_render'
app/controllers/api/moves_controller.rb:8:in `index'
```
I'm not sure where those are generated?

smoothcontract: These come from the AMS gem.. I'll resolve and add coverage for that in the bigger PR. It probably makes sense to close this PR now - but I'll make sure this is covered off in the new PR.

smoothcontract:
Closing in favour of https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/1005 - I moved on a bit too far from this so unfortunately it's now bundled up with the wider serialization gem changes.

PR #1003, opened 2020-10-05
[P4-2304] Newly specced events (part 1)

willfish:
### Jira link

P4-2304

### What?

I have added/removed/altered:

- [x] Adds JourneyPersonBoardsVehicle event
- [x] Adds MoveOperationHMCTS event
- [x] Adds PerCourtExcessiveDelayNotDueToSupplier event
- [x] Adds PerCourtAllDocumentationProvidedToSupplier event
- [x] Adds PerCourtReturnToCustodyAreaFromDock event
- [x] Adds PerCourtCellShareRiskAssessment event
- [x] Adds PerCourtAssignCellInCustody event
- [x] Adds PerCourtReadyInCustody event
- [x] Plumbs together PersonEscortRecord as an eventable

### Why?

I am doing this because:

- These are needed by suppliers and the frontend for the generic event api

### Have you? (optional)

- [x] The event api wiki docs have been updated to include these

- These are not currently in use/plumbed in by suppliers so this is very low risk.


app/models/generic_event/per_court_cell_share_risk_assessment.rb
```diff
@@ -0,0 +1,15 @@
...
+    validates :location_id, presence: true
+
+    def location_id=(location_id)
+      details['location_id'] = location_id
+    end
+
+    def location_id
+      details['location_id']
+    end
+  end
```
smoothcontract, line 14: This looks like a good candidate for extracting to a shared concern perhaps - I see it used in a couple of events?

spec/models/generic_event/journey_admit_through_outer_gate_spec.rb
```diff
@@ -2,5 +2,6 @@
   subject(:generic_event) { build(:event_journey_admit_through_outer_gate) }
 
   it { is_expected.to validate_inclusion_of(:eventable_type).in_array(%w[Journey]) }
-  it { is_expected.to validate_presence_of(:vehicle_reg) }
+
+  it_behaves_like 'an event that specifies a vehicle registration'
```
smoothcontract, line 6: 👍

smoothcontract (commented):
Nice cleanup - I like the refactoring here to DRY up the implementation as more events are added (and there are so many!). Not sure if you want to introduce a pattern for common attributes across different events, but without validation as these seem to be optional in some cases (e.g. `location_id`) ? Maybe also worth moving the various concerns to a dedicated `events` module just for clarity as I doubt these would be useful outside of the events use case.

app/models/generic_event/per_court_cell_share_risk_assessment.rb
willfish: Very good idea 👍 .

smoothcontract (approved):
Discussed offline, moving further DRYing up to a separate PR for clarity, so this is good to go 👍

PR #1005, opened 2020-10-06
P4 2300 replace serializer gems

smoothcontract:
### Jira link

P4-2300

### What?

- [x] Remove `active_model_serializers` gem
- [x] Added `jsonapi-serializer` gem
- [x] Refactored all existing serializer classes to use new gem
- [x] Refactored references to other AMS classes throughout codebase
- [x] Removed `pager_api` gem
- [x] Added new `Serializable` concern to handle API pagination and JSONAPI rendering

### Why?

- The existing AMS gem is essentially deprecated and known to be pretty slow. The newer `jsonapi-serializer` gem is documented as 25x faster and is actively maintained. The hope is that this change will bring substantial performance improvements to most of the existing API endpoints, particularly where large amounts of data is returned.

- The `pager_api` gem doesn't play nicely with the new faster jsonapi gem (it only supports AMS) and is no longer maintained, so reimplementing this within the PECS repo allows further customisation plus test coverage to be added. In addition, we can remove unused features of the gem such as the pagination header links and metadata links.

- The AMS gem previously generated top level `links` section within the JSONAPI responses, but these links were in fact broken as the included page params could not be correctly parsed in the application and generated an error. This has now been resolved by a custom implementation of those top level links.

- The only breaking change introduced by this PR is the removal of `links` section under the `pagination` metadata section. Both suppliers and the front end devs have confirmed this metadata was not used so can be removed safely; it also does not adhere to JSONAPI specifications plus is redundant as pagination links are correctly included in the top level document. See example responses:

Before: https://gist.github.com/smoothcontract/0339d0657ccafe2178f8cfa9322bc964
After: https://gist.github.com/smoothcontract/19e018675116da104a6c1328dcdd9a91

### Have you? (optional)

- This is a wide reaching change affecting every API endpoint so will need extensive testing before release


app/controllers/concerns/serializable.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  def build_serializer(serializable, options)
+    serializer_class = options.delete(:serializer)
+    raise ArgumentError, 'serializer option must be specified' unless serializer_class
+
+    serializer_class.new(serializable, options.slice(:include, :fields, :meta, :links))
+  end
+
+  def paginate_collection(collection, options)
+    options[:page] = params[:page] || 1
```
willfish, line 39:
```suggestion
    options[:page] = params[:page] || FIRST_PAGE
```

smoothcontract: Good catch, I was planning to do that 👍

spec/requests/mock/pagination_controller_spec.rb
```diff
@@ -0,0 +1,167 @@
...
+            first: 'http://www.example.com/mock/data?page=1&per_page=5',
+            prev: nil,
+            next: 'http://www.example.com/mock/data?page=2&per_page=5',
+            last: 'http://www.example.com/mock/data?page=2&per_page=5',
+          },
+        )
+      end
+
+      it 'links to previous page' do
+        get '/mock/data?page=2'
```
smoothcontract, line 111:
```suggestion
        get '/mock/data', params: { page: 2}'
```

spec/requests/mock/pagination_controller_spec.rb
```diff
@@ -0,0 +1,167 @@
...
+            first: 'http://www.example.com/mock/data?page=1&per_page=5',
+            prev: 'http://www.example.com/mock/data?page=1&per_page=5',
+            next: nil,
+            last: 'http://www.example.com/mock/data?page=2&per_page=5',
+          },
+        )
+      end
+
+      it 'links to next page' do
+        get '/mock/data?per_page=2'
```
smoothcontract, line 125:
```suggestion
        get '/mock/data', params: { per_page=2}
```

app/serializers/document_serializer.rb
```diff
@@ -1,24 +1,27 @@
...
+class DocumentSerializer
+  include JSONAPI::Serializer
   include Rails.application.routes.url_helpers
   include ActionView::Helpers::NumberHelper
 
-  attributes :id, :url, :filename, :filesize, :content_type
+  set_type :documents
 
-  def url
+  attributes :id
```
willfish, line 10: I don't think we want the id in the attributes?

app/serializers/move_serializer.rb
```diff
@@ -19,15 +23,17 @@ class MoveSerializer < ActiveModel::Serializer
...
 
   has_one :from_location, serializer: LocationSerializer
   has_one :to_location, serializer: LocationSerializer
-  has_one :prison_transfer_reason, serializer: PrisonTransferReasonSerializer
-  has_many :documents, serializer: DocumentSerializer
-  has_many :court_hearings, serializer: CourtHearingSerializer
-  belongs_to :allocation, serializer: AllocationSerializer
+  has_one :prison_transfer_reason
+  has_many :documents do |object|
+    object.profile&.documents
```
willfish, line 33: 👍

app/serializers/move_serializer.rb
```diff
@@ -19,15 +23,17 @@ class MoveSerializer < ActiveModel::Serializer
              :date_from,
              :date_to
 
-  has_one :person, serializer: PersonSerializer
-  has_one :profile, serializer: ProfileSerializer # <- TODO: update the serializer
+  has_one :person
+  has_one :profile # <- TODO: update the serializer to V2
```
willfish, line 27: This comment is actually redundant now as we ended up creating a V2 move action which just uses the v2 move serializer. We'll just remove this when the suppliers are on v2

app/serializers/notification_serializer.rb
```diff
@@ -1,23 +1,20 @@
...
-  belongs_to :topic, polymorphic: true, key: 'move', if: :move? do |serializer|
-    link :self, serializer.move_url
-    object.topic
-  end
+  attribute :timestamp, &:created_at
 
-  def move?
-    object.topic.is_a?(Move)
-  end
+  belongs_to :topic, polymorphic: true, key: :move, if: proc { |object| move?(object) }, links: {
```
willfish, line 13: Smells like Notifications don't need to have a polymorphic topic. Could you add a method to the `Notification` model that points at the topic:

```ruby
def move
  topic if topic.is_a?(Move)
end
```
 or equivalent and tidy this up a bit? I don't think we need the test of move even since I'm guessing that jsonapi serializer doesn't try to serialize nil?

app/serializers/person_escort_record_serializer.rb
```diff
@@ -1,34 +1,33 @@
-class PersonEscortRecordSerializer < ActiveModel::Serializer
-  belongs_to :profile, record_type: :profile
```
willfish, line 2: Why the record_type, here? I can't seem to find anything in https://www.rubydoc.info/gems/active_model_serializers

willfish: Grepping for it in the ams repo doesn't yield anything, either:

```
william  11:26  ~/R/active_model_serializers  > ag record_type
william  11:26  ~/R/active_model_serializers  >
````

File: app/serializers/person_escort_record_serializer.rb [hunk 9]
```diff
@@ -1,34 +1,33 @@
...
   end
 
-  def status
-    if object.status == 'unstarted'
-      'not_started'
-    else
-      object.status
-    end
+  attribute :status do |object|
+    object.status == 'unstarted' ? 'not_started' : object.status
```
willfish, line 26: I think I prefer the non-ternary form 😛 . Suspect there's a brain difference between developers that prefer horizontal over vertical complexity. No pressure to change back as is style thing.

app/serializers/person_serializer.rb
```diff
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
-class PersonSerializer < ActiveModel::Serializer
+class PersonSerializer
+  include JSONAPI::Serializer
+
+  set_type :people
```
willfish, line 6: Given it's called the JSONAPI gem, I'm genuinely confused about why it doesn't infer the plural form for type 🤷 . Or has Dom been saying we always need plural forms for json:api when we don't 😆

app/serializers/person_serializer.rb
```diff
@@ -11,43 +15,23 @@ class PersonSerializer < ActiveModel::Serializer
...
-  def assessment_answers
+  attribute :assessment_answers do |object|
     object.latest_profile&.assessment_answers || []
   end
 
-  def identifiers
-    [police_national_computer, prison_number, criminal_records_office].compact
-  end
+  attribute :identifiers do |object|
+    %i[police_national_computer prison_number criminal_records_office].each_with_object([]) do |identifier, array|
```
willfish, line 23: How do you feel about using `acc` to indicate the idiomatic name for the accumulator in a fold function body?

app/serializers/prison_transfer_reason_serializer.rb
```diff
@@ -1,5 +1,9 @@
 # frozen_string_literal: true
 
-class PrisonTransferReasonSerializer < ActiveModel::Serializer
+class PrisonTransferReasonSerializer
+  include JSONAPI::Serializer
+
+  set_type :prison_transfer_reasons
+
   attributes :id, :title, :key, :disabled_at
```
willfish, line 8: More with the id in the attributes 🤦

app/serializers/timetable_serializer.rb
```diff
@@ -1,11 +1,12 @@
...
-  belongs_to :location, serializer: LocationSerializer
+  set_type :timetable_entries
 
-  def nomis_type
-    object.type
-  end
+  attributes :start_time, :reason
+  attribute :nomis_type, &:type
+
+  belongs_to :location, &:location
```
willfish, line 11: Not sure why we need to pass the proc with the same key name?

app/serializers/document_serializer.rb
smoothcontract: I wondered about this, it's included in most but not all of the existing serializers - although it shouldn't be needed according to the JSONAPI spec. I'll try removing `:id` from all of the serializers and see what happens 🤔

app/serializers/move_serializer.rb
smoothcontract: Fair enough, I'll remove the comment.

app/serializers/notification_serializer.rb
smoothcontract: Interesting idea, I'll have a look

app/serializers/person_escort_record_serializer.rb
smoothcontract: @cwrw any ideas? It did seem redundant to me so I removed it.

config/database.yml
```diff
@@ -6,7 +6,7 @@ default: &default
 
 development:
   <<: *default
-  database: pecs_move_platform_backend_development
+  database: pecs_staging
```
willfish, line 9: Accidentally committed?

config/initializers/jsonapi_serializer.rb
```diff
@@ -0,0 +1 @@
+ActiveSupport::JSON::Encoding.time_precision = 0
```
willfish, line 1: 🎉

app/serializers/person_serializer.rb
smoothcontract: Rather than `array` you mean?

app/serializers/prison_transfer_reason_serializer.rb
smoothcontract: Yep, but note this example is unchanged so has been around for a while. I do agree it's likely not needed so will aim to clean these all up.

app/serializers/timetable_serializer.rb
smoothcontract: This one is fun - the timetable model doesn't use a traditional `belongs_to :location` relationship, but instead has a custom `location` method. The serializer likes to use `location_id` by default to resolve associations which isn't implemented on the timetable model. So this forces use of the custom `location` method instead - alternatively I could add an implementation for `location_id` on the timetable model. There's another Nomis model (court cases maybe) that has the same pattern.

config/database.yml
smoothcontract: Oh yes, that's a good catch. Will revert.

spec/requests/api/allocations_controller_index_spec.rb
```diff
@@ -88,11 +88,15 @@
...
-          links: {
-            first: '/api/v1/allocations?page=1',
-            last: '/api/v1/allocations?page=2',
-            next: '/api/v1/allocations?page=2',
-          },
+        }
+      end
+      let(:pagination_links) do
+        {
+          self: 'http://www.example.com/api/v1/allocations?page=1&per_page=5',
```
willfish, line 95: Are we okay with fully qualified over relative links? I presume this has been discussed somewhere with Dom et al. Probably I missed it

spec/requests/mock/pagination_controller_spec.rb
```diff
@@ -0,0 +1,167 @@
+# frozen_string_literal: true
```
willfish, line 1: 🎉

spec/requests/api/allocations_controller_index_spec.rb
smoothcontract: These were previously fully qualified with the AMS implementation (don't confuse them with the relative links that were in the metadata section, these are in the top level of the document). Also the links were previously broken with erroneous `page[number]&page[per_page]` parameters - thanks AMS - so this PR fixes that. The front end doesn't actually look at the content of the links anyway, only for presence of those e.g. https://github.com/ministryofjustice/hmpps-book-secure-move-frontend/blob/master/common/services/move.js#L33

The JSONAPI spec seems to suggest full URL's in the example but it's not made explicit as far as I can tell: https://jsonapi.org/format/#document-links

TL:DR; Yes 😄

spec/serializers/person_serializer_spec.rb
```diff
@@ -134,7 +132,7 @@
       end
 
       it 'does not contain an included ethnicity' do
-        expect(result[:included]).to be_nil
+        expect(result[:included]).to be_empty
```
willfish, line 135: This is different behaviour?

smoothcontract: Yes, it returns an empty array now when there are no includes. I don't think that's a big problem but maybe worth keeping an eye on when testing.

swagger/v1/pagination.yaml
```diff
@@ -10,14 +10,3 @@ Pagination:
     total_objects:
       type: integer
       example: 1977
-    links:
```
willfish, line 13: 👍

app/serializers/person_serializer.rb
willfish: Yeah

willfish: No big pressure :)

app/serializers/timetable_serializer.rb
willfish: Wow 🤯

willfish: Okay, well in which case this is a clean solution

spec/requests/api/allocations_controller_index_spec.rb
willfish: Thanks Andrew. Very clear.

spec/serializers/person_serializer_spec.rb
willfish: Yeah. Depending on how clients are coded this may be impactful. Would be good to have a full UAT session with both suppliers to make sure.

willfish:
So my instinct when I thought about this work was to start with full document tests for all of the serializers before I even touched the actual gem. We do things in our tests like `only check these specific attributes`. This framework change could have subtle differences that aren't currently caught by our subset tests.

I think a decent substitute is rolling this out to dev/staging (unfortunately we can't roll out to UAT) and get both suppliers to do a full integration test to validate they get what they expect. If they haven't actually automated their integration tests I guess we're in a position where our fullest test is from the e2e tests. Maybe the best we can get from suppliers is for them to do some manual testing against the major models (move, per, profile, person) and call it a day with testing.

app/serializers/person_serializer.rb
smoothcontract: It is weird, but without this line in each serializer it does indeed infer the singular form from the name, i.e. this would populate a type of `person`. c'est la vie.

app/serializers/person_serializer.rb
smoothcontract: I think I prefer the name `array` TBH

app/serializers/person_serializer.rb
willfish: Yeah. Crazy, but fast, JSONAPI gem

app/serializers/person_serializer.rb
willfish: Okey poke

app/models/profile/assessment_answer.rb
```diff
@@ -39,23 +43,6 @@ def empty?
       assessment_question_id.blank?
     end
 
-    def as_json(_options = {})
```
smoothcontract, line 42: This comes for free now with the `ActiveModel::Model` module 😁

app/serializers/notification_serializer.rb
smoothcontract: I tried this but it didn't gain much - the `proc` is still needed to handle the case of an optional `move` relationship so it ends up polluting the model with serializer concerns. I think I'll leave the existing implementation instead.

willfish: Thanks for trying :)

willfish (approved):
This is just awesome and makes me really happy. You're actually a ninja/robot/super turtle all rolled into one

cwrw (commented):
wow pretty intense, well done! also very thorough and nice cleanup of existing things. I've left a few minor comments, but I'm going to do some testing tomorrow to confirm further but otherwise 👏

app/controllers/concerns/serializable.rb
```diff
@@ -0,0 +1,75 @@
...
+
+private
+
+  FIRST_PAGE = 1
+
+  def build_serializer(serializable, options)
+    serializer_class = options.delete(:serializer)
+    raise ArgumentError, 'serializer option must be specified' unless serializer_class
+
+    serializer_class.new(serializable, options.slice(:include, :fields, :meta, :links))
```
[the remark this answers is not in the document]
smoothcontract: If the serializer key is omitted it throws an error - I didn't add coverage for this but could if it feels useful. I just wanted to ensure that during local development we'd spot easily if the key was missed. If the passed `serializable` object or collection doesn't match the serializer class then we'll get a myriad of errors depending on the exact combination. I'm not sure that's hugely valuable to test though.

app/models/court_case.rb
```diff
@@ -19,6 +19,6 @@ def build_from_nomis(court_case)
   end
 
   def location
-    Location.find_by nomis_agency_id: @agency_id
+    @location ||= Location.find_by nomis_agency_id: @agency_id
```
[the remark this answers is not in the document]
smoothcontract: It'll be called by the serializer, but not sure how many times. It just looked like a good candidate to memoize whilst passing.

app/serializers/court_case_serializer.rb
```diff
@@ -1,25 +1,16 @@
...
-  end
+  attribute :nomis_case_id, &:case_id
+  attribute :nomis_case_status, &:case_status
+  attribute :case_start_date, &:begin_date
+  attribute :case_number, &:case_info_number
 
-  def case_number
-    object.case_info_number
-  end
+  belongs_to :location, &:location
```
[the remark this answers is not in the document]
smoothcontract: This follows the same pattern as `Timetable` serializer - pls see the discussion https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/1005/files#diff-c1ca135968dcc9ccb19a1f757510150eR11

app/serializers/framework_flag_serializer.rb
```diff
@@ -1,5 +1,11 @@
-class FrameworkFlagSerializer < ActiveModel::Serializer
-  belongs_to :framework_question, key: :question
+# frozen_string_literal: true
+
+class FrameworkFlagSerializer
+  include JSONAPI::Serializer
+
+  set_type :framework_flags
+
+  belongs_to :question, serializer: FrameworkQuestionSerializer, &:framework_question
```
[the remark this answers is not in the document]
smoothcontract: Yes, that's where I started but it doesn't work properly with `includes` - basically the key isn't recognised as a valid `includes` so I had to go with this. That feels a bit like a bug with the serializer but this workaround works okay.

spec/models/profile/assessment_answer_spec.rb
```diff
@@ -230,7 +224,7 @@
     it 'instantiates an expected AssessmentAnswer' do
       result = described_class.from_nomis_personal_care_need(personal_care_need, assessment_question, alert_type_description)
 
-      expect(result.as_json).to eq(expected_assessment_answer)
+      expect(result.as_json.symbolize_keys).to eq(expected_assessment_answer)
```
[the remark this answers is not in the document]
smoothcontract: This shouldn't be an issue in practice - JSON representation doesn't have the concept of symbols so won't affect any consumers of the API. There was some inconsistency in the serializer testing throughout the specs - in some cases the serializable_hash output was tested - which returns ruby objects, rather than generating the full JSON representation and then parsing it (a more thorough test). I've tried to make this consistent throughout the tests by always rendering to and parsing from JSON. The conversion of keys to symbols or strings is then just done to minimise noise in the spec changes.

spec/serializers/person_serializer_spec.rb
smoothcontract: I explored this further - it only returns an empty `included` array when the `include` option was passed to the serializer but no valid resources were found to include. In this case we're requesting to include `ethnicity` but then explicitly setting the person ethnicity to `nil`. The output of an empty `included` array seems reasonable in that case, but is a change in behaviour from AMS.

spec/serializers/prison_transfer_reason_serializer_spec.rb
```diff
@@ -26,6 +26,6 @@
   end
 
   it 'contains a `disabled_at` attribute' do
-    expect(result[:data][:attributes][:disabled_at]).to eql disabled_at
+    expect(result[:data][:attributes][:disabled_at]).to eql disabled_at.iso8601
```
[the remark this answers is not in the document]
smoothcontract: That's because previously this spec wasn't testing a full round trip of the serializer (converting to JSON and then parsing back), but instead inspecting Ruby objects in the `serializable_hash`. I've tried to make all of the serializer specs more complete and representative by always generating JSON output and then parsing.

That does mean in some cases (depending on the individual spec) I've had to tweak expectations slightly to check the correct object type is returned - e.g. here we do indeed want dates in `YYYY-MM-DD` format for the JSON output.

spec/serializers/v2/person_serializer_spec.rb
```diff
@@ -64,13 +62,13 @@
       let(:person) { create(:person, ethnicity: nil) }
 
       it 'does not contain an included ethnicity' do
-        expect(result[:included]).to be_nil
+        expect(result[:included]).to be_empty
```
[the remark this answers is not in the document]
smoothcontract: See above thread on this topic - it's just a very slight behaviour change with the new serializer but I don't think will be significant.

spec/services/personal_care_needs/importer_spec.rb
```diff
@@ -42,7 +42,7 @@
     it 'appends an assessment answer with the default values' do
       importer.call
 
-      expect(profile.assessment_answers.first.as_json).to include(
+      expect(profile.assessment_answers.first.as_json.symbolize_keys).to include(
```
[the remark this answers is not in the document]
smoothcontract: `AssessmentAnswers` previously had a custom implementation of `as_json` which (incorrectly) returned all attributes with symbolized keys. The `ActiveModel::Model` implementation instead (correctly) returns stringified keys. Keep in mind that `as_json` is then rendered with `to_json` so the symbol/string distinction completely disappears anyway. I've added the `symbolize_keys` call here to avoid having to change each of the following attributes in the spec.

cwrw (approved):
👏

PR #1008, opened 2020-10-07
[P4-2304] DRY up location validation in generic events

willfish:
### Jira link

P4-2304

### What?

I have added/removed/altered:

- [x] DRY up location validations in generic event STI models

### Why?

I am doing this because:

- DRYer is cleaner
- DRYer means fewer places to change


spec/requests/api/generic_events_controller_create_spec.rb
```diff
@@ -69,6 +69,14 @@
               },
               'code' => 'blank',
             },
+            {
+              'title' => 'Unprocessable entity',
+              'detail' => 'From location id the location relationship you passed has an id that does not exist in our system. please use an existing from location',
+              'source' => {
+                'pointer' => '/data/attributes/from_location_id',
```
willfish, line 76: So I've noticed relationship validations always point at the attributes section of the document when they have their separate section in the request as per the json:api spec. Happy to update this separately using magic and a bit of patience.

spec/models/generic_event/move_redirect_spec.rb
```diff
@@ -14,6 +14,8 @@
 
   it_behaves_like 'a move event'
 
+  it_behaves_like 'an event requiring a location', :to_location_id
+
   it { is_expected.to validate_presence_of(:to_location_id) }
```
cwrw, line 19: this can now be removed

spec/support/event_requiring_a_location.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    it 'attaches a custom message to the generic event' do
+      expect { generic_event.valid? }
+        .to change { generic_event.errors[location_key] }
+        .from([])
+        .to(["The location relationship you passed has an id that does not exist in our system. Please use an existing #{named_location}"])
+    end
+  end
+
+  it "validates #{location_key} existing in the database" do
```
cwrw, line 23: isn't this a duplicate test on line 13?

spec/support/event_requiring_a_location.rb
```diff
@@ -0,0 +1,28 @@
...
+  context "when the #{location_key} does not exist" do
+    let(:named_location) { location_key.to_s.sub('_id', '') }
+
+    before do
+      generic_event.public_send("#{location_key}=", 'flibble')
+    end
+
+    it { is_expected.not_to be_valid }
+
+    it 'attaches a custom message to the generic event' do
```
cwrw, line 15: instead of having this context, maybe we can have one test for the id not existing in the database instead:
```
it "validates #{location_key} existing in the database" do
  generic_event.public_send("#{location_key}=", 'flibble')

  expect(generic_event).not_to be_valid
  expect(generic_event.errors[location_key]).to eq(['The location relationship you passed has....'])
end

```

app/models/concerns/location_validations.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    sti_klass.define_method("#{location_key}=") do |location_id|
+      details[location_key] = location_id
+    end
+
+    validates location_key, presence: true
+    validates_each location_key do |record, attr, value|
+      Location.find(value)
+    rescue ActiveRecord::RecordNotFound
+      record.errors.add(attr, "The location relationship you passed has an id that does not exist in our system. Please use an existing #{location_key.to_s.sub('_id', '')}")
```
cwrw, line 21: Currently for all other instances of `ActiveRecord::RecordNotFound` in the database, we fallback to this error on the API response: https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/controllers/api_controller.rb#L103
To maintain consistency, we should not add a custom validation/error message here, but allow the API controller to handle that error as default

spec/support/event_requiring_a_location.rb
cwrw: On revisiting this with using the `ActiveRecord::RecordNotFound` default error message instead we can test this as follows:
```
  it "validates #{location_key} existing in the database" do
    generic_event.public_send("#{location_key}=", 'flibble')
    
      expect { generic_event.valid? }.to raise_error(ActiveRecord::RecordNotFound, "Couldn't find Location with 'id'=flibble")
  end
```

app/models/concerns/location_validations.rb
willfish: I had debated this myself, actually. I ended up thinking I could give a clearer message to the user in the event of the value not existing. I think on balance the better solution is to be consistent/cause fewer questions about what's going on. It's actually cleaner code, too.

spec/models/generic_event/journey_lockout_spec.rb
```diff
@@ -74,32 +75,5 @@
         expect(described_class.from_event(event)).to be_valid
       end
     end
-
-    context 'when the event does not have locations' do
```
willfish, line 78: This context was verifying the process of building a v2 event from a v1 event where there was a missing location. 

Removing this is fine because if there were events created without a location in the v1 world we would have exploded as the data was coming in: https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/dc6a4b3c88eac51587d802ad3269cd39224a2402/app/services/journey_events/params_validator.rb#L15-L16.

For V1 event endpoints this does reflect that these validations which now exist in the model are now duplicated in both the event params validation services (https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/dry_up_locations/app/services/journey_events/params_validator.rb and https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/dry_up_locations/app/services/move_events/params_validator.rb).

This is fine since we need to handle this behaviour in both kinds of endpoint. We could actually probably just remove the old way of doing it, though probably in a separate PR.

spec/models/generic_event/journey_lodging_spec.rb
```diff
@@ -74,32 +75,5 @@
         expect(described_class.from_event(event)).to be_valid
       end
     end
-
-    context 'when the event does not have locations' do
```
willfish, line 78: This context was verifying the process of building a v2 event from a v1 event where there was a missing location. 

Removing this is fine because if there were events created without a location in the v1 world we would have exploded as the data was coming in at the controller level: https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/dc6a4b3c88eac51587d802ad3269cd39224a2402/app/services/journey_events/params_validator.rb#L15-L16.

For V1 event endpoints this does reflect that these validations which now exist in the model are now duplicated in both the event params validation services (https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/dry_up_locations/app/services/journey_events/params_validator.rb and https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/dry_up_locations/app/services/move_events/params_validator.rb).

This is fine since we need to handle this behaviour in both kinds of endpoint. We could actually probably just remove the old way of doing it, though probably in a separate PR.

cwrw (approved):

PR #1009, opened 2020-10-07
P4-2301 Add NOMIS sync status to Person Escort Record

cwrw:
### Jira link

[P4-2301](https://dsdmoj.atlassian.net/browse/P4-2301)

### What?

- [x] Added new `nomis_sync_status` to Person escort record
- [x] Record status of each NOMIS import when creating the PER
- [x] Surface new jsonb column in Person escort record serializer

### Why?
To surface any NOMIS resource importing failures on PER creation, add a new column on the PER: 'nomis_sync_status' which lists all the resources imported, and their status, timestamp as well as error message if available. This will allow the FE to render a warning message if any NOMIS resources have not been successfully associated to the responses, and to allow the user to check any missing resources in NOMIS itself instead.

### Have you? (optional)

- Changes an api that is used in production, but importing is still not available in production until all features are released

### DIFF
--- a/app/services/framework_nomis_mappings/alerts.rb
+++ b/app/services/framework_nomis_mappings/alerts.rb
@@ -16,12 +16,7 @@ module FrameworkNomisMappings
   private
 
     def imported_alerts
-      @imported_alerts ||= begin
-        alerts = NomisClient::Alerts.get([prison_number])
-        nomis_sync_status.set_success
-
-        alerts
-      end
+      @imported_alerts ||= NomisClient::Alerts.get([prison_number]).tap { nomis_sync_status.set_success }
     rescue Faraday::ConnectionFailed, Faraday::TimeoutError, OAuth2::Error => e
       Rails.logger.warn "Importing Framework alert mappings Error: #{e.message}"
       nomis_sync_status.set_failure(message: e.message)
--- a/app/services/framework_nomis_mappings/alerts.rb
+++ b/app/services/framework_nomis_mappings/alerts.rb
@@ -1,9 +1,10 @@
 module FrameworkNomisMappings
   class Alerts
-    attr_reader :prison_number
+    attr_reader :prison_number, :nomis_sync_status
 
-    def initialize(prison_number:)
+    def initialize(prison_number:, nomis_sync_status: FrameworkNomisMappings::NomisSyncStatus.new(resource_type: 'alerts'))
       @prison_number = prison_number
+      @nomis_sync_status = nomis_sync_status
     end
 
     def call
@@ -15,9 +16,15 @@ module FrameworkNomisMappings
   private
 
     def imported_alerts
-      @imported_alerts ||= NomisClient::Alerts.get([prison_number])
+      @imported_alerts ||= begin
+        alerts = NomisClient::Alerts.get([prison_number])
+        nomis_sync_status.set_success
+
+        alerts
+      end
     rescue Faraday::ConnectionFailed, Faraday::TimeoutError, OAuth2::Error => e
       Rails.logger.warn "Importing Framework alert mappings Error: #{e.message}"
+      nomis_sync_status.set_failure(message: e.message)
 
       []
     end
--- a/app/services/framework_nomis_mappings/importer.rb
+++ b/app/services/framework_nomis_mappings/importer.rb
@@ -2,16 +2,17 @@
 
 module FrameworkNomisMappings
   class Importer
-    attr_reader :person, :framework_responses, :framework_nomis_codes
+    attr_reader :person, :person_escort_record, :framework_responses, :framework_nomis_codes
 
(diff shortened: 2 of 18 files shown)


app/services/framework_nomis_mappings/importer.rb
```diff
@@ -20,14 +21,24 @@ def call
...
         end
+
+        person_escort_record.update(nomis_sync_status: nomis_sync_status)
       end
     end
 
   private
 
+    def nomis_sync_status
+      [
```
cwrw, line 32: I'll set up sending failures to Sentry in next PR

swagger/v1/person_escort_record.yaml
```diff
@@ -33,6 +33,29 @@ PersonEscortRecord:
...
+                type: string
+                enum:
+                  - success
+                  - failed
+              synced_at:
+                type: string
+                format: date-time
+              message:
+                type: string
+          readOnly: true
```
willfish, line 57: 👍

willfish (approved):
So I can't see anything obviously wrong with the code and it does what you say it does.

app/services/framework_nomis_mappings/alerts.rb
```diff
@@ -15,9 +16,15 @@ def call
   private
 
     def imported_alerts
-      @imported_alerts ||= NomisClient::Alerts.get([prison_number])
+      @imported_alerts ||= begin
+        alerts = NomisClient::Alerts.get([prison_number])
+        nomis_sync_status.set_success
+
+        alerts
+      end
```
smoothcontract, line 24: You could get this on one line if you prefer (I think).. something like this

```suggestion
  @imported_alerts = NomisClient::Alerts.get([prison_number]).tap { nomis_sync_status.set_success }
      end
```

Not sure if that's better or not 🤔

app/services/framework_nomis_mappings/nomis_sync_status.rb
```diff
@@ -0,0 +1,32 @@
...
+      @synced_at = Time.zone.now
+    end
+
+    def set_failure(message: nil)
+      @status = FAILED
+      @synced_at = Time.zone.now
+      @message = message
+    end
+
+    def as_json(_options = {})
```
smoothcontract, line 23: You could remove this method and include the `ActiveModel::Model` module instead which would give the same thing for free. Only difference is that the attributes would have stringified names rather than symbols, but I think this is expected behaviour - or at least is how conventional models work.

spec/services/framework_nomis_mappings/nomis_sync_status_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+      )
+    end
+
+    it 'returns the attributes of nomis sync status if set as json' do
+      nomis_sync_status = described_class.new(resource_type: 'some_resource')
+      sync_at_timstamp = Time.zone.now
+      allow(Time).to receive(:now).and_return(sync_at_timstamp)
+      nomis_sync_status.set_success
+
+      expect(nomis_sync_status.as_json).to eq(
```
smoothcontract, line 67: If you do remove the custom `as_json` implementation as mentioned above, this should still pass if you add `.symbolize_keys`

swagger/v1/person_escort_record.yaml
```diff
@@ -33,6 +33,29 @@ PersonEscortRecord:
...
+        nomis_sync_status:
+          type: array
+          items:
+            type: object
+            properties:
+              resource_type:
+                type: string
+                enum:
+                  - alerts
+                  - personcal_care_needs
```
smoothcontract, line 45: `personal_care_needs`

smoothcontract (commented):
Looks good, just some minor things I spotted...

app/services/framework_nomis_mappings/nomis_sync_status.rb
cwrw: talked in person: this approach will hide keys with no values, so agreed to keep current implementation

app/services/framework_nomis_mappings/personal_care_needs.rb
```diff
@@ -16,9 +17,15 @@ def call
   private
 
     def imported_personal_care_needs
-      @imported_personal_care_needs ||= NomisClient::PersonalCareNeeds.get(nomis_offender_numbers: [prison_number], personal_care_types: PERSONAL_CARE_NEED_CODES)
+      @imported_personal_care_needs ||= begin
+        personal_care_needs = NomisClient::PersonalCareNeeds.get(nomis_offender_numbers: [prison_number], personal_care_types: PERSONAL_CARE_NEED_CODES)
```
smoothcontract, line 21: You could use the same `.tap {}` pattern here if you want, but by no means essential and could be done later if you prefer.

smoothcontract (approved):
LGTM - thanks for amending/discussing 👍

PR #1012, opened 2020-10-07
[P4-2304] Adds optional redirect_reasons to the MoveRedirect STI model

willfish:
### Jira link

P4-2304

### What?

I have added/removed/altered:

- [x] Adds redirect reasons enum to MoveRedirect model

### Why?

I am doing this because:

- This is required  by the programme

### Have you? (optional)

- [x] Updated event wiki API docs


martyn-w (approved):
👍

PR #1017, opened 2020-10-08
Sort in memory and add eager load profiles

cwrw:
Avoid doing extra queries which ignore eager loading by sorting in memory, also add missing includes to allocation endpoint. Locally this cut down API calls from over 8seconds to 1.7seconds for 100 allocations


martyn-w (approved):
👍

PR #1019, opened 2020-10-08
Ruby 2.7.2

willfish:
### Jira link

P4-<TODO>

### What?

I have added/removed/altered:

- [x] Bumps ruby 2.7.1 -> 2.7.2

### Why?

This update bundles a more secure version of [webrick](https://www.ruby-lang.org/en/news/2020/09/29/http-request-smuggling-cve-2020-25613/) and disables noisy deprecation warnings - we've already fixed this by
upgrading all of our gems.


cwrw (approved):

willfish:
@martyn-w Are we still preferring that these changes wait until Monday before merging? I think it was Monday. It's actually been a while 🙂

martyn-w:
Tuesdays, when we start the new sprint, if that's ok?

willfish:
Totally fine 🙂.

PR #1020, opened 2020-10-08
Fix diagnostics endpoint

willfish:
### Jira link

P4-<TODO>

### What?

I have added/removed/altered:

- [x] Make diagnostics endpoint pull out GenericEvent rows for journeys and moves

### Why?

See https://sentry.io/organizations/ministryofjustice/issues/1944589456/?project=2014776&referrer=slack

### DIFF
--- a/app/services/diagnostics/move_inspector.rb
+++ b/app/services/diagnostics/move_inspector.rb
@@ -36,11 +36,11 @@ module Diagnostics
         -----------
       ENDMOVEEVENTS
 
-      if move.events.any?
+      if move.generic_events.any?
         @output << "#{'EVENT'.ljust(15)}\t#{'TIMESTAMP'.ljust(27)}\tPARAMS\n"
-        move.events.default_order.each do |event| # NB use each to preserve sort order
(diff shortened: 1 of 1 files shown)


martyn-w:
Not 100% sure but from what I can tell I think this problem might be caused by new events having a `MoveEvent` set in the type field - as opposed to nil. I think the same bug may also affect `lib/tasks/fake_data/journeys.rb` ? Maybe best to fix the event creator to force the type to be nil?

martyn-w (approved):

willfish:
Just for clarity on what I think was happening here:

1. We added type to the Event model (when we were going to add STI to the existing model).
2. Where in the Eventable#process_event module we were creating move_events with `move.move_events.create` this automatically was adding the MoveEvent type to the model and instantiating a `MoveEvent`
3. When reading these events out with `Event.all` or `Move.move_events` the STI factory internal to ActiveRecord was automatically instantiating Event records with a type field of MoveEvent as a MoveEvent model.
4. Removing the MoveEvent model without upgrading the final reference to it in the diagnostics endpoint led to an exception because the type was missing.

PR #1022, opened 2020-10-09
[P4-2304] Newly specced events

willfish:
### Jira link

P4-2304

### What?

I have added/removed/altered:

- [x] Adds PerCourtPreReleaseChecksCompleted event
- [x] Adds PerCourtRelease event
- [x] Adds PerCourtReleaseOnBail event
- [x] Adds PerCourtReturnToCustodyAreaFromVisitorArea event
- [x] Adds PerCourtTakeFromCustodyToDock event
- [x] Adds PerCourtTakeToSeeVisitors event
- [x] Adds PerCourtTask event
- [x] Adds PerGeneric event
- [x] Adds PerMedicalAid event
- [x] Adds PerPrisonerWelfare event

### Why?

I am doing this because:

- These are needed for various business reasons as part of the generic event api

### Have you? (optional)


app/models/generic_event/per_medical_aid.rb
```diff
@@ -0,0 +1,74 @@
...
+    include LocationValidations
+    include PersonEscortRecordEventValidations
+    include SupplierPersonnelNumberValidations
+
+    validates :advised_at, presence: true
+    validates :advised_by, presence: true
+    validates :treated_at, presence: true
+    validates :treated_by, presence: true
+
+    validates_each :advised_at do |record, attr, value|
```
smoothcontract, line 22:
```suggestion
    validates_each :advised_at, :treated_at do |record, attr, value|
```

app/models/generic_event/per_medical_aid.rb
```diff
@@ -0,0 +1,74 @@
...
+      Time.zone.iso8601(value)
+    rescue ArgumentError
+      record.errors.add(attr, 'must be formatted as a valid ISO-8601 date-time')
+    end
+
+    validates_each :treated_at do |record, attr, value|
```
smoothcontract, line 28: This validation can be removed and replaced with above suggested change :)

app/models/generic_event/per_medical_aid.rb
```diff
@@ -0,0 +1,74 @@
...
+      Time.zone.iso8601(value)
+    rescue ArgumentError
+      record.errors.add(attr, 'must be formatted as a valid ISO-8601 date-time')
+    end
+
+    def advised_at=(advised_at)
```
smoothcontract, line 34: Hmm... these events have a very common pattern for getters and setters with the details hash - seems like a perfect use case for some meta programming which would remove all of these boiler plate getters and setters 🤔

smoothcontract (approved):
Really clean implementation, nice job. The DRYing up of location validation you did previous has really been worthwhile 🥇 

I could only spot one immediate potential improvement, but moving forwards it might be nice to consider using a meta programming approach for all the details methods on each event, rather than manually adding a getter and setter method for each one.

So rather than...

```ruby
  def given_at=(given_at)
      details['given_at'] = given_at
    end

    def given_at
      details['given_at']
    end

    def outcome=(outcome)
      details['outcome'] = outcome
    end

    def outcome
      details['outcome']
    end

    def subtype=(subtype)
      details['subtype'] = subtype
    end

    def subtype
      details['subtype']
    end
```

this...

```ruby
details_attributes :given_at, :outcome, :subtype
```

app/models/generic_event/per_medical_aid.rb
willfish: I was having the same thought process, actually. I'll look at sorting this in another PR :).

PR #1023, opened 2020-10-10
P4-2211 Log exceptions when NOMIS mappings fail or new NOMIS codes detected

cwrw:
jira-ticket: https://dsdmoj.atlassian.net/browse/P4-2211

Log a Sentry error if NOMIS mappings are not persisted in bulk, to surface any validation issues when importing NOMIS resources. Also log any new NOMIS codes we receive to surface those values to be actioned and added to our system.

### DIFF
--- a/app/services/framework_nomis_mappings/importer.rb
+++ b/app/services/framework_nomis_mappings/importer.rb
@@ -39,8 +39,9 @@ module FrameworkNomisMappings
     def persist_framework_nomis_mappings
       @persist_framework_nomis_mappings ||= begin
         mappings = alert_mappings.call + personal_care_need_mappings.call + reasonable_adjust_mappings.call
-        # TODO: log any validation failures
         import = FrameworkNomisMapping.import(mappings, all_or_none: true)
+
+        log_exception('FrameworkNomisMapping import validation Error', import.failed_instances.map { |mapping| mapping.errors.messages }) if import.failed_instances.any?
 
         FrameworkNomisMapping.where(id: import.ids)
       end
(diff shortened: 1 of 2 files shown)


app/services/framework_nomis_mappings/importer.rb
```diff
@@ -89,5 +91,16 @@ def nomis_code_ids_to_mappings
     def responses_to_codes
       @responses_to_codes ||= framework_responses.joins(:framework_nomis_codes).select('framework_responses.id as id, framework_nomis_codes.id as nomis_code_id').group_by(&:id)
     end
+
+    def log_exception(description, params)
+      Raven.capture_message(
+        description,
+        extra: {
+          id: person_escort_record.id,
```
martyn-w, line 99: Could `person_escort_record` ever be nil her? I see in the initialize method it is possible. So maybe use `person_escort_record&.id` instead ?

martyn-w (approved):

app/services/framework_nomis_mappings/importer.rb
```diff
@@ -81,6 +82,7 @@ def nomis_code_ids_to_mappings
             end
           elsif mapping_nomis_fallback
             hash[mapping_nomis_fallback.id] = hash[mapping_nomis_fallback.id].to_a + [mapping]
+            log_exception('New NOMIS codes imported', [{ code: mapping.code, type: mapping_nomis_fallback.code_type }])
```
willfish, line 85: Should we only warn for new codes?

willfish (approved):

app/services/framework_nomis_mappings/importer.rb
cwrw: We might need to add custom config to pickup warnings in slack, so best stick to errors for now

PR #1024, opened 2020-10-10
P4-2211 Details optional if NOMIS mappings present on response

cwrw:
jira-ticket: https://dsdmoj.atlassian.net/browse/P4-2211

Do not validate comments on responses if there are any NOMIS mappings attached to the response. To avoid users having to fill in "see above" in required give details boxes when NOMIS mappings are available, make this box optional.


martyn-w (approved):
👍 LGTM

willfish (approved):

PR #1025, opened 2020-10-12
Updates allocation_serializer to use v2 moves

willfish:
### Jira link

P4-

### What?

I have added/removed/altered:

- [x] Altered `AllocationsSerializer` so that it now uses `V2::MoveSerializer`

### Why?

I am doing this because:

- The old serializer pulls on the now-replaced person#latest_profile method which is going to be removed


solidgoldpig (approved):

willfish:
Testing this with @solidgoldpig now

PR #1027, opened 2020-10-12
Removes allocation versioning

willfish:
### Jira link

P4-<TODO>

### What?

I have added/removed/altered:

- [x] Removes versioning in allocations endpoints

### Why?

I am doing this because:

- This versioning had nothing to do with serializers and was only related to whether we supported default includes or not
- Since only the frontend uses this and it's on v2, let's just remove versioning.

### Have you? (optional)


willfish:
We never documented the version difference between v1/v2 allocation includes so nothing to change in swagger docs.

app/controllers/api/v1/allocations_actions.rb
```diff
@@ -1,7 +0,0 @@
-module Api::V1
-  module AllocationsActions
-    def included_relationships
-      IncludeParamHandler.new(params).call || AllocationSerializer::SUPPORTED_RELATIONSHIPS
```
smoothcontract, line 4: Argh!! This is the file that caused us so much grief when testing - it defaults to returning all of the includes if none were specified. Kill it now. Kill it with fire!

smoothcontract (approved):
So nice to see this cleaned up and V2 behaviour get promoted 👏

app/controllers/api/v1/allocations_actions.rb
willfish: 💥

PR #1029, opened 2020-10-12
Add new model and endpoint to return population free spaces

smoothcontract:
### Jira link

P4-2196

### What?

- [x] Adds new `GET /locations_free_spaces` endpoint to support PMU dashboard showing free spaces by location for a given date range
- [x] Extends Locations::Finder service to support filtering by region id and location id, and also adds sorting support.
- [x] Add new `fake_data:create_populations` rake task to populate test data for populations.

### Why?

- The new endpoint makes use of custom `meta` section to provide just enough data to render the front end screen without the additional overhead of returning all associated `Population` resources. Free space count is calculated as Postgres level rather than in Ruby which makes the endpoint fast enough to be used in real time for a lengthy list of locations.
- Custom serializer (based on Location) is used to support sparse attributes (only title is needed for display), custom metadata and omits supplier relationship. This is very performant and avoids any queries for unnecessary data.
- Pagination and additional location filtering supports front end requirements to select appropriate list of locations to display

### Have you? (optional)

- [ ] Updated API docs if necessary - Undecided yet whether new endpoint will be documented as it is intended only for front end use with PMU

- New API endpoint so minimal risk from that
- Default sort order of ascending title is also introduced to locations reference data endpoint. Need to confirm whether this has any material impact on front end or suppliers.


app/models/population.rb
```diff
@@ -0,0 +1,56 @@
...
+          locations_hash[location_id] << free_space_details
+        end
+      end
+    end
+  end
+
+  def self.free_spaces_for_date(locations, date)
+    locations
+      # Join with matching populations for location and given date (if any). Can't use a where clause as there may not be a capacity record
+      .joins(sanitize_sql(['LEFT OUTER JOIN populations p ON p.location_id = locations.id AND p.date = :date', date: date]))
```
willfish, line 44: I think the use of sanitize_sql is discouraged https://apidock.com/rails/ActiveRecord/Base/sanitize_sql/class?

app/models/population.rb
```diff
@@ -0,0 +1,56 @@
...
+          locations_hash[location_id] = [] unless locations_hash.key?(location_id)
+          locations_hash[location_id] << free_space_details
+        end
+      end
+    end
+  end
+
+  def self.free_spaces_for_date(locations, date)
+    locations
+      # Join with matching populations for location and given date (if any). Can't use a where clause as there may not be a capacity record
```
willfish, line 43:
```suggestion
      # Join with matching populations for location and given date (if any). Can't use a where clause as there may not be a population record
```

willfish: Maybe this model could be `LocationCapacityInformation`? I find `Population` less intuitive? Essentially these are descriptive statistics about a location on a given day? Are the locations restricted to prisons or do we report information about other types of location?

app/models/population.rb
```diff
@@ -0,0 +1,56 @@
...
+      # Join with matching (non cancelled) moves to the location on the given date (if any) so we can count them
+      .joins("LEFT OUTER JOIN moves moves_in ON moves_in.to_location_id = locations.id AND moves_in.move_type = 'prison_transfer' AND moves_in.status <> 'cancelled' AND moves_in.date = p.date")
+      # Join with matching (non cancelled) moves from the location on the given date (if any) so we can count them
+      .joins("LEFT OUTER JOIN moves moves_out ON moves_out.from_location_id = locations.id AND moves_out.move_type = 'prison_transfer' AND moves_out.status <> 'cancelled' AND moves_out.date = p.date")
+      # Group by columns used in the free space calculation
+      .group(:id, :'p.id', :usable_capacity, :unlock, :bedwatch, :overnights_in, :overnights_out, :out_of_area_courts, :discharges)
+      # Need to wrap derived columns in pointless Arel.sql call to resolve annoying deprecation warning, even though this is safe :(
+      .pluck(Arel.sql('locations.id, p.id, usable_capacity - unlock - bedwatch - overnights_in - count(moves_in) + overnights_out + out_of_area_courts + discharges + count(moves_out) as free_space'))
```
willfish, line 52: Gnarly sql

app/models/population.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  def free_spaces
+    usable_capacity - unlock - bedwatch - overnights_in - moves_to.count + overnights_out + out_of_area_courts + discharges + moves_from.count
+  end
+
+  def self.free_spaces_date_range(locations, date_range)
+    {}.tap do |locations_hash|
+      date_range.each do |date|
+        free_spaces_for_date(locations, date).each do |row|
```
willfish, line 29: Is it possible/desirable to push the grouping by date into SQL so we only have one db query? I'm also wondering whether we're at a point where properly formatted plain SQL would be easier to read?

app/models/population.rb
```diff
@@ -0,0 +1,56 @@
...
+          location_id, population_id, free_spaces = *row
+
+          free_space_details = { id: population_id, free_spaces: free_spaces } if population_id.present?
```
willfish, line 32: Is it interesting for the client to know which date the given free spaces are on or are we happy for them to just rely on the implicit order?

app/serializers/location_free_spaces_serializer.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class LocationFreeSpacesSerializer
```
willfish, line 3: What's to stop us reusing the existing locations serializer and adding a meta section to that?

app/serializers/location_free_spaces_serializer.rb
```diff
@@ -0,0 +1,15 @@
...
+  include JSONAPI::Serializer
+
+  set_type :locations
+
+  attributes :title
+
+  meta do |object, params|
+    {
+      populations: params[:populations][object.id],
```
willfish, line 12:
```suggestion
      capacity_over_days: params[:capacity_over_days][object.id],
```

app/controllers/api/populations_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+    def date_to
+      Date.strptime(date_range_params[:date_to], '%Y-%m-%d')
+    end
+
+    def validate_date_range_params
+      Populations::ParamsValidator.new(date_range_params, sort_params).validate!
+    end
+
+    def locations
+      @locations ||= Locations::Finder.new(filter_params, sort_params).call
```
willfish, line 44: This feels like it should be part of the location controller or nested under it 🤔 ?

app/models/population.rb
```diff
@@ -0,0 +1,56 @@
+class Population < ApplicationRecord
+  belongs_to :location
+
+  validates :date, presence: true
+  validates :operational_capacity, presence: true, numericality: { greater_than: 0 }
```
willfish, line 5: Presumably a bunch of these have a sensible maximum, too?

app/models/population.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def moves_from
+    location.moves_from.prison_transfer.not_cancelled.where(date: date)
+  end
+
+  def moves_to
+    location.moves_to.prison_transfer.not_cancelled.where(date: date)
+  end
+
+  def free_spaces
```
willfish, line 22: I'm guessing we can remove this? I can't see that it's used anywhere

app/models/population.rb
```diff
@@ -0,0 +1,56 @@
...
+  validates :overnights_in, presence: true, numericality: { greater_than_or_equal_to: 0 }
+  validates :overnights_out, presence: true, numericality: { greater_than_or_equal_to: 0 }
+  validates :out_of_area_courts, presence: true, numericality: { greater_than_or_equal_to: 0 }
+  validates :discharges, presence: true, numericality: { greater_than_or_equal_to: 0 }
+
+  def moves_from
+    location.moves_from.prison_transfer.not_cancelled.where(date: date)
+  end
+
+  def moves_to
```
willfish, line 18: Can we remove this?

app/models/population.rb
```diff
@@ -0,0 +1,56 @@
...
+  validates :operational_capacity, presence: true, numericality: { greater_than: 0 }
+  validates :usable_capacity, presence: true, numericality: { greater_than: 0 }
+  validates :unlock, presence: true, numericality: { greater_than_or_equal_to: 0 }
+  validates :bedwatch, presence: true, numericality: { greater_than_or_equal_to: 0 }
+  validates :overnights_in, presence: true, numericality: { greater_than_or_equal_to: 0 }
+  validates :overnights_out, presence: true, numericality: { greater_than_or_equal_to: 0 }
+  validates :out_of_area_courts, presence: true, numericality: { greater_than_or_equal_to: 0 }
+  validates :discharges, presence: true, numericality: { greater_than_or_equal_to: 0 }
+
+  def moves_from
```
willfish, line 14: Can we remove this?

app/services/locations/finder.rb
```diff
@@ -2,30 +2,58 @@
 
 module Locations
   class Finder
-    attr_accessor :filter_params
+    attr_accessor :filter_params, :order_by, :order_direction
```
willfish, line 5: Feels weird mixing sort/order wording.

app/services/locations/finder.rb
```diff
@@ -2,30 +2,58 @@
...
 
-    def initialize(filter_params)
+    def initialize(filter_params, sort_params = {})
       self.filter_params = filter_params
+      self.order_by = (sort_params[:by] || 'title').to_sym
```
willfish, line 9: I believe @martyn-w said that this is okay to default to sorting by title.

config/routes.rb
```diff
@@ -66,6 +66,8 @@
     end
     resources :framework_responses, only: %i[update]
 
+    get 'locations_free_spaces', to: 'populations#index'
```
willfish, line 69: Yeah, this feels like our controller class should just be `LocationFreeSpacesController`?

lib/tasks/fake_data.rake
```diff
@@ -236,6 +236,28 @@ namespace :fake_data do
     end
   end
 
+  desc 'create fake prison location populations'
```
willfish, line 239: 🎉

spec/models/population_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+  it { is_expected.to validate_presence_of(:date) }
+  it { is_expected.to validate_presence_of(:operational_capacity) }
+  it { is_expected.to validate_presence_of(:usable_capacity) }
+  it { is_expected.to validate_presence_of(:unlock) }
+  it { is_expected.to validate_presence_of(:bedwatch) }
+  it { is_expected.to validate_presence_of(:overnights_in) }
+  it { is_expected.to validate_presence_of(:overnights_out) }
+  it { is_expected.to validate_presence_of(:out_of_area_courts) }
+  it { is_expected.to validate_presence_of(:discharges) }
+  it { is_expected.to validate_numericality_of(:operational_capacity) }
```
willfish, line 15: Can we test the minimum restriction in these validations?

spec/models/population_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+  it { is_expected.to validate_numericality_of(:usable_capacity) }
+  it { is_expected.to validate_numericality_of(:unlock) }
+  it { is_expected.to validate_numericality_of(:bedwatch) }
+  it { is_expected.to validate_numericality_of(:overnights_in) }
+  it { is_expected.to validate_numericality_of(:overnights_out) }
+  it { is_expected.to validate_numericality_of(:out_of_area_courts) }
+  it { is_expected.to validate_numericality_of(:discharges) }
+
+  describe '#moves_from' do
```
willfish, line 24: Think we can remove this describe block now we don't use this method?

spec/models/population_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+    it 'excludes moves from a different location' do
+      location = create(:location, :prison)
+      create(:move, :prison_transfer, date: Date.today)
+      population = create(:population, date: Date.today, location: location)
+
+      expect(population.moves_from).to be_empty
+    end
+  end
+
+  describe '#moves_to' do
```
willfish, line 66: And this one.

spec/models/population_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+    it 'excludes moves from a different location' do
+      location = create(:location, :prison)
+      create(:move, :prison_transfer, date: Date.today)
+      population = create(:population, date: Date.today, location: location)
+
+      expect(population.moves_to).to be_empty
+    end
+  end
+
+  describe '#free_spaces' do
```
willfish, line 108: ...and this one.

spec/models/population_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+    let!(:prison1) { create(:location, :prison) }
+    let!(:prison2) { create(:location, :prison) }
+    let!(:move1) { create(:move, :prison_transfer, from_location: prison1, date: Date.today) }
+    let!(:move2) { create(:move, :prison_transfer, to_location: prison2, date: Date.today) }
+    let!(:move3) { create(:move, :prison_transfer, from_location: prison1, date: Date.tomorrow) }
+    let!(:population1) { create(:population, location: prison1, date: Date.today) }
+    let!(:population2) { create(:population, location: prison2, date: Date.today) }
+    let!(:population3) { create(:population, location: prison1, date: Date.tomorrow) }
+    let!(:population4) { create(:population, location: prison2, date: Date.tomorrow) }
+    let(:date_range) { (Date.yesterday..Date.tomorrow) }
```
willfish, line 142: Should we validate capacity in locations that falls outside of this date range for fullness?

willfish: e.g. create capacity information for yesterday but expect it to be excluded. create capacity information for tomorrow + 1 and expect it to be excluded?

app/models/population.rb
smoothcontract: It will be used in the `GET /populations/:id` endpoint that I'll be working on next, and is also used to verify the request specs

spec/models/population_spec.rb
willfish: Oh I see, you're testing that no capacity information leads to nil. I think we're still missing coverage for the edges of the range query, though

app/models/population.rb
smoothcontract: I explored this but the tradeoff is a massive increase in DB query complexity. This is just about acceptable from a complexity point of view and results in one query per date. The front end queries 5 days ahead, so that's 5 days max. I'm been testing locally with a reasonable volume of data copied from staging and seeing really good performance - returning all of the prisons about once takes around 300ms - a page of 20 at at time is more like 40ms. So I'm happy with the tradeoff here.

spec/models/population_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+    let(:locations) { Location.where(id: [prison1.id, prison2.id]) }
+
+    let(:expected_hash) do
+      {
+        prison1.id => [
+          nil,
+          {
```
willfish, line 149: How do you feel about adding the date for each of these capacity calculation?

app/models/population.rb
smoothcontract: Discussed with Robin and he's happy with this meta data structure.

app/models/population.rb
smoothcontract: On the front end the adopted terminology is population so I've aligned with that just to avoid the impedance mismatch when discussing this. My initial choice during the spike was "capacity" but I've changed to "population" now.

app/models/population.rb
smoothcontract: Well, then ActiveRecord really isn't leaving me many choices here. It moans without this even though it's safe.

app/models/population.rb
smoothcontract: Gnarly enough without trying to get even cleverer and return multiple dates at once!

app/serializers/location_free_spaces_serializer.rb
smoothcontract: I explored this and there are a couple of issues - firstly meta support isn't optional. There are some outstanding issues and PR's on the serializer repo but the maintainer seems unresponsive to adding support for that. So even if the `meta` block returns `nil` then a `meta` key is added to every serialized location which I don't really want to introduce as a change. Secondly we don't care about associated suppliers so I don't want to involve extra queries for that relationship. Thirdly this only has sparse attributes for title since that's all that's required by the front end. TL:DR; this is an optimised serialiser designed to be fast and to deliver just the essential data required by the front end for this specific use case, without introducing breaking changes elsewhere.

app/serializers/location_free_spaces_serializer.rb
smoothcontract: The contract for this has already been agreed with the front end devs.

spec/requests/api/populations_controller_index_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+      let(:expected_json) do
+        {
+          data: [
+            {
+              "id": location.id,
+              "type": 'locations',
+              "attributes": {
+                "title": location.title,
+              },
+              "meta": {
```
willfish, line 38: Shouldn't the meta be at the root? https://jsonapi.org/format/#document-meta

config/routes.rb
smoothcontract: There will be more populations related stuff added in the next few stories I'm working on. I'm reluctant to diverge from the adopted front end terminology too.

app/models/population.rb
smoothcontract: That might be refined while working on the further stories to edit the population data, but so far has only been specified as a mandatory positive integer.

app/models/population.rb
smoothcontract: Nope

app/models/population.rb
smoothcontract: Also nope

app/services/locations/finder.rb
smoothcontract: Yeah, I can't see an issue with that but I flagged as a breaking change just in case.

spec/models/population_spec.rb
smoothcontract: It's used

spec/models/population_spec.rb
smoothcontract: Ditto 😀

spec/models/population_spec.rb
smoothcontract: Also ditto 😀

spec/models/population_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+            free_spaces: population1.free_spaces,
```
smoothcontract, line 150: This is where `free_spaces` method is used at present, and will also be called in the forthcoming `GET /populations/:id` endpoint.

app/models/population.rb
willfish: I'm sure it's fine. It'll explode as we've got tests if they remove the alias in future

app/models/population.rb
willfish: 😢  okay

app/models/population.rb
willfish: Okay. Makes sense.

app/serializers/location_free_spaces_serializer.rb
willfish: Clear. Thanks

app/models/population.rb
willfish: Okay

app/models/population.rb
willfish: Coolio

willfish (approved):
I think it makes sense not to document this api, personally.

willfish:
Also, we're fine to remove the breaking change label I think

spec/models/population_spec.rb
smoothcontract: Yes, I'll add that

spec/models/population_spec.rb
smoothcontract: Let me think about that 😁

spec/requests/api/populations_controller_index_spec.rb
smoothcontract: It's supported both at document and resource level. Here's the relevant section - plus the gem wouldn't support it if it wasn't a thing I guess...

```
In addition, a resource object MAY contain any of these top-level members:

* attributes: an attributes object representing some of the resource’s data.
* relationships: a relationships object describing relationships between the resource and other JSON:API resources.
* links: a links object containing links related to the resource.
* meta: a meta object containing non-standard meta-information about a resource that can not be represented as an attribute or relationship.
```

merlinc (approved):
LGTM

spec/models/population_spec.rb
smoothcontract: Added another population that falls outside the date range 👍

PR #1033, opened 2020-10-13
Cancellation reason and rejection reason should accept a list

cwrw:
The filters for move should accept a list of comma delimited values for cancellation reason or rejection reason. Change param validator to validate two arrays instead of a string value being included in an array.

This needs to be confirmed and tested/released with FE as we don't allow the use of undefined.

### DIFF
--- a/spec/services/moves/params_validator_spec.rb
+++ b/spec/services/moves/params_validator_spec.rb
@@ -28,4 +28,74 @@ RSpec.describe Moves::ParamsValidator do
 
     it { is_expected.not_to be_valid }
   end
+
+  context 'with an empty cancellation reason' do
+    let(:filter_params) { { cancellation_reason: cancellation_reason } }
+    let(:cancellation_reason) { nil }
+
+    it { is_expected.to be_valid }
+  end
+
+  context 'with a correct cancellation reason' do
+    let(:filter_params) { { cancellation_reason: cancellation_reason } }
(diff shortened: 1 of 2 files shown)


spec/services/moves/params_validator_spec.rb
```diff
@@ -28,4 +28,74 @@
...
+  end
+
+  context 'with incorrect cancellation reason' do
+    let(:filter_params) { { cancellation_reason: cancellation_reason } }
+    let(:cancellation_reason) { 'boom' }
+
+    it { is_expected.not_to be_valid }
+  end
+
+  context 'with multiple incorrect cancellation reasons' do
```
willfish, line 60:
```suggestion
  context 'with one incorrect cancellation reason' do
```

spec/services/moves/params_validator_spec.rb
```diff
@@ -28,4 +28,74 @@
...
+  end
+
+  context 'with incorrect rejection reason' do
+    let(:filter_params) { { rejection_reason: rejection_reason } }
+    let(:rejection_reason) { 'boom' }
+
+    it { is_expected.not_to be_valid }
+  end
+
+  context 'with multiple incorrect rejection reasons' do
```
willfish, line 95:
```suggestion
  context 'with one incorrect rejection reason' do
```

willfish (approved):

PR #1035, opened 2020-10-13
Retry transaction when it fails on updating responses

cwrw:
https://dsdmoj.atlassian.net/browse/IM-52
When an application is deployed, and new migrations run, we end up with cached prepared statements which are no
longer valid since the underlying schema has changed. Since we have a rolling deploy, we cannot resolve this by restarting all pods at once after migrations is run. Retry the transaction instead, only once.

Sources used for this: https://medium.com/carwow-product-engineering/activerecord-error-cached-plan-must-not-change-result-type-aca72cdef504
https://flexport.engineering/avoiding-activerecord-preparedstatementcacheexpired-errors-4499a4f961cf

I opted out of having to retry all transactions and only this specific one as we don't know how safe other transactions are now or new ones added in the future, so to avoid that we only retry the offending transaction for now. Happy to hear thoughts around this.

### DIFF
--- a/app/models/framework_response.rb
+++ b/app/models/framework_response.rb
@@ -24,16 +24,7 @@ class FrameworkResponse < VersionedModel
   def update_with_flags!(new_value)
     return unless value != new_value
 
-    ActiveRecord::Base.transaction do
-      update!(value: new_value)
-      rebuild_flags!
-      FrameworkResponse.clear_dependent_values_and_flags!([self])
-
-      # lock the status update to avoid race condition on multiple response patches
-      person_escort_record.with_lock do
-        person_escort_record.update_status!
-      end
-    end
+    ActiveRecord::Base.transaction { update_response_transaction(new_value) }
   rescue FiniteMachine::InvalidStateError
     raise ActiveRecord::ReadOnlyRecord, "Can't update framework_responses because person_escort_record is #{person_escort_record.status}"
   end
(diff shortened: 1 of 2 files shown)


willfish (approved):
🙏

app/models/framework_response.rb
```diff
@@ -112,4 +103,24 @@ def self.recursive_tree
   def set_responded_value
     self.responded = true
   end
+
+  def update_response_transaction(new_value)
+    retried ||= false
```
smoothcontract, line 108: I'm not convinced this will work with a local variable within this method - i.e. as soon as the method exits then `retried` falls out of scope and is eligible for garbage collection. Perhaps better to move this to an instance method in the class instead - the intent and behaviour should then be a bit clearer I think.

app/models/framework_response.rb
```diff
@@ -112,4 +103,24 @@ def self.recursive_tree
...
+    # lock the status update to avoid race condition on multiple response patches
+    person_escort_record.with_lock do
+      person_escort_record.update_status!
+    end
+  rescue ActiveRecord::PreparedStatementCacheExpired
+    # retry transaction once if this transaction fails
+    if retried
+      raise
+    else
+      retried = true
```
smoothcontract, line 122: Change to `@retried = true` or `self.retried = true` when moving to an instance variable

smoothcontract (commented):
Suggest moving `retried` to an instance variable for clarity as it's intended to survive multiple invocations of `update_response_transaction` method, otherwise looks good.

app/models/framework_response.rb
cwrw: so `retried` is memoized so it will remain in scope and the behaviour is as expected. However I've moved it to an instance method for clarity as suggested

PR #1037, opened 2020-10-14
[P4-2349] DRY up getters/setters in events

willfish:
### Jira link

P4-2349

### What?

I have added/removed/altered:

- [x] Adds and uses methods for setting and getting detail and relationship attributes

### Why?

I am doing this because:

- As I've been writing events we noticed a lot of repetition around getters/setters. DRYing up here to make this simpler

### DIFF
--- a/app/controllers/concerns/eventable.rb
+++ b/app/controllers/concerns/eventable.rb
@@ -8,7 +8,7 @@ module Eventable
 
   def process_event(eventables, event_name, event_params)
     [eventables].flatten.each do |eventable|
-      result = create_generic_event(eventable, event_name, event_params)
+      create_generic_event(eventable, event_name, event_params)
       run_event_logs(eventable)
     end
   end
--- a/app/controllers/concerns/eventable.rb
+++ b/app/controllers/concerns/eventable.rb
@@ -8,7 +8,7 @@ module Eventable
 
   def process_event(eventables, event_name, event_params)
     [eventables].flatten.each do |eventable|
-      create_generic_event(eventable, event_name, event_params)
+      result = create_generic_event(eventable, event_name, event_params)
       run_event_logs(eventable)
     end
   end
@@ -43,6 +43,6 @@ private
   def assign_specific_attributes!(attributes, event_params, type)
     attributes[:details] = {}
     attributes[:details].merge!(::GenericEvents::EventSpecificRelationshipsMapper.new(event_params[:relationships]).call)
-    attributes[:details].merge!(event_params[:attributes].slice(*type::DETAILS_ATTRIBUTES))
+    attributes[:details].merge!(event_params[:attributes].slice(*type.details_attributes))
   end
 end
--- a/app/models/concerns/court_cell_validations.rb
+++ b/app/models/concerns/court_cell_validations.rb
@@ -3,14 +3,6 @@ require 'active_support/concern'
 module CourtCellValidations
   extend ActiveSupport::Concern
 
-  def court_cell_number=(court_cell_number)
-    details['court_cell_number'] = court_cell_number
-  end
-
-  def court_cell_number
-    details['court_cell_number']
-  end
-
   included do
     validates :court_cell_number, presence: true
   end
--- a/app/models/generic_event.rb
+++ b/app/models/generic_event.rb
@@ -98,7 +98,7 @@ class GenericEvent < ApplicationRecord
 
   def self.details_attributes(*attributes)
     define_singleton_method(:details_attributes) do
-      class_variable_get('@@details_attributes')
+      instance_variable_get('@details_attributes')
     end
 
     attributes = attributes.each_with_object([]) do |attribute_key, acc|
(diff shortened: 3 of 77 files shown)


app/models/concerns/court_cell_validations.rb
```diff
@@ -3,14 +3,6 @@
...
   extend ActiveSupport::Concern
 
-  def court_cell_number=(court_cell_number)
-    details['court_cell_number'] = court_cell_number
-  end
-
-  def court_cell_number
-    details['court_cell_number']
-  end
-
```
smoothcontract, line 13: Should the concern also specify the `details_attributes` or do you do this for each case where the concern is included, and then call `details_attributes` just once per class? 🤔

app/models/generic_event.rb
```diff
@@ -95,4 +95,46 @@ def self.from_event(event)
...
+    attributes = attributes.each_with_object([]) do |attribute_key, acc|
+      define_method(attribute_key) do
+        details[attribute_key]
+      end
+
+      define_method("#{attribute_key}=") do |attribute_value|
+        details[attribute_key] = attribute_value
+      end
+
+      acc << attribute_key
```
smoothcontract, line 113: I don't think you need this accumulator array, as it just ends up with the original contents of `attributes`.

app/models/generic_event.rb
```diff
@@ -95,4 +95,46 @@ def self.from_event(event)
...
```
smoothcontract, line 135: Same here - the array accumulator is redundant. You could just do

```ruby
attributes.each do |attribute_key|
  define_method...
end
```

app/models/generic_event/move_reject.rb
```diff
@@ -1,42 +1,22 @@
...
       eventable.cancellation_reason = Move::CANCELLATION_REASON_REJECTED
       eventable.cancellation_reason_comment = cancellation_reason_comment
-      eventable.rebook if rebook?
+      eventable.rebook if rebook
     end
 
     def for_feed
       super.tap do |common_feed_attributes|
-        common_feed_attributes['details']['rebook'] = rebook?
+        common_feed_attributes['details']['rebook'] = rebook
```
smoothcontract, line 19: Did you want to add `|| false` here so it's always a boolean value in the feed rather than potentially `nil` ?

app/models/generic_event/per_court_excessive_delay_not_due_to_supplier.rb
```diff
@@ -1,11 +1,9 @@
 class GenericEvent
   class PerCourtExcessiveDelayNotDueToSupplier < GenericEvent
     LOCATION_ATTRIBUTE_KEY = :location_id
-    DETAILS_ATTRIBUTES = %w[
-      subtype
-      vehicle_reg
-      ended_at
-    ].freeze
+
+    details_attributes :subtype, :vehicle_reg, :ended_at, :authorised_by, :authorised_at
```
smoothcontract, line 5: Where did the new `authorised_by` and `authorised_at` attributes come from? Are those intentional?

smoothcontract: Scratch that - those are defined in the included concern 👍

smoothcontract (commented):
This is such a nice refactor - awesome job ❤️ .. I think there's scope for a further slight simplification with the array accumulator thing and I left a couple of queries (some of which I figured out but left for other reviewers that may have the same questions). 👏

app/controllers/concerns/eventable.rb
```diff
@@ -8,7 +8,7 @@ def supplier_id
 
   def process_event(eventables, event_name, event_params)
     [eventables].flatten.each do |eventable|
-      create_generic_event(eventable, event_name, event_params)
+      result = create_generic_event(eventable, event_name, event_params)
```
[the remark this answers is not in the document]
willfish: Feeexed

app/models/concerns/court_cell_validations.rb
willfish: I think it makes sense for them to be decoupled like this, yeah. Even if there is a dependency there 🤔

app/models/generic_event/move_reject.rb
willfish: Yeah, we should surface it.

app/models/concerns/court_cell_validations.rb
smoothcontract: The comment is mostly me rambling whilst trying to understand how this works - it's all good 👍

willfish: Cool beans 🏖️

app/models/generic_event.rb
willfish: Removed :)

smoothcontract (approved):
Lovely work 👍

app/models/generic_event/move_reject.rb
willfish: Surfaced in most recent commit :)

PR #1045, opened 2020-10-15
Remove profile from creation of person escort record

cwrw:
We currently support both profile or move for creating a PER. Switch to only use the move to create a PER, as using a profile is deprecated, we will be associating the profile attached to the move instead. The move is required to access the location and other data required for creating the PER, which isn't available by just passing the profile.

~This is blocked until FE fully switches to using move.~ FE has now fully switch to use move to create the PER

### DIFF
--- a/app/controllers/api/person_escort_records_controller.rb
+++ b/app/controllers/api/person_escort_records_controller.rb
@@ -4,11 +4,10 @@ module Api
   class PersonEscortRecordsController < ApiController
     after_action :send_notification, only: :update
 
-    # TODO: Remove profile id and transition to Move id, for now accept both
     NEW_PERMITTED_PER_PARAMS = [
       :type,
       attributes: [:version],
-      relationships: [profile: {}, move: {}],
+      relationships: [move: {}],
     ].freeze
 
     UPDATE_PERMITTED_PER_PARAMS = [
@@ -17,10 +16,8 @@ module Api
     ].freeze
 
     def create
-      # TODO: Remove profile id and transition to Move id, for now accept both
       person_escort_record = PersonEscortRecord.save_with_responses!(
         version: new_person_escort_record_params.dig(:attributes, :version),
-        profile_id: new_person_escort_record_params.dig(:relationships, :profile, :data, :id),
         move_id: new_person_escort_record_params.dig(:relationships, :move, :data, :id),
       )
 
--- a/app/models/person_escort_record.rb
+++ b/app/models/person_escort_record.rb
@@ -39,15 +39,9 @@ class PersonEscortRecord < VersionedModel
            :confirmed?,
            to: :state_machine
 
-  def self.save_with_responses!(profile_id: nil, version: nil, move_id: nil)
-    # TODO: Remove profile id and transition to Move id, for now accept both
-    if move_id.present?
-      move = Move.find(move_id)
-      profile = move.profile
-    else
-      move = nil
-      profile = Profile.find(profile_id)
-    end
+  def self.save_with_responses!(version: nil, move_id: nil)
+    move = Move.find(move_id)
+    profile = move.profile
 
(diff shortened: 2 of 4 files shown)


swagger/v1/person_escort_record.yaml
```diff
@@ -100,6 +100,7 @@ PersonEscortRecord:
     relationships:
       type: object
       required:
+        - move
         - profile
```
willfish, line 104: Do you want to remove profile as a required relationship reference, here?

willfish:
LGTM. Re previous conversations, I'm not convinced our naming of the profile now makes any sense since it is now only the PTR answers.

swagger/v1/person_escort_record.yaml
cwrw: We do require both a move, and a move's profile to be always present relationships

smoothcontract: This still feels slightly confusing to me - the profile relationship is indirectly through the move, so I don't think it makes sense to specify it here as a required relationship of the PER, particularly as the PER schema makes no mention of the profile.

smoothcontract (commented):
Looks good apart from one suggestion for updating documentation.

swagger/v1/person_escort_record.yaml
cwrw: Those values depend on the context of what we see those required relationships as, A PER cannot exist without both a move and its associated profile. If one is missing the PER creation fails. They both exist as reference ids on the table itself. We cannot remove the `profile_id` column as this supports legacy PERs. However if we see these relationships as what we pass in on creation then I can remove `profile` as well as `framework_responses` from `required.

willfish: Sounds like we might want to migrate all historical PERs to not use the profile id?

willfish: And then we can just do away with the profile as a required relationship in this document?

swagger/v1/person_escort_record.yaml
```diff
@@ -100,9 +100,8 @@ PersonEscortRecord:
     relationships:
       type: object
       required:
-        - profile
+        - move
         - framework
-        - responses
```
willfish, line 105: How comes we're removing responses?

cwrw: see https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/1045#discussion_r508295933

swagger/v1/person_escort_record.yaml
cwrw: A profile can be linked to multiple moves, meaning we can't determine the specific move related to the person escort record, the main reason we now require a Move instead of a Profile to create the PER. I would like to keep profile now, as we require a unique validation on Profile when creating the PER. But In reality we never attached a profile to multiple moves historically, so I don't think migrating might be problematic, but I'd like to keep it for now as later a profile can be linked to multiple moves when wanting to reuse the PER, so after clients implement that change I'm going to revisit the migration.

smoothcontract (approved):
LGTM - thanks for the detailed explanation of the changes to profile relationship 👍

willfish (approved):

swagger/v1/person_escort_record.yaml
martyn-w: Keep in mind that suppliers also create moves and may re-use a single profile across multiple moves -  especially for the same custody period. E.g. move1 to court; court hearing; move2 remand from court. So probably both move and profile are necessary here?

cwrw: Yea I think so, best keep them as is. I've also chatted with the FE and we're happy to keep both relationships for now as well, but only create a PER from a move 👍

smoothcontract (approved):
Looks great 👍

PR #1051, opened 2020-10-16
Add new endpoint POST /populations to support creating new populations

smoothcontract:
### Jira link

P4-1427

### What?

- [x] Add a new endpoint `POST /populations` to create a new population record
- [x] Added a unique index on population date and location to prevent multiple populations being created

### Why?

- All the front end things need this

### Have you? (optional)

- New endpoint, minimal risk

### DIFF
--- a/app/models/population.rb
+++ b/app/models/population.rb
@@ -1,4 +1,4 @@
-class Population < VersionedModel
+class Population < ApplicationRecord
   belongs_to :location
 
   validates :date, presence: true
--- a/app/models/population.rb
+++ b/app/models/population.rb
@@ -1,7 +1,7 @@
 class Population < ApplicationRecord
   belongs_to :location
 
-  validates :date, presence: true
+  validates :date, presence: true, uniqueness: { scope: :location_id }
   validates :operational_capacity, presence: true, numericality: { greater_than: 0 }
   validates :usable_capacity, presence: true, numericality: { greater_than: 0 }
   validates :unlock, presence: true, numericality: { greater_than_or_equal_to: 0 }
--- a/app/models/population.rb
+++ b/app/models/population.rb
@@ -1,4 +1,4 @@
-class Population < ApplicationRecord
+class Population < VersionedModel
   belongs_to :location
 
   validates :date, presence: true
@@ -21,6 +21,14 @@ class Population < ApplicationRecord
 
   def free_spaces
     usable_capacity - unlock - bedwatch - overnights_in - moves_to.count + overnights_out + out_of_area_courts + discharges + moves_from.count
+  end
+
+  def save_uniquely!
+    save!
+    self
+  rescue PG::UniqueViolation, ActiveRecord::RecordNotUnique
+    errors.add(:date, :taken)
+    raise ActiveRecord::RecordInvalid, self
   end
 
   def self.free_spaces_date_range(locations, date_range)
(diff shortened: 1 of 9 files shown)


app/models/population.rb
```diff
@@ -1,4 +1,4 @@
-class Population < ApplicationRecord
+class Population < VersionedModel
```
willfish, line 1: Are we sure we want to add this to versioned models? I'm not sure how important it is to keep a history of changes for the dashboard and that table is already pretty busy?

app/models/population.rb
```diff
@@ -23,6 +23,14 @@ def free_spaces
     usable_capacity - unlock - bedwatch - overnights_in - moves_to.count + overnights_out + out_of_area_courts + discharges + moves_from.count
   end
 
+  def save_uniquely!
+    save!
+    self
+  rescue PG::UniqueViolation, ActiveRecord::RecordNotUnique
```
willfish, line 29: Should we just add a validation that requires that the population is unique?

willfish: Confused about why this can't just be handled in the same way as: https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/968a860c5b92df2b4f87eacaabe8e4cf7086abd3/app/models/move.rb#L99

willfish: Obviously with a simpler approach to validating uniqueness

martyn-w: I agree that a uniqueness validation would be good addition 👍 
but I think the database fallback catch is _also_ required, because the pre-emptive uniqueness validator doesn't always catch clashes (creating moves on different pods etc)

app/models/population.rb
smoothcontract: Hmm... good question. Maybe we don't care about this actually 🤔

app/models/population.rb
smoothcontract: I thought this would be controversial so it's great you all spotted it 😛 . Here's my thinking - the rails unique validations basically don't work as Abeer found out in previous work on the PER it's possible and has been observed for two requests to overlap and defeat Rails uniqueness validations (which essentially perform a query before an update). Without a database level constraint it's therefore possible to get duplicate rows; in the case of the population free spaces view this would lead to incorrect calculations and confusing duplicate data so we don't want that.

The database level constraint is therefore essential to ensure that no duplicates can be created. That means we have to catch database level exceptions and propagate them nicely back to the client (raising the same error message they would have seen with a rails level uniqueness check).

So that leads me to wonder what value the uniqueness check would be adding on top of that - it means performing an additional `SELECT` query before every update and insert of the population table, plus additional code that needs spec coverage, only to provide the same level of uniqueness protection we get solely from the database level constraint. I think it's basically superfluous in this circumstance which is why I left it out. It's less idiomatic than the usual approach but perhaps the usual approach is flawed... 🤯

smoothcontract: Lots of articles on this online but this captures the main points pretty well: https://makandracards.com/makandra/13901-understanding-race-conditions-with-duplicate-unique-keys-in-rails-mysql

app/models/population.rb
smoothcontract: I've removed now - good call raising this 👍

cwrw (approved):
Looks good, discussed in person: I would still add the model validation for documentation and to be idiomatic

willfish (approved):
Looking good, man

PR #1052, opened 2020-10-16
[P4-2351] Impement final event STI models

willfish:
### Jira link

P4-2351

### What?

I have added/removed/altered:

- [x] Adds JourneyChangeVehicle event
- [x] Adds MoveCrossSupplierDropOff event
- [x] Adds MoveCrossSupplierPickUp event
- [x] Adds PersonMoveBookedIntoReceivingEstablishment event
- [x] Adds PerCourtHearing event
- [x] Adds PersonMoveAssault event
- [x] Adds PersonMoveDeathInCustody event
- [x] Adds PersonMoveMajorIncidentOther event
- [x] Adds PersonMoveMinorIncidentOther event
- [x] Adds PersonMovePersonEscaped event
- [x] Adds PersonMovePersonEscapedKPI event
- [x] Adds PersonMoveReleasedError event
- [x] Adds PersonMoveRoadTrafficAccident event
- [x] Adds PersonMoveSeriousInjury event
- [x] Adds PersonMoveUsedForce event
- [x] Adds PersonMoveVehicleBrokeDown event
- [x] Adds PersonMoveVehicleSystemsFailed event
- [x] DRYs up some iso date validations
- [x] Adds postcode validations for vehicle events that require them
- [x] Adds success case test for all newly added events

### Why?

I am doing this because:

- These are needed for timeline, the hub and JPC

### Have you? (optional)

- [x] Adding wiki documentation after this


app/models/generic_event.rb
```diff
@@ -133,4 +149,14 @@ def self.relationship_attributes(*attributes)
       end
     end
   end
+
+  def self.eventable_types(*types)
```
willfish, line 153: Already quite a big PR. Going to refactor other events to use this method instead of the concern.

app/models/generic_event.rb
```diff
@@ -63,6 +67,19 @@ class GenericEvent < ApplicationRecord
     PerGeneric
     PerMedicalAid
     PerPrisonerWelfare
+    PersonMoveAssault
+    PersonMoveBookedIntoReceivingEstablishment
+    PersonMoveDeathInCustody
+    PersonMoveMajorIncidentOther
+    PersonMoveMinorIncidentOther
+    PersonMovePersonEscaped
```
smoothcontract, line 75: 🏃

app/models/generic_event.rb
```diff
@@ -63,6 +67,19 @@ class GenericEvent < ApplicationRecord
...
+    PersonMovePersonEscapedKpi
+    PersonMoveReleasedError
+    PersonMoveRoadTrafficAccident
+    PersonMoveSeriousInjury
+    PersonMoveUsedForce
```
smoothcontract, line 80: 👊

app/models/generic_event/journey_change_vehicle.rb
```diff
@@ -0,0 +1,20 @@
+class GenericEvent
+  class JourneyChangeVehicle < GenericEvent
+    before_create :set_previous_vehicle_registration
+
+    details_attributes :vehicle_reg, :previous_vehicle_reg
```
smoothcontract, line 5: Could this be addressed with a versioned model instead rather than hand rolled versioning for specific attribute?

app/models/generic_event/per_court_hearing.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    validates :is_virtual,       presence: true
+    validates :is_trial,         presence: true
+    validates :court_listing_at, presence: true, iso_date_time: true
+    validates :started_at,       presence: true, iso_date_time: true
+    validates :ended_at,         presence: true, iso_date_time: true
+    validates :agreed_at,        presence: true, iso_date_time: true
+    validates :court_outcome,    presence: true
+
+    validates :is_virtual,       inclusion: [true, false]
```
smoothcontract, line 19: Could combine this with line 11 and the line below with line 12

app/models/generic_event/person_move_death_in_custody.rb
```diff
@@ -0,0 +1,14 @@
+class GenericEvent
```
smoothcontract, line 1: This and the next few classes are identical apart from naming, might be a nice use case for inheritance to avoid repeating the same stuff over and over?

app/validators/iso_date_time_validator.rb
```diff
@@ -0,0 +1,7 @@
+class IsoDateTimeValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    Time.zone.iso8601(value)
+  rescue ArgumentError
+    record.errors.add(attribute, 'must be formatted as a valid ISO-8601 date-time')
+  end
+end
```
smoothcontract, line 7: Awesome, this lets us clean up lots of other validation services in future 🥇

smoothcontract (commented):
This is shaping up really nicely. I like the shared Time and postcode validation a lot, great work there and means we can continue to clean up other services and models that do this so thanks for establishing the pattern. Only real concern is the level of duplication with multiple identical event classes - if those are all extremely similar then it might be a nice opportunity to introduce a parent event e.g. `PersonMoveGeneralEvent` and DRY these specific events up as it seems to only be the event name that differs

app/models/generic_event/journey_change_vehicle.rb
willfish: Just talking this through with Sangeetha. I'd actually like to remove history in the event that belongs to the eventable if possible (or force the supplier to provide it/have the report work it out).

app/models/generic_event/per_court_hearing.rb
willfish: True!

app/models/generic_event/person_move_death_in_custody.rb
willfish: I was thinking that, yeah.

willfish: Will have a play with it

app/models/generic_event/journey_change_vehicle.rb
willfish: Okay, killed the versioning. Client just supplies this value after they've broken down and get a vehicle replacement so they have the information we need for the JPC.

app/models/generic_event/person_move_death_in_custody.rb
willfish: Done

app/models/generic_event/journey_change_vehicle.rb
smoothcontract: Awesome, nice to keep it simple. Great work convincing people with this 👍

smoothcontract (approved):
Nice job, and thanks for taking feedback on board with simplifying requirements and introducing a touch of OO to DRY up the implementation 👏

PR #1056, opened 2020-10-19
P4 2379 moves index serializer deploydev

martyn-w:
### Jira link

P4-2379

### What?

I have added/removed/altered:

- [x] Restricted supported included relationships for /moves endpoint
- [x] Restricted serialization of /moves endpoint
- [x] Linked moves finder includes to those specified in the request

### Why?

I am doing this because:

- to speed up the moves finder and remove N+1 inefficienies, whilst also not pulling more data than is actually required to service the request

### Have you? (optional)

- [x] Updated API docs if necessary TODO

- Changes v2 /moves endpoint api that is used in production - currently only the FE is believed to be using it
- could potentially break pages: needs thorough testing in the front end


app/serializers/prison_transfer_reason_serializer.rb
```diff
@@ -3,7 +3,9 @@
 class PrisonTransferReasonSerializer
   include JSONAPI::Serializer
 
+  INCLUDED_ATTRIBUTES = %i[title key disabled_at].freeze
```
willfish, line 6: Aren't these sparse fields in json:api speak?

martyn-w: Interesting question. We do not support the `fields?` parameter in the API at the moment. Reading up on [spare fieldsets](https://jsonapi.org/format/#fetching-sparse-fieldsets) I think it is fine to give a subset of fields:
> ...If a client does not specify the set of fields for a given resource type, the server MAY send all fields, a subset of fields, or no fields for that resource type.

willfish: I mean the name is obvious enough 👍

app/controllers/api/v1/moves_actions.rb
```diff
@@ -1,7 +1,10 @@
 module Api::V1
   module MovesActions
     def index_and_render
-      paginate moves, serializer: MoveSerializer, include: included_relationships - %w[court_hearings], fields: MoveSerializer::INCLUDED_FIELDS
+      paginate moves,
```
willfish, line 4: I guess we mean to update the v1 move actions?

willfish: Okay, ignore this.

app/controllers/api_controller.rb
```diff
@@ -223,7 +223,25 @@ def validate_include_params
...
+    parts = value.split('.', 2)
+    if parts.length == 1
+      # special hack to allow for the flags table renamed in the backend but not in the FE includes
+      if parts.first == 'flags'
+        :framework_flags
+      else
+        parts.first.to_sym
+      end
+    else
+      { parts.first.to_sym => cast_relationship_to_symbol_hash(parts.last) }
```
willfish, line 243: Ruby doesn't optimise for recursion 😢  so I'd avoid it and prefer mutating state/iterating

willfish: [shakes fist at ruby]

app/controllers/api_controller.rb
```diff
@@ -223,7 +223,25 @@ def validate_include_params
...
   def included_relationships
-    IncludeParamHandler.new(params).call
+    @included_relationships ||= IncludeParamHandler.new(params).call
+  end
+
+  def included_db_relationships
+    @included_db_relationships ||= included_relationships.map { |value| cast_relationship_to_symbol_hash(value) } if included_relationships.present?
+  end
+
+  def cast_relationship_to_symbol_hash(value)
```
willfish, line 233: I guess the responsibility of this method is to convert relationships that are a list of query param includes to those compatible with `ActiveRecord::Base#includes` and `ActiveRecord::Base#joins` calls?  How do you feel about renaming this to `query_include_to_active_record_include_hash` or equivalent?

willfish: For reference the gem we use does this, too: https://github.com/jsonapi-serializer/jsonapi-serializer/blob/master/lib/fast_jsonapi/serialization_core.rb#L155-L161. I'm confused why it doesn't explode 😕  re our need to special snowflake the flags thingy majig.

app/controllers/api_controller.rb
willfish: I actually think this process is complex enough to need unit testing in a utility class.

app/services/moves/finder.rb
```diff
@@ -2,11 +2,9 @@
 
 module Moves
   class Finder
-    attr_reader :filter_params, :ability
+    attr_reader :filter_params, :ability, :db_includes
 
-    MOVE_INCLUDES = [:allocation, :supplier, :court_hearings, :prison_transfer_reason, :original_move, profile: [documents: { file_attachment: :blob }, person_escort_record: [:framework, :framework_responses, framework_flags: :framework_question]], person: %i[gender ethnicity], from_location: %i[supplier_locations], to_location: %i[supplier_locations]].freeze
-
-    def initialize(filter_params, ability, order_params)
+    def initialize(filter_params, ability, order_params, db_includes)
```
smoothcontract, line 7: I think now would be a good time to introduce named parameters for this class - three params was just about tolerable but four would be clearer with named params.

app/controllers/api/moves_controller.rb
```diff
@@ -9,7 +9,10 @@ def index
     end
 
     def csv
-      send_file(Moves::Exporter.new(moves).call, type: 'text/csv', disposition: :inline)
+      # NB: this method should behave the same irrespective of api version / includes etc
+      csv_moves = Moves::Finder.new(filter_params, current_ability, params[:sort] || {},
+                                    [:from_location, :to_location, { profile: :documents }, person: %i[gender ethnicity]]).call
```
smoothcontract, line 14: Might be worth splitting out includes for readability e.g.

```ruby
csv_includes = [:from_location, :to_location, { profile: :documents }, person: %i[gender ethnicity]]
```

That probably provides enough clarity to make the comment on line 12 redundant too.

app/controllers/api/moves_controller.rb
```diff
@@ -27,7 +30,8 @@ def update
   private
 
     def moves
-      @moves ||= Moves::Finder.new(filter_params, current_ability, params[:sort] || {}).call
+      # called by index method only
```
smoothcontract, line 33: I don't think this comment is really needed

app/controllers/api/v1/moves_actions.rb
```diff
@@ -119,6 +122,21 @@ def included_relationships
       IncludeParamHandler.new(params).call || MoveSerializer::SUPPORTED_RELATIONSHIPS
     end
 
+    def included_db_relationships
+      # NB: v1 API needs a hardcoded included_db_relationships as they are not usually provided in the request
```
smoothcontract, line 126: v1 is getting really really annoying now 😠

app/controllers/api/v2/moves_actions.rb
```diff
@@ -151,7 +151,12 @@ def included_relationships
     end
 
     def supported_relationships
-      ::V2::MoveSerializer::SUPPORTED_RELATIONSHIPS
+      # for performance reasons, we support a fewer number of include relationships on the index (and csv) actions
```
smoothcontract, line 154:
```suggestion
      # for performance reasons, we support fewer include relationships on the index action
```

app/controllers/api_controller.rb
```diff
@@ -223,7 +223,25 @@ def validate_include_params
...
+    parts = value.split('.', 2)
+    if parts.length == 1
+      # special hack to allow for the flags table renamed in the backend but not in the FE includes
+      if parts.first == 'flags'
+        :framework_flags
```
smoothcontract, line 238: Ew. It's unfortunate that the naming diverged between the API and the model, this has caused lots of hacky workarounds for the serialization stuff. I wonder if in a future PR we could look at renaming the model instead @cwrw ?

app/controllers/api_controller.rb
smoothcontract: Presumably these hashes are unlikely to be more than 3 layers deep so not sure the recursion thing is a big issue. But I agree this does need strong coverage.

app/serializers/location_serializer.rb
```diff
@@ -3,14 +3,18 @@
 class LocationSerializer
   include JSONAPI::Serializer
 
+  INCLUDED_ATTRIBUTES = %i[
```
smoothcontract, line 6: You can avoid all of this noise by inspecting the `attributes_to_serialize` class method of each serializer - and I think it's much more idiomatic to do this rather than define an array of attributes in every serializer. Plus this will also include attributes that are defined individually which we do in a few places. See:

```ruby
[1] pry(main)> LocationSerializer.attributes_to_serialize
=> {:key=>#<FastJsonapi::Attribute:0x00007fdc9692f868 @conditional_proc=nil, @key=:key, @method=:key>,
 :title=>#<FastJsonapi::Attribute:0x00007fdc9692f818 @conditional_proc=nil, @key=:title, @method=:title>,
 :location_type=>#<FastJsonapi::Attribute:0x00007fdc9692f7c8 @conditional_proc=nil, @key=:location_type, @method=:location_type>,
 :nomis_agency_id=>#<FastJsonapi::Attribute:0x00007fdc9692f778 @conditional_proc=nil, @key=:nomis_agency_id, @method=:nomis_agency_id>,
 :can_upload_documents=>#<FastJsonapi::Attribute:0x00007fdc9692f728 @conditional_proc=nil, @key=:can_upload_documents, @method=:can_upload_documents>,
 :disabled_at=>#<FastJsonapi::Attribute:0x00007fdc9692f6d8 @conditional_proc=nil, @key=:disabled_at, @method=:disabled_at>}
[2] pry(main)> LocationSerializer.attributes_to_serialize.keys
=> [:key, :title, :location_type, :nomis_agency_id, :can_upload_documents, :disabled_at]
```

app/serializers/v2/moves_serializer.rb
```diff
@@ -0,0 +1,58 @@
...
+      moves: INCLUDED_ATTRIBUTES +
+        %i[profile from_location to_location prison_transfer_reason supplier additional_information],
+      profiles: ::V2::ProfileSerializer::INCLUDED_ATTRIBUTES + %i[person],
+      people: ::V2::PersonSerializer::INCLUDED_ATTRIBUTES + %i[gender ethnicity],
+      locations: ::LocationSerializer::INCLUDED_ATTRIBUTES,
+      prison_transfer_reasons: ::PrisonTransferReasonSerializer::INCLUDED_ATTRIBUTES,
+      suppliers: ::SupplierSerializer::INCLUDED_ATTRIBUTES,
+
+      # allocation: %i[to_location from_location moves_count created_at],
+    }.freeze
```
smoothcontract, line 35:
```suggestion
INCLUDED_FIELDS = {
      moves: attributes_to_serialize.keys +
        %i[profile from_location to_location prison_transfer_reason supplier additional_information],
      profiles: ::V2::ProfileSerializer.attributes_to_serialize.keys + %i[person],
      people: ::V2::PersonSerializer.attributes_to_serialize.keys + %i[gender ethnicity],
      locations: ::LocationSerializer.attributes_to_serialize.keys,
      prison_transfer_reasons: ::PrisonTransferReasonSerializer.attributes_to_serialize.keys,
      suppliers: ::SupplierSerializer.attributes_to_serialize.keys,
    }.freeze
```

app/services/moves/finder.rb
```diff
@@ -59,6 +58,12 @@ def apply_filters(scope)
       scope
     end
 
+    def apply_includes(scope)
+      scope = scope.includes(db_includes)
+      # scope = scope.joins(db_includes) # NB: do not join - you may get duplicates
```
smoothcontract, line 63: Remove commented line?

smoothcontract (commented):
This is looking pretty promising, my main concerns are that the approach to defining an array of attributes for each serializer is noisy and not terribly idiomatic - you can avoid that by using e.g. `LocationSerializer.attributes_to_serialize.keys`. That should remove most of the noise in this PR and isolate the bulk of the changes to just the new `MovesSerializer`.

The other concern is that generating db includes from request includes needs to be well tested and guarded against potential malicious requests - I'd like to see some coverage against unknown or omitted includes at least.

app/services/moves/finder.rb
smoothcontract: You can do this all on one line too, so maybe the method is superfluous and could be replaced on line 22 with

```ruby
scope = scope.includes(db_includes)
```

app/services/moves/finder.rb
```diff
@@ -16,10 +14,12 @@ def initialize(filter_params, ability, order_params)
                            # default if no 'by' parameter is date descending
                            :desc
                          end
+      @db_includes = db_includes
     end
 
     def call
       scope = apply_filters(Move)
+      scope = apply_includes(scope)
```
smoothcontract, line 22:
```suggestion
      scope = scope.includes(db_includes)
```

spec/requests/api/moves_controller_index_spec.rb
```diff
@@ -44,7 +44,7 @@
 
           get_moves
 
-          expect(Moves::Finder).to have_received(:new).with({ from_location_id: from_location_id }, ability, {})
+          expect(Moves::Finder).to have_received(:new).with({ from_location_id: from_location_id }, ability, {}, any_args)
```
smoothcontract, line 47: `any_args` is a bit too vague I think - ditto the same thing for v2 below

app/controllers/api_controller.rb
```diff
@@ -223,7 +223,25 @@ def validate_include_params
...
+  end
+
+  def included_db_relationships
+    @included_db_relationships ||= included_relationships.map { |value| cast_relationship_to_symbol_hash(value) } if included_relationships.present?
+  end
+
+  def cast_relationship_to_symbol_hash(value)
+    parts = value.split('.', 2)
+    if parts.length == 1
+      # special hack to allow for the flags table renamed in the backend but not in the FE includes
```
cwrw, line 236: can this comment be updated as this is isn't a rename but rather the choice of how we surface relationships in the API by omitting `framework_` from any PER resources

app/controllers/api_controller.rb
cwrw: There is a bit of misunderstanding to what this is. This isn't a rename issue, but rather how we have chosen to surface PER relationships in the API. We don't prefix any relationship with `framework_`, this includes `responses` as well as `questions`

cwrw: also I wouldn't say it's hacky in serialization, these were decisions we took at the time before mapping exact includes received in requests came in. Serializers provide support for having custom names for relationships, and we've utilised those correctly.

app/serializers/v2/moves_serializer.rb
```diff
@@ -0,0 +1,58 @@
...
+      rejection_reason
+      status
+      time_due
+      updated_at
+    ].freeze
+
+    INCLUDED_FIELDS = {
+      moves: INCLUDED_ATTRIBUTES +
+        %i[profile from_location to_location prison_transfer_reason supplier additional_information],
+      profiles: ::V2::ProfileSerializer::INCLUDED_ATTRIBUTES + %i[person],
```
cwrw, line 28: we've previously seen an issue with having nested includes also duplicated at the higher level, is there a risk here where that will cause similar issues if we introduce new filters?

app/serializers/v2/moves_serializer.rb
```diff
@@ -0,0 +1,58 @@
...
+      people: ::V2::PersonSerializer::INCLUDED_ATTRIBUTES + %i[gender ethnicity],
+      locations: ::LocationSerializer::INCLUDED_ATTRIBUTES,
+      prison_transfer_reasons: ::PrisonTransferReasonSerializer::INCLUDED_ATTRIBUTES,
+      suppliers: ::SupplierSerializer::INCLUDED_ATTRIBUTES,
+
+      # allocation: %i[to_location from_location moves_count created_at],
```
cwrw, line 34: are we not including allocations here? Also I'm wondering why we need the included fields. From the single move serialiser it seems this was to avoid loading too much off the allocation resource, but here those will come in as expected from the serializers anyway, and we might not need this?

app/controllers/api_controller.rb
willfish: Just good to not encourage it, yeah

app/serializers/v2/moves_serializer.rb
martyn-w: now switched to `attributes_to_serialize.keys`

app/services/moves/finder.rb
martyn-w: named parameters now added to finder 👍

app/controllers/api/moves_controller.rb
martyn-w: added `csv_includes` as a constant

app/controllers/api_controller.rb
martyn-w: I've replaced this with a `db_alias` method in `app/services/include_param_handler.rb`

app/serializers/v2/moves_serializer.rb
martyn-w: I don't understand this point?
my aim here is to restrict the serializer for the moves index endpoint to all attributes plus a few select relationships

app/serializers/v2/moves_serializer.rb
martyn-w: I've re-added allocations to the moves serializer; though I wasn't sure if it was being used or not. In any case it is now supported.

app/controllers/api/v1/people_actions.rb
```diff
@@ -76,12 +76,12 @@ def supported_relationships
...
 
     def included_relationships
-      IncludeParamHandler.new(params).call || PersonSerializer::SUPPORTED_RELATIONSHIPS
+      include_params_handler.included_relationships || PersonSerializer::SUPPORTED_RELATIONSHIPS
     end
 
     def other_included_relationships
       # Custom included relationships to avoid the people actions relationships
-      @other_included_relationships ||= IncludeParamHandler.new(params).call || Api::PeopleController::OTHER_SUPPORTED_RELATIONSHIPS
+      @other_included_relationships ||= IncludeParamHandler.new(params).included_relationships || Api::PeopleController::OTHER_SUPPORTED_RELATIONSHIPS
```
cwrw, line 84: can we use `include_params_handler.included_relationships` instead of `IncludeParamHandler.new(params).included_relationships`?

app/serializers/v2/moves_serializer.rb
cwrw: the included field is indeed smaller on the index endpoint, however previously we've seen issues where specific filters might cause larger resources to be pulled in thus introducing some performance issues. I think since you've had a look/rejigged the ready in transit filter, and all the other filters have been tested on the FE we're ok.

cwrw (approved):
Looks good, great work with performance, a minor comment otherwise 👍

martyn-w:
Hi @smoothcontract just responding on your comment about validating `?includes` - we already do this via the `SUPPORTED_RELATIONSHIPS` constants - e.g. asking for `/api/moves?include=something.dodgy` returns:

```
{
  "errors": [
    {
      "title": "Bad request",
      "detail": "[\"something.dodgy\"] is not supported. Valid values are: [\"profile\", \"profile.person\", \"profile.person.ethnicity\", \"profile.person.gender\", \"profile.person_escort_record\", \"profile.person_escort_record.flags\", \"profile.person_escort_record.framework\", \"profile.person_escort_record.responses\", \"from_location\", \"to_location\", \"prison_transfer_reason\", \"supplier\", \"allocation\"]"
    }
  ]
}
```

smoothcontract (approved):
Amazing work @martyn-w and thanks for taking on board all the feedback. Here's hoping this thing flies now! 🥇

PR #1060, opened 2020-10-21
P4-2325 Import `prefill` value from a Framework

cwrw:
jira-ticket: https://dsdmoj.atlassian.net/browse/P4-2353

Add a new column `prefill` to framework questions and import this value when pulling in a new framework. This column will determine if a question response should be prefilled or not when creating the Person escort record, with an existing confirmed person escort record. This value can be `nil` to support add multiple questions which allow dependent questions to determine their `prefill` status.


willfish (approved):
LGTM

PR #1061, opened 2020-10-21
P4-2356 Surface pre-filling fields and relationships

cwrw:
jira-ticket: https://dsdmoj.atlassian.net/browse/P4-2356

To surface if a response has been already prefilled by an existing person escort record similar response, add a new column `prefilled` to framework responses. This value will be set to true if on creation of a person escort record the value is set from a different response.

To capture the previous person escort record the current record is being prefilled from, add a new relationship to `prefill_source` on `person_escort_records`.

Surface this values in the relative serializers and update documentation.

### DIFF
--- a/db/migrate/20201021065138_add_prefill_source_to_person_escort_records.rb
+++ b/db/migrate/20201021065138_add_prefill_source_to_person_escort_records.rb
@@ -1,0 +1,5 @@
+class AddPrefillSourceToPersonEscortRecords < ActiveRecord::Migration[6.0]
+  def change
+    add_reference :person_escort_records, :prefill_source, type: :uuid, index: true, null: true
+  end
+end
--- a/spec/serializers/person_escort_record_serializer_spec.rb
+++ b/spec/serializers/person_escort_record_serializer_spec.rb
@@ -91,13 +91,15 @@ RSpec.describe PersonEscortRecordSerializer do
     expect(result[:data][:relationships][:prefill_source][:data]).to be_nil
   end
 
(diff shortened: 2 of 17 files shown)


db/migrate/20201021065138_add_prefill_source_to_person_escort_records.rb
```diff
@@ -0,0 +1,5 @@
+class AddPrefillSourceToPersonEscortRecords < ActiveRecord::Migration[6.0]
+  def change
+    add_reference :person_escort_records, :prefill_source, type: :uuid, index: true, null: true
```
willfish, line 3: Do we need the index?

spec/serializers/person_escort_record_serializer_spec.rb
```diff
@@ -87,6 +87,19 @@
     )
   end
 
+  it 'contains a nil `prefill_source` relationship if no prefill_source present' do
+    expect(result[:data][:relationships][:prefill_source][:data]).to be_nil
+  end
+
+  it 'contains a`prefill_source` relationship with person_escort_record prefill_source' do
+    person_escort_record.update(prefill_source: create(:person_escort_record))
```
willfish, line 95: Is it worth using a factory girl trait for the prefill source?

willfish (approved):
This is very clear and LGTM

db/migrate/20201021065138_add_prefill_source_to_person_escort_records.rb
cwrw: yea, I think its good practice to index references, even self references

PR #1063, opened 2020-10-21
Update swagger documentation to use Move instead of Profile

cwrw:
To create Person escort record, switch documentation to use move instead of profile, however we still surface a profile in
relationships.


smoothcontract (approved):
LGTM 👍

PR #1066, opened 2020-10-22
[P4-2394] Adds incident fault classification enum to events

willfish:
jira link: https://dsdmoj.atlassian.net/browse/P4-2394

Fault classification is needed on every request to create an incident
event.

- [x] Adds fault classification to the parent incident event model
- [x] Adds fault classification to the tests of the individual incident
  models via shared examples

### DIFF
--- a/app/models/generic_event/incident.rb
+++ b/app/models/generic_event/incident.rb
@@ -3,14 +3,21 @@ class GenericEvent
     LOCATION_ATTRIBUTE_KEY = :location_id
 
     def self.inherited(child)
-      child.details_attributes :supplier_personnel_numbers, :vehicle_reg, :reported_at
+      child.details_attributes :supplier_personnel_numbers, :vehicle_reg, :reported_at, :fault_classification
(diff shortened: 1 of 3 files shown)


app/models/generic_event/incident.rb
```diff
@@ -3,14 +3,21 @@ class Incident < GenericEvent
     LOCATION_ATTRIBUTE_KEY = :location_id
 
     def self.inherited(child)
-      child.details_attributes :supplier_personnel_numbers, :vehicle_reg, :reported_at
+      child.details_attributes :supplier_personnel_numbers, :vehicle_reg, :reported_at, :fault_classification
```
cwrw, line 6: if this is needed on every incident request, should we add a required validation, similar to `SupplierPersonnelNumberValidations`?

willfish: We should, yeah. I've actually documented that it needs to be present so :)

cwrw (approved):
👍

PR #1067, opened 2020-10-22
P4 2338 import prisoner category

martyn-w:
### Jira link

P4-2338

### What?

I have added/removed/altered:

- [x] Imports the prisoner category for people with a nomis_booking_id
- [x] Regenerated wiremock captures

### Why?

I am doing this because:

- so that we can validate and prevent Cat A prisoners from using the BaSM service

- fairly minimal


app/services/profiles/import_prisoner_category.rb
```diff
@@ -0,0 +1,12 @@
+module Profiles
+  class ImportPrisonerCategory
+    def initialize(profile)
+      @profile = profile
+      @nomis_booking_id = @profile&.person&.latest_nomis_booking_id
+    end
+
+    def call
+      @profile.update(GetPrisonerCategoryAttributes.new(@nomis_booking_id).call) if @nomis_booking_id.present?
```
cwrw, line 9: do we need the nomis booking id check here as well as `GetPrisonerCategoryAttributes`? This will avoid updating/clearing the category data if it changes, but if we keep it on `GetPrisonerCategoryAttributes` and remove here we will always update it

app/lib/nomis_client/prisoner_category.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module NomisClient
+  class PrisonerCategory
```
cwrw, line 4: Since this class is querying the generic bookings endpoint, maybe we should avoid naming it `PrisonCatgeory` and go with something more general incase we extend it to pull more attributes later?

lib/tasks/refresh_prisoner_category.rake
```diff
@@ -0,0 +1,15 @@
+namespace :prisoner_category do
+  desc 'Refresh prisoner category'
+  task refresh_data: :environment do
+    profiles = Profile.joins(:person).where.not(people: { latest_nomis_booking_id: nil })
+    count = profiles.count
+    i = 1
+
+    puts "Refreshing #{count} profiles"
+    profiles.find_each do |profile|
```
cwrw, line 9: maybe we can utilise `find_index` here instead?
```
profiles.find_each.with_index do |profile, index|
  ...
end
```

spec/fixtures/files/nomis/get_booking_details_200.json
```diff
@@ -0,0 +1,364 @@
...
+    }
+  ],
+  "csra": "Standard",
+  "category": "Cat B",
+  "categoryCode": "B",
+  "birthPlace": "SCOTLAND",
+  "birthCountryCode": "SCOT",
+  "inOutStatus": "IN",
+  "status": "ACTIVE IN"
+}
```
cwrw, line 364: 🛑  (no new line eof) same in other fixture

app/services/profiles/import_prisoner_category.rb
cwrw: update this is only for the rake task so is ok

app/lib/nomis_client/prisoner_category.rb
```diff
@@ -0,0 +1,24 @@
...
+    class << self
+      def get(nomis_booking_id)
+        attributes_for(
+          get_response(nomis_booking_id),
+        )
+      end
+
+      def get_response(nomis_booking_id)
+        NomisClient::Base.get("/bookings/#{nomis_booking_id}").parsed
```
willfish, line 13: Given how important this is, do we want to propagate a nomis failure and store the result of the `get`/post it to Sentry?

app/services/profiles/get_prisoner_category_attributes.rb
```diff
@@ -0,0 +1,18 @@
...
+    def initialize(nomis_booking_id)
+      @nomis_booking_id = nomis_booking_id
+    end
+
+    def call
+      if @nomis_booking_id.present?
+        NomisClient::PrisonerCategory.get(@nomis_booking_id)
+      else
+        {
+          category: nil,
```
willfish, line 12: I'm wondering whether this service is necessary? If we only merge in the attributes coming back from the client when there is a booking id then I think we get the same behaviour as if these attributes were nil? Same situation for the rake task I guess. And in fact, Looking at the ImportPrisonerCategory service that's exactly what you are doing :).


```ruby
attributes.merge(prisoner_category_attributes) if nomis_booking_id.present?
```

db/migrate/20201021164648_add_category_to_profile.rb
```diff
@@ -0,0 +1,7 @@
+class AddCategoryToProfile < ActiveRecord::Migration[6.0]
+  def change
+    add_column :profiles, :category, :string, null: true
+    add_column :profiles, :category_code, :string, null: true
+    add_index :profiles, :category_code
```
willfish, line 5: Do we strictly need this index? I mean, aren't we going to just guard against permitting creation of moves for a profile for a person of category A as opposed to do a bunch of reporting on the code?

app/lib/nomis_client/prisoner_category.rb
willfish: Maybe we can call it a failure condition if we have a booking id and no category, actually.

willfish: Since I'm guessing all prisoners have a category?

lib/tasks/refresh_prisoner_category.rake
martyn-w: yep, good shout 👍

app/lib/nomis_client/prisoner_category.rb
martyn-w: hmm, yes I ummmmed and arrrred over that, but happy to rename to something more generic (`booking_details`?) if that is the consensus view

spec/fixtures/files/nomis/get_booking_details_200.json
martyn-w: fixed 👍

app/lib/nomis_client/prisoner_category.rb
martyn-w: Unfortunately, no, not all bookings (prisoners) seem to have a category - so we need to handle this gracefully rather than throw an error which could prevent a profile from being saved and a move from taking place. This is the output from the category synchroniser (for staging)  and you can see some have no category:

```
...
processed 975 of 995: fa8c4069-4a06-4cb3-82bf-142d39b1d827 -> Unsentenced
processed 976 of 995: fb567d4d-538b-480d-b6b7-ba520a5fe9ca -> Cat C
processed 977 of 995: fb7349a8-6362-470d-a2f3-35093a388edb -> Unsentenced
processed 978 of 995: fbb11e79-623e-4bd5-8fc7-2b0236f7a4e0 -> Cat C
processed 979 of 995: fbf9e523-6cce-4c75-aa8a-ecd687156c04 -> Cat C
processed 980 of 995: fc409b2e-0194-4d0c-9d76-5778f82f01d8 -> Cat C
processed 981 of 995: fc8fbf74-f24d-4eff-8ee7-d7054153e079 -> Cat B
processed 982 of 995: fc95c683-819b-4d07-afc1-ead49b744982 -> 
processed 983 of 995: fd2f1e07-14e2-491a-9c8f-d21f90821214 -> Unsentenced
processed 984 of 995: fd695b86-4bd6-40c6-9a8a-fe93126d3389 -> Unsentenced
processed 985 of 995: fda552e2-1139-498d-be1d-34c5c645f8e2 -> Cat C
processed 986 of 995: fe4a9556-d064-497d-80fb-81414bf3399c -> Cat C
processed 987 of 995: fe6ea530-eb8f-46bd-a553-5480f984b91f -> Unsentenced
processed 988 of 995: fec9a80c-8b2f-46bb-9469-ed83c28e60e4 -> Cat C
processed 989 of 995: ff00f677-3370-495e-b58c-95c35fc2c440 -> Cat C
processed 990 of 995: ff13213d-72a8-4654-b849-b24a7c3640a9 -> Cat C
processed 991 of 995: ff2b4570-6dce-4416-ab6b-078e4a2df7ad -> 
processed 992 of 995: ffb18270-de31-4c91-abb6-c98651582dba -> 
processed 993 of 995: fff01cea-ea4d-4a6b-b8cd-17584e4e958b -> Unsentenced
processed 994 of 995: fffed775-c645-4665-aec8-f1338bd65e83 -> Cat B
```

db/migrate/20201021164648_add_category_to_profile.rb
martyn-w: Yep, you're right, we don't strictly need this index.

It is here to help address ad hoc questions like "how many Category B prisoners have we moved?". I don't think the index size is too overbearing - it is just over 1mb for the staging database of 53,154 profiles.

app/services/profiles/get_prisoner_category_attributes.rb
martyn-w: yep, the `GetPrisonerCategoryAttributes` service isn't doing very much. I was trying to follow the existing pattern of calling the Nomis client through an intermediate service rather than directly in the controller. But I agree - it would be simply to just disband with this particular service - I'll remove it.

db/migrate/20201021164648_add_category_to_profile.rb
willfish: :+1:

willfish (approved):

PR #1069, opened 2020-10-22
Ensure that metadata is only provided for paginated subset of locations

smoothcontract:
### Jira link

P4-1427

### What?

- [x] Only calculate location free spaces data for paginated subset of locations
- [x] Add missing spec coverage for `LocationFreeSpacesSerializer`

### Why?

- Whilst working on another story to add metadata for allocations I realised that the existing metadata implementation for `GET /location_free_spaces` was calculating free space data for every location, rather than just the ones currently selected according to the pagination parameters. Cleaned this up by adding callback support in the `paginate` method to expose the paginated collection and options prior to instantiating the serializer. Also simplified the structure for passing parameters as only one parameter is currently needed.

- Doesn't affect api output and is not yet used in production

### DIFF
--- a/app/controllers/concerns/serializable.rb
+++ b/app/controllers/concerns/serializable.rb
@@ -8,6 +8,8 @@ module Serializable
 
     paginated_collection = paginate_collection(collection, options)
 
+    yield(paginated_collection, options) if block_given?
+
     options[:meta] = meta(paginated_collection, options)
     options[:links] = links(paginated_collection, options)
     options[:json] = build_serializer(paginated_collection, options)
--- a/spec/serializers/location_free_spaces_serializer_spec.rb
+++ b/spec/serializers/location_free_spaces_serializer_spec.rb
@@ -25,6 +25,10 @@ RSpec.describe LocationFreeSpacesSerializer do
     it 'contains a title attribute' do
       expect(attributes[:title]).to eql location.title
     end
+
+    it 'contains empty meta data' do
+      expect(meta).to eql({ populations: nil })
+    end
   end
 
   context 'with custom params' do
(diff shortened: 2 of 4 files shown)


spec/serializers/location_free_spaces_serializer_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  context 'with custom params' do
+    let(:adapter_options) do
+      { params: {
+        location.id => { foo: 'bar' },
+      } }
+    end
+
+    it 'contains meta data' do
+      expect(meta).to eql({ populations: { foo: 'bar' } })
```
cwrw, line 38: what does meta look like when we don't pass in custom params, is it `nil` or populations `nil`? can we add a test for that?

app/controllers/concerns/serializable.rb
```diff
@@ -8,6 +8,8 @@ def paginate(*args)
 
     paginated_collection = paginate_collection(collection, options)
 
+    yield(paginated_collection, options) if block_given?
```
cwrw, line 11: can you add a test for this? or maybe at the controller level that could cover this and the code above?

smoothcontract: This is covered by the request spec for `GET /location_free_spaces`.

spec/serializers/location_free_spaces_serializer_spec.rb
smoothcontract: Added a spec for this

app/controllers/concerns/serializable.rb
smoothcontract: https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/spec/requests/api/populations_controller_index_spec.rb#L38

cwrw (approved):

app/controllers/concerns/serializable.rb
smoothcontract: https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/spec/requests/mock/pagination_controller_spec.rb

PR #1070, opened 2020-10-22
Add support for custom move count metadata on `GET /allocations`

smoothcontract:
### Jira link

P4-2340

### What?

- [x] Add support for custom move count metadata on `GET /allocations`
- [x] Add new `move_totals` class method to `Allocation` model to calculate a hash of allocation id to total and filled moves count
- [x] Include this additional data as a param when calling `AllocationsSerializer` - this descends from `AllocationSerializer` to add custom `meta` tag.

### Why?

- This will allow the front end to use the provided count of total and filled moves per allocation, rather than having to request all of the associated move and profile resources per allocation and derive this calculation. Once the meta data is implemented within the front end the list of supported includes on the `GET /allocations` endpoint can be trimmed back to something more sensible - and ideally remove support for requesting moves entirely as this is a slow operation.
- This does add an extra query to derive the aggregated totals but it's extremely fast so shouldn't affect production performance significantly - during local testing against a snapshot of staging it takes around 6ms for ~400 allocations.
- It's not possible to add optional `meta` tags with the serializer library we're using so I've added a custom serializer for the `GET /allocations` endpoint; this derives from `AllocationSerializer` so currently has the same attributes and relationships but will likely be modified further to remove support for some includes in future once the front end no longer has a dependency on those.

### Have you? (optional)

- [ ] Updated API docs if necessary - not sure how/if we need to represent `meta` tags in API documentation 🤔 

- Extends existing API with `meta` tags so shouldn't be a breaking change for front end. Have tagged as breaking just in case though so we can test this works without the corresponding front end changes being made.

### DIFF
--- a/app/controllers/api/allocations_controller.rb
+++ b/app/controllers/api/allocations_controller.rb
@@ -8,7 +8,9 @@ module Api
       allocations_params = Allocations::ParamsValidator.new(filter_params, sort_params)
       if allocations_params.valid?
         allocations = Allocations::Finder.new(filters: filter_params, ordering: sort_params, search: search_params).call
-        paginate allocations, serializer: AllocationSerializer, include: included_relationships
+        paginate allocations, serializer: AllocationsSerializer, include: included_relationships do |paginated_allocations, options|
+          options[:params] = paginated_allocations.move_totals
+        end
       else
         render json: { error: allocations_params.errors }, status: :bad_request
       end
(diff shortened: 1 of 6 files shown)


smoothcontract:
Alex has confirmed this won't affect the existing front end master branch (it ignores any extra attributes or keys) so removing the breaking change flag 🎏

cwrw (approved):
looks good, and well annotated :D I think my only comment would be to see testing for multiple moves per allocation, but otherwise 👍

PR #1071, opened 2020-10-23
P4-2354 Prefill Person Escort Records on Creation

cwrw:
jira-ticket: https://dsdmoj.atlassian.net/browse/P4-2354

Allow Person escort records to be pre-filled by a previous, confirmed Person escort record if it exists on the same person.
If the feature flag is enabled for pre-fill, the previous person escort record is retrieved and is set on the `prefill_source` of a Person escort record.
To map previous responses to current responses, question keys are used to build a hash map to the response. This is injected into the response builder for creation.

A response is added to the hash map if a question should be pre-filled. Add multiple items questions are slightly different as the parent question does not hold the pre-fill value, but the dependent questions themselves. This allows flexibility in multiple item questions, where not all questions should adhere to the same rules. To build the multiple item value, dependent questions are traversed to retrieve the pre-fill value, then used to determine if the value should be added to the final responses value for an item.
Feature flag support was also added to tests, to enable testing the Flipper feature flags.

Documentation has already been added for `prefill_source` on the PER and `prefilled` value on responses.

### DIFF
--- a/spec/models/framework_response_spec.rb
+++ b/spec/models/framework_response_spec.rb
@@ -360,6 +360,24 @@ RSpec.describe FrameworkResponse do
 
         expect(child_response.reload.responded).to be(true)
       end
+
+      it 'sets prefilled to false on dependent responses if value changed' do
+        parent_response = create(:collection_response, :details)
+        child_question = create(:framework_question, :checkbox, followup_comment: true, dependent_value: 'Level 1', parent: parent_response.framework_question)
+        child_response = create(:collection_response, :details, framework_question: child_question, parent: parent_response, responded: true, prefilled: true)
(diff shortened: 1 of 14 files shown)


spec/models/framework_response_spec.rb
```diff
@@ -92,72 +92,6 @@
     end
   end
 
-  describe '.requires_value?' do
```
smoothcontract, line 95: Is this now covered elsewhere?

smoothcontract: Ignore that, I can see it's just moved 👍

smoothcontract (approved):
Really nice implementation, splendid stuff 👍

PR #1073, opened 2020-10-23
Add log warn to OAuth2::Errors

Thource:
### Jira link

P4-2128

### What?

I have added/removed/altered:

- [x] Added logging to OAuth2::Errors raised through NomisClient
- [x] Added unit tests to cover OAuth2::Error logging and Faraday error logging

### Why?

I am doing this because:

- It will make it more clear where the error is coming from

### DIFF
--- a/app/lib/nomis_client/base.rb
+++ b/app/lib/nomis_client/base.rb
@@ -29,6 +29,9 @@ module NomisClient
       rescue Faraday::ConnectionFailed, Faraday::TimeoutError => e
         Rails.logger.warn "Nomis Connection Error: #{e.message}"
         raise e
+      rescue OAuth2::Error => e
+        Rails.logger.warn "Nomis OAuth Client Error: #{e.message}"
+        raise e
       end
 
       def token
--- a/spec/lib/nomis_client/base_spec.rb
+++ b/spec/lib/nomis_client/base_spec.rb
@@ -306,4 +306,40 @@ RSpec.describe NomisClient::Base do
       end
     end
   end
+
+  describe '.token_request' do
+    let(:response) { described_class.get(api_endpoint) }
+    let(:api_endpoint) { '/movements/transfers/BXI' }
+    let(:token_expires_at) { 1.hour.from_now.to_i }
+
+    [
+      { error: OAuth2::Error, log_text: 'Nomis OAuth Client Error:' },
+      { error: Faraday::ConnectionFailed, log_text: 'Nomis Connection Error:' },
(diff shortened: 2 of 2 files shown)


willfish (approved):
Seems fine to me

willfish:
Congrats on first PR 🎉

app/lib/nomis_client/base.rb
```diff
@@ -29,6 +29,9 @@ def token_request(method, path, params)
       rescue Faraday::ConnectionFailed, Faraday::TimeoutError => e
         Rails.logger.warn "Nomis Connection Error: #{e.message}"
         raise e
+      rescue OAuth2::Error => e
```
cwrw, line 32: This seems similar to the above rescue block, maybe we can combine them?

Nimphal: The message is specific to the errors, I can't think of a neat way to combine them and retain the message granularity, but keen to learn!

spec/lib/nomis_client/base_spec.rb
```diff
@@ -306,4 +306,40 @@
...
+          allow(Rails).to receive(:logger).and_return(instance_spy('logger'))
+        end
+
+        %i[post put get].each do |method|
+          context "when the request type is #{method}" do
+            before do
+              allow(token).to receive(method).and_raise(oauth2_error)
+            end
+
+            it 'logs additional info, describing what the error is' do
```
cwrw, line 336: Generally I don't think we should be testing logging in the app as its usually superfluous, but happy to hear other's thoughts around this.

app/lib/nomis_client/base.rb
Thource: The only thing I can think of would be something like

```ruby
Rails.logger.warn "Nomis #{e.is_a? OAuth2::Error ? 'Oauth Client' : 'Connection'} Error: #{e.message}"
```

cwrw: That's fair, I was hoping the error message would give us the details of the error itself, but happy to keep as is

spec/lib/nomis_client/base_spec.rb
willfish: I don't think I agree. It's a test that tells us we're logging errors (which is important) and doesn't have much cost for the value it gives us if we no longer log these due to a regression. If we were logging other stuff like random unimportant messages I'd argue that this shouldn't be tested. But the value aspect makes me think this is fine.

PR #1074, opened 2020-10-25
Update ENV variables to latest NOMIS envs

cwrw:
Update example and test ENVs to remove old variables for NOMIS. Also update Heroku app.json to pull in variables.
The variables have been updated on Heroku itself, but I'll cleanup those after the PR goes in.


willfish (approved):
Thanks for doing this.

PR #1076, opened 2020-10-26
IM-60 Retry long transactions if they fail

cwrw:
jira-ticket: https://dsdmoj.atlassian.net/browse/IM-60

When deploying migrations, the database schema becomes invalid. If there were transactions midway during such a deploy a
`ActiveRecord::PreparedStatementCacheExpired` error is thrown. To avoid those errors on deploy, wrap all long running transacions with a custom retriable transaction, which retries once if this error is raised, but fails if it fails another time. All transactions to be retried are safe, but are long running so are prone to raise such errors on deploy.

Reference: https://flexport.engineering/avoiding-activerecord-preparedstatementcacheexpired-errors-4499a4f961cf

### DIFF
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -4,4 +4,17 @@ class ApplicationRecord < ActiveRecord::Base
   self.abstract_class = true
 
   include UpdatedAtRange
+  class << self
+    def retriable_transaction(**options, &block)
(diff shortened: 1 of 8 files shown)


martyn-w (approved):
👍 LGTM

PR #1081, opened 2020-10-27
Add unfilled moves count to allocations meta

smoothcontract:
### Jira link

P4-2340

### What?

- [x] Also include `unfilled` moves count in the allocation `meta` tag for `GET /allocations` endpoint

### Why?

- This allows further simplification of front end code and doesn't impact backend performance so feels like a nicer place to implement this logic.

### Have you? (optional)

- [ ] Updated API docs if necessary - `meta` tags are not documented, still need to consider if this is worth doing 🤔 

- Updates `meta` for production allocations api (which is currently ignored in production), the change ties in with ongoing front end work and has been discussed with @solidgoldpig


cwrw (approved):

solidgoldpig (approved):

cwrw (commented):
do we want to update documentation as well?

smoothcontract:
> do we want to update documentation as well?

The meta stuff isn't documented yet, I'll have a think about that in a follow up PR as there's no established pattern in the swagger docs for those at present.

PR #1085, opened 2020-10-27
P4 2339 validate prisoner category

martyn-w:
### Jira link

P4-2339

### What?

I have added/removed/altered:

- [x] Added validation of prisoner category 

### Why?

I am doing this because:

- Cat-A prisoners (and associated categories H and E) should not be moved by BaSM

### Have you? (optional)

### DIFF
--- a/spec/requests/api/moves_controller_create_v2_spec.rb
+++ b/spec/requests/api/moves_controller_create_v2_spec.rb
@@ -513,6 +513,32 @@ RSpec.describe Api::MovesController do
         end
       end
     end
+
+    context 'when the profile is for a Cat A prisoner' do
+      let(:profile) { create(:profile, :category_a) }
+      let(:errors_422) do
+        [
+          {
+            'title' => 'Unprocessable entity',
+            'detail' => "Profile person is a category 'a' prisoner and cannot be moved using this service",
+            'source' => { 'pointer' => '/data/attributes/profile' },
+            'code' => 'unsupported_prisoner_category',
+          },
+        ]
+      end
+
+      it_behaves_like 'an endpoint that responds with error 422' do
+        before { do_post }
+      end
+    end
+
+    context 'when the profile is for a Cat B prisoner' do
+      let(:profile) { create(:profile, :category_b) }
+
+      it_behaves_like 'an endpoint that responds with success 201' do
+        before { do_post }
+      end
+    end
   end
 
   def do_post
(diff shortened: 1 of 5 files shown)


spec/requests/api/moves_controller_create_v2_spec.rb
```diff
@@ -513,6 +513,32 @@
...
       end
     end
+
+    context 'when the profile is for a Cat A prisoner' do
+      let(:profile) { create(:profile, :category_a) }
+      let(:errors_422) do
+        [
+          {
+            'title' => 'Unprocessable entity',
+            'detail' => "Profile person is a category 'a' prisoner and cannot be moved using this service",
```
Nimphal, line 523: Super minor, but should this be capital `A`?

cwrw: very good question, in our API controller we `.humanize` the detail message, which in turn downcases the `A`:
https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/controllers/api_controller.rb#L182

Nimphal: Ah, that's a bit inconvenient :D We've had to end up implementing words that are exceptions in other projects, so that things like HMP and YOI don't get downcased.

cwrw (approved):
looks good 👍  Is the FE going to be surfacing a similar message to users when they see this error? Also how will this work with existing users in the system that are Cat A, as moves can no longer be updated so can't be cancelled etc.?

spec/requests/api/moves_controller_create_v2_spec.rb
martyn-w: yep its some annoying rails magic :-|

martyn-w:
Hi @cwrw 

> Is the FE going to be surfacing a similar message to users when they see this error?

Yes, the FE will be updated to check the category of the prisoner (P4-2310). However, I see that they want to do that check _before_ the move is created, so that will require a new endpoint, or perhaps extra attributes on an existing endpoint. I'll chat with @solidgoldpig about what he prefers.

> Also how will this work with existing people in the system that are Cat A, as moves can no longer be updated so can't be cancelled etc.?

Good question. I have checked and fortunately we have very few records in our system for Cat-A, and none are for current moves, so we should be fine in that regard.

PR #1090, opened 2020-10-28
P4 2416 categories endpoint

martyn-w:
### Jira link

P4-2416

### What?

I have added/removed/altered:

- [x] Added a new endpoint, `GET /people/{person_id}?include=category` (for v2 only)
- [x] Added a new endpoint, `GET /reference/categories`
- [x] Added supporting tests and swagger docs

### Why?

I am doing this because:

- So the front end is able to check the category of a prisoner prior to creating a move - so that Cat-A prisoners are flagged at the start of the process
- Needed for P4-2310

### Have you? (optional)

- minimal


martyn-w:
Blocked to harmonise with PMU categories work with @smoothcontract

app/serializers/v2/people_serializer.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module V2
```
cwrw, line 3: maybe we can leverage inheritance here to avoid duplication in the serializer, this is a pattern we have adopted with the Allocation serializer

app/models/person.rb
```diff
@@ -66,4 +66,14 @@ def for_feed
   def latest_person_escort_record
     person_escort_records.where(status: 'confirmed').order(confirmed_at: :desc).first
   end
+
+  def category
+    if latest_nomis_booking_id.present?
+      @category ||= begin
```
cwrw, line 72: I've not been involved in the discussions around why we want to expose the category on a person in realtime vs using profiles (which might not be up to date with the latest category as that represents those at that specific time), can you walk me through which dashboards/process we need to pull in this value for?

spec/models/category_spec.rb
```diff
@@ -11,5 +11,8 @@
   it { is_expected.to validate_presence_of(:key) }
   it { is_expected.to validate_uniqueness_of(:key) }
   it { is_expected.to validate_presence_of(:title) }
-  it { is_expected.to validate_inclusion_of(:move_supported).in_array([true, false]) }
```
cwrw, line 14: I've just merge a pr for this 😬 so we can remove this now with the comment

app/serializers/v2/people_serializer.rb
smoothcontract: Agree, that would be much cleaner 👍

app/controllers/api_controller.rb
```diff
@@ -24,6 +24,7 @@ class ApiController < ApplicationController
   rescue_from CanCan::AccessDenied, with: :render_unauthorized_error
   rescue_from ActiveModel::ValidationError, with: :render_validation_error
   rescue_from IncludeParamsValidator::ValidationError, with: :render_include_validation_error
+  rescue_from NotSupportedInOldVersionError, with: :render_not_supported_in_old_version_error
```
smoothcontract, line 27: Love this ❤️

app/controllers/api_controller.rb
```diff
@@ -222,6 +223,20 @@ def render_include_validation_error(exception)
...
 
+  def render_not_supported_in_old_version_error
+    render(
+      json: {
+        errors: [{
+          title: 'Not Supported In Old Version Error',
+          detail: "This endpoint is not supported in version v#{api_version} - please change the ACCEPT header to a newer version",
+        }],
+      },
+      # NB: 406 Not Acceptable (The resource identified by the request is only capable of generating response entities
```
smoothcontract, line 234: Don't think this comment is really necessary - the return code seems fine and is pretty much self documenting

app/models/person.rb
smoothcontract: The use case is that the front end can call `GET /people/:id` endpoint to verify that the prisoner is not in a high risk category and should be prevented from being moved; this can happen early in the UX journey rather than the current approach of raising an error when the new move is created in the backend.

We always want to check the latest prisoner category in nomis for this use case.. however, it would seem reasonable to store the `category_id` in the profile though once we've fetched this data so it can be useful for reporting or other less critical endpoints.

app/serializers/v2/people_serializer.rb
martyn-w: that's a good shout; I'll see if that can work here 👍

smoothcontract: Presumably this is just a rename of the old `PersonSerializer` ?

app/models/person.rb
martyn-w: Yes sure:
When creating a move it would be good to know in advance of creating the move and profile records whether the person being moved is a Cat-A prisoner, and if so don't allow the move. We can't check the profile record because it doesn't yet exist when the move is initiated.

From a UI point of view, my thinking is: 
1) search for the person to be moved
2) select the person record from the list presented
3) check whether they can be moved by calling `GET /people/id?include=category` which does a Nomis lookup
4) if they can be moved, then continue to as normal to fill in the PTR. Otherwise exist with a message about not moving Cat-A prisoners.

spec/models/category_spec.rb
martyn-w: 👍

app/serializers/v2/person_serializer.rb
```diff
@@ -19,10 +19,13 @@ class PersonSerializer
     has_one :ethnicity, serializer: EthnicitySerializer
     has_one :gender, serializer: GenderSerializer
 
+    # NB: we need to lazy load the category to prevent an unnecessary call to Nomis
+    has_one :category, serializer: CategorySerializer, id_method_name: :key, lazy_load_data: true, &:category
```
smoothcontract, line 23: Not sure this needs to have any special treatment - ideally we should persist the category against the profile when the Nomis API call completes. I'd prefer that the lookup happens explicitly in the people controller and the category relationship is a standard `belongs_to :category` (or maybe `has_one` - whichever you feel is more semantically correct 🤔 ) in the `Person` model.

Also this file could be simplified using inheritance as per @cwrw suggestion...

```ruby
module V2
  class PersonSerializer < PeopleSerializer
     has_one :category, serializer: CategorySerializer

    SUPPORTED_RELATIONSHIPS = %w[ethnicity gender profiles category].freeze
  end
end
```

Sorry to say this but I think it would also be good to migrate the existing `category` attribute on the `Person` model to the corresponding `category_id`. This moves closer to an idiomatic approach which is generally preferred.

app/controllers/api/v2/people_actions.rb
```diff
@@ -14,6 +14,10 @@ def create_and_render
       render_person(person, :created)
     end
 
+    def show_and_render
+      render_person(person, :ok)
```
smoothcontract, line 18: See comments further on but I think this is the point at which an explicit refresh of person details from Nomis should be triggered, rather than as a side effect of querying the persons `category`. It would be clearer to understand exactly what's happening and mean more idiomatic implementation of the associated serializer and model. Something like

```ruby
def show_and_render
  person.refresh_from_nomis!
  render_person(person, :ok)
end
```

app/serializers/v2/profile_serializer.rb
```diff
@@ -7,7 +7,7 @@ class V2::ProfileSerializer
 
   attributes :assessment_answers
 
-  belongs_to :person, serializer: ::V2::PersonSerializer
+  belongs_to :person, serializer: ::V2::PeopleSerializer
```
smoothcontract, line 10: I can see the intent here to avoid introducing a potentially breaking change to the existing API with the new category relationship, but it's also stretching things a little to use the `PeopleSerializer` to return a single person. Perhaps this needs a little renaming to better align... or potentially could we check with suppliers that they can tolerate the introduction of a new relationship without breaking things?

So either rename something like this

* existing `person_serializer.rb` <- retain the same name
* new `person_serializer.rb` <- rename to `person_with_category_serializer.rb`

or... just stick with a single person serializer for V2 that includes the new `category` relationship. If this is an idiomatic belongs_to or has_one model relationship (potentially with lazy loading if needed) then I wouldn't think that introduces any performance penalty as we only perform the explicit Nomis lookup when invoking the new `GET /people/:id` endpoint.

profile_serializer.

smoothcontract (changes requested):
Much happier with the direction this is heading in, but I think it could still benefit from a touch more work to use an idiomatic implementation of `category` on the person model. I'd prefer that to be a boring old `has_one` or `belongs_to` relationship in the model that returns whatever `category_id` is currently persisted on the person rather than having a side effect of calling Nomis behind the scenes.

In turn that `category_id` can be updated when we explicitly refresh person details from Nomis, and ideally have the existing `category` attribute both migrated to the corresponding `category_id` and then be removed from the schema.

spec/swagger/swagger_doc_v2.yaml
```diff
@@ -719,3 +719,45 @@
             application/vnd.api+json:
               schema:
                 "$ref": "../v2/post_event_responses.yaml#/422"
+  "/categories":
```
cwrw, line 722: can you also update the documentation to include the new endpoint for getting a single person?

app/models/person.rb
```diff
@@ -66,4 +66,14 @@ def for_feed
...
+                      booking_details = NomisClient::BookingDetails.get(latest_nomis_booking_id)
```
cwrw, line 73: We currently rescue timeout and connection errors, however if NOMIS returns an internal error (surfaced as an oauth error) here the whole step would fail, should we instead gracefully rescue that error with the correct messaging for clients?

spec/swagger/swagger_doc_v2.yaml
```diff
@@ -719,3 +719,45 @@
...
+            application/vnd.api+json:
+              schema:
+                "$ref": "../v2/get_categories_responses.yaml#/404"
+        "200":
+          description: success
+          content:
+            application/vnd.api+json:
+              schema:
+                "$ref": "../v2/get_categories_responses.yaml#/200"
+        "504":
```
cwrw, line 758: will we ever get this error on categories 🤔  I think this will be on the GET person endpoint instead?

app/controllers/api/v2/people_actions.rb
smoothcontract: Discussed this offline, this is fine to leave as is.

app/serializers/v2/person_serializer.rb
smoothcontract: Discussed this offline, the `has_one` implementation of `category` you have here is fine as is, but would still be nice to introduce inheritance for a cleaner serializer.

app/serializers/v2/profile_serializer.rb
smoothcontract: Discussed offline -

smoothcontract (commented):
Discussed offline, after clearing (my) misunderstanding of people vs profile, it makes sense to look to move the **profile** `category` attribute to `category_id` instead; this will be a snapshot of the Nomis details when the profile is created. The **person** category can remain as it is here in this PR with a real time lookup of current Nomis prisoner category so that the front end check is against current data.

Also discussed making the change to profile category/category_id as a separate PR. I would still prefer to see improved naming of the `PersonSerializer`, either to retain a single serializer  that introduces `category` relationship for V2, or to add something like `PersonWithCategorySerializer` so that the naming is explicit. It feels a little unwieldy currently having `PersonSerializer` and `PeopleSerializer` but then not using them consistently in the expected places - i.e. using the `PeopleSerializer` for serializing a single person.

I may also steal the `CategorySerializer` and associated spec from this PR as I need the same thing for the PMU dashboard work 😉

app/serializers/v2/person_with_category_serializer.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module V2
+  class PersonWithCategorySerializer < ::V2::PersonSerializer
```
smoothcontract, line 4: Nicely done ❤️

app/serializers/v2/people_serializer.rb
martyn-w: To clarify, the original `person_serializer` was (temporarily) renamed `people_serializer`. It has now been renamed. back to `person_serializer`.

A new serializer, initially called `person_serializer` has been renamed `person_with_category_serializer` and it inherits from `person_serializer`.

app/models/person.rb
martyn-w: To confirm, we do store category in the profile record. At the moment this is an attribute, `category_code`. In a separate PR I will migrate this to `category_id` (ideally without upsetting the JPC reports).

Also, to confirm: as discussed offline, we will not store category at the person level, only at the profile level.

app/controllers/api_controller.rb
martyn-w: comment removed

martyn-w:
Also, to clarify that the category will likely be fetched twice from Nomis:

1. the first fetch will be before the profile and move record exist, to verify the person to be moved is not in an unmoveable category.
2. the second fetch will be when the profile record is realised and synchronised with Nomis.

app/serializers/v2/person_serializer.rb
martyn-w: The person serializers now have inheritance: `class PersonWithCategorySerializer < ::V2::PersonSerializer`

app/serializers/v2/profile_serializer.rb
martyn-w: the serializers have been renamed 👍

spec/swagger/swagger_doc_v2.yaml
martyn-w: Swagger docs updated to include `GET /people/{person_id}` 👍

spec/swagger/swagger_doc_v2.yaml
martyn-w: eek good spot! I'll fix this

app/models/person.rb
martyn-w: I'm not entirely sure how `OAuth2::Error` Nomis errors should generally be handled... but looking at the court bookings endpoint I've added code to log this to Sentry and then render it as a 504 Bad Gateway error. Is that what you had in mind?

swagger/v1/get_person_responses.yaml
```diff
@@ -0,0 +1,9 @@
+'406':
```
cwrw, line 1: Im not sure this is being used anywhere?

cwrw (approved):
Looks really good, thanks for addressing comments 👍  I only have one minor comment otherwise feel free to merge

app/models/person.rb
cwrw: Yes that perfect,  we should start adopting this pattern anywhere we might be surfacing 500s if NOMIS fails

app/serializers/category_serializer.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class CategorySerializer
+  include JSONAPI::Serializer
+
+  set_type :categories
+
+  attributes :key, :title, :move_supported
```
smoothcontract, line 8: After stealing this file for my PR, I made a slight tweak to use the standard `id` column, moving `key` to an attribute - I think that's a hangover from the original implementation of this endpoint that was creating a category on the fly essentially without a persisted UUID. Also added timestamp attributes and rewrote spec to match. See: https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/1111/files#diff-f3ca654ba2e5de8b9e9882a75527c0ead45a85466759a7b7a242b43c408bd1c5R8

smoothcontract: Actually looks like you've made the same change now, so they're pretty much in sync - just the timestamps missing from this version.

smoothcontract (approved):
Looks good to me, this had a lengthy and difficult birth but it got there in the end. It might be worth updating the PR description as that's a little out of date now but otherwise all is well 👍 👏

app/serializers/category_serializer.rb
martyn-w: OK, thanks -  I think I'll merge this now and let your PR add created_at, updated_at

smoothcontract: sounds good 👍

swagger/v1/get_person_responses.yaml
martyn-w: This is intentional: it is raises a 406 (Not Acceptable) error if a client tries to use `GET /people/{person_id}` in version 1

PR #1091, opened 2020-10-29
Add username and user_type to paper_trail (P4-1776)

Thource:
### Jira link

P4-1776

### What?

I have added/removed/altered:

- [x] Altered user_for_paper_trail to use the X-Current-User header if it's present, if not, it'll fallback the current behaviour
- [x] Added user_type to PaperTrail::Version

### Why?

I am doing this because:

- It gives us everything we need to find more information about the user, as we can find the user's name through the auth service.
- It will work as normal when the X-Current-User header is not present.


app/controllers/api_controller.rb
```diff
@@ -40,9 +40,15 @@ def doorkeeper_application_owner
   end
 
   def user_for_paper_trail
+    return headers['X-Current-User'] if headers['X-Current-User'].present?
+
     current_user.owner_id
   end
 
+  def info_for_paper_trail
+    { user_type: headers['X-Current-User'].present? ? 'user' : 'supplier' }
```
martyn-w, line 49: unfortunately, this is not a safe assumption. Instead use the `doorkeeper_application_owner` in the ApiController. This will be nil for the front end or the supplier (geo / serco) for an api client.

db/migrate/20201029093313_add_user_type_to_versions.rb
```diff
@@ -0,0 +1,5 @@
+class AddUserTypeToVersions < ActiveRecord::Migration[6.0]
+  def change
+    add_column :versions, :user_type, :string
```
martyn-w, line 3: rather than user type, maybe this could be `supplier_id` ? nil for a FE user, otherwise the id of the supplier for a BE user

Thource: If I were to change this to `supplier_id`, would `user_for_paper_trail` want changing to return `X-Current-User or nil`? Currently, it's returning the supplier id

martyn-w:
good start  - see my comment on using a mock request spec rather than a controller spec

db/migrate/20201029093313_add_user_type_to_versions.rb
martyn-w: oh sorry, I didn't spot that. We shouldn't mix strings with uuids - I think `user_for_paper_trail` should be a string, and should be the `headers['X-Current-User']` or nil.

We may want to encourage suppliers to also send the user in the future, so this field represent either an MoJ user or a supplier user.

Thource:
I made a mistake... please see https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/1100

PR #1092, opened 2020-10-29
Add capability of running every/each feed on a specific date

willfish:
### Jira link

P4-2420

### What?

Enables running each report feed for the hub and jpc on a specific date.

This change enables specifying the REPORT_ON_DATE for all or for each of
the feeds

### Why?

We need this because of a reporting error in which migrations created
updates to every profile in the profiles table in production. We need to
rerun the report feeds that failed to generate for the relevant date.


mafonso (approved):
LGTM

PR #1093, opened 2020-10-29
Removes negation prefixing on one of the enum options on incidents

willfish:
The incident enum was causing a warning when eager loading in the rails
console on production. Declutter the console output to be a good boy
scout

https://blog.bigbinary.com/2019/03/06/rails-6-adds-negative-scopes-on-enum.html


willfish:
I've already updated the corresponding event validation descriptions to replace not_supplier with was_not_supplier

mafonso (approved):

PR #1095, opened 2020-10-29
P4 2409 import hospitals

martyn-w:
### Jira link

P4-2409

### What?

I have added/removed/altered:

- [x] Add hospital location type

### Why?

I am doing this because:

- so we can import hospitals

- TODO: run `rake reference_data:create_locations` on all environments after deployment


cwrw (approved):

cwrw (commented):
oops forgot to add that we need to add this to the swagger documentation as well for location

cwrw (approved):
looks good, thanks for addressing

PR #1096, opened 2020-10-30
Use created_at field in range query when retrieving immutable resources

willfish:
### Jira link

P4-<TODO>

### What?

Notification and Event resources can reasonably be expected not to
change once they've been created.

We should control for these resources being in the feed using the
created_at rather than the updated_at since arbitrary updates to these
records might result from rake task migrations, etc, creating a large amount of data in the feed for that day.

- [x] Add created_at scope
- [x] Simplify scope adding to ApplicationRecord
- [x] Move to using created at for Notification feed
- [x] Move to using created at for Event feed

### Why?

This is one way in which we can protect ourselves from large numbers of updates to the feed for these resources.

### DIFF
--- a/app/services/feeds/notification.rb
+++ b/app/services/feeds/notification.rb
@@ -1,14 +1,14 @@
 module Feeds
   class Notification
-    def initialize(updated_at_from = nil, updated_at_to = nil)
-      @updated_at_from = updated_at_from || Time.zone.yesterday.beginning_of_day
-      @updated_at_to = updated_at_to || Time.zone.yesterday.end_of_day
+    def initialize(created_at_from = nil, created_at_to = nil)
+      @created_at_from = created_at_from || Time.zone.yesterday.beginning_of_day
+      @created_at_to = created_at_to || Time.zone.yesterday.end_of_day
 
       @feed = []
     end
 
     def call
-      ::Notification.includes(:notification_type).updated_at_range(@updated_at_from, @updated_at_to).find_each do |notification|
+      ::Notification.created_at_range(@created_at_from, @created_at_to).find_each do |notification|
         @feed << notification.for_feed.to_json
       end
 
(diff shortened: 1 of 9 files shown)


app/services/feeds/notification.rb
```diff
@@ -1,14 +1,14 @@
...
+    def initialize(created_at_from = nil, created_at_to = nil)
+      @created_at_from = created_at_from || Time.zone.yesterday.beginning_of_day
+      @created_at_to = created_at_to || Time.zone.yesterday.end_of_day
 
       @feed = []
     end
 
     def call
-      ::Notification.includes(:notification_type).updated_at_range(@updated_at_from, @updated_at_to).find_each do |notification|
+      ::Notification.created_at_range(@created_at_from, @created_at_to).find_each do |notification|
```
martyn-w, line 11: just wondering - do you need `.includes(:notification_type)` ?

lib/tasks/feeds/event.rake
```diff
@@ -3,34 +3,14 @@ namespace :feeds do
...
-      feed = Feeds::Event.new(updated_at_from, updated_at_to).call
+      feed = Feeds::Event.new(created_at_from, created_at_to).call
       CloudDataFeed.new.write(feed, 'events.jsonl', report_on_date)
     else
       feed = Feeds::Event.new.call
       CloudDataFeed.new.write(feed, 'events.jsonl')
     end
   end
-
-  desc 'Exports the JSON starting from a certain date (inclusive)'
```
martyn-w, line 17: just wondering why this task is no longer required?

martyn-w (approved):
👍  LGTM

app/services/feeds/notification.rb
willfish: We don't. It was not needed in the first place. It was added because we originally didn't know that the id of the notification type was actually revealing by itself.

lib/tasks/feeds/event.rake
willfish: I could have left it but I actually don't think it will be used now that we can regenerate reports for a specific day. This was originally needed to backfill all events after adding a missing supplier id to all events.

martyn-w: 👍

PR #1097, opened 2020-10-30
Add schema migration and model for categories

smoothcontract:
### Jira link

P4-2416 P4-1421

### What?

- [x] Create category table and associated model, factory, specs.
- [x] Add relationship between category and locations, profiles.

### Why?

- This will be used to determine prisoner category (via profile) in order to block moving category A prisoners. It will also be used to provide a breakdown by prison (location) category in the PMU dashboards. This PR adds a skeleton implementation of the model for now to unblock work on both streams.

See also: https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/1090

### Have you? (optional)

- [ ] Updated API docs if necessary - Not in this PR, will come when associated endpoints are added/updated

- Changes data structures but new relationships are optional so do not effect existing behaviour

### DIFF
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -24,6 +24,8 @@ class Profile < VersionedModel
 
   before_validation :set_assessment_answers
 
+  # TODO: uncomment this when category attribute is removed from model
+  # belongs_to :category, optional: true
   belongs_to :person, touch: true
 
   has_many :moves, dependent: :nullify
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_10_22_072140) do
+ActiveRecord::Schema.define(version: 2020_10_30_095815) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "citext"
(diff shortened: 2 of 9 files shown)


db/schema.rb
```diff
@@ -309,6 +318,8 @@
     t.string "key", null: false
     t.datetime "disabled_at"
     t.boolean "can_upload_documents", default: false, null: false
+    t.uuid "category_id"
```
martyn-w, line 321: the assumption here is that there is 1 category per location - it would be worth double-checking this with @Nimphal 's expert in the MoJ

smoothcontract: Jade has confirmed with the PMU that each location has only one category - and that's what the location category will be used for so I think it makes sense to go with this structure until that assertion changes. That's preferable to adding a has_many relationship just in case we might need it I think. We should go with what we're told really rather than try to second guess.

app/models/profile.rb
```diff
@@ -24,6 +24,8 @@ class Profile < VersionedModel
 
   before_validation :set_assessment_answers
 
+  # TODO: uncomment this when category attribute is removed from model
+  # belongs_to :category, optional: true
```
smoothcontract, line 28: @martyn-w I had to comment this for now as it conflicts with the existing `category` attribute.

martyn-w: 👍  sure

db/schema.rb
martyn-w: MoJ expert has confirmed there can be a mixture of prisoner categories in a single prison. E.g. a Cat B prison can hold Cat B and Cat C prisoners. But it should not hold prisoners with a rating higher than the prison - e.g. a Cat B prison should not hold Cat A prisoners.

martyn-w (approved):
👍  LGTM

PR #1098, opened 2020-10-30
Don't create fake data that should be reference data

smoothcontract:
### Jira link

P4-<TODO>

### What?

- [x] Removed fake data from rake task that should really be reference data

### Why?

- This rake task is inconsistent and couldn't easily be used in combination with the reference data rake task; the majority of reference information comes either from YAML files or via Nomis, so it makes little sense to delete this and replace with random stuff from Faker. However it is useful to retain the creation of fake moves, allocations, journeys etc. as these are all transactional and should simulate use of the PECS system.

- The goal here is that a working local development environment can be setup by running: `rails reference_data:create_all fake_data:recreate_all`.


martyn-w (approved):
Long live the revolution

willfish (approved):

PR #1099, opened 2020-11-02
[P4-2422] Surface event relationships: part 1

willfish:

The goal of this PR is to enable surfacing generic event relationships
that are currently stored in the details attribute to be in the
relationship section in the serialized generic event.

In order to achieve that we have to first know the type of each
relationship so we add type information to the relationship_attributes
call.

We also need a method that takes all of that information for the current
event STI instance and pulls out the relationships that we're
interested in for consumption by the frontend later.

- [x] Updates GenericEvent#relationship_attributes dsl to also take a
  type
- [x] Updates all GenericEvent STI classes to pass their relationship
  type where necessary
- [x] Adds GenericEvent#relationships method to render relationship
  information
- [x] Updates all tests to reflect change to relationship attributes
- [x] Removes some now-redundant relationship methods which were used to
  retrieve the relationship

### DIFF
--- a/app/models/generic_event.rb
+++ b/app/models/generic_event.rb
@@ -107,6 +107,19 @@ class GenericEvent < ApplicationRecord
     feed
   end
 
+  def relationships
+    return {} unless self.class.instance_variable_defined?(:@relationship_attributes)
+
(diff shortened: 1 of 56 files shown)


martyn-w (approved):
👍 LGTM

PR #1100, opened 2020-11-02
P4 1776 add user info to papertrail

Thource:
### Jira link

P4-1776

### What?

I have added/removed/altered:

- [x] Altered user_for_paper_trail to use the X-Current-User header or `nil`
- [x] Added supplier_id to PaperTrail::Version

### Why?

I am doing this because:

- It gives us everything we need to find more information about the user, as we can find the user's name through the auth service.
- It will work as normal when the X-Current-User header is not present.


app/controllers/api_controller.rb
```diff
@@ -40,7 +40,11 @@ def doorkeeper_application_owner
   end
 
   def user_for_paper_trail
-    current_user.owner_id
+    request.headers['X-Current-User']
```
Nimphal, line 43: Is this the intended? This changes the current behaviour, so if the header is not present (which it won't be), then the user will nil, rather than maintain current logic and set it to the owner_id.

Thource: This was suggested by @martyn-w, but I'm open to all suggestions.

Maybe I should make the migration go through previous paper trail objects and if the paper trail user is a supplier id, then set the user to nil and backfill the supplier_id field?

Nimphal: I am probably out of the loop here, I was more thinking something like `request.headers['X-Current-User'] || current_user.owner_id` so it falls back on current behaviour. But I'm sure @martyn-w will advise :)

db/migrate/20201102120213_add_supplier_id_to_versions.rb
```diff
@@ -0,0 +1,5 @@
+class AddSupplierIdToVersions < ActiveRecord::Migration[6.0]
+  def change
+    add_column :versions, :supplier_id, :string
```
martyn-w, line 3: supplier_id is a uuid - not a string - and probably should have a foreign key constraint. How about:

```
    add_column : versions, : supplier_id, :uuid, null: true
    add_foreign_key :versions, :suppliers, column: :supplier_id
```

Thource: Ah, I didn't realise there was a special field type for uuids, thanks!

app/controllers/api_controller.rb
martyn-w: interesting point, but rather than fall back, I think it will be cleaner to migrate the old values (whodunit) to the new field (supplier_id). So I think this implementation is good 👍

martyn-w:
Adding a note about the intended behaviour here for posterity:

> Hi @Adam Sharp (cc @Nev),  - just wanting to the clarify the intended behaviour in your PR
> **currently**
> user_for_paper_trail is the supplier_id for a supplier user, or nil for a front end user. This is stored in the database in the versions table as whodunit
>
> **changed in this PR**
> user_for_paper_trail is the user as supplied by the FE request.headers['X-Current-User'] and will be stored in the db as versions.whodunit . If the header is not supplied, it will be nil.
> info_for_paper_trail will store the supplier_id in field versions.supplier_id if it is available, otherwise nil.
>  Is that correct ? (edited) 
> 
> Adam Sharp  14:58
> That’s correct, it also adds the ability for suppliers to also send a X-Current-User, if they choose to, in the future
> 
> 
> Martyn Whitwell  15:04
> Ok, I think that behaviour sounds reasonable to me :+1:
> I would add one more task: moving the old supplier_ids in  versions.whodunit  to the new field versions.supplier_id. Maybe :thinking_face: this should be done as a data migration, something like this:
> `UPDATE versions SET supplier_id = whodunit WHERE whodunit IN (SELECT id FROM suppliers);`

martyn-w (approved):
👍 Looks good to me

PR #1102, opened 2020-11-02
Add new service and YAML file to import category reference data

smoothcontract:
### Jira link

P4-1421

### What?

- [x] Add new `Categories::Importer` service
- [x] Add new `categories.yaml` reference data file and associated rake task to import (`rails reference_data:create_locations`)

### Why?

- This data will be used to categories locations within the PMU dashboard and also to associate prisoners with their correct category so that moves for category A prisoners can be prevented. The reference data is therefore a combination of the known categories within Nomis plus some additional ones used within PMU to group locations.

- Adds new data that is not yet used (future PR's will add this category data to PMU locations dashboard)

### DIFF
--- a/lib/tasks/data/categories.yml
+++ b/lib/tasks/data/categories.yml
@@ -1,0 +1,208 @@
+---
+  A:
+    title: Category A
+    move_supported: false
+    locations:
+  B:
+    title: Category B
+    move_supported: true
+    locations:
+    - FKI
+    - FNI
+    - GHI
+    - GTI
+    - IWI
+    - LLI
+    - LGI
+    - RHI
+    - SLI
+    - WDI
+    - WRI
+    - MRI
+    - WHI
+  BTH:
+    title: Category B Therapeutic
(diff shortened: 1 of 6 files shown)


lib/tasks/data/categories.yml
```diff
@@ -0,0 +1,208 @@
...
+    title: Parole Granted
+    move_supported: true
+    locations:
+  H:
+    title: Cat A Hi
+    move_supported: false
+    locations:
+  HI:
+    title: High
+    move_supported: true
```
martyn-w, line 135: lets confirm with DPS that `HI` moves should be supported

martyn-w (approved):
👍 LGTM

lib/tasks/data/categories.yml
smoothcontract: Okay. I think I'll merge this as is for now and then amend the yaml file if needed as that will be a small tweak and I need this merged to unblock the next story.

PR #1103, opened 2020-11-02
Finalise non-functional refactor of eventable validations

willfish:
### What?

I have added/removed/altered:

- [x] Use eventable_types dsl method in Move and Journey events
- [x] Remove now-unused model concerns

### Why?

I am doing this because:

- This was a refactor that was too large to finalise in another PR.


app/models/generic_event/move_cancel.rb
```diff
@@ -1,13 +1,7 @@
 class GenericEvent
   class MoveCancel < GenericEvent
-    DETAILS_ATTRIBUTES = %w[
```
willfish, line 3: These were replaced by attributes defined at class evaluation time.

app/models/generic_event/move_cross_supplier_pick_up.rb
```diff
@@ -6,7 +6,5 @@ class MoveCrossSupplierPickUp < GenericEvent
     validates_each :previous_move_id do |_record, _attr, value|
       Move.find(value)
     end
-
-    include MoveEventValidations
```
willfish, line 10: This duplicates the eventable validations added at the top of the file.

smoothcontract (approved):
LGTM 👍

PR #1104, opened 2020-11-02
P4 1776 add user info to papertrail

Thource:
### Jira link

P4-1776

### What?

I have added/removed/altered:

- [x] Added a migration that alters PaperTrail::Versions, populating the supplier_id field with the whodunnit field, if the whodunnit field is a valid Supplier id

### Why?

I am doing this because:

- It will make the past PaperTrail::Versions be in line with the future ones


db/migrate/20201102160605_split_whodunnit_from_versions.rb
```diff
@@ -0,0 +1,31 @@
+class SplitWhodunnitFromVersions < ActiveRecord::Migration[6.0]
+  def up
+    count = 0
+
+    PaperTrail::Version.all.each do |version|
```
willfish, line 5: You can do the following to enable only loading 1000 into memory at a time :). The table is actually massive so it's possible that we end up loading a ridiculous number of records into memory all at once.

```suggestion
    PaperTrail::Version.all.find_each do |version|
```

db/migrate/20201102160605_split_whodunnit_from_versions.rb
```diff
@@ -0,0 +1,31 @@
...
+      next if version.whodunnit.blank?
+
+      begin
+        Supplier.find(version.whodunnit)
```
willfish, line 9: I reckon it's preferable to have potentially broken references in the data than to execute a query for every single row in the versions table. Looking at the count estimates you can see that there are quite a lot of rows:

```ruby
irb(main):006:0> result =  ActiveRecord::Base.connection.execute("SELECT reltuples as approximate_row_count FROM pg_class WHERE relname = 'versions';")
=> #<PG::Result:0x0000556cd25f8428 status=PGRES_TUPLES_OK ntuples=1 nfields=1 cmd_tuples=1>
irb(main):007:0> result.to_a
=> [{"approximate_row_count"=>5224700.0}]
```

willfish: I actually think it may be simpler to drop into SQL directly and just let the database do the work 🤔 . You can also only select data that has a known supplier id as per Martyn's query: 

```
# up
UPDATE versions SET supplier_id = whodunit WHERE whodunit IN (SELECT id FROM suppliers)
# down
UPDATE versions SET supplier_id = null WHERE whodunit IN (SELECT id FROM suppliers)
```

willfish:
```ruby
def up
  ActiveRecord::Base.connection.execute(
    'UPDATE versions SET supplier_id = whodunit WHERE whodunit IN (SELECT id FROM suppliers)'
  )
end

def down
  ActiveRecord::Base.connection.execute(
    'UPDATE versions SET supplier_id = null WHERE whodunit IN (SELECT id FROM suppliers)'
  )
end
```

Unfortunately there isn't an index for the whodunnit column so the WHERE clause won't speed us up at all (probably wouldn't anyway given we probably have valid whodunit values for most of the rows.

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_11_02_120213) do
+ActiveRecord::Schema.define(version: 2020_11_02_160605) do
```
willfish, line 13: This feels like it shouldn't affect the schema 🤔 .

willfish: Happy to be taught otherwise :)

Thource: The only way around this that I can think of is to have it as a rake task and manually run it, but I'm happy to be told otherwise also :D

db/migrate/20201102160605_split_whodunnit_from_versions.rb
Thource: Using this raw SQL, this migration takes 1.7s to run for 100,012 versions, that seems fast enough to me :)

willfish: Nice :).

db/schema.rb
Thource: Migration has now been replaced with a rake task, which can be used as so:

```
rake paper_trail:populate_with_suppliers
rake paper_trail:populate_with_suppliers\[up\]
rake paper_trail:populate_with_suppliers\[down\]
```

If no argument is passed, it will default to `up`.

lib/tasks/populate_paper_trail_with_suppliers.rake
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+namespace :paper_trail do
+  desc 'Populate PaperTrail::Versions with Supplier ids found in the whodunnit field'
+  task :populate_with_suppliers, %i[up_down] => :environment do |_, args|
```
willfish, line 5: Wonder whether this api would be simpler by just having two tasks?

willfish:
Presume you've run this locally? 😄

willfish (approved):
LGTM

Thource:
> Presume you've run this locally? 😄

Of course! 😄 
```
adamsharp@L0830 hmpps-book-secure-move-api % rake paper_trail:populate_with_suppliers        
Populated 100,011 PaperTrail::Versions with Supplier ids.
adamsharp@L0830 hmpps-book-secure-move-api % rake paper_trail:populate_with_suppliers\[down\]
Reverted 100,011 PaperTrail::Versions, replacing whodunnit with Supplier id.
```

PR #1107, opened 2020-11-03
[P4-2422] Adds support for dynamic anonymous event STI serializers

willfish:
### Jira link

P4-2422

### What?

The goal of this PR is to achieve dynamic serializers which support
includes and relationships for every event STI class which default to V2
relationships where possible.

In this vain we:

- [x] Update the `GenericEvent.relationship_attributes` DSL to support
  dynamic classes
- [x] Consume them in the `GenericEventController#create` endpoint
- [x] Consume them in the `V2::MoveSerializer`
- [x] Add tests for each of the consumers for events which define
  relationship_attributes and for those that don't define relationship_attributes

This enables `GET /api/events?include=from_location` if we need too. But mainly enables relationships that are currently in the jsonb details field to be surfaced automatically in the relationships section of the json:api document for the given event.

It automatically works on a collection and the classes are memoized only once (any additional object penalty affects only the startup of the application and only once).

### Why?

I am doing this because:

- This is for json:api compliance
- This will reduce the amount of work for the frontend

### DIFF
--- a/app/models/generic_event.rb
+++ b/app/models/generic_event.rb
@@ -143,13 +143,38 @@ class GenericEvent < ApplicationRecord
     end
   end
 
+  # NB: Majority of events will use this serializer rather than the anonymous class.
+  def self.serializer
+    GenericEventSerializer
+  end
+
   # Relationship attributes live against the details but are expected in the json:api relationship section
   # so are defined separately
   def self.relationship_attributes(attributes)
     define_singleton_method(:relationship_attributes) do
       instance_variable_get('@relationship_attributes')
     end
+
     instance_variable_set('@relationship_attributes', attributes)
+
+    define_singleton_method(:serializer) do
+      @serializer ||=
+        Class.new(GenericEventSerializer).tap do |serializer|
+          relationship_attributes.each do |attribute_key, attribute_type|
+            # NB: Default to V2 relationship serializer and fallback to unversioned serializer. Never support v1
+            relation_serializer = "V2::#{attribute_type.to_s.singularize.camelize}Serializer"
+            relation_serializer = if const_defined?(relation_serializer)
+                                    relation_serializer
+                                  else
+                                    relation_serializer.gsub('V2::', '')
+                                  end
+
+            named_relationship_key = attribute_key.to_s.sub('_id', '')
+            serializer.set_type :events
+            serializer.has_one named_relationship_key, serializer: relation_serializer.constantize
+          end
+        end
+    end
 
     attributes.each do |attribute_key, attribute_type|
       named_relationship_key = attribute_key.to_s.sub('_id', '')
--- a/app/serializers/v2/move_serializer.rb
+++ b/app/serializers/v2/move_serializer.rb
@@ -29,7 +29,8 @@ module V2
     has_one :to_location,            serializer: LocationSerializer
 
     has_many :court_hearings, serializer: CourtHearingSerializer
-    has_many :timeline_events, serializer: GenericEventSerializer, &:all_events_for_timeline
+
+    has_many :timeline_events, serializer: ->(record, _params) { record.class.serializer }, &:all_events_for_timeline
 
     belongs_to :allocation, serializer: AllocationSerializer
     belongs_to :original_move, serializer: V2::MoveSerializer
--- a/spec/requests/api/generic_events_controller_create_spec.rb
+++ b/spec/requests/api/generic_events_controller_create_spec.rb
@@ -93,13 +93,150 @@ RSpec.describe Api::GenericEventsController do
       expect { do_post }.to change { GenericEvent.find_by(eventable_id: move.id, eventable_type: 'Move') }.from(nil).to(be_a(GenericEvent))
     end
 
-    it 'returns serialized data' do
-      do_post
+    context 'when the event has no additional relationships' do
+      let(:event_attributes) do
+        {
+          occurred_at: '2020-11-04T14:25:48+00:00',
+          recorded_at: '2020-11-04T14:25:48+00:00',
+          notes: 'Flibble',
+          details: {
+            cancellation_reason: 'made_in_error',
+            cancellation_reason_comment: 'It was a mistake',
+          },
+          event_type: 'MoveCancel',
+        }
+      end
 
-      event = GenericEvent.last
-      resource_to_json = JSON.parse(GenericEventSerializer.new(event).serializable_hash.to_json)
+      let(:data) do
+        {
+          type: 'events',
+          attributes: event_attributes,
+          relationships: {
+            eventable: { data: { type: 'moves', id: move.id } },
+          },
+        }
+      end
 
-      expect(response_json).to eq resource_to_json
+      it 'returns serialized data' do
+        do_post
+
+        event = GenericEvent.last
+        resource_to_json = JSON.parse(event.class.serializer.new(event).serializable_hash.to_json)
+
+        expect(response_json).to eq resource_to_json
+      end
+
+      it 'returns the serialized relationships of the event' do
+        do_post
+
+        event = GenericEvent.last
+        resource_to_json = JSON.parse(event.class.serializer.new(event).serializable_hash.to_json)
+
+        expected_relationships = {
+          'eventable' => { 'data' => { 'id' => event.eventable.id, 'type' => 'moves' } },
+        }
+
+        expect(resource_to_json.dig('data', 'relationships')).to eq(expected_relationships)
+      end
+    end
+
+    context 'when the event has an additional relationship with no v2 implementation' do
+      let(:event_attributes) do
+        {
+          occurred_at: '2020-11-04T14:31:51+00:00',
+          recorded_at: '2020-11-04T14:31:51+00:00',
+          notes: 'Flibble',
+          details: {
+            reason: 'no_space',
+            authorised_at: '2020-11-04T14:31:51+00:00',
+            authorised_by: 'PMU',
+          },
+          event_type: 'MoveLockout',
+        }
+      end
+
+      let(:data) do
+        {
+          type: 'events',
+          attributes: event_attributes,
+          relationships: {
+            eventable: { data: { type: 'moves', id: move.id } },
+            from_location: { data: { type: 'locations', id: move.from_location.id } },
+          },
+        }
+      end
+
+      it 'returns serialized data' do
+        do_post
+
+        event = GenericEvent.last
+        resource_to_json = JSON.parse(event.class.serializer.new(event).serializable_hash.to_json)
+
+        expect(response_json).to eq resource_to_json
+      end
+
+      it 'returns the serialized from_location of the event' do
+        do_post
+
+        event = GenericEvent.last
+        resource_to_json = JSON.parse(event.class.serializer.new(event).serializable_hash.to_json)
+
+        expected_relationships = {
+          'eventable' => { 'data' => { 'id' => event.eventable.id, 'type' => 'moves' } },
+          'from_location' => { 'data' => { 'id' => event.from_location.id, 'type' => 'locations' } },
+        }
+        expect(resource_to_json.dig('data', 'relationships')).to eq(expected_relationships)
+      end
+    end
+
+    context 'when the event has an additional relationship with a v2 implementation' do
+      before do
+        allow(V2::MoveSerializer).to receive(:new).and_call_original
+      end
+
+      let(:previous_move) { create(:move) }
+
+      let(:event_attributes) do
+        {
+          occurred_at: '2020-11-04T14:31:51+00:00',
+          recorded_at: '2020-11-04T14:31:51+00:00',
+          notes: 'Flibble',
+          event_type: 'MoveCrossSupplierPickUp',
+        }
+      end
+
+      let(:data) do
+        {
+          type: 'events',
+          attributes: event_attributes,
+          relationships: {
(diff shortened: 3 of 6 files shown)


app/models/generic_event.rb
```diff
@@ -143,14 +143,43 @@ def self.details_attributes(*attributes)
...
+            # NB: Default to V2 relationship serializer and fallback to unversioned serializer. Never support v1
+            relation_serializer = "V2::#{attribute_type.to_s.singularize.camelize}Serializer"
+            relation_serializer = if const_defined?(relation_serializer)
+                                    relation_serializer
+                                  else
+                                    relation_serializer.gsub('V2::', '')
+                                  end
+
+            named_relationship_key = attribute_key.to_s.sub('_id', '')
+            serializer.set_type :events
```
willfish, line 175: This is necessary as the type is accessed via instance variable in the parent and needs to be set in the child.

app/serializers/v2/move_serializer.rb
```diff
@@ -29,7 +29,8 @@ class MoveSerializer
     has_one :to_location,            serializer: LocationSerializer
 
     has_many :court_hearings, serializer: CourtHearingSerializer
-    has_many :timeline_events, serializer: GenericEventSerializer, &:all_events_for_timeline
+
+    has_many :timeline_events, serializer: ->(record, _params) { record.class.serializer }, &:all_events_for_timeline
```
willfish, line 33: The serializer option on relationships can be a callable. We need this because the serializer is different and defined against each STI class.

willfish:
Separately I'm going to be squishing the relationships out of the details in the `GenericEventSerializer` class which will propagate to all children.

willfish:
NB: No generic event gets serialised and consumed currently. The aim of this is to be done in the frontend so it's okay to do this in parts :).

willfish:
**Context**

Three stakeholders and their consumption patterns:

- suppliers (only create events)
- frontend (create and read events)
- hub (only read a special view of events via the feed - see for_feed method on each feed record type)

v1 event endpoints don't return a serialized event so suppliers and the frontend that currently use them to create events aren't going
they're not going to be affected by this change.

Since we weren't returning events before, this is a completely new thing for the frontend, only, and wasn't a problem we encountered previously.

All I'm really doing is coming up with a cunning way to make the event responses during creation (but only in the v2 world) and when reading the timeline of events for the frontend

__I understood that the suppliers were vexed by the previous events approach with lots of different endpoints, which is why there's been so much refactoring to introduce generic events via a simpler API.__

1. This only effects v2 world (e.g. the generic event api and the timeline for events since events were never serialized in the v1 world).
2. This isn't actually any more complex from a supplier point of view and is actually consistent with how they create events in the v2 world currently (see request body, here: https://github.com/ministryofjustice/hmpps-book-secure-move-api/wiki/Journey-Events#request-example)
3. As an aside we're actually contractually obliged to be json:api compliant

__So with that in mind, this now seems to be taking a step back in the other direction with specific serializers (i.e. different API responses) for some of the events.__

1. We already have different responses for every event depending on what the details contain based on the request body the supplier is pushing.

__In supplier Java land that's going to mean them adding additional classes to handle this response, and I suspect more bleating from them__

1. Suppliers can just do away with events responses currently. If they need to handle them then they'd need to do something - they definitely just fire and forget currently. I'm also guessing java supports generics/equivalent for json responses varying.

app/models/generic_event.rb
```diff
@@ -143,14 +143,43 @@ def self.details_attributes(*attributes)
...
     instance_variable_set('@relationship_attributes', attributes)
 
+    define_singleton_method(:serializer) do
+      @serializer ||=
+        begin
+          serializer = Class.new(GenericEventSerializer)
+
+          relationship_attributes.each do |attribute_key, attribute_type|
+            # NB: Default to V2 relationship serializer and fallback to unversioned serializer. Never support v1
+            relation_serializer = "V2::#{attribute_type.to_s.singularize.camelize}Serializer"
```
smoothcontract, line 167: Might be nice to use a separate variable for this rather than reassigning the cla

app/models/generic_event.rb
smoothcontract: Really not convinced you need this line - in `GenericEventSerializer` there is `set_type :events` method call, Maybe I'm wrong - could you try without and see if it still works? I couldn't spot any coverage that proves this is needed (unless there's some existing stuff) so might be worth adding an explicit check on `type`.

app/models/generic_event.rb
```diff
@@ -143,14 +143,43 @@ def self.details_attributes(*attributes)
...
   def self.relationship_attributes(attributes)
     define_singleton_method(:relationship_attributes) do
       instance_variable_get('@relationship_attributes')
     end
+
     instance_variable_set('@relationship_attributes', attributes)
 
+    define_singleton_method(:serializer) do
+      @serializer ||=
+        begin
```
smoothcontract, line 162:
```suggestion
        Class.new(GenericEventSerializer).tap |serializer|
```

app/models/generic_event.rb
```diff
@@ -143,14 +143,43 @@ def self.details_attributes(*attributes)
...
+                                  else
+                                    relation_serializer.gsub('V2::', '')
+                                  end
+
+            named_relationship_key = attribute_key.to_s.sub('_id', '')
+            serializer.set_type :events
+            serializer.has_one named_relationship_key, serializer: relation_serializer.constantize
+          end
+
+          serializer
```
smoothcontract, line 179: Can remove if using `tap` pattern

spec/requests/api/generic_events_controller_create_spec.rb
```diff
@@ -93,13 +93,67 @@
...
+        expect(resource_to_json.dig('data', 'relationships')).to eq(expected_relationships)
+      end
+    end
+
+    context 'when the event has an additional relationship' do
+      let(:event_attributes) do
+        attributes_for(:event_move_lockout).tap do |attributes|
+          attributes.except!(:eventable)
+          attributes[:details] = attributes[:details].except(:from_location_id)
+          attributes[:event_type] = 'MoveLockout'
```
smoothcontract, line 132: This is starting to require a little bit of mental gymnastics to figure out what the expected attributes will be - perhaps clearer to make this explicit even if it means repeating some of the attributes again.

app/models/generic_event.rb
```diff
@@ -143,14 +143,43 @@ def self.details_attributes(*attributes)
...
+          relationship_attributes.each do |attribute_key, attribute_type|
+            # NB: Default to V2 relationship serializer and fallback to unversioned serializer. Never support v1
+            relation_serializer = "V2::#{attribute_type.to_s.singularize.camelize}Serializer"
+            relation_serializer = if const_defined?(relation_serializer)
+                                    relation_serializer
+                                  else
+                                    relation_serializer.gsub('V2::', '')
+                                  end
+
+            named_relationship_key = attribute_key.to_s.sub('_id', '')
```
smoothcontract, line 174: There's quite a bit of stuff going on here that might be worth adding coverage at unit level - would be good to see specs for V1/V2 serializers being returned depending on which things we have defined.

smoothcontract (commented):
Looking really good and is a very elegant implementation - thanks also for the background context to understand the motivation for this. Only significant comment is that I'd like to see a little more unit level coverage for the new dynamic serializer behaviour; the happy path coverage at request level is great but there a few other code branches (e.g. for undefined V2 serializers) that would be good to exercise.

app/models/generic_event.rb
willfish: I'll pull out the exact line of code in the serializer core that explains this. I didn't add it for fun unfortunately. There might be another way of achieving it (overriding a method in the parent and using the method lookup tree, for example)

app/models/generic_event.rb
willfish: Re testing: I am exercising every serializer implicitly in a successful request for every event. I could add an explicit test of the serializer results for one that implements a v2 and one that doesn't have a v2?

I could also dump the result of every event serialization process to a fixture file for each event and verify that the result is what we expect?

spec/requests/api/generic_events_controller_create_spec.rb
willfish: Fixed by being explicit about attributes.

willfish:
@smoothcontract

Re set_type, the variable lives on the singleton class in an instance variable and is accessed via an attr_reader which doesn't seem to propagate to the parent. We could do:

```ruby
serializer.define_method(:record_type) do
  super # or equivalent
end
```

See:


```ruby
# Note that the parent has a record_type
[6] pry(GenericEvent::PerCourtTakeFromCustodyToDock)> GenericEventSerializer.record_type
=> :events
# And the child doesn't
[7] pry(GenericEvent::PerCourtTakeFromCustodyToDock)> serializer.record_type
=> nil
# You can see it differently when listing instance_variables with pry's ls command:
[8] pry(GenericEvent::PerCourtTakeFromCustodyToDock)> ls GenericEventSerializer
constants: SUPPORTED_RELATIONSHIPS
Object.methods: yaml_tag
FastJsonapi::SerializationCore::ClassMethods#methods: attributes_hash  get_included_records  id_from_record  id_hash  links_hash  meta_hash  parse_includes_list  record_cache_options  record_hash  relationships_hash
FastJsonapi::ObjectSerializer::ClassMethods#methods:
  add_relationship  attributes  cache_options           create_relationship       fetch_polymorphic_option  has_one    is_collection?  meta                   run_key_transform  set_id             set_type    validate_includes!
  attribute         belongs_to  compute_id_method_name  deprecated_cache_options  has_many                  inherited  link            reflected_record_type  serializer_for     set_key_transform  use_hyphen
GenericEventSerializer.methods:
  attributes_to_serialize              cachable_relationships_to_serialize=  cache_store_options   data_links=         record_id    record_type=                 transform_method                       uncachable_relationships_to_serialize=
  attributes_to_serialize=             cache_store_instance                  cache_store_options=  meta_to_serialize   record_id=   relationships_to_serialize   transform_method=
  cachable_relationships_to_serialize  cache_store_instance=                 data_links            meta_to_serialize=  record_type  relationships_to_serialize=  uncachable_relationships_to_serialize
instance variables: @attributes_to_serialize  @cachable_relationships_to_serialize  @record_type  @reflected_record_type  @relationships_to_serialize  @uncachable_relationships_to_serialize
[9] pry(GenericEvent::PerCourtTakeFromCustodyToDock)> ls serializer
Object.methods: yaml_tag
FastJsonapi::SerializationCore::ClassMethods#methods: attributes_hash  get_included_records  id_from_record  id_hash  links_hash  meta_hash  parse_includes_list  record_cache_options  record_hash  relationships_hash
FastJsonapi::ObjectSerializer::ClassMethods#methods:
  add_relationship  attributes  cache_options           create_relationship       fetch_polymorphic_option  has_one    is_collection?  meta                   run_key_transform  set_id             set_type    validate_includes!
  attribute         belongs_to  compute_id_method_name  deprecated_cache_options  has_many                  inherited  link            reflected_record_type  serializer_for     set_key_transform  use_hyphen
GenericEventSerializer.methods:
  attributes_to_serialize              cachable_relationships_to_serialize=  cache_store_options   data_links=         record_id    record_type=                 transform_method                       uncachable_relationships_to_serialize=
  attributes_to_serialize=             cache_store_instance                  cache_store_options=  meta_to_serialize   record_id=   relationships_to_serialize   transform_method=
  cachable_relationships_to_serialize  cache_store_instance=                 data_links            meta_to_serialize=  record_type  relationships_to_serialize=  uncachable_relationships_to_serialize
instance variables: @attributes_to_serialize  @cache_store_instance  @cache_store_options  @meta_to_serialize  @record_id  @reflected_record_type  @relationships_to_serialize  @transform_method  @uncachable_relationships_to_serialize
```

app/models/generic_event.rb
smoothcontract: It seems strange that the inherited type is working fine for this class - https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/serializers/allocations_serializer.rb - and tested explicitly here https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/spec/requests/api/allocations_controller_index_spec.rb#L30

Quite the mystery 👻 🐕 🚗

app/models/generic_event.rb
smoothcontract: A simple unit level spec for this model that tests V2 fallback behaviour is all that's missing I think - happy with the higher level coverage of each generate serializer. The unit level spec also acts as living documentation so I think is beneficial to explicitly cover this behaviour.

app/models/generic_event.rb
smoothcontract: Here's a simpler spec that shows the same thing: https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/spec/serializers/allocations_serializer_spec.rb#L17

I wonder if the behaviour you're seeing is just down to the dynamic class definitions rather than static approach for allocations. It's more academic curiosity at this point to be honest, happy for the line to be left in place and we can ponder this some other time.

smoothcontract: Discussed offline and solved the mystery - in the case of allocations serializer the `type` is defaulted to `allocations` entirely based on the class name, and the `type` is indeed not inherited from the ancestor. Move along people, nothing to see here 😁

smoothcontract (approved):
All looks good to me, thanks for the amendments 👍

PR #1108, opened 2020-11-04
Remove inclusion test for boolean column on categories

cwrw:
To avoid warning message when running tests, where inclusion is not possible to fully test on booleans, remove test.
The warning message in full:
```
************************************************************************
Warning from shoulda-matchers:

You are using `validate_inclusion_of` to assert that a boolean column
allows boolean values and disallows non-boolean ones. Be aware that it
is not possible to fully test this, as boolean columns will
automatically convert non-boolean values to boolean ones. Hence, you
should consider removing this test.
************************************************************************
```


smoothcontract (approved):
LGTM - I was doing the same thing on my branch having noticed the noisy spec output but it's good to merge this independently so everyone has a more peaceful day 😀

PR #1110, opened 2020-11-04
Ensure the right version of tzdata is used in the docker container

Nimphal:
### Jira link

P4-XXXX

### What?

I have added/removed/altered:

- [x] Include the full version of tzdata in the Docker image so the 'London' timezone works correctly in the docker containers. Follows this SO answer https://stackoverflow.com/questions/64581249/rails-activesupport-time-and-timezone-error-on-alpine-linux

### Why?

I am doing this because:

- The API is using the wrong time and that has impact on the data we store and communicate to suppliers.

### Have you? (optional)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,6 +43,10 @@ RUN addgroup -g $APPUID -S appgroup && \
 
 RUN apk add --update --no-cache git tzdata postgresql-dev
 
+# Fix incompatibility with slim tzdata from 2020b onwards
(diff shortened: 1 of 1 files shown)


cwrw (approved):
👍  this looks like it affected quite a few people with similar setup, I look forward to a fix from a future release of `tzinfo` so we can remove this

Nimphal:
Definitely!

martyn-w:
Ticket: https://dsdmoj.atlassian.net/browse/P4-2440

PR #1111, opened 2020-11-04
Add support for including categories within PMU dashboard endpoint

smoothcontract:
### Jira link

P4-2297 P4-2416

### What?

- [x] Add support for `include=category` and sorting by category title (and then location title) to the `/locations_free_spaces` API endpoint

### Why?

- The PMU dashboard of free spaces per location should be broken down by category. Reference data for each category and associated mapping of prisons to categories has been completed previously. This PR leverages that data and adds support for `include=category` query string parameter which the front end can now specify.

- Note that the `Locations::Finder` service has been updated to follow the new pattern from `Moves::Finder` of eager fetching only the required relations to match the requested list of includes - this both avoids an N+1 query when categories are included, and needlessly fetching all categories when they are not included. Also added support for sorting by category so that the data can be returned to the front end in a sensible order.

### Have you? (optional)

- [ ] Updated API docs if necessary - The PMU dashboard endpoint is intentionally not documented as it's only useful to the front end rather than suppliers

- Extends free spaces endpoint but this is a non destructive change and also use of this is hidden behind a feature flag in the front end

### DIFF
--- a/app/serializers/category_serializer.rb
+++ b/app/serializers/category_serializer.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class CategorySerializer
+  include JSONAPI::Serializer
+
+  set_type :categories
+  set_id :key
+
+  attributes :title, :move_supported
+end
--- a/app/serializers/category_serializer.rb
+++ b/app/serializers/category_serializer.rb
@@ -4,7 +4,6 @@ class CategorySerializer
   include JSONAPI::Serializer
 
   set_type :categories
-  set_id :key
 
-  attributes :title, :move_supported
+  attributes :key, :title, :move_supported, :created_at, :updated_at
 end
--- a/app/services/locations/finder.rb
+++ b/app/services/locations/finder.rb
@@ -2,25 +2,29 @@
 
 module Locations
   class Finder
-    attr_accessor :filter_params, :order_by, :order_direction
+    attr_accessor :filter_params, :order_by, :order_direction, :active_record_relationships
 
-    def initialize(filter_params, sort_params = {})
+    def initialize(filter_params:, sort_params: {}, active_record_relationships: [])
       self.filter_params = filter_params
       self.order_by = (sort_params[:by] || 'title').to_sym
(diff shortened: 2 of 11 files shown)


app/services/locations/finder.rb
```diff
@@ -2,25 +2,29 @@
...
   private
 
     def apply_ordering(scope)
-      case @order_by
+      case order_by
       when :title
-        scope.order(@order_by => @order_direction)
+        scope.order(title: order_direction)
+      when :category
+        scope.left_joins(:category).order('categories.title' => order_direction, 'locations.title' => order_direction)
```
smoothcontract, line 27: The left join here is mostly for backwards compatibility and also just in case we have a location without a category; it should still be returned in the response but will have a nil category.

app/services/locations/finder.rb
```diff
@@ -2,25 +2,29 @@
 
 module Locations
   class Finder
-    attr_accessor :filter_params, :order_by, :order_direction
+    attr_accessor :filter_params, :order_by, :order_direction, :active_record_relationships
 
-    def initialize(filter_params, sort_params = {})
+    def initialize(filter_params:, sort_params: {}, active_record_relationships: [])
```
smoothcontract, line 7: Swapped to named parameters which makes the PR a little noisy but is worth the improved clarity I think

app/serializers/category_serializer.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class CategorySerializer
+  include JSONAPI::Serializer
+
+  set_type :categories
+  set_id :key
```
smoothcontract, line 7: @martyn-w @merlinc Just been testing this locally and realised that this should really be returning the UUID `id` column as `id` so that it matches with the relationships in the rest of the JSONAPI payload. `key` will instead move to an attribute. I'll update the serializer and spec accordingly.

merlinc (commented):
Seems ok to me :)

martyn-w (approved):
👍 LGTM

PR #1113, opened 2020-11-05
[P4-2422]  Remove relationships from details in GenericEventSerializer

willfish:

Now we're compliant and have relationships in the correct json:api
location of the response when surfacing events we need to remove these
relationships from the details attribute dynamically for all events

I've done this separately to reduce review complexity.

Leaving this would be confusing for the frontend

### DIFF
--- a/app/serializers/generic_event_serializer.rb
+++ b/app/serializers/generic_event_serializer.rb
@@ -5,7 +5,7 @@ class GenericEventSerializer
 
   set_type :events
 
-  attributes :occurred_at, :recorded_at, :notes, :details
+  attributes :occurred_at, :recorded_at, :notes
 
   has_one :eventable, serializer: ->(record, _params) { SerializerVersionChooser.call(record.class) }
   has_one :supplier
@@ -14,5 +14,15 @@ class GenericEventSerializer
 
   attribute :event_type do |object|
     object.type.try(:gsub, 'GenericEvent::', '')
+  end
+
+  attribute :details do |record, _params|
+    record.details.deep_dup.tap do |details|
+      if record.class.instance_variable_defined?(:@relationship_attributes)
+        record.class.relationship_attributes.each do |attribute_key, _attribute_type|
+          details.delete(attribute_key)
+        end
+      end
+    end
   end
 end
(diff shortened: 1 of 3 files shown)


app/serializers/generic_event_serializer.rb
```diff
@@ -14,4 +14,14 @@ class GenericEventSerializer
   attribute :event_type do |object|
     object.type.try(:gsub, 'GenericEvent::', '')
   end
+
+  attribute :details do |record, _params|
+    record.details.deep_dup.tap do |details|
```
willfish, line 19: I'm using `dup` because I don't want to mutate the underlying attributes that are referenced when generating the relationships in the serializer. Our serializer gem sequence goes:

1. render attributes ->
2. render relationships

In terms of the `deep_` in  `deep_dup` we never nest attributes in the event details jsonb (this was actually one of the desired changes in the v2 events model). Regardless, I'm doing a `deep_dup` rather than a `dup` to avoid breaking future event implementations that are outside of my control since I won't be here.

spec/requests/api/generic_events_controller_create_spec.rb
```diff
@@ -237,6 +247,16 @@
         }
         expect(resource_to_json.dig('data', 'relationships')).to eq(expected_relationships)
       end
+
+      it 'excludes the previous_move from the details in the response' do
```
willfish, line 251: Tests with a V2 serializer.

spec/requests/api/generic_events_controller_create_spec.rb
```diff
@@ -187,6 +187,16 @@
         }
         expect(resource_to_json.dig('data', 'relationships')).to eq(expected_relationships)
       end
+
+      it 'excludes the from_location from the details in the response' do
```
willfish, line 191: Tests with a fallback (non-v2) serializer.

cwrw:
I've lacking a bit of context regarding this change, I was wondering why we would still have any relationship attributes in the details, is this just for v1, where v2 now won't have this issue?

willfish:
So the details jsonb field is where all event-specific information lives @cwrw. That is for attributes and relationships. What we're doing at the presentation layer, only, is removing the relationships from the details blob that is being returned to the client to be consistent with json:api. We've already duplicated the relationships in the response document to be json:api compliant [here](https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/1107). This PR just removes them from the details attribute being returned by the `GenericEventSerializer`. I did them separately as there was already quite a lot of complexity in the original PR.

cwrw (approved):

smoothcontract (approved):
Nice 👍

PR #1114, opened 2020-11-05
[P4-2422] Removes presenter method

willfish:

We ended up leaning on our json:api serializer gem to render the
relationships. This enables us to automatically support includes on each
of the anonymous serializers and is significantly cleaner.

This does mean that the method removed in this commit is no longer used.


cwrw (approved):

smoothcontract (approved):
Nice cleanup 👍

PR #1115, opened 2020-11-05
[P4-2422] Support smarter polymorphic serializer choice for eventable

willfish:

We need to make sure that eventables that are being serialized as
includes and as relationships for the frontend are picking the correct
serializer version.

In order to achieve this we:

- [x] Extract choice of serializer version into `SerializerVersionChooser` service and add tests
- [x] Consume `SerializerVersionChooser` when picking the serializer
  class for polymorphic eventables in the `GenericEventSerializer`
- [x] Consume `SerializerVersionChooser` when picking the relation serializer
  class for STI relations

### DIFF
--- a/app/models/generic_event.rb
+++ b/app/models/generic_event.rb
@@ -159,19 +159,12 @@ class GenericEvent < ApplicationRecord
 
     define_singleton_method(:serializer) do
       @serializer ||=
-        Class.new(GenericEventSerializer).tap do |serializer|
+        Class.new(GenericEventSerializer).tap do |new_serializer_class|
           relationship_attributes.each do |attribute_key, attribute_type|
-            # NB: Default to V2 relationship serializer and fallback to unversioned serializer. Never support v1
-            relation_serializer = "V2::#{attribute_type.to_s.singularize.camelize}Serializer"
-            relation_serializer = if const_defined?(relation_serializer)
-                                    relation_serializer
-                                  else
-                                    relation_serializer.gsub('V2::', '')
-                                  end
-
             named_relationship_key = attribute_key.to_s.sub('_id', '')
-            serializer.set_type :events
-            serializer.has_one named_relationship_key, serializer: relation_serializer.constantize
+
+            new_serializer_class.set_type :events
+            new_serializer_class.has_one named_relationship_key, serializer: SerializerVersionChooser.call(attribute_type.to_s.singularize.camelize)
           end
         end
     end
--- a/app/models/generic_event.rb
+++ b/app/models/generic_event.rb
@@ -164,7 +164,7 @@ class GenericEvent < ApplicationRecord
             named_relationship_key = attribute_key.to_s.sub('_id', '')
 
             new_serializer_class.set_type :events
-            new_serializer_class.has_one named_relationship_key, serializer: SerializerVersionChooser.call(attribute_type.to_s.singularize.camelize)
+            new_serializer_class.has_one named_relationship_key, serializer: SerializerVersionChooser.call(attribute_type)
           end
         end
     end
--- a/app/services/serializer_version_chooser.rb
+++ b/app/services/serializer_version_chooser.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+class SerializerVersionChooser
+  DEFAULT_API_VERSION = 'V2'
+
+  def self.call(interpolatable)
+    serializer = "#{DEFAULT_API_VERSION}::#{interpolatable.to_s.singularize.camelize}Serializer"
+
+    serializer = if const_defined?(serializer)
+                   serializer
+                 else
+                   serializer.sub("#{DEFAULT_API_VERSION}::", '')
+                 end
+
+    serializer.constantize
+  end
+end
--- a/spec/services/serializer_version_chooser_spec.rb
+++ b/spec/services/serializer_version_chooser_spec.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe SerializerVersionChooser do
+  subject(:chooser) { described_class.call(interpolatable) }
+
+  context 'when a version 2 serializer exists' do
+    let(:interpolatable) { Move }
+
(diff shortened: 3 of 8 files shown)


app/models/generic_event.rb
```diff
@@ -161,17 +161,10 @@ def self.relationship_attributes(attributes)
       @serializer ||=
         Class.new(GenericEventSerializer).tap do |serializer|
```
smoothcontract, line 162: How do you feel about renaming this to something more explicit e.g. `new_serializer_class` ?

app/models/generic_event.rb
```diff
@@ -161,17 +161,10 @@ def self.relationship_attributes(attributes)
...
-                                    relation_serializer
-                                  else
-                                    relation_serializer.gsub('V2::', '')
-                                  end
-
             named_relationship_key = attribute_key.to_s.sub('_id', '')
+
             serializer.set_type :events
-            serializer.has_one named_relationship_key, serializer: relation_serializer.constantize
+            serializer.has_one named_relationship_key, serializer: SerializerVersionChooser.call(attribute_type.to_s.singularize.camelize)
```
smoothcontract, line 167: It would be really nice to do this conversion inside the new chooser class if possible...

```suggestion
            new_serializer_class.has_one named_relationship_key, serializer: SerializerVersionChooser.call(attribute_type)
```

app/services/serializer_version_chooser.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class SerializerVersionChooser
+  DEFAULT_API_VERSION = 'V2'
+
+  def self.call(interpolatable)
+    serializer = "V2::#{interpolatable}Serializer"
```
smoothcontract, line 7:
```suggestion
    serializer = "V2::#{interpolatable.to_s.singularize.camelize}Serializer"
```

Tested this in IRB and seems to work fine passing in either `:move` symbol or `Move` class.

smoothcontract: Also the constant doesn't seem to be used - could you add that rather than `V2` here and on line 12 pls?

app/services/serializer_version_chooser.rb
```diff
@@ -0,0 +1,17 @@
...
+
+    serializer = if const_defined?(serializer)
+                   serializer
+                 else
+                   serializer.sub('V2::', '')
+                 end
+
+    serializer.constantize
```
smoothcontract, line 15: Could you add coverage for an undefined V1 serializer here too? I'm curious what the behaviour would be - whatever it is it's probably useful to document via a spec. e.g. `SerializerVersionChooser.call(:foo)`

smoothcontract: Never mind, I see you already have this covered 👍

spec/services/serializer_version_chooser_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    let(:interpolatable) { Location }
+
+    it { is_expected.to eq(LocationSerializer) }
+  end
+
+  context 'when the serializer does not exist' do
+    let(:interpolatable) { Notifier }
+
+    it { expect { chooser }.to raise_error(NameError, /uninitialized constant NotifierSerializer/) }
+  end
```
smoothcontract, line 24: Maybe add another spec here (if you follow my suggestion to support passing in a symbol and/or a class)

```
context 'when the serializer is passed as a symbol' do
  let(:interpolatable) { :move }
    ...
```

smoothcontract (commented):
Like the idea of this, just a few small things to potentially tidy up slightly please 👍

app/models/generic_event.rb
willfish: Fine by me :)

app/models/generic_event.rb
willfish: That's fine for everything except the `AllocationsSerializer` which goes against the singular grain of `<model>Serializer` naming. We're unlikely to ever create an event about an Allocation record though? Since it's just a bag of individual moves?

willfish: We can always make it simpler to handle that situation later, though.

app/services/serializer_version_chooser.rb
willfish: Done

spec/services/serializer_version_chooser_spec.rb
willfish: Also done

app/models/generic_event.rb
```diff
@@ -159,19 +159,12 @@ def self.relationship_attributes(attributes)
...
-                                  else
-                                    relation_serializer.gsub('V2::', '')
-                                  end
-
             named_relationship_key = attribute_key.to_s.sub('_id', '')
-            serializer.set_type :events
-            serializer.has_one named_relationship_key, serializer: relation_serializer.constantize
+
+            new_serializer_class.set_type :events
+            new_serializer_class.has_one named_relationship_key, serializer: SerializerVersionChooser.call(attribute_type.to_s.singularize.camelize)
```
smoothcontract, line 167: You don't need to append `.to_s.singularize.camelize` as that's now happening inside the service 😀 
```suggestion
            new_serializer_class.has_one named_relationship_key, serializer: SerializerVersionChooser.call(attribute_type)
```

spec/services/serializer_version_chooser_spec.rb
smoothcontract: Nice 👍

smoothcontract (commented):
Looks good - just one more small change I spotted pls 🙏

app/models/generic_event.rb
smoothcontract: We have singular and plural versions of both move and allocation serializers now, but I would imagine that an event only relates to a singular resource so this shouldn't be a problem.

app/models/generic_event.rb
willfish: Thanks. My derp

smoothcontract (approved):
All looks good 👍

PR #1118, opened 2020-11-06
P4 2339 migrate category attribute to model

martyn-w:
### Jira link

P4-2339

### What?

I have added/removed/altered:

- [x] Switched profile model from category_code attributes to category model
- [x] Add rake task to migrate category_code to category model _without_ touching the updated_at flag

### Why?

I am doing this because:

- to tidy up profile/categories relationships
- we shouldn't touch profile updated_at otherwise JPC reports will fail

### Have you? (optional)


cwrw (commented):
I was wondering about how often categories change or get updated. Will we ever risk pulling in a category for someone that we don't have in our Categories list, meaning we miss assigning an id to a profile correctly?

app/controllers/api/profiles_controller.rb
```diff
@@ -50,12 +50,12 @@ def update
...
       profile_attributes[:documents] = documents unless document_attributes.nil?
-      profile_attributes.merge!(category_attributes)
+      profile_attributes[:category] = category if person.latest_nomis_booking_id.present?
       profile_attributes
     end
 
-    def category_attributes
-      NomisClient::BookingDetails.get(person.latest_nomis_booking_id)
+    def category
+      Category.find_by(key: NomisClient::BookingDetails.get(person.latest_nomis_booking_id)[:category_code])
```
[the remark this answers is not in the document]
martyn-w: yep, that is a good point, I'll consolidate that code 👍

martyn-w:
> I was wondering about how often categories change or get updated. Will we ever risk pulling in a category for someone that we don't have in our Categories list, meaning we miss assigning an id to a profile correctly?

For the moving people side of things, the important thing is that we maintain a list of categories which we cannot move (Cat-A etc). Generally any changes to the list of categories we cannot move will be planned and co-ordinated. 

If new categories are created in Nomis for which we do not have a mapping, people will be imported as uncategorised and there will be no impact on their move-ability (as we allow all moves except for the pre-determined set of categories for which they forbidden).

lib/tasks/refresh_prisoner_category.rake
```diff
@@ -6,7 +6,12 @@ namespace :prisoner_category do
 
     puts "Refreshing #{count} profiles"
     profiles.find_each.with_index do |profile, i|
-      Profiles::ImportPrisonerCategory.new(profile).call
+      latest_nomis_booking_id = profile&.person&.latest_nomis_booking_id
+      if latest_nomis_booking_id.present?
+        profile.category = Categories::FindByNomisBookingId.new(latest_nomis_booking_id).call
```
cwrw, line 11: I think assigning a category this way actually updates the record?

martyn-w: 😱 arghhhhh 😱

martyn-w: @cwrw I've double-checked and the assignment _does not_ save the record - so I think this is fine:

```
[39] pry(main)> p.category
=> nil
[40] pry(main)> p.category = Category.first
  Category Load (0.4ms)  SELECT "categories".* FROM "categories" ORDER BY "categories"."id" ASC LIMIT $1  [["LIMIT", 1]]
=> #<Category:0x00007f97c881a9e0 id: "1759804b-f958-44f9-b085-fd99bf7fbdbd", key: "CYOI", title: "Category C / YOI", move_supported: true, created_at: Tue, 03 Nov 2020 11:17:58 GMT +00:00, updated_at: Tue, 03 Nov 2020 11:17:58 GMT +00:00>
[41] pry(main)> p.updated_at
=> Fri, 06 Nov 2020 16:00:45 GMT +00:00
[42] pry(main)> p.save
   (0.3ms)  BEGIN
  Profile Update (0.7ms)  UPDATE "profiles" SET "updated_at" = $1, "category_id" = $2 WHERE "profiles"."id" = $3  [["updated_at", "2020-11-06 16:01:13.895585"], ["category_id", "1759804b-f958-44f9-b085-fd99bf7fbdbd"], ["id", "6388e8bf-9845-466a-8540-8f31d800fea1"]]
  PaperTrail::Version Create (0.4ms)  INSERT INTO "versions" ("item_type", "item_id", "event", "object", "created_at") VALUES ($1, $2, $3, $4, $5) RETURNING "id"  [["item_type", "Profile"], ["item_id", "6388e8bf-9845-466a-8540-8f31d800fea1"], ["event", "update"], ["object", "---\nid: 6388e8bf-9845-466a-8540-8f31d800fea1\nperson_id: 0003a13f-a6e4-4b33-aa57-60c172f993e7\ncreated_at: &1 2020-11-06 16:00:45.836746000 Z\nupdated_at: *1\nassessment_answers: \"[]\"\ncategory_id: \n"], ["created_at", "2020-11-06 16:01:13.895585"]]
  Person Update (0.4ms)  UPDATE "people" SET "updated_at" = $1 WHERE "people"."id" = $2  [["updated_at", "2020-11-06 16:01:13.899660"], ["id", "0003a13f-a6e4-4b33-aa57-60c172f993e7"]]
  PaperTrail::Version Create (0.3ms)  INSERT INTO "versions" ("item_type", "item_id", "event", "object", "created_at") VALUES ($1, $2, $3, $4, $5) RETURNING "id"  [["item_type", "Person"], ["item_id", "0003a13f-a6e4-4b33-aa57-60c172f993e7"], ["event", "update"], ["object", "---\nid: 0003a13f-a6e4-4b33-aa57-60c172f993e7\ncreated_at: 2020-07-09 17:49:53.898335000 Z\nupdated_at: 2020-11-06 16:01:13.899660000 Z\nnomis_prison_number: \nprison_number: \ncriminal_records_office: \npolice_national_computer: 0TFIZ1/VZ\nfirst_names: Jett\nlast_name: Price\ndate_of_birth: 1955-06-02\ngender_additional_information: ''\nethnicity_id: 379ced2e-a3f2-44dd-8ebe-e935e600ad68\ngender_id: 1b227566-ef41-4516-b41d-9aec6651243a\nlast_synced_with_nomis: \nlatest_nomis_booking_id: \n"], ["created_at", "2020-11-06 16:01:13.899660"]]
   (5.7ms)  COMMIT
=> true
[43] pry(main)> p.updated_at
=> Fri, 06 Nov 2020 16:01:13 GMT +00:00
```

cwrw: ah your right, its only on the has many, single ones don't get updated, phew! It caught me out a bit when updating flags but this is fine 👍

cwrw (approved):
Thanks for the refactor looks really good 👍

app/services/categories/find_by_nomis_booking_id.rb
```diff
@@ -0,0 +1,19 @@
+module Categories
+  class FindByNomisBookingId
+    attr_reader :nomis_booking_id
+
+    def initialize(nomis_booking_id)
+      @nomis_booking_id = nomis_booking_id
+    end
+
+    def call
+      Category.find_by(key: booking_details[:category_code]) if nomis_booking_id.present?
```
smoothcontract, line 10: @martyn-w is there is a possibility that `booking_details` could be nil here? If so might be worth adding a spec in a follow up PR just to explore the behaviour for that scenario and potentially add a nil guard if needed. I guess for now this will just return a 500 error which might not be the friendliest behaviour for the front end 🤔

martyn-w: Hi @smoothcontract I don't see how that could happen?: the `booking_details` variable will either be the an object like this: 
`{category: 'foo', category_code: 'F'}` 
or, if not found, an object like this:
`{category: nil, category_code: nil}` 

see the `app/lib/nomis_client/booking_details.rb` library

cwrw: Im not sure that could happen if we get a successful response from NOMIS, as we are wrapping it within our own hash and keys: https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/lib/nomis_client/booking_details.rb#L27
If NOMIS itself does raise an error, we now rescue it in the API controller and return an appropriate message https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/controllers/api_controller.rb#L32

PR #1121, opened 2020-11-09
Add `completed_at` timestamp on PER when status transitions to complete

cwrw:
To capture stats on when users are *first* completing a PER, add a new timestamp to capture the first date/time a user completes the PER. If the PER moves back to in progress then completed again we don't overwrite the timestamp with the new value, but maintain the old one. This is only internal for now so won't be surfaced in serializers.


martyn-w (approved):
👍

PR #1122, opened 2020-11-09
Remove prefill feature flag

cwrw:
Pre-filling has been live for a while now without incident. Remove feature flag from codebase to make this feature permanent.


martyn-w (approved):
👍

PR #1126, opened 2020-11-10
Add metadata to GET /allocations/:id endpoint

smoothcontract:
### Jira link

P4-2363

### What?

- [x] Add `meta` tag to the `GET /allocations/:id` endpoint

### Why?

- This replicates the same metadata included for each allocation as part of the GET /allocations endpoint, allowing some code cleanup in the front end.

### Have you? (optional)

- Adding `meta` tags is a backwards compatible change; clients can ignore it if not required


app/models/allocation.rb
```diff
@@ -92,6 +92,14 @@ def refresh_status_and_moves_count!
     save!
   end
 
+  def move_totals
```
cwrw, line 95: is there a way we can somehow adapt `self.move_totals` to support both a single and multiple allocations to avoid duplication?

cwrw: Or maybe the instance method can invoke the class method on an active record relation from the single allocation

smoothcontract: Its's a good question and is something I considered but decided against as although the logic for the the metadata rules is the same in both single and multiple cases, the implementation has to be widely different to be performant. The class version of this method returns a hash of allocation id's for each set of metadata, so would need quite a bit of munging both in calling and consuming the class method for it to be shared, and overall that would look worse and potentially more confusing than keeping them separate.

spec/requests/api/allocations_controller_show_spec.rb
```diff
@@ -0,0 +1,91 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Api::AllocationsController do
```
smoothcontract, line 5: This was an RSwag test previously so I took the opportunity to rewrite it as a conventional request spec, however it's basically testing the same thing but with more explicit checking of the response.

spec/swagger/swagger_doc_v1.yaml
```diff
@@ -284,6 +284,48 @@
...
+        - "$ref": "../v1/content_type_parameter.yaml#/ContentType"
+        - "$ref": "../v1/allocation_id_parameter.yaml#/AllocationId"
+        - "$ref": "../v1/allocation_include_parameter.yaml#/AllocationIncludeParameter"
+      responses:
+        "200":
+          description: success
+          content:
+            application/vnd.api+json:
+              schema:
+                "$ref": "../v1/get_allocation_responses.yaml#/200"
```
cwrw, line 304: Could you amend `get_allocation_responses.yaml` as its not structured correctly thus when viewing the swagger docs generated it's coming back as `{}` for the 200 response

cwrw (approved):
Looks good, just a minor comment for swagger

spec/swagger/swagger_doc_v1.yaml
smoothcontract: Good catch, well spotted and now resolved 👍

PR #1131, opened 2020-11-11
Ensure that empty Person reference values are saved as null

smoothcontract:
### Jira link

P4-2456

### What?

- [x] Added `auto_strip_attributes` gem and annotations to `Person` model
- [x] Added `data_maintenance:fix_blank_person_references` rake task

### Why?

- We (or rather suppliers) don't want empty strings for Person reference values to be stored as empty string, but should instead be stored as `null. Added `auto_strip_attributes` gem as the easiest way to achieve this as it's likely we will want to add this to more models over time. The gem is very minimal (only 100 LOC) and well maintained so I think is justified to use. Also added a rake task to clean up historic values to be run in each environment after the fix has been deployed.

https://github.com/holli/auto_strip_attributes

- Rake task changes production data but only converts empty strings to null so is safe to run, and has been tested locally.

### DIFF
--- a/lib/tasks/data_maintenance.rake
+++ b/lib/tasks/data_maintenance.rake
@@ -28,4 +28,12 @@ namespace :data_maintenance do
   task fix_nil_allocations_estate: :environment do
     Allocation.where(estate: nil).update_all(estate: :adult_male)
   end
+
+  desc 'fix blank (empty string) person references that should be stored as null'
+  task fix_blank_person_references: :environment do
+    Person.where(nomis_prison_number: '').update_all(nomis_prison_number: nil)
+    Person.where(prison_number: '').update_all(prison_number: nil)
+    Person.where(criminal_records_office: '').update_all(criminal_records_office: nil)
+    Person.where(police_national_computer: '').update_all(police_national_computer: nil)
+  end
 end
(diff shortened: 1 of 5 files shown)


lib/tasks/data_maintenance.rake
```diff
@@ -28,4 +28,12 @@ namespace :data_maintenance do
   task fix_nil_allocations_estate: :environment do
     Allocation.where(estate: nil).update_all(estate: :adult_male)
   end
+
+  desc 'fix blank (empty string) person references that should be stored as null'
```
martyn-w, line 32: Might this trigger a JPC/CDI catastrophe by causing all the updated_at flags of related records to be touched?

smoothcontract: No, Rails doesn't update the timestamps when using `.update_all` so this won't cause a hub meltdown. https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-update_all

> Note: As Active Record callbacks are not triggered, this method will not automatically update updated_at/updated_on columns.

martyn-w: jolly good 👍

martyn-w (approved):
👍 LGTM

PR #1132, opened 2020-11-12
Fix method name for rendering NOMIS error

cwrw:
To fix https://sentry.io/organizations/ministryofjustice/issues/2024795868/?project=2014776&referrer=slack
Amend method name and add test for NOMIS OAuth2 errors when NOMIS responds with a 500


smoothcontract (approved):
Nicely done 👍

PR #1134, opened 2020-11-12
P4-2447 Add `assessmentable` association to responses

cwrw:
### Jira link

[P4-2447](https://dsdmoj.atlassian.net/browse/P4-2447)

### What?

I have added/removed/altered:

- [x] Added a new polymorphic association to Framework responses called assessmentable
- [x] Renamed all associations to assessmentable whilst keeping backward compatibility to `person_escort_record`
- [x] Exposed assessmentable in framework response serializer and added appropriate includes
- [x] Add rake task to backfill data on existing framework responses

### Why?

To add the Youth Risk Assessment, we need to be able to support multiple types of assessments on framework responses. Adding a new polymorphic association helps achieve this goal. Since this is a breaking change, we will first roll out support for this new relationship, whilst maintaining the old one. After we backfill all the data on production using the rake task included, and the frontend client starts utilising the new relationship, we can folllowup with a PR to remove the old relationship.

### Have you? (optional)

- Hopefully this shouldn't affect existing data, but rather adds a new relationship whilst keeping the old one

### DIFF
--- a/app/models/framework_question.rb
+++ b/app/models/framework_question.rb
@@ -22,13 +22,13 @@ class FrameworkQuestion < VersionedModel
   has_many :framework_responses
   has_and_belongs_to_many :framework_nomis_codes, autosave: true
 
-  def build_responses(question: self, person_escort_record:, questions:, previous_responses: {})
-    response = build_response(question, person_escort_record, previous_responses[question.key])
+  def build_responses(question: self, assessmentable:, questions:, previous_responses: {})
+    response = build_response(question, assessmentable, previous_responses[question.key])
     return response if question.dependents.empty? || question.question_type == 'add_multiple_items'
 
     question.dependents.each do |dependent_question|
       # NB: to avoid extra queries use original set of questions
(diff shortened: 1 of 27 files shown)


app/models/framework_question.rb
```diff
@@ -62,6 +62,6 @@ def build_response(question, person_escort_record, previous_response = nil)
         FrameworkResponse::String
       end
 
-    klass.new(framework_question: question, person_escort_record: person_escort_record, value: previous_response, prefilled: previous_response.present?)
+    klass.new(framework_question: question, assessmentable: assessmentable, person_escort_record: assessmentable, value: previous_response, prefilled: previous_response.present?)
```
martyn-w, line 65: is the `person_escort_record:` key still needed?

martyn-w: my bad - I guess this is required temporarily whilst this change is rolled out

martyn-w (approved):
👍  looks fine to me - mostly a rename of `person_escort_record` to `assessmentable`

PR #1135, opened 2020-11-13
Keep old implementation to rely on person escort record relationship

cwrw:
Revert to using old association, as if we use assessmentable now, between the time we run the rake task users will have a glitch where person escort record will not return any responses. Only switch to assessmentable when backfill is complete.


martyn-w (approved):
👍

PR #1136, opened 2020-11-13
new locations for suppliers

martyn-w:
New supplier mappings for locations

### DIFF
--- a/lib/tasks/data/supplier_locations_go_live.yml
+++ b/lib/tasks/data/supplier_locations_go_live.yml
@@ -1,7 +1,7 @@
 effective_from: "2020-08-29"
 effective_to: null
 suppliers:
-  geoamey:	
+  geoamey:
     - ACI
     - AGI
     - BCI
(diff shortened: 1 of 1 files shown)


lib/tasks/data/supplier_locations_go_live.yml
```diff
@@ -1,7 +1,7 @@
 effective_from: "2020-08-29"
 effective_to: null
 suppliers:
-  geoamey:	
+  geoamey:
```
cwrw, line 4: just to confirm we don't need to do this across two files? `supplier_locations.yml`?

cwrw (approved):
Looks good, just a question around needing to replicate those across the two data files (although the other one might be deprecated now)

lib/tasks/data/supplier_locations_go_live.yml
martyn-w: yep, updating the old file isn't necessary

PR #1137, opened 2020-11-13
Use PER instead of assessmentable for updating framework responses

cwrw:
To avoid any glitches in the FE before backfilling, use old column on framework bulk updates as well.


martyn-w (approved):
👍 LGTM

PR #1139, opened 2020-11-16
P4-2248 Stop using `person_escort_record` relationship on framework responses

cwrw:
jira-ticket: https://dsdmoj.atlassian.net/browse/P4-2448

Stop using the `person_escort_record` relationship on responses, although we still populate the `person_escort_record` value on creation. This is due to our deployment strategy, where we need to support both old and new pods, and the old pods will still be reading from the old column, so to avoid any glitched this change needs to be staggered. After this is fully deployed we can finally get rid of the PER column safely.


martyn-w (approved):
👍 LGTM

PR #1140, opened 2020-11-16
P4-2248 Stop writing to person_escort_record_id in responses

cwrw:
* Make the `person_escort_record_id` column optional, but keep for now to support staggered release
* Stop writing to the `person_escort_record_id` column when populating responses


smoothcontract (approved):
LGTM 👍

PR #1141, opened 2020-11-16
P4-2248 Remove person_escort_record_id from framework responses

cwrw:
jira-ticket: https://dsdmoj.atlassian.net/browse/P4-2448

Since we no longer need to populate/use the `person_escort_record_id` on framework responses, remove this column.


smoothcontract (approved):
LGTM 👍

PR #1148, opened 2020-11-17
Add custom serializer for prefill source with fewer fields

smoothcontract:
### Jira link

P4-2480

### What?

- [x] Use a custom lightweight serializer for person escort record prefill source

### Why?

- The front end currently only requires the `confirmed_at` attribute for the previously filled person escort record. However the prefill source is currently using the same `PersonEscortRecordSerializer` which requires lots of additional queries to fetch associated framework questions and responses, none of which are then consumed by the front end client.

- Since prefilling was introduced, observed performance in production has dipped significantly as fetching a move is now effectively returning two complete PER resources. The new lightweight serializer for the prefilled PER is effectively a subset of the PER serializer, but contains only those attributes which can be served directly from the prefill PER record and without any additional queries. This should support the existing front end code with no changes, yet perform much faster in the production environment.

### DIFF
--- a/app/serializers/person_escort_record_serializer.rb
+++ b/app/serializers/person_escort_record_serializer.rb
@@ -8,7 +8,7 @@ class PersonEscortRecordSerializer
   belongs_to :profile, serializer: V2::ProfileSerializer
   belongs_to :move, serializer: V2::MoveSerializer
   belongs_to :framework
-  belongs_to :prefill_source, serializer: PersonEscortRecordSerializer
+  belongs_to :prefill_source, serializer: PrefillSourceSerializer
 
   has_many :responses, serializer: FrameworkResponseSerializer do |object|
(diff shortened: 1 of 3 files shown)


smoothcontract:
Just as a side note, this approach is within the permitted JSONAPI specs, specifically this section https://jsonapi.org/format/#fetching-sparse-fieldsets

> If a client does not specify the set of fields for a given resource type, the server MAY send all fields, a subset of fields, or no fields for that resource type.

Implementing full support for sparse fieldset params is a fair bit of work and unlikely to happen before the end of the project, so this is essentially a quick compromise.

cwrw (approved):
looks good 👍

smoothcontract:
@cwrw and @teneightfive tested on a Heroku review app and also ran end to end test suite against this branch, and PR prefilling is all working as expected.

PR #1152, opened 2020-11-18
P4-2446 Use exact Framework name on creation of PER

cwrw:
### Jira link

[P4-2446](https://dsdmoj.atlassian.net/browse/P4-2446)

### What?

- Look for specific framework name on PER creation
- Amend framework FactoryBot fixture
- Rename framework file names
- Extend tests to check for different frameworks

### Why?

In preparation of introducing the Youth Risk Assessment, look for exact name of PER framework on PER creation. This requires amending the FactoryBot fixture to be more representative in naming the framework, however to avoid validation errors of uniqueness of name/version across different tests, find or create the framework on initialisation.

I've also renamed some fixture tests files for importing frameworks, to be more representative of real framework data we will be importing.


smoothcontract (approved):
LGTM 👍

PR #1153, opened 2020-11-18
[P4-2424] Add MoveRequested and MoveProposed events for timeline

willfish:
### Jira link

P4-2424

### What?

These events need to be generated to indicate the starting point of the
move in the timeline and supplement the events that Martyn added for
Journeys which are now deprecated/not used anywhere.

### Why?

- This is needed to indicate the initial starting point in the events timeline


martyn-w (approved):
👍 LGTM

PR #1157, opened 2020-11-19
[P4-2424] Fix supplier on new move events

willfish:
### Jira link

P4-2424

### What?

I have added/removed/altered:

- [x] Altered the supplier that gets associated with a move initial state event to be based on the application owner rather than the supplier of the move.

### Why?

I am doing this because:

- The UI needs to know the original creator of the move rather than the inferred supplier picked by the `SupplierChooser`

### DIFF
--- a/app/controllers/api/v2/moves_actions.rb
+++ b/app/controllers/api/v2/moves_actions.rb
@@ -166,7 +166,7 @@ module Api::V2
         recorded_at: now,
         notes: 'Automatically generated event',
         details: {},
-        supplier_id: move.supplier_id,
+        supplier_id: doorkeeper_application_owner&.id,
       }
 
       if move.proposed?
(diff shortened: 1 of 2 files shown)


app/controllers/api/v2/moves_actions.rb
```diff
@@ -166,7 +166,7 @@ def create_event(move)
         recorded_at: now,
         notes: 'Automatically generated event',
         details: {},
-        supplier_id: move.supplier_id,
+        supplier_id: doorkeeper_application_owner&.id,
```
martyn-w, line 169: the field `supplier_id` sounds like it is misnamed: 
* is it the supplier who will perform the move?
* the user (which may or may not be a supplier) who created the move?
if the latter, then a different name might be more appropriate (e.g. created_by)

willfish: We're going to merge this and fix this after we know the frontend user token that created the event as part of the `created_by` attribute that we intend to add to all events later on.

willfish: Had a conversation with Martyn on a call just now.

martyn-w (commented):
👍  good for now

martyn-w (approved):

PR #1159, opened 2020-11-19
[P4-2507] Adjust supplier on move initial status events

willfish:
### Jira link

P4-2507

### What?

I have added/removed/altered:

- [x] Removes automatic creation of old event model events
- [x] Updates fake data generator to generate initial status events
- [x] Adds rake task for enumerating and correcting initial status events (`MoveProposed`, `MoveRequested`) that have the wrong supplier_id

### Why?

This corrects a UI issue for the frontend where the presented data on the timeline indicates the wrong information about who created the move.
It also corrects cruft that should have been removed previously.

### DIFF
--- a/app/controllers/api/journeys_controller.rb
+++ b/app/controllers/api/journeys_controller.rb
@@ -147,11 +147,11 @@ module Api
         supplier_id: supplier.id,
       }
 
-      if action_name == "update"
+      if action_name == 'update'
         GenericEvent::JourneyUpdate.create!(event_attributes)
       end
 
-      if action_name == "create"
+      if action_name == 'create'
         GenericEvent::JourneyCreate.create!(event_attributes)
       end
     end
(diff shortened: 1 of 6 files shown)


app/controllers/api/journeys_controller.rb
```diff
@@ -136,22 +136,24 @@ def update_journey_attributes
...
-      event.update!(generic_event_id: generic_event.id)
+      now = Time.zone.now.iso8601
+
+      event_attributes = {
+        eventable: journey,
+        occurred_at: now,
+        recorded_at: now,
+        notes: 'Automatically generated event',
+        details: {},
+        supplier_id: supplier.id,
```
willfish, line 147: This `supplier` getter is different from the one we have for the MoveProposed and MoveRequested events. It gives preference to the id as an attribute coming from the json:api relationship and only falls back on the supplier coming from the application owner.

This override could be confusing the meaning of this id (indicating who owns instead of who created the event)

willfish: I'm being cheeky getting this related/unrelated change in @martyn-w . If it's too clutterful for review I'll remove it :).

martyn-w (approved):
👍 LGTM

PR #1160, opened 2020-11-20
Use custom event type for PER completion notifications

smoothcontract:
### Jira link

P4-2485

### What?

- [x] Revised implementation of PER notifications to support a dedicated PER completion event type

### Why?

- Previously PER completions were triggering move update notifications for all moves related to the per profile; this is a legacy approach predating the new direct relationship between a PER and the associated move. We also want to use a dedicated event type of `confirm_person_escort_record` for PER completion notifications.

- Suppliers are not yet making use of PER webhooks, so this PR aims to improve the implementation prior to it being integrated with suppliers and then being harder to change.

### DIFF
--- a/app/jobs/prepare_move_notifications_job.rb
+++ b/app/jobs/prepare_move_notifications_job.rb
@@ -56,6 +56,6 @@ private
       'update' => 'update_move',
       'update_status' => 'update_move_status',
       'destroy' => 'destroy_move',
-    }[action_name]
+    }[action_name] || action_name
   end
 end
--- a/app/jobs/prepare_person_escort_record_notifications_job.rb
+++ b/app/jobs/prepare_person_escort_record_notifications_job.rb
@@ -7,25 +7,6 @@ class PreparePersonEscortRecordNotificationsJob < ApplicationJob
     person_escort_record = PersonEscortRecord.find(topic_id)
     return unless (move = person_escort_record.move)
 
-    [move.supplier || move.suppliers].flatten.each do |supplier|
-      supplier.subscriptions.kept.each do |subscription|
-        next unless subscription.enabled? && subscription.callback_url.present?
-
-        NotifyWebhookJob.perform_later(
-          notification_id: build_notification(subscription, NotificationType::WEBHOOK, move).id,
-          queue_as: move_queue_priority(move),
-        )
-      end
-    end
-  end
-
-private
-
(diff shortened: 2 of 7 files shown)


app/jobs/prepare_person_escort_record_notifications_job.rb
```diff
@@ -3,9 +3,29 @@
...
   include QueueDeterminer
 
-  def perform(topic_id:, action_name:, **_)
-    PersonEscortRecord.find(topic_id).profile.moves.find_each do |move|
-      PrepareMoveNotificationsJob.perform_now(topic_id: move.id, action_name: action_name, queue_as: move_queue_priority(move))
+  def perform(topic_id:, **_)
+    person_escort_record = PersonEscortRecord.find(topic_id)
+    return unless (move = person_escort_record.move)
+
+    [move.supplier || move.suppliers].flatten.each do |supplier|
```
cwrw, line 10: There is a bit of  duplication here, is there something we can do to choose which notification we want and passing it in to `PrepareMoveNotificationsJob` instead?

cwrw: Update: I just confirmed we will need to send both webhooks and emails unfortunately for this notification

smoothcontract: 👍

app/jobs/prepare_move_notifications_job.rb
```diff
@@ -56,6 +56,6 @@ def event_type(action_name)
       'update' => 'update_move',
       'update_status' => 'update_move_status',
       'destroy' => 'destroy_move',
-    }[action_name]
+    }[action_name] || action_name
```
cwrw, line 59: What's the reason you chose not to add `confirm_person_escort_record` to the as a key, I guess its not related to move?

cwrw (approved):
looks good, just a question around including the action as part of moves or not

app/jobs/prepare_move_notifications_job.rb
smoothcontract: I decided it was fairly pointless adding a mapping for `confirm_person_escort_record` here, and instead allowed the event type to be passed in directly by adding this fallback. The existing mappings are for move related actions; the PER completion is piggy backing onto this functionality but doesn't fit with the same model of actions really.

PR #1161, opened 2020-11-20
[P4-1981] Adds location relationships for the event feed

willfish:
### Jira link

P4-1981

### What?

Locations have their own way of appearing in the feed via the Location#for_feed method. We need to make sure that all new events that will be getting created by the suppliers are serialised correctly for the hub by using this method. The approach I've taken is to tidy all existing for_feed methods that define a location and consolidate them under a module for this specific purpose. This cleans up quite a bit of code :tada:.

I have added/removed/altered:

- [x] Added module for feeds that serialize locations for the hub and consume it in all GenericEvent classes that have a location relationship

### Why?

These are needed for the hub.

### DIFF
--- a/app/models/concerns/location_feed.rb
+++ b/app/models/concerns/location_feed.rb
@@ -1,0 +1,16 @@
+module LocationFeed
+  def for_feed
+    super.tap do |common_feed_attributes|
+      location_key = self.class::LOCATION_ATTRIBUTE_KEY.to_s
+      location_method = location_key.sub('_id', '')
+      prefix = location_key == 'location_id' ? nil : location_key.sub('_location_id', '')
+
+      feed_details = common_feed_attributes['details']
+        .deep_dup
+        .except(location_key)
+      feed_details.merge!(public_send(location_method).for_feed(prefix: prefix))
+
+      common_feed_attributes['details'] = feed_details
+    end
+  end
+end
--- a/app/models/concerns/location_feed.rb
+++ b/app/models/concerns/location_feed.rb
@@ -5,12 +5,10 @@ module LocationFeed
       location_method = location_key.sub('_id', '')
       prefix = location_key == 'location_id' ? nil : location_key.sub('_location_id', '')
 
-      feed_details = common_feed_attributes['details']
-        .deep_dup
-        .except(location_key)
-      feed_details.merge!(public_send(location_method).for_feed(prefix: prefix))
+      updated_location = public_send(location_method).for_feed(prefix: prefix)
 
-      common_feed_attributes['details'] = feed_details
+      common_feed_attributes['details'].merge!(updated_location)
+      common_feed_attributes['details'].delete(location_key)
     end
   end
 end
--- a/app/models/generic_event/move_lockout.rb
+++ b/app/models/generic_event/move_lockout.rb
@@ -4,10 +4,11 @@ class GenericEvent
 
     details_attributes :authorised_at, :authorised_by, :reason
     relationship_attributes from_location_id: :locations
-
     eventable_types 'Move'
+
     include AuthoriserValidations
     include LocationValidations
+    include LocationFeed
 
     enum reason: {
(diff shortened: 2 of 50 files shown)


app/models/concerns/location_feed.rb
```diff
@@ -0,0 +1,16 @@
...
+      location_key = self.class::LOCATION_ATTRIBUTE_KEY.to_s
+      location_method = location_key.sub('_id', '')
+      prefix = location_key == 'location_id' ? nil : location_key.sub('_location_id', '')
+
+      feed_details = common_feed_attributes['details']
+        .deep_dup
+        .except(location_key)
+      feed_details.merge!(public_send(location_method).for_feed(prefix: prefix))
+
+      common_feed_attributes['details'] = feed_details
```
smoothcontract, line 13: This cloning and merging stuff seems slight overkill - could the `details` part not be updated in situ instead?

```suggestion
      updated_location_mapping = public_send(location_method).for_feed(prefix: prefix))
      common_feed_attributes['details'].delete(location_key).merge!(updated_location_mapping)
```

or perhaps the `delete` isn't even needed as the same key is being overwritten (I think).. so

```suggestion
      updated_location_mapping = public_send(location_method).for_feed(prefix: prefix))
      common_feed_attributes['details'].merge!(updated_location_mapping)
```

The name `updated_location_mapping` may need some thought - I'm assuming this returns a key/value pair. This feels clearer to follow to me (if it works!).

app/models/generic_event/move_lockout.rb
```diff
@@ -27,17 +28,6 @@ def from_location
       Location.find_by(id: from_location_id)
     end
 
-    def for_feed
-      super.tap do |common_feed_attributes|
-        common_feed_attributes['details'] = from_location.for_feed(prefix: 'from')
-        common_feed_attributes['details'].merge!(
-          'authorised_at' => authorised_at,
```
smoothcontract, line 34: These attributes don't get replaced by the new shared module - were they removed intentionally?

app/models/generic_event/move_redirect.rb
```diff
@@ -24,13 +25,6 @@ def trigger
       eventable.move_type = move_type if move_type.present?
     end
 
-    def for_feed
-      super.tap do |common_feed_attributes|
-        common_feed_attributes['details'] = to_location.for_feed(prefix: 'to')
-        common_feed_attributes['details']['move_type'] = move_type if move_type.present?
```
smoothcontract, line 30: This attribute doesn't get replaced by the new shared module - was it removed intentionally?

smoothcontract (commented):
Nice refactor, but I'm not sure if a few extra unrelated attributes were unintentionally removed too? Might be worth having a look at the shared module to see if the hash updates can be made more readable too - see suggestions.

app/models/generic_event/move_lockout.rb
willfish: They come in for free since we're simply deleting the location key rather than building the hash from scratch (see line 32 where we're assigning the details anew from the result of calling the `from_location.for_feed`

app/models/generic_event/move_redirect.rb
willfish: So this is a bit silly. Essentially the move type if it's set to anything will get set again in the details. If it's not set it will not get set. It's basically doing nothing of value and was clearly an early brain fart.

smoothcontract (approved):
LGTM and thanks for updating 👍

app/models/generic_event/move_redirect.rb
smoothcontract: Thanks for the explanation 👍

app/models/generic_event/move_lockout.rb
smoothcontract: Coolio 👍

app/models/concerns/location_feed.rb
smoothcontract: Looks nicer now, thanks for cleaning up 👍

PR #1164, opened 2020-11-24
Remove superfluous relationships and attributes from GET /moves endpoint

smoothcontract:
### Jira link

P4-2467

### What?

- [x] Added new `ConditionalRelationships` module to support conditional `has_many` and `has_one` methods
- [x] Refactored existing `MovesSerializer` and selected child serializers to leverage the new conditional resources approach
- [x] Updated/added coverage for new serializers
- [x] Dropped support for `profile.person_escort_record.framework` and `profile.person_escort_record.responses` includes within `GET /moves` endpoint.

### Why?

- We have been trying a number of techniques to improve API performance when serialising a graph of resources. The aim is to support a reasonable set of includes for each endpoint and to ensure that the associated resources are retrieved efficiently without superfluous or N+1 queries. There are known performance issues when including `profile.person_escort_record.flags` as part of a `GET /moves` request that need to be addressed.

- Previous approaches have used a hard coded list of relationships to render in the `MovesSerializer` - this works to a point if all of those relationships are passed in the `include` query string. If those are omitted then for a `has_one` or `has_many` relationship we see N+1 SQL queries as the id's for those resources cannot be retrieved directly from the parent record (move). If a `belongs_to` relationship is omitted from the `include` query string there is no additional SQL query as the child record id is present on the move record.

- This PR introduces conditional relationships, namely `has_one_if_included` and `has_many_if_included`. These relationships are only rendered in the JSONAPI output if they are explicitly requested in the `include` query string parameter. Some careful plumbing is needed to make the serializer aware of the original query string; this is added via the `paginate` and `render_json` methods from the `Serializable` concern; the `include` query string is parsed and added as a custom `included` param in the serializer constructor. These params are then available when rendering child serializers which allows us to check whether the relationship should be rendered.

- Effectively this gives the same net result as using `lazy_load_data: true` option for `has_many` and `has_one` relationships. Unfortunately the implementation of this is broken within the jsonapi-serializer gem and we see orphaned resources without the correct relationships defined. This produces broken JSONAPI output that cannot be correctly deserialized so we (for now at least) need to avoid using that approach.

- As the serializer `params` are now used to support `included`, some existing serializers for location free spaces and allocations needed a slight change to their use of `params`. These have now been namespaced rather than including parameter data at the root level. This allows simultaneous use of the conditional relationships and custom parameters.

- The change has only been applied to the `v2` `GET /moves` endpoint but this PR lays the groundwork for rolling the approach out across other endpoints over time. Client behaviour should be unaffected as the same response is returned if the correct `include` query string is passed, however this will need to be extensively tested.

### Have you? (optional)

### DIFF
--- a/app/controllers/api/allocations_controller.rb
+++ b/app/controllers/api/allocations_controller.rb
@@ -8,7 +8,7 @@ module Api
       allocations_params = Allocations::ParamsValidator.new(filter_params, sort_params)
       if allocations_params.valid?
         paginate allocations, serializer: AllocationsSerializer, include: included_relationships, fields: AllocationsSerializer::INCLUDED_FIELDS do |paginated_allocations, options|
-          options[:params] = paginated_allocations.move_totals
+          options[:params] = { totals: paginated_allocations.move_totals }
         end
       else
         render json: { error: allocations_params.errors }, status: :bad_request
--- a/app/controllers/api/populations_controller.rb
+++ b/app/controllers/api/populations_controller.rb
@@ -6,7 +6,7 @@ module Api
 
     def index
       paginate locations, serializer: LocationFreeSpacesSerializer, include: included_relationships do |paginated_locations, options|
-        options[:params] = Population.free_spaces_date_range(paginated_locations, (date_from..date_to))
+        options[:params] = { spaces: Population.free_spaces_date_range(paginated_locations, (date_from..date_to)) }
       end
     end
 
--- a/app/controllers/concerns/serializable.rb
+++ b/app/controllers/concerns/serializable.rb
@@ -34,12 +34,14 @@ private
     serializer_class = options.delete(:serializer)
     raise ArgumentError, 'serializer option must be specified' unless serializer_class
 
-    options[:params] = included_params(options)
+    options[:params] = merge_included_params(options)
 
     serializer_class.new(serializable, options.slice(:include, :fields, :meta, :links, :params))
   end
 
-  def included_params(options)
+  def merge_included_params(options)
+    # This converts a string or array of includes to a unique list of individual symbols for each level
+    # e.g. "profile.person,profile.person_escort_record" => [:profile, :person, :person_escort_record]
     includes_list = [options[:include]].flatten.compact
     unique_resources = includes_list.map { |i| i.split('.') }.flatten.uniq.map(&:to_sym)
 
--- a/app/controllers/concerns/serializable.rb
+++ b/app/controllers/concerns/serializable.rb
@@ -10,8 +10,8 @@ module Serializable
 
     yield(paginated_collection, options) if block_given?
 
-    options[:meta] = meta(paginated_collection, options)
-    options[:links] = links(paginated_collection, options)
+    options[:meta] = pagination_meta(paginated_collection, options)
+    options[:links] = pagination_links(paginated_collection, options)
     options[:json] = build_serializer(paginated_collection, options)
 
     render options.slice(:json, :status)
@@ -34,7 +34,18 @@ private
     serializer_class = options.delete(:serializer)
     raise ArgumentError, 'serializer option must be specified' unless serializer_class
 
+    options[:params] = included_params(options)
+
     serializer_class.new(serializable, options.slice(:include, :fields, :meta, :links, :params))
+  end
+
+  def included_params(options)
+    includes_list = [options[:include]].flatten.compact
+    unique_resources = includes_list.map { |i| i.split('.') }.flatten.uniq.map(&:to_sym)
+
+    options.fetch(:params, {}).merge(
+      included: unique_resources,
+    )
   end
 
   def paginate_collection(collection, options)
@@ -44,7 +55,7 @@ private
     collection.page(options[:page]).per(options[:per_page])
   end
 
-  def meta(collection, options)
+  def pagination_meta(collection, options)
     options.fetch(:meta, {}).merge(
       pagination:
       {
@@ -56,7 +67,7 @@ private
   end
 
   # Based on https://github.com/rails-api/active_model_serializers
-  def links(collection, options)
+  def pagination_links(collection, options)
     no_pages = collection.total_pages.zero?
 
     options.fetch(:links, {}).merge(
--- a/app/serializers/move_serializer.rb
+++ b/app/serializers/move_serializer.rb
@@ -22,7 +22,7 @@ class MoveSerializer
              :date_from,
              :date_to
 
-  belongs_to :person
+  has_one :person
   belongs_to :profile
   belongs_to :from_location, serializer: LocationSerializer
   belongs_to :to_location, serializer: LocationSerializer
--- a/app/serializers/move_serializer.rb
+++ b/app/serializers/move_serializer.rb
@@ -22,16 +22,17 @@ class MoveSerializer
              :date_from,
              :date_to
 
-  has_one :person
-  has_one :profile
+  belongs_to :person
+  belongs_to :profile
+  belongs_to :from_location, serializer: LocationSerializer
+  belongs_to :to_location, serializer: LocationSerializer
+  belongs_to :prison_transfer_reason
 
-  has_one :from_location, serializer: LocationSerializer
-  has_one :to_location, serializer: LocationSerializer
-  has_one :prison_transfer_reason
   has_many :documents do |object|
     object.profile&.documents
   end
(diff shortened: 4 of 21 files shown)


app/controllers/api/allocations_controller.rb
```diff
@@ -8,7 +8,7 @@ def index
       allocations_params = Allocations::ParamsValidator.new(filter_params, sort_params)
       if allocations_params.valid?
         paginate allocations, serializer: AllocationsSerializer, include: included_relationships, fields: AllocationsSerializer::INCLUDED_FIELDS do |paginated_allocations, options|
-          options[:params] = paginated_allocations.move_totals
+          options[:params] = { totals: paginated_allocations.move_totals }
```
smoothcontract, line 11: This is just namespacing the existing `params` to avoid a clash with new `included` param.

app/controllers/api/populations_controller.rb
```diff
@@ -6,7 +6,7 @@ class PopulationsController < ApiController
 
     def index
       paginate locations, serializer: LocationFreeSpacesSerializer, include: included_relationships do |paginated_locations, options|
-        options[:params] = Population.free_spaces_date_range(paginated_locations, (date_from..date_to))
+        options[:params] = { spaces: Population.free_spaces_date_range(paginated_locations, (date_from..date_to)) }
```
smoothcontract, line 9: This is just namespacing the existing `params` to avoid a clash with new `included` param.

app/controllers/concerns/serializable.rb
```diff
@@ -10,8 +10,8 @@ def paginate(*args)
 
     yield(paginated_collection, options) if block_given?
 
-    options[:meta] = meta(paginated_collection, options)
-    options[:links] = links(paginated_collection, options)
+    options[:meta] = pagination_meta(paginated_collection, options)
```
smoothcontract, line 13: Renamed these methods for improved clarity

app/serializers/v2/moves_serializer.rb
```diff
@@ -23,14 +24,9 @@ class MovesSerializer
     set_type :moves
 
     INCLUDED_FIELDS = {
-      moves: attributes_to_serialize.keys +
-        %i[profile from_location to_location prison_transfer_reason supplier],
-      profiles: ::V2::ProfileSerializer.attributes_to_serialize.keys + %i[person person_escort_record],
+      # TODO: remove these and replace with conditional relationships in relevant serializer
```
smoothcontract, line 27: I didn't want to increase the scope of this PR further and make potentially breaking changes to person and location serializers as those are used in several other API endpoints which would mean lots of additional testing. Left as a future TODO.

app/serializers/v2/moves_serializer.rb
```diff
@@ -23,14 +24,9 @@ class MovesSerializer
...
       people: ::V2::PersonSerializer.attributes_to_serialize.keys + %i[gender ethnicity],
       locations: ::LocationSerializer.attributes_to_serialize.keys,
-      prison_transfer_reasons: ::PrisonTransferReasonSerializer.attributes_to_serialize.keys,
-      suppliers: ::SupplierSerializer.attributes_to_serialize.keys,
-      allocations: ::AllocationSerializer.attributes_to_serialize.keys,
```
smoothcontract, line 33: It turns out omitting resources completely from this list means that the serializer renders everything (all attributes and relationships). Since we now use conditional relationships and custom lightweight serializers there's no need to specify which fields to render.

spec/requests/api/moves_controller_index_v2_spec.rb
```diff
@@ -23,7 +23,7 @@
     let!(:moves) { create_list :move, 2 }
 
     it_behaves_like 'an endpoint that responds with success 200' do
-      before { do_get }
+      before { do_get('?include=from_location') }
```
smoothcontract, line 26: We have a shared schema for `Move` which lists `from_location` as a required relationship, so it needs to be explicitly included here so that it gets rendered conditionally and  the response validates against the schema.

app/serializers/move_serializer.rb
```diff
@@ -22,16 +22,17 @@ class MoveSerializer
              :date_from,
              :date_to
 
-  has_one :person
-  has_one :profile
+  belongs_to :person
```
cwrw, line 25: I think this is a `has_one through` relationship on the move model

app/serializers/person_escort_records_serializer.rb
```diff
@@ -0,0 +1,12 @@
...
+
+class PersonEscortRecordsSerializer
+  include JSONAPI::Serializer
+  include JSONAPI::ConditionalRelationships
+
+  set_type :person_escort_records
+
+  has_many_if_included :flags, serializer: FrameworkFlagsSerializer, &:framework_flags
+
+  attributes :confirmed_at, :created_at, :nomis_sync_status
```
cwrw, line 11: do we not need a `SUPPORTED_RELATIONSHIPS` for flags here?

spec/serializers/person_escort_records_serializer_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    it 'contains a`flags` relationship with framework response flags' do
+      flag = create(:framework_flag)
+      create(:string_response, assessmentable: person_escort_record, framework_flags: [flag])
+
+      expect(result[:data][:relationships][:flags][:data]).to contain_exactly(
+        id: flag.id,
+        type: 'framework_flags',
+      )
+    end
```
cwrw, line 51: could you add a test for `included` section also?

app/serializers/v2/profiles_serializer.rb
```diff
@@ -0,0 +1,14 @@
...
+  include JSONAPI::Serializer
+  include JSONAPI::ConditionalRelationships
+
+  set_type :profiles
+
+  attributes :assessment_answers
+
+  belongs_to :person, serializer: ::V2::PersonSerializer
+
+  has_one_if_included :person_escort_record, serializer: PersonEscortRecordsSerializer
```
cwrw, line 13: can you add the `SUPPORTED_RELATIONSHIPS` here as well

spec/serializers/v2/profiles_serializer_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+    end
+
+    it 'contains a`person_escort_record` relationship with person escort record' do
+      person_escort_record = create(:person_escort_record, profile: profile)
+
+      expect(result[:data][:relationships][:person_escort_record][:data]).to eq({
+        id: person_escort_record.id,
+        type: 'person_escort_records',
+      })
+    end
```
cwrw, line 74: same here around the `result['included']` test

spec/serializers/v2/moves_serializer_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    it { expect(result).to include_json(expected_json) }
+    it { expect(result[:included]).to be_nil }
+  end
+
+  context 'with all supported includes' do
+    let(:options) { { include: described_class::SUPPORTED_RELATIONSHIPS } }
+
+    it 'contains all included relationships' do
+      expect(result[:included].map { |r| r[:type] })
```
cwrw, line 54: can you include a PER with flags here?

app/serializers/move_serializer.rb
smoothcontract: Good catch - I'll revert this one.

app/serializers/person_escort_records_serializer.rb
smoothcontract: No because it doesn't get rendered directly from a controller

app/serializers/v2/profiles_serializer.rb
smoothcontract: It's not needed as we don't have a profiles controller that would make use of the constant.

spec/serializers/person_escort_records_serializer_spec.rb
smoothcontract: Not sure what you mean - `result[:included]` doesn't exist - included is only present in the params which doesn't get exposed directly in the rendered JSON output.

spec/serializers/v2/moves_serializer_spec.rb
smoothcontract: will do

cwrw (approved):

PR #1166, opened 2020-11-25
P4-2449 Support multiple assessments

cwrw:
### Jira link

[P4-2449](https://dsdmoj.atlassian.net/browse/P4-2449)

### What?

I have added/removed/altered:

- Add parent assessment controller and allow PER controller to inherit from it
- Add assessmentable concern and include in PER model
- Rename validator and state machine classes to a more generic name

### Why?

To add support for the Youth risk assessment, which will initially reuse all logic from the Person Escort Record, extract shared logic into abstract parent classes and concerns.

Serializer reuse will be done in a separate pr.


app/models/concerns/framework_assessmentable.rb
```diff
@@ -0,0 +1,159 @@
...
+
+    has_many :framework_responses, as: :assessmentable, dependent: :destroy
+    has_many :generic_events, as: :eventable, dependent: :destroy # NB: polymorphic association
+
+    belongs_to :framework
+    has_many :framework_questions, through: :framework
+    has_many :framework_flags, through: :framework_responses
+    belongs_to :profile
+    belongs_to :move, optional: true
+    belongs_to :prefill_source, class_name: assessment_klass.to_s, optional: true # rubocop:disable Rails/ReflectionClassName
```
cwrw, line 32: Not sure how to get around this without disabling, see https://github.com/rubocop-hq/rubocop-rails/issues/106

app/controllers/api/person_escort_records_controller.rb
```diff
@@ -1,68 +1,15 @@
...
-    def render_person_escort_record(person_escort_record, status)
-      render_json person_escort_record, serializer: PersonEscortRecordSerializer, include: included_relationships, status: status
+    def assessment_class
+      @assessment_class ||= PersonEscortRecord
     end
 
-    def send_notification
-      Notifier.prepare_notifications(topic: person_escort_record, action_name: 'update')
+    def assessment_serializer
+      @assessment_serializer ||= PersonEscortRecordSerializer
```
cwrw, line 12: remove memoization

app/services/framework_assessments/params_validator.rb
```diff
@@ -1,12 +1,12 @@
...
 
-module PersonEscortRecords
+module FrameworkAssessments
   class ParamsValidator
     include ActiveModel::Validations
 
     attr_reader :status
 
-    validates :status, inclusion: { in: [PersonEscortRecord::PERSON_ESCORT_RECORD_CONFIRMED] }
+    validates :status, inclusion: { in: %w[confirmed] }
```
cwrw, line 9: use correct constant

app/models/concerns/framework_assessmentable.rb
cwrw: move out to specific models

willfish (approved):

spec/support/a_framework_assessment.rb
```diff
@@ -0,0 +1,712 @@
+RSpec.shared_examples 'a framework assessment' do |assessment_type|
```
smoothcontract, line 1: This approach seems fine, only minor concern is that it uses a mixture throughout of both `assessment_type` (e.g. `:person_escort_record`) and `described_class`. I think I'd prefer to only see one of those used, with a slight preference towards `assessment_type` which would allow this to be potentially included in other specs. Admittedly that's unlikely to happen but it would mean this reads more cleanly too I think.

```suggestion
RSpec.shared_examples 'a framework assessment' do |assessment_type|
  subject { create(assessment_type) }
  
  let(:assessment_class) { assessment_type.to_s.classify }
  
  ...
```

..and then use `assessment_class` rather than `described_class` throughout. Thoughts?

smoothcontract (approved):
Really nice refactor, just one minor niggle which I'm happy to discuss if you don't agree with - don't have a strong opinion around that 🥇

smoothcontract (approved):
LGTM 👍

PR #1167, opened 2020-11-26
[P4-2507] Avoid timestamp changes to events and eventables when tweaking supplier

willfish:
### Jira link

P4-2507

### What?

We need to disable timestamp and touch updates to the updated_at of both the event itself, the eventable and all of its descendants when tweaking the supplier on `MoveRequested` and `MoveProposed` events

I have added/removed/altered:

- [x] Updated the supplier tweaking rake task to disable updates to the event and eventable updated_at fields.

### Why?

I am doing this because:

- Don't overrun reports whilst correcting the supplier on `MoveRequested` and `MoveProposed` events.

### DIFF
--- a/lib/tasks/events.rake
+++ b/lib/tasks/events.rake
@@ -38,6 +38,8 @@ namespace :events do
       events_unchanged: 0,
       move_proposed_events: 0,
       move_requested_events: 0,
+      event_updated_at_changed: 0,
+      eventable_updated_at_changed: 0,
     }
 
     events.find_each do |event|
@@ -61,7 +63,17 @@ namespace :events do
       end
 
       if supplier_different
-        event.update(supplier_id: initial_version&.supplier_id) unless dry_run
+        event.record_timestamps = false
+        ActiveRecord::Base.no_touching do
+          event_updated_at = event.updated_at
+          eventable_updated_at = event.eventable.updated_at
+
+          event.update(supplier_id: initial_version&.supplier_id) unless dry_run
+
+          report[:event_updated_at_changed]     += 1 unless event.reload.updated_at == event_updated_at
+          report[:eventable_updated_at_changed] += 1 unless event.reload.eventable.updated_at == eventable_updated_at
+        end
+        event.record_timestamps = true
       end
     end
 


willfish:
```
DRY_RUN: false

{
  "total_events": 700,
  "events_changed": 0,
  "events_unchanged": 700,
  "move_proposed_events": 358,
  "move_requested_events": 342,
  "event_updated_at_changed": 0,
  "eventable_updated_at_changed": 0
}
```

lib/tasks/events.rake
```diff
@@ -61,7 +63,17 @@ namespace :events do
       end
 
       if supplier_different
-        event.update(supplier_id: initial_version&.supplier_id) unless dry_run
+        ActiveRecord::Base.record_timestamps = false
```
willfish, line 66: Required not to update the event's timestamps

lib/tasks/events.rake
```diff
@@ -61,7 +63,17 @@ namespace :events do
...
+        ActiveRecord::Base.no_touching do
```
willfish, line 67: Required not to update the event's descendants timestamps

lib/tasks/events.rake
```diff
@@ -61,7 +63,17 @@ namespace :events do
...
+          event_updated_at = event.updated_at
+          eventable_updated_at = event.eventable.updated_at
+
+          event.update(supplier_id: initial_version&.supplier_id) unless dry_run
```
cwrw, line 71: if we move to use `update_all` on a list of the events that match, we could avoid disabling things as that method  does not trigger Active Record callbacks or validations?

willfish: I guess we could. Not sure what the advantage of skipping the validations would be?

cwrw: It is mainly to avoid the callbacks of the touch model and updated at without having to disable them explicitly, not the validations.

cwrw: However is validations a concern here though?

cwrw: spoke in person, won't use `update_all`

cwrw (approved):
Looks good, only thing is could you change `ActiveRecord::Base.record_timestamps = true` and `ActiveRecord::Base.record_timestamps = false` to `event.record_timestamps = true` and `event.record_timestamps = false`, after feel free to merge

PR #1169, opened 2020-11-26
Remove support for including allocation on v1/v2 GET /moves and v1 GET /move

smoothcontract:
### Jira link

P4-2476

### What?

- [x] Cleaned up unused `allocation` includes for v1 GET /moves/:id endpoint
- [x] Cleaned up unused `allocation` includes for v1 GET /moves endpoint
- [x] Cleaned up unused `allocation` includes for v2 GET /moves endpoint

### Why?

- Suppliers have no need to see allocation details as part of a move so removing support completely under v1 API should not impact them. The v2 API still retains support for including an allocation when retrieving one move, but not multiple moves.
Including allocation details with a list of moves is extremely slow so has been replaced with custom metadata for moves counts under v2 GET /moves.

- We don't want API clients to inadvertently make use of this include so it's good to remove support and cleanup associated code.

### Have you? (optional)

- Should be no risk if suppliers and front end are not using this include. Flagging as a breaking change however to remind us to test this assumption!

### DIFF
--- a/app/serializers/v2/moves_serializer.rb
+++ b/app/serializers/v2/moves_serializer.rb
@@ -26,7 +26,6 @@ module V2
       # TODO: remove these and replace with conditional relationships in relevant serializer
       people: ::V2::PersonSerializer.attributes_to_serialize.keys + %i[gender ethnicity],
       locations: ::LocationSerializer.attributes_to_serialize.keys,
-      allocations: ::AllocationSerializer.attributes_to_serialize.keys,
     }.freeze
 
     SUPPORTED_RELATIONSHIPS = %w[
@@ -39,7 +38,6 @@ module V2
       to_location
       prison_transfer_reason
       supplier
-      allocation
     ].freeze
 
     belongs_to :from_location, serializer: ::LocationSerializer
@@ -47,6 +45,5 @@ module V2
     belongs_to :profile, serializer: V2::ProfilesSerializer
     belongs_to :prison_transfer_reason, serializer: PrisonTransferReasonSerializer
     belongs_to :supplier, serializer: SupplierSerializer
-    belongs_to :allocation, serializer: AllocationSerializer
   end
 end
--- a/app/serializers/v2/moves_serializer.rb
+++ b/app/serializers/v2/moves_serializer.rb
@@ -45,5 +45,6 @@ module V2
     belongs_to :profile, serializer: V2::ProfilesSerializer
     belongs_to :prison_transfer_reason, serializer: PrisonTransferReasonSerializer
     belongs_to :supplier, serializer: SupplierSerializer
+    belongs_to :allocation, serializer: AllocationSerializer
   end
 end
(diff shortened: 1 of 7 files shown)


spec/serializers/move_serializer_spec.rb
```diff
@@ -230,7 +168,7 @@
 
     context 'without an original_move' do
       it 'contains an empty original_move' do
-        expect(result_data[:relationships][:allocation]).to eq(data: nil)
+        expect(result_data[:relationships][:original_move]).to eq(data: nil)
```
smoothcontract, line 171: Fixing a previous copy/paste slipup here

app/serializers/v2/moves_serializer.rb
```diff
@@ -39,14 +38,12 @@ class MovesSerializer
...
       supplier
-      allocation
     ].freeze
 
     belongs_to :from_location, serializer: ::LocationSerializer
     belongs_to :to_location, serializer: ::LocationSerializer
     belongs_to :profile, serializer: V2::ProfilesSerializer
     belongs_to :prison_transfer_reason, serializer: PrisonTransferReasonSerializer
     belongs_to :supplier, serializer: SupplierSerializer
-    belongs_to :allocation, serializer: AllocationSerializer
```
smoothcontract, line 50: I'm wondering whether it's useful to leave this relationship in place so that clients could potentially make a separate API call for the relevant individual allocation 🤔

teneightfive: Does this remove the relationship completely? I think it would be good to preserve the relationships but agree that we can remove the include.

In my mind, I think it's easier to understand if the resource structure is the same both when asking for 1 or many.

If we change the structure (attributes/relationships) I've found it can lead to confusion when trying to develop.

smoothcontract: Yes, I think I agree. The danger is including an allocation, but referencing the allocation relationship doesn't have any performance overhead as we already have the allocation_id available on the move when serializing. I'll reinstate this relationship.

spec/serializers/move_serializer_spec.rb
```diff
@@ -152,63 +152,12 @@
...
-
-      it 'contains an included allocation' do
-        expect(result[:included]).to eq(expected_json)
-      end
-    end
-
-    context 'without an allocation' do
-      let(:adapter_options) { { include: MoveSerializer::SUPPORTED_RELATIONSHIPS } }
-
-      it 'contains an empty allocation' do
```
cwrw, line 205: can we keep this relationship test for no allocation?

cwrw (approved):
looks good 👍 , just a minor comment

spec/serializers/move_serializer_spec.rb
smoothcontract: I don't think it's very valuable without the included allocation; it's just testing the serializer library behaviour so I cleaned this up.

PR #1170, opened 2020-11-26
Removes unused code

willfish:
### Jira link

P4-<TODO>

### What?

When we migrated all of our `Event` rows to `GenericEvent` ones we needed a way to map the attributes from 1 event type to the other. We no longer need this behaviour since we don't create/use V1 Events anymore.

I have added/removed/altered:

- [x] Removes unused `GenericEvent.from_event` methods and references to them

### Why?

I am doing this because:

- These methods are no longer used


smoothcontract (approved):
Hell yes! 💯

PR #1171, opened 2020-11-27
Removes now-unused event model

willfish:
### Jira link

P4-<TODO>

### What?

We've migrated from the Event to the GenericEvent model to capture events that happen around the move service. Now that we're fully migrated we no longer need to keep two separate models.

I have added/removed/altered:

- [x] Removes unused event model and references to it

### Why?

I am doing this because:
- It's unused and is just adding cognitive overhead

- I am not going to remove the table. That can and should be done separately.


app/controllers/api/allocation_events_controller.rb
```diff
@@ -8,8 +8,8 @@ def cancel
       validate_params!(cancel_params)
 
       allocation.transaction do
-        allocation.cancel(**cancellation_details)
-        process_event(allocation.moves, Event::CANCEL, cancel_move_params)
+        allocation.cancel(cancellation_details)
```
smoothcontract, line 11: Not sure about this change - I think it was there to resolve a Ruby deprecation warning. See https://blog.saeloun.com/2019/10/07/ruby-2-7-keyword-arguments-redesign.html#ruby-27

app/models/allocation.rb
```diff
@@ -48,7 +48,6 @@ class Allocation < VersionedModel
   belongs_to :from_location, class_name: 'Location'
   belongs_to :to_location, class_name: 'Location'
   has_many :moves, inverse_of: :allocation, dependent: :destroy, autosave: true
-  has_many :events, as: :eventable, dependent: :destroy
```
smoothcontract, line 51: Should this be replaced with generic_events rather than being removed entirely? It looks like we only create allocation events on the underlying moves currently when the allocation is cancelled, but adding move level events might be something we want to do in future. I guess it's fine to remove this until that requirement surfaces 🤔

app/models/event.rb
```diff
@@ -1,58 +0,0 @@
-class Event < ApplicationRecord
```
smoothcontract, line 1: I agree that it's not wise to remove the underlying table now, but would be good to capture this action in a Jira ticket, Github issue or a code TODO somehow so it doesn't get forgotten forever.

app/models/journey.rb
```diff
@@ -36,7 +36,6 @@ class Journey < ApplicationRecord
   belongs_to :from_location, class_name: 'Location'
   belongs_to :to_location, class_name: 'Location'
 
-  has_many :events, as: :eventable, dependent: :destroy # NB: polymorphic association
   has_many :generic_events, as: :eventable, dependent: :destroy # NB: polymorphic association
```
smoothcontract, line 39: I think the comment here can be dropped now - it's not adding any value TBH

app/models/move.rb
```diff
@@ -85,7 +85,6 @@ class Move < VersionedModel
   has_many :journeys, -> { default_order }, dependent: :restrict_with_exception, inverse_of: :move
   has_many :court_hearings, dependent: :restrict_with_exception
 
-  has_many :events, as: :eventable, dependent: :destroy # NB: polymorphic association
   has_many :generic_events, as: :eventable, dependent: :destroy # NB: polymorphic association
```
smoothcontract, line 88: Drop the comment here too pls

smoothcontract (approved):
Nice cleanup, good to see cruft being removed. Some very minor comments, the main point being to document the need to remove the old events table in future 👍

app/controllers/api/allocation_events_controller.rb
willfish: oh interesting. okay, will remove

app/models/journey.rb
willfish: Not sure it ever added value, tbh :)

app/models/event.rb
willfish: Re the TODO I was thinking I'd rename the GenericEvent model -> Event and inidicate in the table name. Since we're not doing that it makes sense to capture it in a ticket I think since there isn't a logical place to say it

app/models/allocation.rb
willfish: We're not referencing the relationship. Am happy to add it in if we want it to flag this up in the code

willfish: Hang on, this is actually wrong. Moves have events not Allocations

willfish: Allocations cancel calls iterate through the Allocation#moves and generate MoveCancel events for them. So this was actually always wrong

smoothcontract: Agreed 👍

PR #1173, opened 2020-11-27
Reduce spam logging of Papertrail versions

smoothcontract:
### Jira link

P4-2516

### What?

- [x] Stop versioning `FrameworkResponse` model
- [x] Don't version changes to `updated_at` column

### Why?

- There's no need to version FrameworkResponses - we don't care about those while the PER is in draft status and once completed the PER becomes read only so those rows are essentially immutable.
For other versioned resources there's no need to record a new version if only a timestamp is modified; this can happen when the `updated_at` attribute is updated in isolation, or if the model is touched.
This change now means that updating resources won't trigger a cascade of associated versions via the "touch model". In production this should speed up endpoints that modify data as we won't need as many writes to the 7 million row versions table.

- Minimal risk, but should improve performance

### DIFF
--- a/app/models/framework_response.rb
+++ b/app/models/framework_response.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class FrameworkResponse < VersionedModel
+class FrameworkResponse < ApplicationRecord
   ValueTypeError = Class.new(StandardError)
 
   validates :type, presence: true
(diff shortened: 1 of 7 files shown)


willfish (approved):
I was literally about to push this change :grin:

willfish (approved):

cwrw (commented):
I'm not seeing the framework response change?, otherwise the rest looks ok

smoothcontract:
> I'm not seeing the framework response change?, otherwise the rest looks ok

Good catch.. I made this change locally but somehow omitted it from the PR.. probably too much flipping between branches 🤦 Added in now.

cwrw (approved):

PR #1176, opened 2020-11-30
P4-2520 Add editable field to Person Escort Record

cwrw:
jira-ticket: https://dsdmoj.atlassian.net/browse/P4-2520

Surface value to indicate if an assessment's responses are editable or not. An assessment is editable if it has not been confirmed, and the move status is either booked or requested.

For older assessments that do not have moves associated, support editable on confirmed/unconfirmed only.


app/models/person_escort_record.rb
```diff
@@ -7,8 +7,20 @@ def self.framework_name
...
+
+    move_status_editable?
+  end
+
 private
 
+  def move_status_editable?
+    return true if move.blank?
+
+    move.requested? || move.booked?
```
cwrw, line 21: This will cause one extra small query to the move on the PER serializer. If a PER is requested through a move endpoint no extra query will need to done as it will look up the cached parent.

app/models/person_escort_record.rb
```diff
@@ -7,8 +7,20 @@ def self.framework_name
...
+  def editable
+    return false if confirmed?
+
+    move_status_editable?
+  end
+
 private
 
+  def move_status_editable?
+    return true if move.blank?
```
cwrw, line 19: this is to support old PERs which don't have a move attached

smoothcontract (approved):
This looks good - it's nice that we have a separate lightweight PER serializer when returning a list of moves, so this new attribute doesn't add any overhead there. Implementation seems very straightforward too 👍

smoothcontract (approved):
Thanks for changing to predicate methods in framework model 👍

PR #1177, opened 2020-11-30
P4 2459 fake gps data generator

martyn-w:
### Jira link

P4-2459

### What?

I have added/removed/altered:

- [x] Added rake task to generate a fake GPS track for a journey: `rake 'fake_data:create_gps_track[XXX]'`
- [ ] TODO: Write tests of fake gps data generator if necessary

### Why?

I am doing this because:

- so we can have some semi-realistic data to test GPS tracks

- minimal


lib/tasks/fake_data/gps.rb
```diff
@@ -0,0 +1,118 @@
+require_relative 'city.rb'
```
willfish, line 1:
```suggestion
require_relative 'city.'
```

willfish (approved):

mafonso (approved):
LGTM

PR #1180, opened 2020-12-01
P4-2450 Add Youth Risk assessment table and model

cwrw:
### Jira link

[P4-2450](https://dsdmoj.atlassian.net/browse/P4-2450)

### What?

- Added new Youth Risk Assessment table and model
- Moved out logic for optional move to PER model with tests
- Add Youth Risk Assessment associations to other models, similar to the PER.

### Why?

To support a new framework assessment, add new table and model  for the Youth Risk assessment, which extends the assessmentable concern. The youth risk assessment will always have an association to a move, so separate those concerns from the PER, and move out of shared tests.


app/models/person.rb
```diff
@@ -72,6 +73,10 @@ def latest_person_escort_record
     person_escort_records.where(status: 'confirmed').order(confirmed_at: :desc).first
   end
 
+  def latest_youth_risk_assessment
+    youth_risk_assessments.where(status: 'confirmed').order(confirmed_at: :desc).first
```
cwrw, line 77: attempting to dry this up looks more complicated, so I chose to keep as is

spec/models/person_spec.rb
```diff
@@ -128,4 +129,34 @@
       expect(person.latest_person_escort_record).to eq(newest_person_escort_record)
     end
   end
+
+  describe '#latest_youth_risk_assessment' do
```
cwrw, line 133: move to shared example to reuse for both PER and youth assessments

Nimphal (approved):
Nice work!

spec/factories/framework_assessment.rb
```diff
@@ -1,20 +1,17 @@
...
 FactoryBot.define do
-  factory :person_escort_record do
-    association(:framework)
-    association(:profile)
-
+  factory :framework_assessmentable do
     after(:build, &:initialize_state)
 
     trait :with_responses do
       association(:framework, :with_questions)
```
cwrw, line 8: add name to framework as this is shared

app/models/person_escort_record.rb
```diff
@@ -2,6 +2,8 @@ class PersonEscortRecord < VersionedModel
   include FrameworkAssessmentable
 
   belongs_to :prefill_source, class_name: 'PersonEscortRecord', optional: true
+  # To support legacy PERs without a move, allow the association to be optional
+  belongs_to :move, optional: true
```
smoothcontract, line 6: Is there a plan to clean up legacy data at some point... maybe with a rake task that finds a move with a date around the same per created_at date perhaps? Would be nice to not have to think about two different possible relationships between PER, profile and move to reduce future cognitive load.

app/models/profile.rb
```diff
@@ -32,6 +32,7 @@ class Profile < VersionedModel
 
   has_many :moves, dependent: :nullify
   has_one :person_escort_record
+  has_one :youth_risk_assessment
```
smoothcontract, line 35: This relationship confuses me now.. I was under the impression that the relationship between PER and Profile was purely historic, and that PER's should always now belong to move. Also that PER's don't get reused across moves (which is why we have prefill functionality)... so this relationship seems to be inconsistent as we can have a profile with many moves.. 🤔 Am I missing something in my understanding?

smoothcontract: I think it would be cleaner for a youth risk assessment to belong solely to a move, and have no direct relationship with a profile - we can always obtain the profile via the move if needed. Perhaps we can discuss over a call...

app/models/person.rb
```diff
@@ -22,6 +22,7 @@ class Person < VersionedModel
   has_many :profiles, dependent: :destroy
   has_many :moves, through: :profiles
   has_many :person_escort_records, through: :profiles
+  has_many :youth_risk_assessments, through: :profiles
```
smoothcontract, line 25: Through moves instead?

smoothcontract (commented):
I'm pretty confused by the relationships between assessment, profile, person and moves - and have a niggling doubt that something isn't quite consistent there. Can we discuss on a call please?

app/models/person.rb
smoothcontract: I think you can use the auto generated scopes from the enum here, so this should work:

```ruby
  youth_risk_assessments.confirmed.order(confirmed_at: :desc).first
```

app/models/profile.rb
smoothcontract: Discussed over slack and agreed that this is fine, if not immediately obvious. An assessment belongs to a move and a profile. If the profile is then used for a subsequent move, the assessment can be accessed via move.profile.assessment, but the assessment will still refer to the original move. Agreed to document this behaviour in the Github wiki as it may not be obvious to clients of the API.

app/models/person.rb
cwrw: Generated scopes don't work on a collection, but rather single records, so couldn't utilise it like this unfortunately. Im using the constant now however instead of the string

app/models/person_escort_record.rb
cwrw: That's definitely a good idea. These issues weren't prominent previously, but I think now we might need to do it, to avoid all those special cases and complication in supporting the legacy PERs. I'll add tickets for this to be tackled soon 👍

app/models/person.rb
cwrw: discussed in person: this should remain the same

app/models/profile.rb
cwrw: updated https://github.com/ministryofjustice/hmpps-book-secure-move-api/wiki/Person-Escort-Record#reuse-the-per-for-a-different-move if you want to have a look

smoothcontract: Thanks for updating the wiki, it's clear now 👍

spec/support/a_framework_assessment.rb
```diff
@@ -42,65 +41,65 @@
     end
 
     it 'returns error if move is not associated to a profile' do
-      create(:framework, version: '1.2.1')
+      create(:framework, name: assessment_type.to_s.dasherize, version: '1.2.1')
```
smoothcontract, line 44: I think it might read more nicely and be less repetitive if this were assigned to a `let` instead? Something like

```suggestion
      let(:framework_name) { assessment_type.to_s.dasherize }
      
      ...
      
      create(:framework, name: framework_name, version: '1.2.1')
```

smoothcontract (approved):
Looks good, just one minor suggestion to improve test readability 👍

spec/support/a_framework_assessment.rb
cwrw: I would prefer keeping it defined in the context of each spec, mainly due to multiple variations of the framework being defined across each specs

PR #1182, opened 2020-12-02
P4-2415 fix duplicate move validation

martyn-w:
### Jira link

P4-2415

### What?

I have added/removed/altered:

- [x] Ensure duplicate moves are identified by matching people rather than by matching profiles
- [x] Exempt cancelled, proposed and nil profiles from duplicate checks

### Why?

I am doing this because:

- to ensure duplicate moves are correctly identified and reported to the frontend
- reduce the number of errors raised which are actually about duplicate moves

### Have you? (optional)

- [ ] TODO: run front end end-to-end tests

### DIFF
--- a/app/models/move.rb
+++ b/app/models/move.rb
@@ -152,12 +152,23 @@ class Move < VersionedModel
     cancellation_reason == CANCELLATION_REASON_REJECTED
   end
 
-  def existing
-    self.class.not_cancelled.find_by(date: date, profile_id: profile_id, from_location_id: from_location_id, to_location_id: to_location_id)
+  def existing_moves
+    Move
+        .joins(:profile)
+        .where('profiles.person_id = ?', profile.person_id)
+        .not_cancelled
+        .not_proposed
+        .where(
+          from_location_id: from_location_id,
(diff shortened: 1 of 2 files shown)


teneightfive (approved):
👍

app/models/move.rb
```diff
@@ -152,12 +152,23 @@ def rejected?
     cancellation_reason == CANCELLATION_REASON_REJECTED
   end
 
-  def existing
-    self.class.not_cancelled.find_by(date: date, profile_id: profile_id, from_location_id: from_location_id, to_location_id: to_location_id)
+  def existing_moves
+    Move
```
cwrw, line 156: maybe this should be memoized to avoid doing duplicate queries?

cwrw (approved):

app/models/move.rb
martyn-w: @cwrw good question; I was in two minds about this: on the one hand, memoising is more efficient. On the other hand, the memoised method won't catch a newly created duplicate that were added during the lifetime of the move instance.

I've now memoised - lets see how that works - or if unexpected duplicates slip through

PR #1183, opened 2020-12-02
Add support for medical and critical events

smoothcontract:
### Jira link

P4-2425

### What?

- [x] Added new `classification` attribute to `GenericEvent` model (populated automatically when creating events)
- [x] Added rake task to back fill `classification` for all existing events
- [x] Added support for new `incident_events` includes on `GET /moves` and `GET /moves/:id` endpoints

### Why?

- Medical events occur against a PER, and incident events (aka red flag events) occur against a move. These need to be exposed efficiently in the API to allow relevant flags to be shown in the front end when listing moves. This means minimising SQL queries and the rows retrieved to keep the endpoint performant.

### Have you? (optional)

- Extends includes for existing API, but is backwards compatible

### DIFF
--- a/app/serializers/generic_event_serializer.rb
+++ b/app/serializers/generic_event_serializer.rb
@@ -2,15 +2,14 @@
 
 class GenericEventSerializer
   include JSONAPI::Serializer
+  include JSONAPI::ConditionalRelationships
 
   set_type :events
 
-  attributes :event_type, :occurred_at, :recorded_at, :notes
+  attributes :event_type, :classification, :occurred_at, :recorded_at, :notes
 
-  # TODO: should this/these be a belongs_to instead?
-  # TODO consider using has_one_if_included to lazy load these associations
-  has_one :eventable, serializer: ->(record, _params) { SerializerVersionChooser.call(record.class) }
-  has_one :supplier
+  belongs_to :eventable, serializer: ->(record, _params) { SerializerVersionChooser.call(record.class) }
+  belongs_to :supplier
 
   SUPPORTED_RELATIONSHIPS = %w[eventable].freeze
 
--- a/app/serializers/generic_event_serializer.rb
+++ b/app/serializers/generic_event_serializer.rb
@@ -5,16 +5,14 @@ class GenericEventSerializer
 
   set_type :events
 
-  attributes :occurred_at, :recorded_at, :notes
+  attributes :event_type, :occurred_at, :recorded_at, :notes
 
+  # TODO: should this/these be a belongs_to instead?
+  # TODO consider using has_one_if_included to lazy load these associations
   has_one :eventable, serializer: ->(record, _params) { SerializerVersionChooser.call(record.class) }
   has_one :supplier
 
   SUPPORTED_RELATIONSHIPS = %w[eventable].freeze
-
-  attribute :event_type do |object|
-    object.type.try(:gsub, 'GenericEvent::', '')
-  end
 
   attribute :details do |record, _params|
     record.details.deep_dup.tap do |details|
--- a/app/serializers/important_events_serializer.rb
+++ b/app/serializers/important_events_serializer.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class ImportantEventsSerializer
+  include JSONAPI::Serializer
+
+  set_type :events
+
+  attributes :event_type, :classification
+end
--- a/app/serializers/v2/move_serializer.rb
+++ b/app/serializers/v2/move_serializer.rb
@@ -3,6 +3,7 @@
 module V2
   class MoveSerializer
     include JSONAPI::Serializer
+    include JSONAPI::ConditionalRelationships
 
     set_type :moves
 
@@ -30,7 +31,8 @@ module V2
 
     has_many :court_hearings, serializer: CourtHearingSerializer
 
-    has_many :timeline_events, serializer: ->(record, _params) { record.class.serializer }, &:all_events_for_timeline
+    has_many_if_included :timeline_events, serializer: ->(record, _params) { record.class.serializer }, &:all_events_for_timeline
+    has_many_if_included :important_events, serializer: ImportantEventsSerializer, &:important_events
 
     belongs_to :allocation, serializer: AllocationSerializer
     belongs_to :original_move, serializer: V2::MoveSerializer
@@ -64,6 +66,7 @@ module V2
       allocation
       original_move
       supplier
+      important_events
       timeline_events
       timeline_events.eventable
     ].freeze
--- a/app/serializers/v2/moves_serializer.rb
+++ b/app/serializers/v2/moves_serializer.rb
@@ -35,12 +35,11 @@ module V2
       profile.person.gender
       profile.person_escort_record
       profile.person_escort_record.flags
-      profile.person_escort_record.medical_events
       from_location
       to_location
       prison_transfer_reason
       supplier
-      critical_events
+      important_events
     ].freeze
 
     belongs_to :from_location, serializer: ::LocationSerializer
@@ -50,6 +49,6 @@ module V2
     belongs_to :supplier, serializer: SupplierSerializer
     belongs_to :allocation, serializer: AllocationSerializer
 
-    has_many_if_included :critical_events, serializer: ImportantEventsSerializer
+    has_many_if_included :important_events, serializer: ImportantEventsSerializer, &:important_events
   end
 end
--- a/app/serializers/v2/moves_serializer.rb
+++ b/app/serializers/v2/moves_serializer.rb
@@ -3,6 +3,7 @@
 module V2
   class MovesSerializer
     include JSONAPI::Serializer
+    include JSONAPI::ConditionalRelationships
 
     attributes  :additional_information,
                 :cancellation_reason,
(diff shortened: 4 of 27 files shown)


app/serializers/generic_event_serializer.rb
```diff
@@ -5,17 +5,15 @@ class GenericEventSerializer
...
 
+  # TODO: should this/these be a belongs_to instead?
+  # TODO consider using has_one_if_included to lazy load these associations
   has_one :eventable, serializer: ->(record, _params) { SerializerVersionChooser.call(record.class) }
   has_one :supplier
 
   SUPPORTED_RELATIONSHIPS = %w[eventable].freeze
 
-  attribute :event_type do |object|
-    object.type.try(:gsub, 'GenericEvent::', '')
```
smoothcontract, line 16: Moved this logic to the model to allow reuse

app/serializers/important_events_serializer.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class ImportantEventsSerializer
+  include JSONAPI::Serializer
+
+  set_type :events
+
+  attributes :event_type, :classification
```
smoothcontract, line 8: @solidgoldpig Any critical or medical events will be serialized with just these minimal attributes; enough to build a list of flags in the UI hopefully.

app/serializers/generic_event_serializer.rb
```diff
@@ -5,17 +5,15 @@ class GenericEventSerializer
 
   set_type :events
 
-  attributes :occurred_at, :recorded_at, :notes
+  attributes :event_type, :occurred_at, :recorded_at, :notes
 
+  # TODO: should this/these be a belongs_to instead?
```
willfish, line 10: Can you just change it without all the specs exploding?

app/serializers/important_events_serializer.rb
willfish, line 8: Are we 100% sure we don't more attributes than this?

solidgoldpig: We need the event's id, that's for sure

solidgoldpig: Will the timeline events also have the classification?

lib/tasks/data_maintenance.rake
```diff
@@ -36,4 +36,19 @@ namespace :data_maintenance do
     Person.where(criminal_records_office: '').update_all(criminal_records_office: nil)
     Person.where(police_national_computer: '').update_all(police_national_computer: nil)
   end
+
+  desc 'fix generic event classification for existing data'
+  task fix_generic_event_classifications: :environment do
```
willfish, line 41:
```ruby
GenericEvent::PerMedicalAid.update_all(classification: 'medical')

critical_event_types = [
  'GenericEvent::PersonMoveRoadTrafficAccident',
  'GenericEvent::PersonMovePersonEscaped',
  'GenericEvent::PersonMoveUsedForce',
  'GenericEvent::PersonMoveMajorIncidentOther',
  'GenericEvent::PersonMoveSeriousInjury',
  'GenericEvent::PersonMoveMinorIncidentOther',
  'GenericEvent::PersonMoveDeathInCustody',
  'GenericEvent::PersonMoveAssault',
  'GenericEvent::PersonMovePersonEscapedKpi',
  'GenericEvent::PersonMoveReleasedError',
]

GenericEvent.where(type: types).update_all(classification: 'critical')
```

app/serializers/important_events_serializer.rb
willfish: That comes automatically because of JSONAPI

willfish:
Is the requirement to surface the classification for the timeline or is it to surface it for the dashboard?

For the timeline, it makes sense to me not to have a separate serialized thing that we return with the same type and colliding with includes for the timeline events (as I understand it, includes are unique by type and id and with the current implementation if Alex requests both I'm not sure which one - the one with fewer attributes/the one with more attributes wins).

willfish:
Is there some way we can leverage sparse fields and filters rather than having a completely separate serializer?

app/serializers/important_events_serializer.rb
smoothcontract: Yes, I've added classification to those too 👍

smoothcontract:
> Is there some way we can leverage sparse fields and filters rather than having a completely separate serializer?

It's possible but we've adopted the approach of custom lightweight serializers for other things; this ends up being a simpler and clearer approach although does mean we now have a lot of serializers.

smoothcontract:
> Is the requirement to surface the classification for the timeline or is it to surface it for the dashboard?
> 
> For the timeline, it makes sense to me not to have a separate serialized thing that we return with the same type and colliding with includes for the timeline events (as I understand it, includes are unique by type and id and with the current implementation if Alex requests both I'm not sure which one - the one with fewer attributes/the one with more attributes - wins).

Suport for `important_events` has been added to `GET /moves` and `GET /moves/:id`. The latter endpoint also supports `timeline_events` but won't be used in combination with `important_events`.

lib/tasks/data_maintenance.rake
smoothcontract: 👍

app/serializers/generic_event_serializer.rb
smoothcontract: Done

app/serializers/v2/move_serializer.rb
```diff
@@ -30,7 +31,8 @@ class MoveSerializer
 
     has_many :court_hearings, serializer: CourtHearingSerializer
 
-    has_many :timeline_events, serializer: ->(record, _params) { record.class.serializer }, &:all_events_for_timeline
+    has_many_if_included :timeline_events, serializer: ->(record, _params) { record.class.serializer }, &:all_events_for_timeline
```
martyn-w, line 34: good shout 👍 - I didn't know about `has_many_if_included`

app/serializers/v2/moves_serializer.rb
```diff
@@ -38,6 +39,7 @@ class MovesSerializer
       to_location
       prison_transfer_reason
       supplier
+      important_events
```
martyn-w, line 42: Do you have a sense of how this will perform? on production the moves table is ~200k and the events table is ~400k rows.

Might this lead to the same joining flags for multiple moves performance issue we had a couple of weeks back?

lib/tasks/data_maintenance.rake
martyn-w: Running this task may be tricky on prod - you'll be updating 400k rows. Best to test it on pre-prod first.

martyn-w (approved):
👍 LGTM - but see comment on performance of joining moves to events

spec/serializers/important_events_serializer_spec.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe ImportantEventsSerializer do
+  subject(:serializer) { described_class.new(event) }
+
+  let(:event) { create(:event_move_approve) }
```
cwrw, line 8: maybe we should use an important event here instead (either incident or medical) to show that the classification will come back as either of those?

spec/serializers/v2/move_serializer_spec.rb
```diff
@@ -60,15 +60,14 @@
...
-    let!(:event) { create(:event_move_cancel, eventable: move) }
 
     let(:adapter_options) { { include: described_class::SUPPORTED_RELATIONSHIPS } }
 
     before { ActiveStorage::Current.host = 'http://www.example.com' } # This is used in the serializer
 
     it 'contains all included relationships' do
       expect(result[:included].map { |r| r[:type] })
-        .to match_array(%w[people ethnicities genders locations locations profiles moves documents prison_transfer_reasons court_hearings suppliers events moves])
+        .to match_array(%w[people ethnicities genders locations locations profiles moves documents prison_transfer_reasons court_hearings suppliers])
```
cwrw, line 70: not sure why the extra event was removed?

app/serializers/v2/moves_serializer.rb
smoothcontract: It's really fast - adds two extra queries which we can't really avoid but no N+1's. We could get the move incident events with just one query, but the front end guys wanted the PER medical event to also be included which therefore means eager loading the PER to achieve that. That said the PER is also included pretty much the whole time anyway, so shouldn't make things worse.

Flags were bad because of the indirect N:N relationship to PER via framework responses, of which there are many. This is a much simpler relationship dealing with far less data (very few moves or PER's will have these incident or medical flags) so the relevant queries are fast.

spec/serializers/important_events_serializer_spec.rb
smoothcontract: I don't think it makes much difference to be honest; the classification behaviour already has coverage at unit level. But happy to change either way.

app/serializers/v2/moves_serializer.rb
martyn-w: 👍 jolly good

spec/serializers/v2/move_serializer_spec.rb
smoothcontract: The `timeline_events` stuff which returns `events` is already covered extensively a little lower down in this same file so no sense testing it again here.

cwrw (approved):
tested together, looks great 👍

spec/serializers/important_events_serializer_spec.rb
smoothcontract: Updated

PR #1184, opened 2020-12-03
P4-2451 Add Youth Risk Assessment routes and controllers

cwrw:
### Jira link

[P4-2451](https://dsdmoj.atlassian.net/browse/P4-2451)

### What?

- Added new routes for Youth risk assessment
- Added new controller for Youth risk assessment
- Updated assessment serializers to inherit from parent assessment serializer class
- Added new Youth risk assessment serializer
- Added includes of Youth risk assessment to moves and profiles

### Why?

Surface new routes for Youth risk assessment, by adding new controller which inherits from framework assessments. The serializers will inherit from a parent framework assessment serializer, only for single assessments, as there is no requirement to expose youth risk assessment on the get multiple moves endpoint.

Prefilling unfortunately needs to follow a similar pattern as we cannot pass in a block for the type of serializer. 
Notifications for youth risk assessments will be picked up in a different PR, so tests have been marked for now as pending.

### Have you? (optional)

### DIFF
--- a/app/serializers/framework_assessment_serializer.rb
+++ b/app/serializers/framework_assessment_serializer.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+class FrameworkAssessmentSerializer
+  include JSONAPI::Serializer
+
+  belongs_to :profile, serializer: V2::ProfileSerializer
+  belongs_to :move, serializer: V2::MoveSerializer
+  belongs_to :framework
+
+  has_many :responses, serializer: FrameworkResponseSerializer do |object|
+    object.framework_responses.includes(:framework_flags, :framework_nomis_mappings, framework_question: [:framework, dependents: :dependents])
+  end
+  has_many :flags, serializer: FrameworkFlagSerializer do |object|
+    object.framework_flags.includes(framework_question: :dependents)
+  end
+
+  attributes :confirmed_at, :created_at, :nomis_sync_status
+
+  attribute :version do |object|
+    object.framework.version
+  end
+
+  attribute :status do |object|
+    object.status == 'unstarted' ? 'not_started' : object.status
+  end
+
+  attribute :editable, &:editable?
+
+  meta do |object|
+    { section_progress: object.section_progress }
+  end
+
+  SUPPORTED_RELATIONSHIPS = %w[
+    move
+    framework
+    profile.person
+    responses.question
+    responses.nomis_mappings
+    responses.question.descendants.**
+    flags
+    prefill_source
+  ].freeze
+end
--- a/app/serializers/framework_assessment_serializer.rb
+++ b/app/serializers/framework_assessment_serializer.rb
@@ -3,8 +3,6 @@
 class FrameworkAssessmentSerializer
   include JSONAPI::Serializer
 
-  belongs_to :profile, serializer: V2::ProfileSerializer
-  belongs_to :move, serializer: V2::MoveSerializer
   belongs_to :framework
 
   has_many :responses, serializer: FrameworkResponseSerializer do |object|
--- a/app/serializers/framework_assessment_serializer.rb
+++ b/app/serializers/framework_assessment_serializer.rb
@@ -18,6 +18,8 @@ class FrameworkAssessmentSerializer
     object.framework.version
   end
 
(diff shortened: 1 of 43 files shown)


spec/requests/api/framework_responses_controller_bulk_update_spec.rb
```diff
@@ -296,4 +296,77 @@
       end
     end
   end
+
+  describe 'PATCH /youth_risk_assessments/:youth_risk_assessment_id/framework_responses' do
```
cwrw, line 300: added a few tests (not all) to show that this endpoint is working successfully

spec/requests/api/youth_risk_assessments_controller_update_spec.rb
```diff
@@ -0,0 +1,154 @@
...
+            govuk_notify_response:
+            instance_double(Notifications::Client::ResponseNotification, id: SecureRandom.uuid),
+          ),
+        )
+      end
+
+      before do
+        allow(Faraday).to receive(:new).and_return(faraday_client)
+        allow(MoveMailer).to receive(:notify).and_return(notify_response)
+        perform_enqueued_jobs(only: [PreparePersonEscortRecordNotificationsJob, NotifyWebhookJob, NotifyEmailJob]) do
```
cwrw, line 120: this wil be updated in the next PR, for now mark as pending

app/serializers/youth_risk_assessment_serializer.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class YouthRiskAssessmentSerializer < FrameworkAssessmentSerializer
+  belongs_to :profile, serializer: V2::ProfileSerializer
```
cwrw, line 4: this was required due to autoloading the mirror relationship in v2 serializers, which wasn't working as expected. When we move off v2 this can go back to the parent class

martyn-w:
Very minor comment - could you remove the `/v1/` from the api paths in the RSpecs - we want to turn off v1 as soon as we can and definitely don't want new endpoints to use it.

app/serializers/framework_assessment_serializer.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  attributes :confirmed_at, :created_at, :nomis_sync_status
+
+  attribute :version do |object|
+    object.framework.version
+  end
+
+  attribute :status do |object|
+    object.status == 'unstarted' ? 'not_started' : object.status
```
martyn-w, line 22: I'm sure there is a story behind this mapping... ...but is it possible to rename `unstarted` to `not_started` instead?

martyn-w (approved):
👍 looks good to me, but would be good to get a review from another dev

app/serializers/framework_assessment_serializer.rb
cwrw: Unfortunately due to enum values throwing a warning if their name is prepended with a "not_", we kept the enum name as "unstarted" however it doesn't read well being surfaced, so we return "not_started" instead. I've added a comment to avoid confusion.

martyn-w: 👍

spec/serializers/framework_assessment_serializer_spec.rb
```diff
@@ -0,0 +1,127 @@
...
+      id: response.id,
+      type: 'framework_responses',
+    )
+  end
+
+  it 'contains an empty `flags` relationship if no flags present' do
+    expect(result[:data][:relationships][:flags][:data]).to be_empty
+  end
+
+  it 'contains a`flags` relationship with framework response flags' do
```
smoothcontract, line 65: Missing a space before `flags` 🤣

smoothcontract (approved):
Awesome stuff - nice refactor to introduce inheritance for the shared behaviour across both assessment types. Can't spot any issues with it ⭐

app/serializers/youth_risk_assessment_serializer.rb
cwrw: I'll a comment so to explain and a so we don't forget to do when we move off 👍

smoothcontract (approved):
LGTM 👍

PR #1185, opened 2020-12-03
Tidy event specs

willfish:
### Jira link

P4-<TODO>

### What?

Tidies up the event creation success case specs

### Why?

Because this is friendly to future developers that work on events.


spec/models/generic_event/move_notify_premises_of_arrival_in30_mins_spec.rb
```diff
@@ -1,5 +1,5 @@
 RSpec.describe GenericEvent::MoveNotifyPremisesOfArrivalIn30Mins do
-  subject(:generic_event) { build(:event_move_notify_premises_of_arrival_in_30_mins) }
+  subject(:generic_event) { build(:event_move_notify_premises_of_arrival_in30_mins) }
```
willfish, line 2: This is what you get when you do:

```ruby
'MoveNotifyPremisesOfArriveIn30Mins'.underscore
```

Seemed simpler and more beneficial not to mess with my underscoring and just renaming the factory.

cwrw (approved):

PR #1187, opened 2020-12-03
Remove MPS28 (West End Custody Suite) from Serco

martyn-w:
### Jira link

P4-<TODO>

### What?

I have added/removed/altered:

- [x] Removed MPS28 West End Custody Suite from Serco

### Why?

I am doing this because:

- the programme have asked for it to be removed


cwrw (approved):

lib/tasks/data/supplier_locations.yml
```diff
@@ -319,7 +319,6 @@ suppliers:
     - MPS25
```
cwrw, line 319: maybe this is the right time to deprecate this file since its no longer in use

martyn-w: yep, good call 👍

PR #1188, opened 2020-12-03
P4-2521 Change notifications for assessments to be generic

cwrw:
jira-ticket: https://dsdmoj.atlassian.net/browse/P4-2521

To support  multiple assessment webhooks and notifications, change the notifier to accept any assessment with its class name, and send a confirmation webhook according to the name of the assessment confirmed


spec/models/profile_spec.rb
```diff
@@ -3,7 +3,7 @@
 require 'rails_helper'
 
 RSpec.describe Profile, type: :model do
-  xit { is_expected.to belong_to(:category).optional }
+  it { is_expected.to belong_to(:category).optional }
```
cwrw, line 6: just something we forgot, not related to pr

martyn-w:
I think we might need to do this as two separate deployments, to avoid problems with pods/sidekiq calling the old job vs calling the new job. I'd suggest:

1. add the new job (PrepareAssessmentNotificationsJob) and update code to call it
2. deploy
3. remove old job (PreparePersonEscortRecordNotificationsJob)
4. deploy

martyn-w:
👍 looks good - but delete `PreparePersonEscortRecordNotificationsJob` in a separate deployment

smoothcontract (approved):
This all looks fine, only one thing to keep an eye on when deleting delayed jobs; it's theoretically possible to have a job queued in sidekiq that uses this class, so when that gets inflated from redis to run it won't work as the class no longer exists.

To be really safe it's perhaps better practice to leave the original class in the code for a while with a TODO marking it for future deletion. Alternatively we need to be sure the sidekiq queue's are empty when deploying this change, and that nobody completes a PER while this is being deployed to all of the pods.

cwrw:
Thanks both for the review, and the excellent spot regarding deploys. Im closing this in favour of the following PRs to be staggered into production:
1) Add new notifier: https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/1193
2) Switch to using new notifier: https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/1194
3) Remove old notifier: https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/1195

PR #1189, opened 2020-12-03
Revert change to filtering immutable feed resources

willfish:
### Jira link

P4-<TODO>

### What?

Reverts change to filtering immutable feed resources.

### Why?

I am doing this because:

- They're not actually immutable and can change and should be up-to-date
for the hub and JPC


cwrw (commented):
looks good, can you update the corresponding rake tasks for events and notifications as well?

willfish:
Ah yeah. Good spot thanks @cwrw

cwrw (approved):

PR #1191, opened 2020-12-04
Adds rake task to backfill missing events

willfish:
### Jira link

P4-<TODO>

### What?

I have added/removed/altered:

- [ ] Adds rake task for backfilling initial state events for all moves in the moves table that don't have them already.

### Why?

I am doing this because:

- These are needed for a more complete UI and for contract deliverables with the hub

### DIFF
--- a/lib/tasks/event.rake
+++ b/lib/tasks/event.rake
@@ -2,7 +2,7 @@
 
 namespace :events do
   INITIAL_STATE_EVENTS = %w[GenericEvent::MoveProposed GenericEvent::MoveRequested].freeze
-  INITIAL_STATUSES = %[proposed requested]
+  INITIAL_STATUSES = %(proposed requested)
 
   desc 'Backfills MoveProposed and MoveRequested events since start of time'
   task add_missing_initial_state_events: :environment do
@@ -31,7 +31,7 @@ namespace :events do
           # The first time PaperTrail stores the initial version is on the first update event
           # And if there have been no updates then the initial version is the current version of the Move stored in the database moves table
           initial_move = move.versions.find_by(event: 'update')&.reify
-          initial_move = move unless initial_move.present?
+          initial_move = move if initial_move.blank?
 
           # We take the create event supplier_id to reflect the user that initially created the first move
           initial_version = move.versions.find_by(event: 'create')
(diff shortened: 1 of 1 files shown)


lib/tasks/event.rake
```diff
@@ -0,0 +1,85 @@
...
+          # The first time PaperTrail stores the initial version is on the first update event
+          # And if there have been no updates then the initial version is the current version of the Move stored in the database moves table
+          initial_move = move.versions.find_by(event: 'update')&.reify
+          initial_move = move if initial_move.blank?
+
+          # We take the create event supplier_id to reflect the user that initially created the first move
+          initial_version = move.versions.find_by(event: 'create')
+          # Due to a large versions table, we never successfully populated the supplier_id for every version.
+          # Handle this by falling back to thei whodunnit field
+          initial_supplier = initial_version.supplier_id || initial_version.whodunnit
```
martyn-w, line 40: check that `initial_version.whodunnit` is a supplier UUID, as it might be a username or other string

lib/tasks/event.rake
```diff
@@ -0,0 +1,85 @@
...
+
+          event_attributes = {
+            eventable: initial_move,
+            created_at: initial_move.created_at,
+            updated_at: initial_move.created_at,
+            occurred_at: initial_move.created_at,
+            recorded_at: initial_move.created_at,
+            notes: 'Automatically generated event',
+            details: {},
+            supplier_id: initial_supplier,
```
martyn-w, line 56: to be clear: is the `supplier_id` in the generic event table the supplier who _created_ the event OR the supplier who "owns" the event ?

If the latter, then I guess the supplier_id of the event is always the move.supplier_id ?

And if so, my previous comment about about casting whodunnit to supplier_id is null and void?

lib/tasks/event.rake
```diff
@@ -0,0 +1,85 @@
...
+      moves_with_initial_state_events: 0,
+      moves_without_initial_state_events: 0,
+      moves_with_invalid_initial_state: 0,
+      proposed_events_to_be_created: 0,
+      requested_events_to_be_created: 0,
+      events_created: 0,
+    }
+    GenericEvent.record_timestamps = false
+    GenericEvent.no_touching do
+      Move.all.find_each do |move|
```
martyn-w, line 24: is `.all` necessary here? in older versions of rails that would instantiate every move in the database into memory

PR #1192, opened 2020-12-04
Validate Youth Risk Assessment move `from_location` on creation

cwrw:
Add validation on the youth risk assessment to only being created if a move `from_location` is either a secure training centre (STC) or a secure children's home (SCH).

This requires some changes to the shared tests, where now we allow passing in a `from_location` to the respective assessment specs.
This also means we will never import NOMIS mappings on the youth risk assessment, as for NOMIS mappings to be imported, the from location should be a prison. I have moved out the specific tests to the PER now instead.

### DIFF
--- a/app/models/youth_risk_assessment.rb
+++ b/app/models/youth_risk_assessment.rb
@@ -4,6 +4,8 @@ class YouthRiskAssessment < VersionedModel
   belongs_to :prefill_source, class_name: 'YouthRiskAssessment', optional: true
   belongs_to :move
 
+  validate :move_from_location
+
   def self.framework_name
     'youth-risk-assessment'
   end
@@ -12,5 +14,9 @@ private
 
   def previous_assessment
     @previous_assessment ||= profile&.person&.latest_youth_risk_assessment
+  end
+
+  def move_from_location
+    errors.add(:move, "'from_location' must be from either a secure training centre or a secure children's home") unless move&.from_location&.secure_training_centre? || move&.from_location&.secure_childrens_home?
   end
 end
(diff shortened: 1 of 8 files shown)


martyn-w (approved):
👍 LGTM

app/models/youth_risk_assessment.rb
```diff
@@ -13,4 +15,8 @@ def self.framework_name
   def previous_assessment
     @previous_assessment ||= profile&.person&.latest_youth_risk_assessment
   end
+
+  def move_from_location
+    errors.add(:move, "'from_location' must be from either a secure training centre or a secure children's home") unless move&.from_location&.secure_training_centre? || move&.from_location&.secure_childrens_home?
```
smoothcontract, line 20: It's not possible to create a move without a from_location (it's a mandatory field) so this could possibly be simplified with fewer `&` 🤔

spec/models/youth_risk_assessment_spec.rb
```diff
@@ -3,7 +3,36 @@
...
+    end
+
+    it 'is invalid if the move from location is not from an sch or stc' do
+      location = create(:location, :prison)
+      move = create(:move, from_location: location)
+      youth_risk_assessment = build(:youth_risk_assessment, move: move)
+
+      expect(youth_risk_assessment).not_to be_valid
+      expect(youth_risk_assessment.errors.messages[:move]).to eq(["'from_location' must be from either a secure training centre or a secure children's home"])
+    end
```
smoothcontract, line 34: Could you add a test for `move` being nil please? There's some defensive coding against this in the `move_from_location` method but it doesn't look like it's all either necessary or being exercised.

smoothcontract (approved):
Looks good, just some minor comments on defensive coding around move and from_location being present (which should always be the case, so may be overkill).

app/models/youth_risk_assessment.rb
cwrw: If a move is indeed empty, all validations will run irrespective on the youth assessment. If we remove those checks we will get an error of `from_location` being called on `nil`. I've added a test as suggested below to confirm the behaviour and to test against it, making sure we get the correct validation error messages

PR #1194, opened 2020-12-04
P4-2521 Switch to using generic assessment job on assessment confirm action

cwrw:
jira-ticket: https://dsdmoj.atlassian.net/browse/P4-2451

Switch both youth risk assessment and PER to use generic assessment job notifier


martyn-w (approved):
👍

PR #1195, opened 2020-12-04
P4-2521 Remove old deprecated PER notifier

cwrw:
jira-ticket: https://dsdmoj.atlassian.net/browse/P4-2451

Since this is no longer used, remove old PER notifier class and spec


martyn-w (approved):
👍

PR #1197, opened 2020-12-04
P4 2318 metrics generator spike

martyn-w:
### Jira link

P4-2318

### What?

I have added/removed/altered:

- [x] Generates move metrics in multiple formats and stores in S3
- [x] Rake task `metrics:export` to run generator

### Why?

I am doing this because:

- so we can generate metrics for moves

### Have you? (optional)

- [ ] TODO: add cron job and S3 bucket for deployment projects
- [ ] TODO: Added environment variables to [deployment repository](https://github.com/ministryofjustice/hmpps-book-secure-move-api-deploy)

- minimal - does not touch existing code


app/models/metrics/base_metric.rb
```diff
@@ -0,0 +1,116 @@
...
+        var
+      end
+    end
+
+    def calculate_all_values
+      if values.nil? || timestamp.nil? || timestamp + interval < Time.zone.now
+        @values = {}.tap do |v|
+          columns.each do |column|
+            rows.each do |row|
+              v[value_key(column, row)] = calculate(column, row)
```
cwrw, line 90: I can envision a bit of data inconsistency here if I understood the flow correctly: we will calculate each query according to the required row/column values. Due to the time it will take for each query to compute, we might get different data due to the time elapsed between each row. Maybe there is something here around getting the data first frozen at that required time, then parsing the data in memory instead?

cwrw: Happy to discuss/confirm over a call might be easier

martyn-w: Hmmm... that is a very interesting point and something I didn't consider. Lets have a chat about it.

app/models/metrics/base_metric.rb
```diff
@@ -0,0 +1,116 @@
...
+      to_csv: 'data.csv',
+      to_fixed_key_json: 'data-fixed_key.json',
+      to_datasette_json: 'data-datasette.json',
+      to_d3_json: 'data-d3.json',
+    }.freeze
+
+    def setup_metric(metric)
+      @label = metric[:label]
+      @file = metric[:file]
+      @interval = metric[:interval]
```
cwrw, line 21: How often will the cron be triggered on production, as currently on our dev environment its set to every 30 minutes which I assume is just for testing, so I'm wondering how the timing that will affect the interval, as if its run every 30 minutes we won't need to set this?

martyn-w: the cache interval per metric is there as a feature we may want to adjust and tweak in the future, to allow for some metrics to be more cached than other metrics.

At the moment the cronjob on dev is set to run every 30 mins, but I think that may shorten (and perhaps the intervals for some metrics may lengthen) once I get sense of how the data will be used and how long it takes to run.

app/models/metrics/base_metric.rb
cwrw: I've had a look through/tested most of the queries required for the metrics and there was a recurring theme: we want to count the number of moves grouped by a column. 

For instance testing one of the metric generators for `CountsByMoveType`, we will have to run the following queries:
```
[5] pry(main)> Metrics::Moves::CountsByMoveType.new.to_datasette_json
   (127.4ms)  SELECT COUNT(*) FROM "moves" WHERE "moves"."move_type" = $1  [["move_type", "court_appearance"]]
   (108.7ms)  SELECT COUNT(*) FROM "moves" WHERE "moves"."move_type" = $1  [["move_type", "court_other"]]
   (144.3ms)  SELECT COUNT(*) FROM "moves" WHERE "moves"."move_type" = $1  [["move_type", "hospital"]]
   (113.2ms)  SELECT COUNT(*) FROM "moves" WHERE "moves"."move_type" = $1  [["move_type", "police_transfer"]]
   (109.8ms)  SELECT COUNT(*) FROM "moves" WHERE "moves"."move_type" = $1  [["move_type", "prison_recall"]]
   (108.5ms)  SELECT COUNT(*) FROM "moves" WHERE "moves"."move_type" = $1  [["move_type", "prison_remand"]]
   (124.4ms)  SELECT COUNT(*) FROM "moves" WHERE "moves"."move_type" = $1  [["move_type", "prison_transfer"]]
   (111.3ms)  SELECT COUNT(*) FROM "moves" WHERE "moves"."move_type" = $1  [["move_type", "video_remand"]]
```
One for each column of types. Instead we can run the following single query which would be much more performant and retrieves all the information we require:
```
[8] pry(main)> Move.group(:move_type).count
   (139.2ms)  SELECT COUNT(*) AS count_all, "moves"."move_type" AS moves_move_type FROM "moves" GROUP BY "moves"."move_type"
=> {"court_appearance"=>139770,
 "court_other"=>371,
 "hospital"=>25,
 "police_transfer"=>1238,
 "prison_recall"=>24731,
 "prison_remand"=>12697,
 "prison_transfer"=>26010,
 "video_remand"=>6922}
```
I suggest remodelling the metric generator to first fetch this data, and then manipulate this structure to accommodate the required data representation, rather than perform a query on each column/row type. 
The following queries can be done on the other metric generators example:
```
Move.group(:move_type).count
Move.group(:move_type, :status).count
Move.group(:status).count
Move.group(:supplier).count
Move.group(:supplier, :status).count
Move.group(:supplier, :move_type).count
```
Regarding the timeframe queries, something similar which includes the time range query and the grouping can be used example:
```
Move.where('date >= ?', '2020-12-09').group(:status).count
```
Which would cut down those queries drastically as there are quite a few there. Happy to discuss on a call. Also having a simpler query can help DRY up some of the code as well.

lib/tasks/metrics.rake
```diff
@@ -0,0 +1,30 @@
...
+    }.flatten
+
+    s3 = Aws::S3::Resource.new
+    bucket = s3.bucket(ENV['S3_METRICS_BUCKET_NAME'])
+
+    METRIC_CLASSES.each do |metric_class|
+      metric_class.new.tap do |metric|
+        puts metric.label
+        metric_class::FORMATS.each do |format_key, format_file|
+          bucket.object("#{metric.file}/#{format_file}").tap do |obj|
```
cwrw, line 18: something to keep in mind but not necessarily for this iteration: is there a need to keep multiple versions of the data or maybe as a fail safe at least one older version, as we can append a timestamp to the filename meaning we don't overwrite older files, and keep a maximum of 2 (or more depending on the need)

lib/tasks/metrics.rake
```diff
@@ -0,0 +1,30 @@
...
+  desc 'Exports metrics in all the popular formats to s3'
+  task export: :environment do
+    abort 'Please set the env-var S3_METRICS_BUCKET_NAME' if ENV['S3_METRICS_BUCKET_NAME'].blank?
+
+    # get list of all available metrics
+    METRIC_CLASSES = [Metrics::Moves].map { |namespace|
+      namespace.constants.map { |c| namespace.const_get(c) if namespace.const_get(c).is_a? Class }.compact
+    }.flatten
+
+    s3 = Aws::S3::Resource.new
```
cwrw, line 11: Instead of defining out own s3 bucket file writer, we can utilise the following class: https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/lib/cloud_data_feed.rb which is doing the same thing (except for the naming)

lib/tasks/metrics.rake
cwrw: spoke in person but for posterity: S3 supports file versioning where we have the ability to turn that on if need be, so we don't need to hand roll our own implementation

lib/tasks/metrics.rake
cwrw: spoke in person: the references class doesn't support what we need here as the file structure is different and this is used heavily in feeds. It might be safer to just create our own class with tests to simplify the logic in the rake task

spec/models/metrics/moves/counts_by_time_bin_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      create(:move, date: Date.yesterday)
+      create(:move, date: Date.today)
+      create(:move, date: Date.tomorrow)
+      create(:move, date: 4.days.from_now)
+      create(:move, date: 8.days.from_now)
+    end
+
+    it 'computes the metric' do
+      expect(metric.calculate('total', yesterday)).to be(1)
+      expect(metric.calculate('total', next_7_days)).to be(3)
```
cwrw, line 31: shouldn't this be 4 🤔

app/models/metrics/moves/counts_by_move_type_status.rb
```diff
@@ -0,0 +1,34 @@
...
+          values: Move.statuses.values,
+        },
+      }.freeze
+
+      def initialize
+        setup_metric(METRIC)
+      end
+
+      def calculate_row(row_status)
+        Move
```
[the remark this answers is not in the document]
cwrw: Update: realised there isn't an implementation for this type of metric, so we can live with the few extra queries

cwrw (approved):
Looks good and tested locally, just a few queries around the time bins otherwise 👍

app/models/metrics/time_bins.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    COMMON_TIME_BINS = [
+      TimeBin.new('the past inc today', nil, 0),
+      TimeBin.new('past 30 days inc today', -29, 0),
+      TimeBin.new('past 7 days inc today', -6, 0),
+      TimeBin.new('yesterday', -1, -1),
+      TimeBin.new('today', 0, 0),
+      TimeBin.new('tomorrow', 1, 1),
+      TimeBin.new('next 7 days inc today', 0, 6),
+      TimeBin.new('next 30 days inc today', 0, 30),
```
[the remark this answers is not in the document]
martyn-w: oops, good spot. Actually, I think I am going to change these so the future timebins exclude today.

spec/models/metrics/moves/counts_by_time_bin_spec.rb
martyn-w: nope - it should be 3, i.e.
* today + 
* tomorrow + 
* 4.days.from_now

it should _not_ include the move 8.days.from_now.

However, I think I will change the definition of next 7 days to exclude today, reducing this from 3 to 2.

app/models/metrics/moves/counts_by_move_type_status.rb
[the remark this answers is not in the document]
martyn-w: yes, saving that for a particularly rainy day.

cwrw (approved):

PR #1202, opened 2020-12-08
Temporary solution to eager load youth risk assessment in a move

cwrw:
The assessments eager loading is a can of worms that needs to be addressed properly. I will add a ticket in the backlog but for now this will stave off many many bad queries on a move.

I paired with @smoothcontract around removing this include entirely and fixing the serializers, but it requires quite a bit of work so we will go with this solution for now.


smoothcontract (approved):
Looks good for now until we can develop an approach to dynamically include things that are needed. 👍

PR #1204, opened 2020-12-08
Surface the status on the PER serializer

cwrw:
Add the status to the PER seriazlier on the moves endpoint, as this is needed to show when the PER is completed to show flags under a person's name.


martyn-w (approved):
👍

PR #1206, opened 2020-12-09
Add fake data rake tasks to create move incidents and PER medical events

smoothcontract:
### Jira link

P4-2425

### What?

- [x] Add new `fake_data:create_incident_events` and `fake_data:create_medical_events` tasks

### Why?

- Suppliers will eventually create these events via the API, but in the meantime we need some sample data for testing. As this will run on production mode environments the factory fixtures unfortunately can't be used as none of the required test group gems are bundled.


martyn-w (approved):
👍 LGTM

PR #1208, opened 2020-12-10
Add framework responses back to the version model

cwrw:
We would like to track who answers/edits responses which helps us calculate the people responsible for completing sections in the PER for auditing purposes. Add back the version model to responses for now to have this information available and tracked, so we can parse it later.

### DIFF
--- a/app/models/framework_response.rb
+++ b/app/models/framework_response.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class FrameworkResponse < ApplicationRecord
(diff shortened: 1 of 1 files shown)


Nimphal (approved):

martyn-w:
Is this really a requirement now ? we removed it a couple of weeks ago because it wasn't required and was filling up the main database

Nimphal:
It turns out it is a requirement indeed. We might end up not using the data from this database, but it's the quickest way to start recording it right now.

PR #1210, opened 2020-12-11
P4-2378 Add `requires_youth_risk_assessment` attribute to profiles

cwrw:
jira-ticket: https://dsdmoj.atlassian.net/browse/P4-2378

If a person is moving from a YOI, the client is required to ask the user if the person needs a youth risk assessment if their age is between 18 and 19. The answer can be stored on the profile as a new attribute which is a boolean representing if the user selects true or false. The attribute remains nil if the question is not answered.

The logic around when to ask this question will be with the client, which will be looking at the move from location, and at the age of the person, being between 18 - 19.

### DIFF
--- a/app/controllers/api/profiles_controller.rb
+++ b/app/controllers/api/profiles_controller.rb
@@ -26,6 +26,7 @@ module Api
     PROFILE_ATTRIBUTES = [
       :type,
       attributes: [
+        :requires_youth_risk_assessment,
         assessment_answers: [
           %i[
             key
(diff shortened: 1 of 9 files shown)


smoothcontract (approved):
Looks good. I struggled initially to understand why this adds the new attribute to v1 rather than just v2 of the API but it seems that the `ProfileController` doesn't have any version specific logic included, nor do we have v2 specific swagger documentation for profiles.

With that context I think it's okay to add a new attribute to v1 & v2 profiles; although it's new it's also optional so shouldn't cause any supplier issues if the new attribute is not specified in a request and we would hope that introducing a new attribute in the response payload will not break their integration with the API🤞 .. it might be a good idea to inform them of this change before deploying though.

Also could you tag this as a breaking change pls?

martyn-w (approved):
👍

PR #1211, opened 2020-12-11
Make sure tests use timezones in time specific tests

cwrw:
Allow tests to utilise timezones, as running tests from a different timezone is causing issues if these aren't set.

### DIFF
--- a/spec/models/activity_spec.rb
+++ b/spec/models/activity_spec.rb
@@ -32,7 +32,7 @@ RSpec.describe Activity do
     let(:expected_attributes) do
       {
         id: nomis_scheduled_event['eventId'],
-        start_time: Time.parse(nomis_scheduled_event['startTime']),
+        start_time: Time.parse(nomis_scheduled_event['startTime']), # rubocop:disable Rails/TimeZone
         type: 'Prison Activities',
         reason: nomis_scheduled_event['eventTypeDesc'],
         agency_id: nomis_scheduled_event['locationCode'],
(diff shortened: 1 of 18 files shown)


martyn-w (approved):
👍 LGTM

spec/models/activity_spec.rb
```diff
@@ -32,7 +32,7 @@
     let(:expected_attributes) do
       {
         id: nomis_scheduled_event['eventId'],
-        start_time: Time.parse(nomis_scheduled_event['startTime']),
+        start_time: Time.parse(nomis_scheduled_event['startTime']), # rubocop:disable Rails/TimeZone
```
cwrw, line 35: From the comment on the activity class:
Nomis startTime does not include a zone so we shouldn't use our local zone with Time.zone.parse and arbitraily add a timezone

spec/models/nomis_court_hearing_spec.rb
```diff
@@ -21,7 +21,7 @@
     let(:expected_attributes) do
       {
         id: nomis_court_hearing['id'],
-        start_time: Time.parse(nomis_court_hearing['dateTime']),
+        start_time: Time.zone.parse(nomis_court_hearing['dateTime']),
```
smoothcontract, line 24: As this is also coming from Nomis, does the same caveat apply as above with `activity_spec` ?

smoothcontract (commented):
Just one query on another spec related to Nomis but otherwise looks good 👍

spec/models/nomis_court_hearing_spec.rb
cwrw: good question, it looks like this only applies to activity, as in the court hearing we are explicitly passing in the zone:
https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/models/nomis_court_hearing.rb#L11

smoothcontract: Cool, thanks for checking 👍

smoothcontract (approved):
LGTM 👍

PR #1212, opened 2020-12-11
Add new `yoi` attribute to locations

smoothcontract:
### Jira link

P4-2374

### What?

- [x] Add new boolean `yoi` flag to `Location` model
- [x] Include new attribute in serialized locations
- [x] Support `filter[yoi]=false` and `filter[yoi]=true` when fetching a list of locations
- [x] Add new rake task `reference_data:update_locations` to correctly set new flag values

### Why?

- Some institutions have YOI as well as adult populations. Within Nomis these are generally held with the prison location type, which means we can't perform YOI specific logic. In an effort to avoid these data issues from Nomis, the new attribute of YOI can be applied independently of the location type and is used to identify the 5 current YOI locations. The rake task and update service sets the correct YOI value for these known locations. `Locations::Finder service` has been extended to support filtering with/without this attribute to allow the front end to get a list of relevant YOI locations from the reference data end point. Finally the `LocationSerializer` now includes the new attribute.

### Have you? (optional)

- Introduces a new attribute in the serialized locations, so has a knock on effect to various endpoints but will be backwards compatible as it's a new attribute. Suppliers will need to be informed of this addition.

### DIFF
--- a/app/services/locations/updater.rb
+++ b/app/services/locations/updater.rb
@@ -5,8 +5,10 @@ module Locations
     YOI_NOMIS_AGENCIES = %w[WYI WNI CKI PRI FMI].freeze
 
     def self.call
-      Location.update_all(young_offender_institution: false)
-      Location.where(nomis_agency_id: YOI_NOMIS_AGENCIES).update_all(young_offender_institution: true)
+      Location.transaction do
+        Location.update_all(young_offender_institution: false)
+        Location.where(nomis_agency_id: YOI_NOMIS_AGENCIES).update_all(young_offender_institution: true)
+      end
     end
   end
 end
--- a/app/services/locations/updater.rb
+++ b/app/services/locations/updater.rb
@@ -5,8 +5,8 @@ module Locations
     YOI_NOMIS_AGENCIES = %w[WYI WNI CKI PRI FMI].freeze
 
     def self.call
-      Location.update_all(yoi: false)
-      Location.where(nomis_agency_id: YOI_NOMIS_AGENCIES).update_all(yoi: true)
+      Location.update_all(young_offender_institution: false)
+      Location.where(nomis_agency_id: YOI_NOMIS_AGENCIES).update_all(young_offender_institution: true)
     end
   end
 end
--- a/app/services/locations/updater.rb
+++ b/app/services/locations/updater.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+module Locations
+  class Updater
+    YOI_NOMIS_AGENCIES = %w[WYI WNI CKI PRI FMI].freeze
+
+    def self.call
+      Location.update_all(yoi: false)
+      Location.where(nomis_agency_id: YOI_NOMIS_AGENCIES).update_all(yoi: true)
+    end
+  end
+end
--- a/app/controllers/api/reference/locations_controller.rb
+++ b/app/controllers/api/reference/locations_controller.rb
@@ -23,7 +23,7 @@ module Api
         render_json location, serializer: LocationSerializer, include: included_relationships, status: status
       end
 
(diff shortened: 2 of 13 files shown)


martyn-w:
I'm not a big fan of TLAs - could we name the attribute something else? maybe `young_offender_institution` ?

app/services/locations/updater.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module Locations
+  class Updater
+    YOI_NOMIS_AGENCIES = %w[WYI WNI CKI PRI FMI].freeze
```
martyn-w, line 5: lets move the actual list into a yaml file, rather than hard-coded

app/services/locations/updater.rb
```diff
@@ -0,0 +1,12 @@
...
+
+    def self.call
```
martyn-w, line 7: rather than a separate process, could this be done as part of the rake task `reference_data:create_locations` - that gets run relatively regularly and creates / updates / disables locations

app/services/locations/updater.rb
```diff
@@ -0,0 +1,12 @@
...
+      Location.update_all(yoi: false)
```
cwrw, line 8: maybe best to wrap those in a transaction to avoid having that millisecond where all location YOIs are false 👹  or if there is a failure etc.

app/services/locations/updater.rb
smoothcontract: I'm not convinced by this - it would still be hardcoded but in yaml form, and mean slightly more complex code to achieve the same thing. We also have other existing data services that set up data without using a yaml file. Either way it means a PR and a deployment to invoke a change to the data.

app/services/locations/updater.rb
smoothcontract: I prefer to keep it as a separate process but will tack it onto the end of the create locations task too so that it's automatically invoked when importing new locations, just in case any of those happen to be a YOI.

app/services/locations/updater.rb
smoothcontract: Okay will do

app/services/locations/updater.rb
martyn-w: that sounds fine 👍

app/services/locations/updater.rb
martyn-w: moving it to a yaml (e.g. `YAML.safe_load(File.read('./lib/tasks/data/young_offender_institutions.yml'))`) would help keep all our reference data in one place, rather than having to scour the code for the list when the inevitable changes come through. I don't think we have other reference data which is hard coded?

smoothcontract: There are quite a few actually... see:
https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/services/allocation_complex_cases/importer.rb 
https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/services/personal_care_needs/importer.rb
https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/services/nomis_alerts/importer.rb
https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/services/identifier_types/importer.rb
https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/services/genders/importer.rb
https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/services/assessment_questions/importer.rb

I believe it's important to keep things as simple as possible, so really don't agree with moving this array of 5 things to a YAML file. If over time the needs of the service become more complex and we need additional data this can be revised of course but right now this is the simplest thing that works (tm) 😁

martyn-w: ok, fair enough 👍

martyn-w (approved):
👍 LGTM

cwrw (approved):
👍

PR #1213, opened 2020-12-11
P4-2558 Guard against question type change on pre-filling assessments

cwrw:
jira-ticket: https://dsdmoj.atlassian.net/browse/P4-2558

If a question type changes across frameworks, but they share a question key name, do not pre-fill the response value from the previous question. Since the value type validation occurs at setting the value attribute, rescue from this error and return an unpre-filled response instead.

An example of changing the question type would be removing the ability of setting the details section on radio buttons, meaning this changes responses from an object to a string question response.


martyn-w (approved):
👍

PR #1219, opened 2020-12-16
Update metadata for location free spaces endpoint

smoothcontract:
### Jira link

P4-2566

### What?

- [x] Revised `Population` metadata generation to return location free spaces to separately include transfer in and out counts

### Why?

- This now includes a count of transfers (moves) in and out as distinct keys in the metadata, so this necessitated a slight change in the behaviour. Previously if a population model was not present for a given date and location then the corresponding hash value was `nil`. With this change the hash is always generated so that `transfers_in` and `transfers_out` can be populated, but `id` and `free_spaces` keys will be `nil`.

- This allows the PMU dashboard to be updated in the front end so that transfer counts can be shown even if there is not yet a corresponding population record. As the behaviour of the generated metadata is amended this is a breaking change so will need to be released in coordination with the front end, however this is not yet a production feature.

### Have you? (optional)

- [ ] Updated API docs if necessary - This endpoint is intentionally undocumented as it's intended for front end use only

- Changes an api that is used in production (but currently sits behind a feature flag in the front end). No impact on suppliers.

### DIFF
--- a/app/models/population.rb
+++ b/app/models/population.rb
@@ -35,9 +35,14 @@ class Population < ApplicationRecord
     {}.tap do |locations_hash|
       date_range.each do |date|
         free_spaces_for_date(locations, date).each do |row|
-          location_id, population_id, free_spaces = *row
+          location_id, population_id, transfers_in, transfers_out, free_spaces_excluding_transfers = *row
 
-          free_space_details = { id: population_id, free_spaces: free_spaces } if population_id.present?
+          free_space_details = {
+            id: population_id,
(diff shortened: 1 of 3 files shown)


merlinc (commented):
I'm 99% sure that I've covered all the edge cases for this in the frontend app - I check for the deeper properties now, and I also hacked the response from the staging server using Charles and nothing broke.

We should be able to see if anything breaks if this gets deployed to staging and things break. We don't have E2E tests for this yet (deadlines, deadlines), so let's take a look once it's merged.

cwrw (approved):

PR #1221, opened 2020-12-17
Add youth offender institution to list of accepted location for assessment

cwrw:
If the from location is a youth offender institution, allow a youth risk assessment to be created

### DIFF
--- a/app/models/youth_risk_assessment.rb
+++ b/app/models/youth_risk_assessment.rb
@@ -17,6 +17,6 @@ private
   end
 
   def move_from_location
-    errors.add(:move, "'from_location' must be from either a secure training centre or a secure children's home") unless move&.from_location&.secure_training_centre? || move&.from_location&.secure_childrens_home?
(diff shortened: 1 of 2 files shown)


teneightfive (approved):

app/models/youth_risk_assessment.rb
```diff
@@ -17,6 +17,6 @@ def previous_assessment
   end
 
   def move_from_location
-    errors.add(:move, "'from_location' must be from either a secure training centre or a secure children's home") unless move&.from_location&.secure_training_centre? || move&.from_location&.secure_childrens_home?
+    errors.add(:move, "'from_location' must be from either a secure training centre or a secure children's home") unless move&.from_location&.secure_training_centre? || move&.from_location&.secure_childrens_home? || move&.from_location&.young_offender_institution
```
smoothcontract, line 20: It might be a nice option in future to move this logic to the `Location` model but this is absolutely fine now. Perhaps if the complexity grows any further....

smoothcontract (approved):
Looks good. How much fun is it typing `young_offender_institution` rather than just `yoi` 😁

PR #1224, opened 2020-12-18
New supplier locations

martyn-w:
New locations <--> supplier mappings
based on the latest spreadsheets from the programme

### DIFF
--- a/lib/tasks/data/supplier_locations.yml
+++ b/lib/tasks/data/supplier_locations.yml
@@ -463,7 +463,6 @@ suppliers:
     - GNI
     - HBI
     - HCI
-    -
     - HNT1
     - HNT2
     - HNT3
--- a/lib/tasks/data/supplier_locations.yml
+++ b/lib/tasks/data/supplier_locations.yml
@@ -10,35 +10,65 @@ suppliers:
     - BWI
     - CFI
     - CHE1
+    - CHE10
     - CHE2
     - CHE3
+    - CHE4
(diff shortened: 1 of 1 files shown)


lib/tasks/data/supplier_locations.yml
```diff
@@ -229,7 +399,7 @@ suppliers:
     - CAM4
     - CAM5
     - CDI
-    - CITWMC
```
martyn-w, line 232: `CITWMC` is Westminster Mag Court and should not have been in this list as its a court

lib/tasks/data/supplier_locations.yml
```diff
@@ -243,51 +413,99 @@ suppliers:
...
+    - GCS2
+    - GCS3
+    - GCS4
     - GCS5
+    - GCS6
     - GMI
     - GNI
     - HBI
     - HCI
+    -
```
smoothcontract, line 466: Not sure if having this blank line will cause an issue? Probably best to delete just to be safe.

smoothcontract (commented):
Looks good - loads of new locations! Just one extra line which I think it would be safest to remove.

lib/tasks/data/supplier_locations.yml
martyn-w: whoops, that should not be there, good spot

smoothcontract (approved):
LGTM 👍

PR #1227, opened 2020-12-21
feature: Populate created_by on events with current user or supplier

Nimphal:
### Jira link

P4-<1716>

### What?

I have added/removed/altered:
- [x] GenericEvent model to return and serialize created_by
- [x] Creation of Move related events to use the current user passed by the frontend in the created_by column for the event

### Why?

I am doing this because:
We want to show which user did an action that triggered an event from the frontend.

### Have you? (optional)

### DIFF
--- a/app/controllers/api/generic_events_controller.rb
+++ b/app/controllers/api/generic_events_controller.rb
@@ -5,6 +5,7 @@ module Api
       attributes: [
         :event_type,
         :occurred_at,
+        :created_by,
         :recorded_at,
         :notes,
         details: {},
--- a/app/controllers/api/generic_events_controller.rb
+++ b/app/controllers/api/generic_events_controller.rb
@@ -5,7 +5,6 @@ module Api
       attributes: [
         :event_type,
         :occurred_at,
-        :created_by,
         :recorded_at,
         :notes,
         details: {},
@@ -31,6 +30,7 @@ module Api
 
         attributes.merge!('supplier' => doorkeeper_application_owner) if doorkeeper_application_owner
         attributes.merge!('eventable' => eventable) if eventable_params
+        attributes.merge!('created_by' => created_by) if created_by
         attributes['details'] = attributes['details']&.slice(*event_type.constantize.details_attributes) || {}
 
         attributes['details'].merge!(event_specific_relationships) if event_specific_relationships.any?
@@ -58,10 +58,6 @@ module Api
 
     def event_type
       @event_type ||= 'GenericEvent::' + event_params.dig('attributes', 'event_type')
-    end
-
-    def created_by
-      current_user&.owner&.name || 'unknown'
     end
 
     def event_specific_relationships
--- a/app/controllers/api/journeys_controller.rb
+++ b/app/controllers/api/journeys_controller.rb
@@ -145,6 +145,7 @@ module Api
         notes: 'Automatically generated event',
         details: {},
         supplier_id: supplier.id,
+        created_by: user_for_paper_trail || doorkeeper_application_owner&.name,
       }
 
       if action_name == 'update'
--- a/app/controllers/api/journeys_controller.rb
+++ b/app/controllers/api/journeys_controller.rb
@@ -145,7 +145,7 @@ module Api
         notes: 'Automatically generated event',
         details: {},
         supplier_id: supplier.id,
-        created_by: user_for_paper_trail || doorkeeper_application_owner&.name,
+        created_by: created_by,
       }
 
       if action_name == 'update'
--- a/app/controllers/api/move_events_controller.rb
+++ b/app/controllers/api/move_events_controller.rb
@@ -8,13 +8,13 @@ module Api
     before_action :validate_idempotency_key
     around_action :idempotent_action
 
-    APPROVE_PARAMS = [:type, attributes: %i[timestamp date create_in_nomis created_by]].freeze
-    CANCEL_PARAMS = [:type, attributes: %i[timestamp cancellation_reason cancellation_reason_comment notes created_by]].freeze
-    LOCKOUT_PARAMS = [:type, attributes: %i[timestamp notes created_by], relationships: { from_location: {} }].freeze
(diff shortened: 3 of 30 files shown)


app/controllers/api/journeys_controller.rb
```diff
@@ -145,6 +145,7 @@ def create_event
         notes: 'Automatically generated event',
         details: {},
         supplier_id: supplier.id,
+        created_by: user_for_paper_trail || doorkeeper_application_owner&.name,
```
cwrw, line 148: To avoid duplication of the created by logic, maybe we can move this into a helper method which consolidates this logic, maybe in the api controller or anywhere more relevant with something like this:
```
def created_by
  user_for_paper_trail || doorkeeper_application_owner&.name || 'unknown'
end
```
Then we can reuse this everywhere?

cwrw: Can you also add a test to check that the journey events set by the controller will have the created at?

app/controllers/concerns/eventable.rb
```diff
@@ -6,6 +6,10 @@ def supplier_id
     @supplier_id ||= current_user.owner&.id
   end
 
+  def created_by
+    user_for_paper_trail || current_user&.owner&.name || 'unknown'
```
cwrw, line 10: since we are adding the created by to all events, is there a shared example we can add a test to, to verify this behaviour?

cwrw:
There are a few tasks around creating fake events, might be beneficial to add this new attribute to those as well, also to update the swagger/wiki docs

app/services/generic_events/common_params_validator.rb
```diff
@@ -12,7 +12,7 @@ class CommonParamsValidator
       journeys
     ].freeze
 
-    attr_reader :occurred_at, :recorded_at, :event_type, :eventable_type
+    attr_reader :occurred_at, :recorded_at, :event_type, :eventable_type, :created_by
```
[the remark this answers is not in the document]
Nimphal: We do not, good spot! Removed.

app/controllers/api/generic_events_controller.rb
```diff
@@ -5,6 +5,7 @@ class GenericEventsController < ApiController
       attributes: [
         :event_type,
         :occurred_at,
+        :created_by,
```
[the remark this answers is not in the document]
Nimphal: We don't need that created_by anymore, it uses the abstracted method from the api_controller now. I am not 100% familiar with it either, but seeing as there's a journey events controller and move events controller, the generic one would be used by all other event types (profile, person) which do not need special behaviour yet. Having realised that, I also realised it wasn't actually setting it correctly, so have updated.

app/controllers/api/move_events_controller.rb
```diff
@@ -8,13 +8,13 @@ class MoveEventsController < ApiController
     before_action :validate_idempotency_key
     around_action :idempotent_action
 
-    APPROVE_PARAMS = [:type, attributes: %i[timestamp date create_in_nomis]].freeze
-    CANCEL_PARAMS = [:type, attributes: %i[timestamp cancellation_reason cancellation_reason_comment notes]].freeze
-    LOCKOUT_PARAMS = [:type, attributes: %i[timestamp notes], relationships: { from_location: {} }].freeze
-    REDIRECT_PARAMS = [:type, attributes: %i[timestamp notes move_type], relationships: { to_location: {} }].freeze
-    REJECT_PARAMS = [:type, attributes: %i[timestamp rejection_reason cancellation_reason_comment rebook]].freeze
+    APPROVE_PARAMS = [:type, attributes: %i[timestamp date create_in_nomis created_by]].freeze
```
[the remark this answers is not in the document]
Nimphal: Gone!

app/controllers/api/journeys_controller.rb
Nimphal: Have moved it to the api controller, couldn't think of a better place either, and definitely knew I should move it even as I was creating the PR :D

app/controllers/api/generic_events_controller.rb
```diff
@@ -30,6 +30,7 @@ def event_attributes
 
         attributes.merge!('supplier' => doorkeeper_application_owner) if doorkeeper_application_owner
         attributes.merge!('eventable' => eventable) if eventable_params
+        attributes.merge!('created_by' => created_by) if created_by
```
Nimphal, line 33: Uses the api controller set `created_by`, so no need for the deleted method below.

app/controllers/api/journeys_controller.rb
Nimphal: Added tests for the journey events controllers as well :)

app/controllers/concerns/eventable.rb
Nimphal: Unsure there is a shared example, but I think I've covered all places where the behaviour header -> api controller -> individual controller -> model is tested.

app/controllers/api_controller.rb
```diff
@@ -45,6 +45,10 @@ def user_for_paper_trail
     request.headers['X-Current-User']
   end
 
+  def created_by
+    user_for_paper_trail || doorkeeper_application_owner&.name
```
cwrw, line 49: are we not going to add `|| 'unknown'` here (noticed this was an option previously?

app/controllers/api/journeys_controller.rb
```diff
@@ -145,6 +145,7 @@ def create_event
         notes: 'Automatically generated event',
         details: {},
         supplier_id: supplier.id,
+        created_by: created_by,
```
cwrw, line 148: could you add a test in the journey's controller create/update specs as well, as the ones in the journeys **events** controller are inheriting from the eventable concern

cwrw (approved):
thanks for addressing comments, looks good, just a few minor questions otherwise 👍

app/controllers/api_controller.rb
Nimphal: it was an option when the event came from the frontend, it shouldn't be a thing now, or if it is, leaving created_by empty is probably better.

PR #1230, opened 2020-12-22
P4-2537 Surface journeys through moves endpoint

cwrw:
### Jira link

[P4-2537](https://dsdmoj.atlassian.net/browse/P4-2537)

### What?

- Added optional journeys to moves endpoint relationships
- Added lightweight journeys serializer

### Why?

Add an optional journeys relationship to the moves endpoint, which surfaces a journey's attributes including the vehicle registration number which is required for the dashboards on the FE application. Adding a journeys relationship rather than a meta section means we can optionally include this when required, rather than always rendering a meta section, which will always have to load the journeys. However this means that the FE will need to look for the latest journey for a move. 

Only concern is the number of journeys that come back, as that might impact render time due to view serialisation, however the query time is quite fast.

### Have you? (optional)

### DIFF
--- a/swagger/v2/move.yaml
+++ b/swagger/v2/move.yaml
@@ -146,3 +146,6 @@ Move:
         timeline_events:
           $ref: "../v1/event_reference.yaml#/EventReference"
           description: Aliased events about this record and its nested relationships
+        journeys:
+          $ref: "../v1/journey_reference.yaml#/JourneyReference"
+          description: The journeys associated with this move
--- a/app/serializers/v2/journeys_serializer.rb
+++ b/app/serializers/v2/journeys_serializer.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+module V2
+  class JourneysSerializer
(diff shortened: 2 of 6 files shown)


Nimphal (approved):

swagger/v2/move.yaml
```diff
@@ -146,3 +146,6 @@ Move:
         timeline_events:
           $ref: "../v1/event_reference.yaml#/EventReference"
           description: Aliased events about this record and its nested relationships
+        journeys:
+          $ref: "../v1/journey_reference.yaml#/JourneyReference"
```
Nimphal, line 150: I assume there's no v2 file?

cwrw: we only have v2 files for the different documentation between v1 and v2. If there is only one it currently lives in v1

cwrw:
A revised approach was taken regarding surfacing this, closing this pr in favour of an upcoming pr to surface things in optional meta section

PR #1237, opened 2020-12-29
P4-2537 Add metadata section to moves resource

cwrw:
### Jira link

[P4-2537](https://dsdmoj.atlassian.net/browse/P4-2537)

### What?

- Add ability to pass in `meta` query param
- Add vehicle registration field in meta section in the move resource for moves index endpoint

### Why?

To surface vehicle registrations for moves, to be used on the FE dashboards, add a meta section to the moves resources only on the moves index endpoint. The vehicle registration is optional and can be included in the new `meta` query string param. If no meta query param is included, we still get a meta section, but it will be empty, as there is no option to make the section itself optional.
 
### Have you? (optional)

### DIFF
--- a/app/controllers/api_controller.rb
+++ b/app/controllers/api_controller.rb
@@ -262,6 +262,10 @@ private
     @included_relationships ||= include_params_handler.included_relationships
   end
 
+  def meta_fields
+    @meta_fields ||= include_params_handler.meta_fields
+  end
+
   def active_record_relationships
     @active_record_relationships ||= include_params_handler.active_record_relationships
   end
--- a/app/controllers/api_controller.rb
+++ b/app/controllers/api_controller.rb
@@ -263,7 +263,7 @@ private
   end
 
   def meta_fields
-    @meta_fields ||= include_params_handler.meta_fields
+    @meta_fields ||= include_params_handler.meta_fields || []
   end
 
   def active_record_relationships
--- a/app/models/move.rb
+++ b/app/models/move.rb
@@ -216,7 +216,12 @@ class Move < VersionedModel
   end
 
   def vehicle_registration
-    journeys.max_by(&:client_timestamp)&.vehicle&.dig('registration')
+    # Process in memory to avoid n+1 queries in serializers
+    journeys
+      .reject(&:cancelled?)
+      .max_by(&:client_timestamp)
+      &.vehicle
+      &.dig('registration')
   end
 
 private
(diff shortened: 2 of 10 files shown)


app/models/move.rb
```diff
@@ -215,6 +215,10 @@ def important_events
     incident_events + (profile&.person_escort_record&.medical_events || [])
   end
 
+  def vehicle_registration
+    journeys.max_by(&:client_timestamp)&.vehicle&.dig('registration')
```
martyn-w, line 219: not sure this will always return the correct answer... might be a good idea to exclude cancelled journeys?

Note also that the van (and thus registration) could change during the lifetime of the move. So the van coming to collect the person from the start might have a different registration to the van delivering the _same_ person to the destination

martyn-w (approved):
👍 LGTM

app/controllers/api_controller.rb
```diff
@@ -262,6 +262,10 @@ def included_relationships
     @included_relationships ||= include_params_handler.included_relationships
   end
 
+  def meta_fields
+    @meta_fields ||= include_params_handler.meta_fields
```
smoothcontract, line 266: It might be slightly nicer to return an empty array here rather than `nil` if no params were passed, that then makes using this easier without a nil guard in `index_and_render`. Or if that doesn't appeal then perhaps add another method e.g.

```ruby
def meta_include?(name)
  meta_fields&.include?(name.to_s)
end
```

app/models/move.rb
smoothcontract, line 219: Is there a way to do this at SQL level so that only one matching row is returned? That would be more performant then returning all journeys and iterating through them, if it's possible.

app/serializers/v2/moves_serializer.rb
```diff
@@ -23,6 +23,13 @@ class MovesSerializer
 
     set_type :moves
 
+    meta do |object, params|
+      metadata = {}
+
+      metadata.merge!(vehicle_registration: object.vehicle_registration) if params[:vehicle_registration]
+      metadata
```
smoothcontract, line 30:
```suggestion
      {}.tap do |metadata|
        metadata.merge!(vehicle_registration: object.vehicle_registration) if params[:vehicle_registration]
      end
```

smoothcontract (commented):
Just some minor things, but looks pretty good. Also would be nice to see some coverage with multiple journeys showing that the most recent non cancelled registration number is the one returned with the move.

app/models/move.rb
cwrw: Unfortunately the serializer eager loads the journeys before getting to this point, so when we attempt to do this at the SQL level, we get an n+1 query. The performance shouldn't be impacted as there aren't a large number of journeys per move, just a few

smoothcontract (approved):
LGTM 👍

PR #1239, opened 2020-12-30
P4-2538 Surface ETA and collection time in moves meta data

cwrw:
### Jira link

[P4-2538](https://dsdmoj.atlassian.net/browse/P4-2538)
[P4-2539](https://dsdmoj.atlassian.net/browse/P4-2539)

### What?

- Add new type of event classification: notifications
-  Add task to backfill all notification events with correct classification type
- Surface ETA and collection times in meta data for moves on index moves endpoint
- Add task to generate fake data for notification events

### Why?

To allow the FE to build dashboards and group by vehicles expected collection time for outgoing moves, and ETA for incoming moves, surface those values from the relevant events that are attached to those moves in the metadata. The event corresponding to expected collection time is `MoveNotifyPremisesOfExpectedCollectionTime`, which contain an `expected_at` timestamp. The event corresponding to ETA is `MoveNotifyPremisesOfEta`, which also contains an `expected_at` timestamp as well.

To avoid eager loading all events on the index moves endpoint, add a new classification: notification events, which includes all events that a supplier sends to inform of expected times. To help with testing on staging, add a fake data generator for notification events.

### Have you? (optional)

### DIFF
--- a/app/models/generic_event/move_notify_premises_of_arrival_in_30_mins.rb
+++ b/app/models/generic_event/move_notify_premises_of_arrival_in_30_mins.rb
@@ -1,9 +1,5 @@
 class GenericEvent
-  class MoveNotifyPremisesOfArrivalIn30Mins < GenericEvent
+  class MoveNotifyPremisesOfArrivalIn30Mins < Notification
     eventable_types 'Move'
-
-    def event_classification
-      :notification
-    end
   end
 end
--- a/app/models/generic_event/move_notify_premises_of_arrival_in_30_mins.rb
+++ b/app/models/generic_event/move_notify_premises_of_arrival_in_30_mins.rb
@@ -1,5 +1,9 @@
 class GenericEvent
   class MoveNotifyPremisesOfArrivalIn30Mins < GenericEvent
     eventable_types 'Move'
+
+    def event_classification
+      :notification
+    end
   end
 end
--- a/app/models/move.rb
+++ b/app/models/move.rb
@@ -225,6 +225,14 @@ class Move < VersionedModel
       &.dig('registration')
   end
 
+  def expected_time_of_arrival
+    notification_events.select { |event| event.type == 'GenericEvent::MoveNotifyPremisesOfEta' }.max_by(&:occurred_at)&.expected_at
+  end
+
+  def expected_collection_time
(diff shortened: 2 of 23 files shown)


app/models/generic_event/move_notify_premises_of_arrival_in_30_mins.rb
```diff
@@ -1,5 +1,9 @@
 class GenericEvent
   class MoveNotifyPremisesOfArrivalIn30Mins < GenericEvent
     eventable_types 'Move'
+
+    def event_classification
```
smoothcontract, line 5: Another option would be to add an interim ancestor class e.g. `GenericMoveNotifyEvent` and add this code there, making the three existing move notify classes then inherit from that. There's a similar approach to this already used for `Incident` so fits quite nicely with the existing OO style and means less repeated code. Up to you as it's mostly a style preference though...

app/models/move.rb
```diff
@@ -219,6 +220,14 @@ def vehicle_registration
     journeys.max_by(&:client_timestamp)&.vehicle&.dig('registration')
   end
 
+  def expected_time_of_arrival
+    notification_events.select { |event| event.type == 'GenericEvent::MoveNotifyPremisesOfEta' }.max_by(&:occurred_at)&.expected_at
+  end
+
+  def expected_collection_time
+    notification_events.select { |event| event.type == 'GenericEvent::MoveNotifyPremisesOfExpectedCollectionTime' }.max_by(&:occurred_at)&.expected_at
```
smoothcontract, line 228: Would be nice if both of these could use SQL approach to return a single event row for each case if possible.   You mentioned a potential N+1 query issue here - I'm guessing that occurs when returning multiple moves and to mitigate that you're also eager loading `notification_events` for moves? If so then that's probably at least worth mentioning in a comment for these two methods...

smoothcontract (commented):
Looks good, just some minor comments

app/models/move.rb
cwrw: Tested again and unfortunately doing this in SQL is indeed causing an n+1, I added a comment for posterity

smoothcontract (approved):
LGTM 👍

PR #1248, opened 2021-01-05
P4-2058 Add PER confirmation event

cwrw:
### Jira link

[P4-2058](https://dsdmoj.atlassian.net/browse/P4-2058)

### What?

I have added/removed/altered:

- [x] Add a new PER confirmation event
- [x] Add a callback on confirming a PER to create a new confirmation event

### Why?

To surface who and when a PER is confirmed on a move's timeline, add a PER confirmation event which captures who confirms a PER and when as a callback on a successful confirmation of the PER, through the PATCH endpoint.

This event will appear when including a move's timeline events.

### Have you? (optional)

- [] Updated API docs if necessary: I'll update https://github.com/ministryofjustice/hmpps-book-secure-move-api/wiki/Event-Documentation once this is merged

### DIFF
--- a/app/controllers/api/person_escort_records_controller.rb
+++ b/app/controllers/api/person_escort_records_controller.rb
@@ -2,6 +2,8 @@
 
 module Api
   class PersonEscortRecordsController < FrameworkAssessmentsController
+    after_action :create_confirmation_event, only: :update # rubocop:disable LexicallyScopedActionFilter
+
   private
 
     def assessment_class
@@ -10,6 +12,23 @@ module Api
 
     def assessment_serializer
       PersonEscortRecordSerializer
+    end
+
+    def create_confirmation_event
+      now = Time.zone.now.iso8601
+
+      event_attributes = {
+        eventable: assessment,
+        occurred_at: now,
+        recorded_at: now,
+        notes: 'Automatically generated event',
+        details: {
+          confirmed_at: assessment.confirmed_at.iso8601,
+        },
+        created_by: created_by,
+      }
+
+      GenericEvent::PerConfirmation.create!(event_attributes)
     end
   end
 end
--- a/app/models/generic_event.rb
+++ b/app/models/generic_event.rb
@@ -52,6 +52,7 @@ class GenericEvent < ApplicationRecord
     MoveReject
     MoveRequested
     MoveStart
+    PerConfirmation
     PerCourtAllDocumentationProvidedToSupplier
     PerCourtAssignCellInCustody
     PerCourtCellShareRiskAssessment
--- a/app/models/generic_event/per_confirmation.rb
+++ b/app/models/generic_event/per_confirmation.rb
@@ -1,0 +1,8 @@
+class GenericEvent
+  class PerConfirmation < GenericEvent
+    details_attributes :confirmed_at
+    include PersonEscortRecordEventValidations
+
+    validates :confirmed_at, presence: true
+  end
+end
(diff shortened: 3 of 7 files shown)


app/controllers/api/person_escort_records_controller.rb
```diff
@@ -11,5 +18,22 @@ def assessment_class
...
+
+      event_attributes = {
+        eventable: assessment,
+        occurred_at: now,
+        recorded_at: now,
+        notes: 'Automatically generated event',
+        details: {
+          confirmed_at: assessment.confirmed_at.iso8601,
+        },
+        created_by: created_by,
```
cwrw, line 33: I didn't add supplier id, as seems optional on events, and I don't think we need it here. It might be tricky trying to get a supplier at this point and will add unnecessary queries.

martyn-w: I think(?) suppliers will create PERs - and so might be useful to capture which supplier confirmed the PER. You could get the supplier id directly in the controller (method `info_for_paper_trail` in ApplicationController)

app/controllers/api/person_escort_records_controller.rb
```diff
@@ -11,5 +18,22 @@ def assessment_class
     def assessment_serializer
       PersonEscortRecordSerializer
     end
+
+    def create_confirmation_event
+      now = Time.zone.now.iso8601
```
martyn-w, line 23: should the occurred_at/recorded_at be passed by the client to the API, rather than assumed by the API to have occurred at the same time as the PER was updated. E.g. what if there is lag in the supplier network and it takes time (minutes...days) for the event to reach us?

cwrw: hmm very interesting question. As an initial iteration I would suggest we internally set the timestamps, similar to [journeys](https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/controllers/api/journeys_controller.rb#L139) and [moves](https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/controllers/api/v2/moves_actions.rb#L164) although those might never be created by suppliers (not sure). 
If we were to implement this however there are two approaches:
1) Do not separate the confirmed at timestamp from the recorded at: allow a client to pass in the confirmed at value and we'll use that to set the recorded at/confirmed at. This will look better on the timeline and is more inline with the PER interface and doesn't surface any internal callbacks we might be doing. However we will have to change the way we set confirmed at which is currently done in the state machine
2) Separate the confirmed at timestamp from the recorded at: either by sending the recorded at and setting the confirmed at at the time internally, or sending both, and setting those respectively. However this will surface some internal callback data I'm not keen on on the interface. 

But I would like to avoid this for now, mainly for the sake of avoiding the amendment of the PER PATCH API interface, until this might become a problem. Let me know your thoughts

martyn-w: sure, that sounds reasonable for now 👍

martyn-w (approved):

smoothcontract (approved):
LGTM - just one comment re: rubocop if it's possible to resolve without adding extra code?

app/controllers/api/person_escort_records_controller.rb
Nimphal: Bit late to this party, but the `created_by` will be the user if it's come from the frontend or the doorkeeper application owner name if not, so it will say which supplier did it if directly through the API.

PR #1249, opened 2021-01-05
Change the type of a PER confirmed error to return a code in API response

cwrw:
To include a code in a PER confirmed error, raise a RecordInvalid error instead,
and set the code to 'invalid_status', to allow clients to look at the code and not the description of the error. This type of error is handled differently in the API controller and adds a code and pointer.

Old error:
```
{
    "errors": [
        {
            "title": "Invalid status",
            "detail": "Validation failed: Status can't update to 'confirmed' from 'confirmed'"
        }
    ]
}
```

New error format:
```
{
    "errors": [
        {
            "title": "Unprocessable entity",
            "detail": "Status can't update to 'confirmed' from 'confirmed'",
            "source": {
                "pointer": "/data/attributes/status"
            },
            "code": "invalid_status"
        }
    ]
}
```

### DIFF
--- a/spec/requests/api/person_escort_records_controller_update_spec.rb
+++ b/spec/requests/api/person_escort_records_controller_update_spec.rb
@@ -73,8 +73,14 @@ RSpec.describe Api::PersonEscortRecordsController do
 
         it_behaves_like 'an endpoint that responds with error 422' do
           let(:errors_422) do
-            [{ 'title' => 'Invalid status',
-               'detail' => "Validation failed: Status can't update to 'confirmed' from 'in_progress'" }]
+            [
+              {
+                'title' => 'Unprocessable entity',
+                'detail' => "Status can't update to 'confirmed' from 'in progress'",
+                'source' => { 'pointer' => '/data/attributes/status' },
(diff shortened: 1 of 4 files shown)


spec/requests/api/person_escort_records_controller_update_spec.rb
```diff
@@ -73,8 +73,14 @@
...
         it_behaves_like 'an endpoint that responds with error 422' do
           let(:errors_422) do
-            [{ 'title' => 'Invalid status',
-               'detail' => "Validation failed: Status can't update to 'confirmed' from 'in_progress'" }]
+            [
+              {
+                'title' => 'Unprocessable entity',
+                'detail' => "Status can't update to 'confirmed' from 'in progress'",
+                'source' => { 'pointer' => '/data/attributes/status' },
+                'code' => 'invalid_status',
```
cwrw, line 81: not sure about this code name, let me know of any alternative thoughts

teneightfive (approved):

martyn-w (approved):
👍 LGTM

PR #1251, opened 2021-01-06
Add endpoint to provide pre-populated defaults for new population record

smoothcontract:
### Jira link

P4-2576

### What?

- [x] Added a new `GET /populations/new` endpoint

### Why?

- This doesn't create a new record, but returns default values to the front end with selected information copied from the most recent population details for the same location. This then allows the front end to provide a form with pre-populated information for the user to amend as needed and then call the existing `POST /populations` endpoint to actually create the new population record. This saves users time and avoids them having to manually look up and copy/paste previous figures.

### Have you? (optional)

- New endpoint so minimal risk

### DIFF
--- a/spec/models/population_spec.rb
+++ b/spec/models/population_spec.rb
@@ -151,6 +151,51 @@ RSpec.describe Population do
     end
   end
 
+  describe '.new_with_defaults' do
+    subject(:new_population) { described_class.new_with_defaults(location: location, date: date) }
+
+    let(:location) { create(:location, :prison) }
+    let(:date) { Date.today }
+
+    context 'with a previous population record for same location' do
+      let!(:previous_population) { create(:population, location: location, date: date - 1.day) }
+
+      it 'populates details from previous record' do
+        expect(new_population).to have_attributes({
+          id: nil,
+          location_id: location.id,
+          date: date,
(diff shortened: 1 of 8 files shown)


spec/models/population_spec.rb
```diff
@@ -151,6 +151,51 @@
...
+          bedwatch: previous_population.bedwatch,
+          overnights_in: previous_population.overnights_in,
+          overnights_out: previous_population.overnights_out,
+          out_of_area_courts: previous_population.out_of_area_courts,
+          discharges: previous_population.discharges,
+        })
+      end
+    end
+
+    context 'without a previous population record' do
```
cwrw, line 180: can you also add a test for checking that it selects the latest previous population?

swagger/v1/population.yaml
```diff
@@ -1,3 +1,100 @@
...
+          oneOf:
+          - type: string
+          - type: 'null'
+          format: date-time
+          description: Timestamp of when the population was last created or updated
+          readOnly: true
+        created_at:
+          oneOf:
+          - type: string
+          - type: 'null'
```
cwrw, line 78: can you add the format here for date-time

cwrw (approved):
looks good, just a few minor comments

PR #1255, opened 2021-01-07
P4-2590 Add POST endpoints

cwrw:
### Jira link

[P4-2590](https://dsdmoj.atlassian.net/browse/P4-2590)

### What?

I have added/removed/altered:
- Added a new POST moves and allocations filtered endpoint
- Added a new moves and allocations filtered controller inheriting from the main controllers
-  Amended the API controller to support extending underscored controller action modules
- Added a moves filtered action module which includes the main moves action module for V2

### Why?

Our FE client is sending large filter query params, which exceed the query param limit, forcing them to batch multiple requests to get the result required. Allow filters to be POSTed to the API for both moves and allocations endpoints, whilst keeping other parameters the same as previously. 

For the Moves controller, we have had to add a new filtered moves actions module which includes the moves action module with all the required methods as this was not automatically loading anything even if we explicitly extended the moves action class itself.

### Have you? (optional)

- Updated API docs if necessary: I will not be adding this to the documentation as this won't be used by anyone other than the FE client	

- Adds a new endpoint and doesn't amend existing endpoints


app/controllers/api/v2/moves_filtered_actions.rb
```diff
@@ -0,0 +1,5 @@
+module Api::V2
```
cwrw, line 1: I had to add this module as the `Api::V2::MovesActions` was not being extended even when I explicitly added the following line to the filtered controller itself: `extended Api::V2::MovesActions`

app/controllers/api_controller.rb
```diff
@@ -286,7 +286,7 @@ def api_version
   def extend_versioned_controller_actions
     version = "Api::V#{api_version || DEFAULT_API_VERSION}"
 
-    actions_module = "#{controller_name.capitalize}Actions"
+    actions_module = "#{controller_name.camelize}Actions"
```
cwrw, line 289: this was not working with controller names like `moves_filtered`

app/controllers/api/allocations_filtered_controller.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module Api
+  class AllocationsFilteredController < AllocationsController
+  private
+
+    PERMITTED_CREATE_FILTER_PARAMS = [
```
martyn-w, line 7: "create" feels a bit mis-named, as nothing is actually being created...  How about `PERMITTED_FILTER_PARAMS` ?

app/controllers/api/v2/moves_filtered_actions.rb
```diff
@@ -0,0 +1,5 @@
...
+  module MovesFilteredActions
+    include Api::V2::MovesActions
```
martyn-w, line 3: Are all the methods in `Api::V2::MovesActions` required? I think only the `index_and_render` method is necessary? its probably not wise to include `create_and_render`, `update_and_render` etc

app/controllers/api/moves_filtered_controller.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module Api
+  class MovesFilteredController < MovesController
+  private
+
+    PERMITTED_CREATE_FILTER_PARAMS = [
```
smoothcontract, line 7: Same comment applies here with naming - ideally make these consistent across both new controllers, but definitely avoid using "create" in the name as that feels misleading.

smoothcontract (commented):
The choice of routing for this and associated introduction of new controllers feels unwieldy to me. I think the code would look simpler with a new custom action in each of the existing controllers, which also means the workaround in `ApiController` wouldn't be necessary. It's also likely to be nicer to work with all of the allocations and moves stuff in a single controller rather than splitting them out like this. The external routes can remain the same, but e.g. `POST /api/allocations/filtered` would route to `filtered` action within the existing `AllocationsController`. I appreciate that we're far from the ideal idiomatic path here, but nested controllers like this are a much better fit for related/child models rather than just a custom action against moves/allocations.

Thoughts?

smoothcontract (approved):
This does look simpler without introducing the new controllers and modules, so nice job and thanks for revisiting this 👍

smoothcontract (approved):
LGTM 👍

PR #1257, opened 2021-01-07
P4-2589 Use value_type column value in response serializer

cwrw:
jira-ticket: https://dsdmoj.atlassian.net/browse/P4-2589

Switch to using persisted value type on responses, to avoid having to reach into framework questions. This is the second part of a phased deploy, after all value types have been backfilled on all ENVs


Thource (approved):

smoothcontract (approved):
Nice refactor, this should help performance a bit more 👍

PR #1261, opened 2021-01-11
Set value type when creating framework responses

cwrw:
We currently slice response and dependent response values before setting them in bulk when creating an assessment. Make sure `value_type` is in the selected fields to be set on responses as well as dependent responses


martyn-w (approved):
👍 LGTM

smoothcontract (approved):
LGTM 👍

PR #1267, opened 2021-01-13
Add a unique index to the profile id on youth risk assessments

cwrw:
Alls well that ends well

To prevent two youth risk assessments being created for the same profile add db level unique index.

### DIFF
--- a/db/migrate/20210113092719_add_unique_index_on_youth_risk_assessment.rb
+++ b/db/migrate/20210113092719_add_unique_index_on_youth_risk_assessment.rb
@@ -1,0 +1,6 @@
+class AddUniqueIndexOnYouthRiskAssessment < ActiveRecord::Migration[6.0]
+  def change
+    remove_index :youth_risk_assessments, :profile_id
+    add_index :youth_risk_assessments, :profile_id, unique: true
+  end
+end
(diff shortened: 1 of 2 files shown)


Nimphal (approved):

db/migrate/20210113092719_add_unique_index_on_youth_risk_assessment.rb
```diff
@@ -0,0 +1,6 @@
+class AddUniqueIndexOnYouthRiskAssessment < ActiveRecord::Migration[6.0]
+  def change
+    remove_index :youth_risk_assessments, :profile_id
+    add_index :youth_risk_assessments, :profile_id, unique: true
```
martyn-w, line 4: might be best to check for duplicates, otherwise this will fail

cwrw: good call, Ive checked on prod and staging, and there are no duplicates
```
YouthRiskAssessment.select(:profile_id).group(:profile_id).having("count(*) > 1")
=> #<ActiveRecord::Relation []>
```

martyn-w: jolly good 👍

martyn-w (approved):

PR #1269, opened 2021-01-13
Determine unlock and discharges from Nomis for new population endpoint

smoothcontract:
### Jira link

P4-2575

### What?

- [x] Add a new `Populations::DefaultsFromNomis` service and associated `NomisClient` subclasses to use Nomis API calls to determine `unlock` and `discharges` counts when providing default values for a new `Population` to the frontend via `GET /populations/new` endpoint

### Why?

- Previously these values were copied from the previous days population record, but the values can be determined via Nomis instead which should be up to date. Unfortunately given the limitations of the Nomis API this means making three separate API calls synchronously and then combining the output of these together to get the correct values. Nomis should be fast enough in production to do this while the front end is waiting for the response - during dev testing the endpoints are reasonably fast, and production is thought to be faster.

- I did explore options for making these 3 calls in parallel, but that's actually not a trivial task and is best achieved with one of a few gems designed for asynchronous HTTP calls. On balance it didn't seem worthwhile to add another dependency; if the observed performance is unsatisfactory then this can be revisited. This is a nice reference I found: https://pawelurbanek.com/ruby-concurrent-requests

- The new population API is not yet used in production, so minimal risk.


app/services/populations/defaults_from_nomis.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module Populations
+  class DefaultsFromNomis
+    def self.call(location, date)
+      nomis_agency_id = location.nomis_agency_id
+      assigned_cells = NomisClient::Rollcount.get(nomis_agency_id, false)
```
martyn-w, line 7: small point: it might be a bit clearer if the service could be invoked with
` NomisClient::Rollcount.get(nomis_agency_id, :assigned)`
and
` NomisClient::Rollcount.get(nomis_agency_id, :unassigned)`
rather than true/false

martyn-w (approved):
👍  LGTM - minor comment on using symbols rather than a boolean for assigned/unassigned

app/services/populations/defaults_from_nomis.rb
smoothcontract: Hmm... not really sure how that would work in practice. I could add an `assigned_or_unassigned` param to the `get` method and then check the value of that `== :assigned` but it doesn't feel like it would be cleaner that way.

If the boolean param is an issue my preference would be to split this into two methods named `get_assigned` and `get_unassigned` within the `Rollcount` class, but that might cause some knock on issues working with the Nomis Oauth test stubs. I can experiment with that if the boolean param is a big issue?

martyn-w: its a small concern so not worth spending much time over. But if `get_assigned` and `get_unassigned` is possible that would be good too 👍 
my worry is it is not obvious what `NomisClient::Rollcount.get(nomis_agency_id, false)` means without looking at the definition of the method.

smoothcontract: Okay, so if the concern is around argument clarity I'll change them to named arguments instead as the simplest fix. Some people have a dislike of any boolean arguments at all so I thought that was what you meant. Updates soon!

martyn-w: yep, named args would be much clearer 👍

app/services/populations/defaults_from_nomis.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module Populations
+  class DefaultsFromNomis
+    def self.call(location, date)
+      nomis_agency_id = location.nomis_agency_id
+      assigned_cells = NomisClient::Rollcount.get(agency_id: nomis_agency_id, unassigned: false)
+      unassigned_cells = NomisClient::Rollcount.get(agency_id: nomis_agency_id, unassigned: true)
```
martyn-w, line 8: excellent 👍

martyn-w (approved):
👍 LGTM

PR #1270, opened 2021-01-13
P4-2564 Improve serializer performance for assessments

cwrw:
### Jira link

[P4-2564]( https://dsdmoj.atlassian.net/browse/P4-2564)

### What?

- Change eager loading includes on move endpoint
- Add eager loading includes on assessments
- Improve assessment serializers to avoid always eager loading resources
- Improve assessment creation performance

### Why?

To improve the performance of assessments, and to avoid always setting up queries for the worst case scenario, only eager load required and included resources. This means we can cleanup the serializers in question.

Whilst testing there were some n+1 queries occurring on latest assessment, so now that has been cleaned up to look at the pre-fill source.

I avoided fixing up more serializers to avoid too much change in one PR. There is further work to be done on the meta section which is also a big offender, which I will be tackling separately. 

### DIFF
--- a/app/controllers/api/framework_assessments_controller.rb
+++ b/app/controllers/api/framework_assessments_controller.rb
@@ -54,7 +54,9 @@ module Api
     end
 
     def assessment
-      @assessment ||= assessment_class.find(params[:id])
+      @assessment ||= assessment_class
+        .includes(active_record_relationships)
+        .find(params[:id])
     end
 
     def render_assessment(assessment, status)
--- a/app/controllers/api/framework_assessments_controller.rb
+++ b/app/controllers/api/framework_assessments_controller.rb
@@ -16,12 +16,12 @@ module Api
     ].freeze
 
     def create
-      assessment = assessment_class.save_with_responses!(
+      new_assessment = assessment_class.save_with_responses!(
         version: new_assessment_params.dig(:attributes, :version),
         move_id: new_assessment_params.dig(:relationships, :move, :data, :id),
       )
 
-      render_assessment(assessment, :created)
+      render_assessment(assessment(new_assessment.id), :created)
     end
 
     def update
@@ -53,10 +53,10 @@ module Api
       assessment_serializer::SUPPORTED_RELATIONSHIPS
     end
 
-    def assessment
+    def assessment(id = params[:id])
       @assessment ||= assessment_class
         .includes(active_record_relationships)
-        .find(params[:id])
+        .find(id)
     end
 
     def render_assessment(assessment, status)
--- a/app/models/concerns/framework_assessmentable.rb
+++ b/app/models/concerns/framework_assessmentable.rb
@@ -150,9 +150,9 @@ module FrameworkAssessmentable
 
   def previous_responses
     @previous_responses ||= begin
-      return {} unless previous_assessment
+      return {} unless prefill_source
 
-      previous_assessment.framework_responses.includes(framework_question: :dependents).each_with_object({}) do |response, hash|
+      prefill_source.framework_responses.includes(framework_question: :dependents).each_with_object({}) do |response, hash|
         value = response.prefill_value
         next if value.blank?
 
--- a/app/serializers/framework_assessment_serializer.rb
+++ b/app/serializers/framework_assessment_serializer.rb
@@ -2,15 +2,12 @@
 
 class FrameworkAssessmentSerializer
   include JSONAPI::Serializer
+  include JSONAPI::ConditionalRelationships
 
   belongs_to :framework
 
-  has_many :responses, serializer: FrameworkResponseSerializer do |object|
-    object.framework_responses.includes(:framework_flags, :framework_nomis_mappings, framework_question: [:framework, dependents: :dependents])
-  end
-  has_many :flags, serializer: FrameworkFlagSerializer do |object|
-    object.framework_flags.includes(framework_question: :dependents)
-  end
+  has_many_if_included :responses, serializer: FrameworkResponseSerializer, &:framework_responses
+  has_many_if_included :flags, serializer: FrameworkFlagSerializer, &:framework_flags
 
   attributes :confirmed_at, :created_at, :nomis_sync_status
 
--- a/app/serializers/framework_flag_serializer.rb
+++ b/app/serializers/framework_flag_serializer.rb
@@ -5,7 +5,7 @@ class FrameworkFlagSerializer
 
   set_type :framework_flags
 
-  belongs_to :question, serializer: FrameworkQuestionSerializer, object_method_name: 'framework_question', id_method_name: 'framework_question_id'
+  belongs_to :question, serializer: FrameworkQuestionSerializer, object_method_name: :framework_question, id_method_name: :framework_question_id
 
   attributes :flag_type, :title, :question_value
 
--- a/app/serializers/framework_flag_serializer.rb
+++ b/app/serializers/framework_flag_serializer.rb
@@ -5,7 +5,7 @@ class FrameworkFlagSerializer
 
   set_type :framework_flags
 
-  belongs_to :question, serializer: FrameworkQuestionSerializer, &:framework_question
+  belongs_to :question, serializer: FrameworkQuestionSerializer, object_method_name: 'framework_question', id_method_name: 'framework_question_id'
 
   attributes :flag_type, :title, :question_value
 
--- a/app/serializers/framework_response_serializer.rb
+++ b/app/serializers/framework_response_serializer.rb
@@ -7,7 +7,7 @@ class FrameworkResponseSerializer
   set_type :framework_responses
 
   belongs_to :assessment, polymorphic: true, &:assessmentable
-  belongs_to :question, serializer: FrameworkQuestionSerializer, object_method_name: 'framework_question', id_method_name: 'framework_question_id'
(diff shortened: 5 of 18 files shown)


app/controllers/api/framework_assessments_controller.rb
```diff
@@ -16,12 +16,12 @@ class FrameworkAssessmentsController < ApiController
...
 
     def create
-      assessment = assessment_class.save_with_responses!(
+      new_assessment = assessment_class.save_with_responses!(
         version: new_assessment_params.dig(:attributes, :version),
         move_id: new_assessment_params.dig(:relationships, :move, :data, :id),
       )
 
-      render_assessment(assessment, :created)
+      render_assessment(assessment(new_assessment.id), :created)
```
cwrw, line 24: we can't include resources on a single resource, which is forcing us to find it again, happy for any alternative ideas on how to do this

smoothcontract: Hmm... interesting. So I think I follow the code here - `save_with_responses` is creating a new framework instance, then persisting it with all of the associated framework responses etc. So at this point you have the corresponding record object, but can't then call `.includes()` on it.

If you think about it, at that point all of the object relationships should already have been loaded into memory, so there's no need to retrieve it from the database again. I suspect that this change is actually performing more work now than before 🤔

app/controllers/api/framework_assessments_controller.rb
```diff
@@ -53,8 +53,10 @@ def supported_relationships
       assessment_serializer::SUPPORTED_RELATIONSHIPS
     end
 
-    def assessment
-      @assessment ||= assessment_class.find(params[:id])
+    def assessment(id = params[:id])
+      @assessment ||= assessment_class
+        .includes(active_record_relationships)
+        .find(id)
```
smoothcontract, line 59: These changes are helpful for the `show` method but probably not for `create` I think - so perhaps you don't need to pass in an optional `id` argument and this could be simplified further to

```suggestion
def assessment
    @assessment ||= assessment_class.includes(active_record_relationships).find(params[:id])
end
```

app/models/concerns/framework_assessmentable.rb
```diff
@@ -150,9 +150,9 @@ def previous_assessment
 
   def previous_responses
     @previous_responses ||= begin
-      return {} unless previous_assessment
+      return {} unless prefill_source
```
smoothcontract, line 153: Could you give a little context for this change pls - seems unrelated to performance tuning?

app/serializers/framework_assessment_serializer.rb
```diff
@@ -2,15 +2,12 @@
 
 class FrameworkAssessmentSerializer
   include JSONAPI::Serializer
+  include JSONAPI::ConditionalRelationships
```
smoothcontract, line 5: 👍

app/serializers/framework_assessment_serializer.rb
```diff
@@ -2,15 +2,12 @@
...
   belongs_to :framework
 
-  has_many :responses, serializer: FrameworkResponseSerializer do |object|
-    object.framework_responses.includes(:framework_flags, :framework_nomis_mappings, framework_question: [:framework, dependents: :dependents])
-  end
-  has_many :flags, serializer: FrameworkFlagSerializer do |object|
-    object.framework_flags.includes(framework_question: :dependents)
-  end
+  has_many_if_included :responses, serializer: FrameworkResponseSerializer, &:framework_responses
+  has_many_if_included :flags, serializer: FrameworkFlagSerializer, &:framework_flags
```
smoothcontract, line 10: Yay, much cleaner 👏

app/serializers/framework_flag_serializer.rb
```diff
@@ -5,7 +5,7 @@ class FrameworkFlagSerializer
 
   set_type :framework_flags
 
-  belongs_to :question, serializer: FrameworkQuestionSerializer, &:framework_question
+  belongs_to :question, serializer: FrameworkQuestionSerializer, object_method_name: 'framework_question', id_method_name: 'framework_question_id'
```
smoothcontract, line 8: I never quite understood this approach vs just calling the method as a block argument. Did you find this improves the DB queries? Also would this work with symbols for the method names - might look slightly cleaner if so.

app/serializers/framework_response_serializer.rb
```diff
@@ -2,13 +2,14 @@
...
   include JSONAPI::Serializer
+  include JSONAPI::ConditionalRelationships
 
   set_type :framework_responses
 
   belongs_to :assessment, polymorphic: true, &:assessmentable
-  belongs_to :question, serializer: FrameworkQuestionSerializer, &:framework_question
-  has_many :flags, serializer: FrameworkFlagSerializer, &:framework_flags
-  has_many :nomis_mappings, serializer: FrameworkNomisMappingSerializer, &:framework_nomis_mappings
+  belongs_to :question, serializer: FrameworkQuestionSerializer, object_method_name: 'framework_question', id_method_name: 'framework_question_id'
```
smoothcontract, line 10: Same as above - use symbols for method names if possible?

app/services/include_param_handler.rb
```diff
@@ -41,6 +41,14 @@ def db_alias(name)
...
     when 'questions'
       :framework_questions
+    when 'question'
+      :framework_question
+    when 'descendants'
+      :dependents
+    when '**'
+      { dependents: :dependents }
+    when 'nomis_mappings'
+      :framework_nomis_mappings
```
smoothcontract, line 51: This is evolving into a very useful pattern, I like it 👍

smoothcontract (commented):
Great progress on this, I left a few minor comments but mainly around the `create` changes which may not be necessary, and perhaps detrimental. Happy to discuss on a call.

app/controllers/api/framework_assessments_controller.rb
cwrw: So through testing unfortunately it was actually ignoring all the loaded records and fetching them again when loading the serializer. But because we didn't include the resources the n+1 queries were substantial. By adding only a find with the includes after the creation we avoid all the many queries

app/controllers/api/framework_assessments_controller.rb
cwrw: please see above comment

app/models/concerns/framework_assessmentable.rb
cwrw: this was calling previous assessment every time regardless of memoization, meaning it would call it around 53 times (depending on the framework response count). I realised we already set this value on the record and save it as prefill source. If we call prefill source here we avoid all the extra queries

app/serializers/framework_flag_serializer.rb
cwrw: I was hoping to avoid this, but it seems if we include something such as a map for the real name of relationship i.e. `&:framework_question` it overwrites the default serializer methods, and ignores everything loaded into memory and retrieves it again causing n+1 queries. By passing in explicitly the object names/id methods we allow the JSON serializer to lookup those values normally. I'll try to see if it works with symbols

app/models/concerns/framework_assessmentable.rb
smoothcontract: Awesome, nice explanation 👍

app/serializers/framework_flag_serializer.rb
smoothcontract: Interesting... then I wonder if this approach could be adopted a bit more widely and it may mitigate the redundant reloading you saw in `create` controller method above 🤔

app/services/include_param_handler.rb
```diff
@@ -41,6 +41,14 @@ def db_alias(name)
       :framework_flags
     when 'questions'
       :framework_questions
+    when 'question'
+      :framework_question
+    when 'descendants'
+      :dependents
+    when '**'
```
cwrw, line 48: write comment on deprecation

app/controllers/api/framework_assessments_controller.rb
smoothcontract: Explored this in person - @cwrw is correct regarding the strange `includes` behaviour for a record that's already (in theory) loaded. This fix means one additional simple query to fetch the assessment again, but that mitigates a *lot* of potential further N+1 queries so is an acceptable approach - I can't think of anything better than this that would work either. All good 👍

app/controllers/api/framework_assessments_controller.rb
smoothcontract: Discussed in person - keeping above `create` changes, so these can stay as is too.

app/serializers/framework_flag_serializer.rb
smoothcontract: Discussed in person - using the block syntax does indeed cause a torrent of N+1 queries so this is a better approach to adopt, but it only seems to help with `belongs_to` relationships and does not improve `has_many`. Given that we have a mixture of both relationships across the serializer graph we can't fix the N+1 loading everywhere like this so need to rely on eager loading the relevant active record relationships where appropriate. Moving forward, we can adopt this pattern for other `belongs_to` serializer relationships that need to present a different name in the serialized output.

smoothcontract (approved):
Walked through this in person - great improvement to query the requested includes rather than rely on worst case scenario. Would be great to see this adopted in the front end now and to only request includes that will be displayed in the UI. In the meantime this will improve performance for supplier API calls 👏

PR #1272, opened 2021-01-14
Clean up legacy Rswag test approach

smoothcontract:
### What?

- [x] Remove/replace legacy `spec/swagger/api/*_spec.rb` integration test approach with request specs
- [x] Remove redundant spec support files relating to rswag 
- [x] Clean up build pipeline for generating Swagger API docs
- [x] Clean up Rspec configuration to avoid duplicated test output

### Why?

- We have migrated to using explicit API definitions in `swagger_doc_v1/v2.yaml` rather than the cumbersome RSwag DSL within integration specs. This change removes the last few remnants of the old testing/documentation approach so that we now have a single consistent pattern to follow. Added a corresponding v1 `GET /move/id` spec to replace the RSwag spec (v2 is already well covered). The other integration specs were already covered with request specs but had some documentation holes left which I've filled.

- In addition, this allows further cleanup around RSwag (although we still need to retain the gem for now). Generating swagger documentation via the Circle CI pipeline can now use the standard rswag rake task, with no need to explicitly run individual integration specs. Also there's no need to explicitly specify Rswag formatter for Rspec configuration, this removes the annoying duplicated output when running specs.

### Have you? (optional)

- No API changes


spec/swagger/swagger_doc_v2.yaml
```diff
@@ -289,46 +289,6 @@
             application/vnd.api+json:
               schema:
                 "$ref": "../v2/patch_people_responses.yaml#/422"
-  "/people/{person_id}/images":
```
smoothcontract, line 292: Moved this to v1 docs (with a few improvements) as it's supported (and identical) across both versions of the API. Not entirely sure how it ended up in the v2 docs.

cwrw (approved):
great cleanup, also this fixes the double rspec results! 👏

cwrw (approved):

PR #1274, opened 2021-01-15
Fix filters bug

cwrw:
If a filter is present but empty in a query parameter, it is interpreted as an empty string. When splitting an empty string the result is an empty array. This means that we query an empty array instead of nil for filters which does not return the intended results. Instead we should return nil when the string is empty to query any filter and expect values that are nil to be returned.

```
[3] pry(main)> ''.split(',')
=> []
[4] pry(main)> Move.where(to_location_id: nil).count
   (8.9ms)  SELECT COUNT(*) FROM "moves" WHERE "moves"."to_location_id" IS NULL
=> 3320
[5] pry(main)> Move.where(to_location_id: []).count
   (0.6ms)  SELECT COUNT(*) FROM "moves" WHERE 1=0
=> 0
[6] pry(main)> Move.where(cancellation_reason: []).count
   (0.6ms)  SELECT COUNT(*) FROM "moves" WHERE 1=0
=> 0
[7] pry(main)> Move.where(cancellation_reason: nil).count
   (14.6ms)  SELECT COUNT(*) FROM "moves" WHERE "moves"."cancellation_reason" IS NULL
=> 28222
[8] pry(main)> Move.where(rejection_reason: []).count
   (0.5ms)  SELECT COUNT(*) FROM "moves" WHERE 1=0
=> 0
[9] pry(main)> Move.where(rejection_reason: nil).count
   (16.7ms)  SELECT COUNT(*) FROM "moves" WHERE "moves"."rejection_reason" IS NULL
=> 43062
```

### DIFF
--- a/app/services/allocations/finder.rb
+++ b/app/services/allocations/finder.rb
@@ -73,7 +73,9 @@ module Allocations
     end
 
     def split_params(name)
-      filter_params[name]&.split(',')
+      return if filter_params[name].blank?
+
+      filter_params[name].split(',')
     end
 
     def apply_location_filters(scope)
--- a/app/services/locations/finder.rb
+++ b/app/services/locations/finder.rb
@@ -40,7 +40,9 @@ module Locations
     end
 
     def split_params(name)
-      filter_params[name]&.split(',')
+      return if filter_params[name].blank?
+
+      filter_params[name].split(',')
     end
 
     def apply_supplier_filters(scope)
--- a/app/services/moves/finder.rb
+++ b/app/services/moves/finder.rb
@@ -59,7 +59,9 @@ module Moves
     end
 
     def split_params(name)
-      filter_params[name]&.split(',')
+      return if filter_params[name].blank?
+
+      filter_params[name].split(',')
     end
 
     def apply_filter(scope, param_name)
--- a/app/services/v2/people/finder.rb
+++ b/app/services/v2/people/finder.rb
@@ -16,7 +16,9 @@ module V2
     private
 
       def split_params(name)
-        filter_params[name]&.split(',')
+        return if filter_params[name].blank?
+
+        filter_params[name].split(',')
       end
 
       def apply_filters(scope)
--- a/spec/services/moves/finder_spec.rb
+++ b/spec/services/moves/finder_spec.rb
@@ -341,9 +341,10 @@ RSpec.describe Moves::Finder do
 
       context 'with empty cancellation reason' do
         let(:filter_params) { { cancellation_reason: '' } }
+        let!(:prison_recall_move) { create :move, :prison_recall }
 
         it 'returns only moves without a cancellation reason' do
-          expect(results).to be_empty
+          expect(results).to contain_exactly(prison_recall_move)
         end
       end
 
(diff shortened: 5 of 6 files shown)


app/services/moves/finder.rb
```diff
@@ -59,7 +59,9 @@ def apply_filters(scope)
     end
 
     def split_params(name)
-      filter_params[name]&.split(',')
+      return if filter_params[name].blank?
```
smoothcontract, line 62: I'd prefer that we have consistent behaviour with this across all of the endpoints that support a comma delimited list - so either we don't do this anywhere, or it should also be done in the corresponding `split_params` method in `Locations::Finder`, `Allocations::Finder` and `People::Finder`.

I'm also not entirely clear why the front end can't just omit empty filters in the request and then no changes would be necessary? 🤔

cwrw: sure I've added them, we're going to test this, and if all is well I'll add all the necessary tests.

I think its mostly a bug/unintentional behaviour on our side, if I send a filter with an empty value I expect records to come back which have this value as empty rather than none.

app/services/allocations/finder.rb
```diff
@@ -73,6 +73,8 @@ def apply_date_range_filters(scope)
     end
 
     def split_params(name)
+      return if filter_params[name].blank?
```
martyn-w, line 76: how about `filter_params[name].presence&.split(',')`

app/services/moves/finder.rb
smoothcontract: Okay, that's cool 👍

app/services/allocations/finder.rb
smoothcontract: I think I prefer the code as @cwrw has it - it's easier to read and understand quickly

app/services/allocations/finder.rb
```diff
@@ -73,6 +73,8 @@ def apply_date_range_filters(scope)
...
+
       filter_params[name]&.split(',')
```
smoothcontract, line 78: You could remove the `&` here now as it's guarded by the new line above.

```suggestion
      filter_params[name].split(',')
```

app/services/locations/finder.rb
```diff
@@ -40,6 +40,8 @@ def apply_filters(scope)
     end
 
     def split_params(name)
+      return if filter_params[name].blank?
+
       filter_params[name]&.split(',')
```
smoothcontract, line 45: Remove `&`

app/services/v2/people/finder.rb
```diff
@@ -16,6 +16,8 @@ def call
     private
 
       def split_params(name)
+        return if filter_params[name].blank?
+
         filter_params[name]&.split(',')
```
smoothcontract, line 21: Remove `&`

app/services/allocations/finder.rb
```diff
@@ -73,7 +73,9 @@ def apply_date_range_filters(scope)
     end
 
     def split_params(name)
-      filter_params[name]&.split(',')
+      return if filter_params[name].blank?
+
+      filter_params[name].split(',')
```
cwrw, line 78: can't add tests for this as we will never have nil values for locations/status on allocations

app/services/locations/finder.rb
```diff
@@ -40,7 +40,9 @@ def apply_filters(scope)
     end
 
     def split_params(name)
-      filter_params[name]&.split(',')
+      return if filter_params[name].blank?
```
cwrw, line 43: can't add tests here either as those filters are all for ids

smoothcontract (approved):
LGTM 👍

PR #1275, opened 2021-01-15
Supplier metrics

martyn-w:
### Jira link

P4-<TODO>

### What?

I have added/removed/altered:

- [x] Added metrics filtered to suppliers
- [ ] ~Refactored metrics to use class inheritance~

### Why?

I am doing this because:

- so we can calculate more things

- none, not used yet


cwrw (commented):
At a first glance, I would suggest we avoid tying the class directories to the actual suppliers. If we were to add other suppliers or change them in the future this would not scale well, as we'll need to duplicate many more classes. 

I guess this is the PR relating to your question recently around dynamically generating classes vs building them, where we went with avoiding the dynamic approach. Having more context now I'm wondering if there is a middle ground where we don't need to define suppliers behind their own class name/directory or dynamically generate them. Instead we introduce this at the rake task level, where we loop through all suppliers and just rerun the same classes by passing in the supplier scope for moves to the existing ones. Happy to jump on a call for discussion around suggestions, let me know your thoughts

martyn-w:
Hi @cwrw yes, I was also a little worried about explicitly naming code objects after suppliers. Lets have a chat about your rake-task idea after standup.

martyn-w:
Now updated to allow a specific supplier filter to be applied to each move metric, without creating specific supplier-metric-classes

spec/factories/supplier.rb
```diff
@@ -3,5 +3,15 @@
 FactoryBot.define do
   factory :supplier do
     sequence(:name) { |n| "Test Supplier #{n}" }
+
+    trait :geoamey do
```
cwrw, line 7: good to have those I think but we aren't using them in this pr anymore, wonder if we should remove or keep them for future use

martyn-w: I promise to use them in the future :-)

app/models/metrics/moves/counts_by_status_100_day.rb
```diff
@@ -0,0 +1,35 @@
...
+          interval: 6.hours, # we don't need to re-calculate this report very often (and it takes a while to calculate)
+          columns: {
+            name: 'status',
+            field: :itself,
+            values: Move.statuses.values << TOTAL,
+          },
+          rows: {
+            name: 'date',
+            field: :iso8601,
+            values: -> { ((Time.zone.today - 100.days)..(Time.zone.yesterday)).to_a.reverse },
```
cwrw, line 21: I've just been testing this locally, and this is running a query for every day, meaning we will make 100 requests per scope of supplier/moves, could we instead try and use one query for the date? 
An example:
```
moves_geoamey/counts_by_status_100_day/data.csv   (0.8ms)  SELECT COUNT(*) AS count_all, "moves"."status" AS moves_status FROM "moves" WHERE "moves"."supplier_id" = $1 AND "moves"."date" = $2 GROUP BY "moves"."status"  [["supplier_id", "bccf4d77-984d-560c-9ffd-9badbb9157ca"], ["date", "2021-01-26"]]
   (0.7ms)  SELECT COUNT(*) AS count_all, "moves"."status" AS moves_status FROM "moves" WHERE "moves"."supplier_id" = $1 AND "moves"."date" = $2 GROUP BY "moves"."status"  [["supplier_id", "bccf4d77-984d-560c-9ffd-9badbb9157ca"], ["date", "2021-01-25"]]
   (0.6ms)  SELECT COUNT(*) AS count_all, "moves"."status" AS moves_status FROM "moves" WHERE "moves"."supplier_id" = $1 AND "moves"."date" = $2 GROUP BY "moves"."status"  [["supplier_id", "bccf4d77-984d-560c-9ffd-9badbb9157ca"], ["date", "2021-01-24"]]
   (0.7ms)  SELECT COUNT(*) AS count_all, "moves"."status" AS moves_status FROM "moves" WHERE "moves"."supplier_id" = $1 AND "moves"."date" = $2 GROUP BY "moves"."status"  [["supplier_id", "bccf4d77-984d-560c-9ffd-9badbb9157ca"], ["date", "2021-01-23"]]
   (0.9ms)  SELECT COUNT(*) AS count_all, "moves"."status" AS moves_status FROM "moves" WHERE "moves"."supplier_id" = $1 AND "moves"."date" = $2 GROUP BY "moves"."status"  [["supplier_id", "bccf4d77-984d-560c-9ffd-9badbb9157ca"], ["date", "2021-01-22"]]
   (0.6ms)  SELECT COUNT(*) AS count_all, "moves"."status" AS moves_status FROM "moves" WHERE "moves"."supplier_id" = $1 AND "moves"."date" = $2 GROUP BY "moves"."status"  [["supplier_id", "bccf4d77-984d-560c-9ffd-9badbb9157ca"], ["date", "2021-01-21"]]
   ....
```

app/models/metrics/moves/counts_by_move_type.rb
```diff
@@ -2,31 +2,32 @@ module Metrics
...
+          label: 'Move counts by move type',
+          file: 'counts_by_move_type',
+          interval: 5.minutes,
+          columns: {
+            name: 'move_type',
+            field: :itself,
+            values: Move.move_types.values << TOTAL,
+          },
+          rows: {
+            name: COUNT,
```
cwrw, line 19: talked in person but for posterity, in json files we are getting an extra column for counts example: 
```
"data": [
    {
      "count": "count",
      "court_appearance": 16151,
      "court_other": 1,
      "hospital": 4,
      "police_transfer": 3,
      "prison_recall": 1849,
      "prison_remand": 18,
      "prison_transfer": 9944,
      "video_remand": 2,
      "total": 27972
    }
  ]
```

app/models/metrics/moves/counts_by_status_100_day.rb
smoothcontract: A count grouped by date (maybe truncating a date/time if needed) should do that, but might mean dropping to lower level SQL, or perhaps using Arel with `pluck` - that will return essentially a huge array that you can pick through as needed. I use a technique like this for the location free spaces calculation and it's really performant: https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/models/population.rb#L87

app/models/metrics/moves/counts_by_move_type.rb
martyn-w: now fixed 👍

app/models/metrics/moves/counts_by_status_100_day.rb
martyn-w: now improved to use group_by(:status, :date)

app/models/metrics/base_metric.rb
```diff
@@ -84,23 +90,28 @@ def evaluate(var)
...
-              sleep(rand(0..0.1).round(2))
-            else
-              # do calculation a cell at a time
-              columns.each do |column|
-                v[value_key(column, row)] = calculate(column, row)
+        if respond_to?(:calculate_table)
+          @values = calculate_table
+          pause
+        else
+          @values = {}.tap do |v|
```
smoothcontract, line 97: How about moving this inner block of code to a new method named `calculate_individual_rows` - it doesn't look like it would need any additional parameters to be passed in so would be easy to refactor. The main benefit of that is that it makes it much easier to see the overall high level logic within the `calculate_all_values` method at a glance.

app/models/metrics/base_metric.rb
```diff
@@ -124,5 +135,25 @@ def value_key(column, row)
...
+    def build_file(file, supplier)
+      if supplier.present?
+        "#{file}-#{supplier.key}"
+      else
+        file
+      end
+    end
+
+    def pause
+      sleep(rand(0..0.1).round(2))
```
smoothcontract, line 156: Maybe add `if Rails.env.production?` suffix to this method? Seems like there would be no need to artificially slow down development environment or test suite.

app/models/metrics/moves/counts_by_status_100_day.rb
smoothcontract: Nice 👍

lib/tasks/metrics.rake
```diff
@@ -11,22 +11,29 @@ namespace :metrics do
...
+            else
+              puts ' skipped'
+            end
           end
-        end
 
-        # sleep for a short while to avoid overloading the system
-        sleep(rand(0..0.5).round(2))
+          # sleep for a short while to avoid overloading the system
+          sleep(rand(0..0.5).round(2))
```
smoothcontract, line 35: Again, maybe only do this in production environments - perhaps split out the `pause` method so it can be reused across these classes? Then again I suppose this is less crucial as the rake task itself doesn't have coverage I presume?

lib/tasks/metrics.rake
```diff
@@ -11,22 +11,29 @@ namespace :metrics do
       namespace.constants.map { |c| namespace.const_get(c) if namespace.const_get(c).is_a? Class }.compact
     }.flatten
 
+    SUPPLIERS = Supplier.all.to_a << nil
```
smoothcontract, line 14: What's the purpose of including an extra `nil` supplier in the array? Is that to generate aggregate results for all suppliers? If so might be worth a comment to mention that.

smoothcontract (commented):
Looking good, this is much improved now. I left a few more minor suggestions if you want to take a look but nothing drastic at all. 👏

app/models/metrics/base_metric.rb
martyn-w: good point - so I'll add a check for `Rails.env.test?` to not slow down tests, but I think it is necessary to have a very short pause in dev/staging/uat/pre-prod/production environments to allow the database time to recover from each query.

lib/tasks/metrics.rake
martyn-w: Note added 👍

lib/tasks/metrics.rake
martyn-w: the rake task doesn't have coverage, however I've an env check anyway

app/models/metrics/base_metric.rb
martyn-w: now separated into a `calculate_individual_rows` method

smoothcontract (approved):
LGTM 👍

PR #1277, opened 2021-01-18
P4-2588 Add `section`, `responded_by`, and `responded_at` to framework responses

cwrw:

This is part of multiple PRs to help with performance and capturing of user data.

To help simplify the assessment section completion calculation, add a new field `section` to framework responses, which although causes denormalisation in the data, will help with this calculation and will avoid reaching into framework questions.
The section will be added once on response creation and does not need to be edited after. Also add a backfill task to fill in all older section values on legacy responses. This value does not need to be surfaced through the serializer as its only for internal calculation.

I've also added the `responded_by` and `responded_at` values to responses, to start recording who responds to questions and when. This will help with future work with surfacing who has completed different sections. I'm not going to surface those values either yet.

### DIFF
--- a/app/models/framework_response.rb
+++ b/app/models/framework_response.rb
@@ -24,7 +24,7 @@ class FrameworkResponse < VersionedModel
   def update_with_flags!(new_value:, responded_by: nil, responded_at: nil)
     return unless value != new_value
 
-    ApplicationRecord.retriable_transaction { update_response_transaction(new_value, responded_by, responded_at) }
+    ApplicationRecord.retriable_transaction { update_response_transaction(new_value: new_value, responded_by: responded_by, responded_at: responded_at) }
   rescue FiniteMachine::InvalidStateError
     raise ActiveRecord::ReadOnlyRecord, "Can't update framework_responses because assessment is #{assessmentable.status}"
   end
@@ -111,7 +111,7 @@ private
     self.responded = true
   end
 
-  def update_response_transaction(new_value, responded_by, responded_at)
+  def update_response_transaction(new_value:, responded_by:, responded_at:)
     update!(value: new_value, responded_by: responded_by, responded_at: responded_at)
     rebuild_flags!
     FrameworkResponse.clear_dependent_values_and_flags!([self])
--- a/app/models/framework_response.rb
+++ b/app/models/framework_response.rb
@@ -21,10 +21,10 @@ class FrameworkResponse < VersionedModel
 
   after_validation :set_responded_value, on: :update
 
-  def update_with_flags!(new_value)
+  def update_with_flags!(new_value:, responded_by: nil, responded_at: nil)
     return unless value != new_value
 
-    ApplicationRecord.retriable_transaction { update_response_transaction(new_value) }
+    ApplicationRecord.retriable_transaction { update_response_transaction(new_value, responded_by, responded_at) }
   rescue FiniteMachine::InvalidStateError
     raise ActiveRecord::ReadOnlyRecord, "Can't update framework_responses because assessment is #{assessmentable.status}"
   end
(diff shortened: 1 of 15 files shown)


app/services/framework_responses/bulk_updater.rb
```diff
@@ -48,7 +53,7 @@ def build_value_updates_list
 
     def apply_bulk_response_changes(updated_responses)
       # Bulk update all modified response values
-      FrameworkResponse.import(updated_responses, validate: false, on_duplicate_key_update: { conflict_target: [:id], columns: %i[value_text value_json responded] })
+      FrameworkResponse.import(updated_responses, validate: false, on_duplicate_key_update: { conflict_target: [:id], columns: %i[value_text value_json responded responded_by responded_at] })
```
smoothcontract, line 56: Just double checking - `section` is not included here because the section does not change after the framework question gets created?

app/models/framework_response.rb
```diff
@@ -111,8 +111,8 @@ def set_responded_value
     self.responded = true
   end
 
-  def update_response_transaction(new_value)
-    update!(value: new_value)
+  def update_response_transaction(new_value, responded_by, responded_at)
```
smoothcontract, line 114: Maybe use named arguments here for consistency with `update_with_flags` ?

db/schema.rb
```diff
@@ -239,6 +239,9 @@
     t.uuid "assessmentable_id"
     t.string "assessmentable_type"
     t.string "value_type"
+    t.string "section"
```
smoothcontract, line 242: Might be worth adding a DB level null constraint to this (and probably `value_type` too) in a future PR once all of the data has been backfilled?

lib/tasks/refresh_framework_responses.rake
```diff
@@ -12,4 +12,13 @@ namespace :framework_responses do
       FrameworkResponse.import(response_batch.to_a, validate: false, timestamps: false, all_or_none: true, on_duplicate_key_update: %i[value_type])
     end
   end
+
+  desc 'Populate section on framework_responses'
+  task populate_section: :environment do
+    # Updates the response section in batches of 1000
+    FrameworkResponse.where(section: nil).includes(:framework_question).in_batches.each do |response_batch|
+      response_batch.each { |response| response.section = response.framework_question.section }
+      FrameworkResponse.import(response_batch.to_a, validate: false, timestamps: false, all_or_none: true, on_duplicate_key_update: %i[section])
```
smoothcontract, line 21: 👏 for `timestamps: false`

spec/requests/api/framework_responses_controller_bulk_update_spec.rb
```diff
@@ -7,11 +7,14 @@
     include_context 'with supplier with spoofed access token'
 
     subject(:bulk_update_framework_responses) do
+      Timecop.freeze(recorded_timestamp)
```
smoothcontract, line 10: Perhaps use block version instead? It feels more robust somehow than adding the explicit `Timecop.return` line 🤔 

```suggestion
      Timecop.freeze(recorded_timestamp) do
        ...
      end
```

spec/requests/api/framework_responses_controller_update_spec.rb
```diff
@@ -24,9 +24,13 @@
       }
     end
     let(:includes) {}
+    let(:recorded_timestamp) { Time.zone.parse('2020-10-07 01:02:03').iso8601 }
 
     before do
+      Timecop.freeze(recorded_timestamp)
```
smoothcontract, line 30: See above comment

smoothcontract (approved):
Great foundational work for future simplification and performance improvements - awesome stuff. Left some minor comments, happy to go through those and discuss but nothing that would block progress on this 👏

app/services/framework_responses/bulk_updater.rb
cwrw: yes, section is only added on creation of responses, and will never change

db/schema.rb
cwrw: done in https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/1281

PR #1278, opened 2021-01-19
Add new reference data endpoints for frameworks

smoothcontract:
### Jira link

P4-2506

### What?

- [x] Added new `GET /api/reference/frameworks` endpoint to return a list of all frameworks (No include support)
- [x] Added new `GET /api/reference/frameworks/:id` endpoint to return a single framework and all associated questions, dependent questions and flags.
- [x] Updated `README` documentation to cover generation of Swagger documentation during local development.

### Why?

- Suppliers will be able to use these endpoints to get full details of the frameworks and cache as reference data in order to build assessments and create those via existing endpoints (currently id's for questions, flags etc. are provided manually to the suppliers)

- Note that suppliers will need to specify the correct set of `includes` for the `show` endpoint to return all relevant framework data without any N+1 queries i.e. `GET /api/reference/frameworks/:id?include=questions,questions.descendants.**,questions.flags,questions.descendants.flags`. The same data is returned if `questions.descendants.flags` is omitted from the `includes` list but this does trigger a few extra N+1 queries; would be nice to resolve that if possible but not essential 🤔 

### Have you? (optional)

- New API so minimal risk


app/controllers/api/reference/frameworks_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+
+module Api
+  module Reference
+    class FrameworksController < ApiController
+      def index
+        paginate Framework.versioned.all, serializer: FrameworkSerializer
+      end
+
+      def show
+        render_json framework, serializer: FrameworkSerializer, include: included_relationships, status: ok
```
cwrw, line 11:
```suggestion
        render_json framework, serializer: FrameworkSerializer, include: included_relationships, status: :ok
```

app/serializers/framework_question_serializer.rb
```diff
@@ -11,6 +11,7 @@ class FrameworkQuestionSerializer
   attributes :key, :section, :question_type, :options, :response_type
 
   has_many_if_included :descendants, serializer: FrameworkQuestionSerializer, &:dependents
+  has_many_if_included :flags, serializer: FrameworkFlagSerializer, &:framework_flags
```
cwrw, line 14: When calling single assessment endpoints, and including the following: `?include=flags,responses.question.descendants.**` this is causing an n+1 query. It is also returning unnecessary flag data, for those endpoints making them much larger. Maybe there is something we can do with either different serializers or custom relationships?

app/serializers/framework_serializer.rb
```diff
@@ -2,14 +2,18 @@
...
   set_type :frameworks
 
-  has_many :questions, &:framework_questions
+  has_many_if_included :questions, &:framework_questions
 
   attributes :name, :version
 
   SUPPORTED_RELATIONSHIPS = %w[
     questions
+    questions.descendants.**
```
cwrw, line 15: since there might be a bit of rejig to serializers maybe we can include all flags on the question level to avoid having to call descendants as well? (unless this still causes an n+1 after rejigg)

cwrw (commented):
left a few comments, happy to discuss over a call as well

app/serializers/framework_serializer.rb
```diff
@@ -2,14 +2,18 @@
 
 class FrameworkSerializer
   include JSONAPI::Serializer
+  include JSONAPI::ConditionalRelationships
 
   set_type :frameworks
 
-  has_many :questions, &:framework_questions
+  has_many_if_included :questions, &:framework_questions
```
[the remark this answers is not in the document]
smoothcontract: I've refactored this now to use dedicated serializers just for these new endpoints; that solves any unwanted side effects with the existing endpoints.

smoothcontract:
@cwrw I've refactored this now to use dedicated serializers just for these new endpoints; that solves any unwanted side effects with the existing endpoints. Also added missing request spec coverage and updated API documentation, plus some improved README docs as a bonus.

README.md
```diff
@@ -118,6 +118,38 @@ The path to the frameworks folder and the semantic version can be specified.
 
 ## Documentation
 
+### Swagger API documentation
```
cwrw, line 121: 👏

swagger/v1/framework.yaml
```diff
@@ -34,8 +34,6 @@ Framework:
           readOnly: true
     relationships:
       type: object
-      required:
-      - questions
```
smoothcontract, line 38: I had to make this change to support the `GET /reference/frameworks` endpoint; this renders frameworks without questions which then breaks the JSON validation. The other way around that would be to add a new definition for the alternate lightweight version of framework but this felt cleaner 🤔

spec/swagger/swagger_doc_v1.yaml
```diff
@@ -3427,6 +3427,91 @@
...
+      parameters:
+        - "$ref": "../v2/accept_type_parameter.yaml#/Accept"
+        - "$ref": "../v1/content_type_parameter.yaml#/ContentType"
+      responses:
+        "200":
+          description: success
+          content:
+            application/vnd.api+json:
+              schema:
+                "$ref": "../v1/get_frameworks_responses.yaml#/200"
```
cwrw, line 3446: This is returning relationships with questions, maybe its better if this as empty as we won't allow includes on the list endpoint

cwrw (approved):
Looks good, a minor comment on the swagger documentation.
So the only thing we previously discussed is the descendants +flags issue, where we have to include all this: `?include=questions.descendants.**,questions.flags,questions.descendants.flags` to avoid n+1 queries for flags and descendant questions. I think for the future we might have to revisit the `**` part of the includes and work with the FE to remove it as it is no longer needed since we moved towards the new serializer. But otherwise 👍

spec/requests/api/reference/frameworks_controller_show_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+        attributes: {
+          name: framework.name,
+          version: framework.version,
+        },
+        relationships: {},
+      }
+    end
+
+    before do
+      get "/api/v1/reference/frameworks/#{framework_id}", headers: headers, params: params
```
martyn-w, line 31: no need for `v1` in the URL

spec/swagger/swagger_doc_v1.yaml
smoothcontract: Okay, so this means I will need to duplicate `Framework` definition to have a variant of that without the associated questions.

cwrw (approved):

smoothcontract:
> Looks good, a minor comment on the swagger documentation.
> So the only thing we previously discussed is the descendants +flags issue, where we have to include all this: `?include=questions.descendants.**,questions.flags,questions.descendants.flags` to avoid n+1 queries for flags and descendant questions. I think for the future we might have to revisit the `**` part of the includes and work with the FE to remove it as it is no longer needed since we moved towards the new serializer. But otherwise 👍

Noted, will look to address this in future 👍

PR #1280, opened 2021-01-20
P4-2588 Add `section_progress` to assessments

cwrw:

This is the second PR to improve performance. This comes after https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/1277 has moved section to framework responses, and all sections have been backfilled.

Add a new column `section_progress` to both assessment tables, to avoid having to do this calculation each time we invoke the PER serialiser. 

First we add the column, and on creation calculate the section progress, then on recalculate on each PATCH either bulk or single. I have tested all PER endpoints and this does not make the calls any worse, as we are making use of already loaded responses on creation, and on PATCH updates we already call the relevant records to calculate the status, so it won't make things worse.

Also add a backfill task to add sections to all existing assessments. After all this is deployed we can move off calculation in the serialiser and just checking the column value instead.

### DIFF
--- a/app/models/concerns/framework_assessmentable.rb
+++ b/app/models/concerns/framework_assessmentable.rb
@@ -44,6 +44,7 @@ module FrameworkAssessmentable
     ApplicationRecord.retriable_transaction do
       self.prefill_source = previous_assessment
       save!
+
       questions = framework_questions.includes(:dependents).index_by(&:id)
       questions.values.each do |question|
         next unless question.parent_id.nil?
--- a/app/models/concerns/framework_assessmentable.rb
+++ b/app/models/concerns/framework_assessmentable.rb
@@ -44,7 +44,6 @@ module FrameworkAssessmentable
     ApplicationRecord.retriable_transaction do
       self.prefill_source = previous_assessment
       save!
-
       questions = framework_questions.includes(:dependents).index_by(&:id)
       questions.values.each do |question|
         next unless question.parent_id.nil?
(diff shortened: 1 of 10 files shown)


app/models/concerns/framework_assessmentable.rb
```diff
@@ -65,17 +66,18 @@ def import_nomis_mappings!
     FrameworkNomisMappings::Importer.new(assessmentable: self).call
   end
 
-  def section_progress
-    sections_to_responded.group_by(&:section).map do |section, responses|
+  def calculate_section_progress(responses: required_responses)
```
cwrw, line 69: Allow optionally passing in responses, to avoid having to calculate required responses on creation of assessments, which if invoked will load framework questions as well. This is because responses are empty and we don't require to do any calculations on creation.

app/models/concerns/framework_assessmentable.rb
```diff
@@ -125,10 +127,6 @@ def framework_name
     end
   end
 
-  def sections_to_responded
```
cwrw, line 128: we no longer need this as sections are now on responses

spec/support/a_framework_assessment.rb
```diff
@@ -346,23 +355,6 @@
       expect { assessment.build_responses! }.to raise_error(ActiveRecord::RecordInvalid)
     end
 
-    it 'does not persist any responses if there are any invalid assessment records' do
```
cwrw, line 349: this test isn't valid anymore as we will never get a wrong type of response, as we are explicitly defining those values ourselves

app/serializers/framework_assessment_serializer.rb
```diff
@@ -24,7 +24,7 @@ class FrameworkAssessmentSerializer
   attribute :editable, &:editable?
 
   meta do |object|
-    { section_progress: object.section_progress }
+    { section_progress: object.calculate_section_progress }
```
smoothcontract, line 27: I'm a bit confused by this - isn't the calculation already persisted in the new `section_progress` attribute of the framework, so it can just be serialized directly as an attribute?

smoothcontract (commented):
Looks great and should be significantly faster, plus simpler implementation. I did spot one change in the serializer which may be a mistake though...

app/serializers/framework_assessment_serializer.rb
smoothcontract: ...unless you're planning to make this change in a future PR once all the data has been backfilled?

smoothcontract: As expected, you're waaaaay ahead of me and this is covered in https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/1281 😁

smoothcontract (approved):
Awesome 👍

martyn-w (approved):

PR #1281, opened 2021-01-20
P4-2588 Use section progress in serializer

cwrw:

This is the third PR for improving performance.
After all section progress columns are backfilled, use the section progress column instead in assessment serializer and multiple moves PER serializer, to allow section progress to be viewed on FE dashboards. Since responses are created after assessments, explicitly call section calculation in GET specs.

Also add null restrictions and validations to framework response database columns now we ensure they are always populated.

### DIFF
--- a/db/migrate/20210120101859_add_null_fields_to_framework_responses.rb
+++ b/db/migrate/20210120101859_add_null_fields_to_framework_responses.rb
@@ -1,0 +1,6 @@
+class AddNullFieldsToFrameworkResponses < ActiveRecord::Migration[6.0]
+  def change
+    change_column :framework_responses, :section, :string, null: false
+    change_column :framework_responses, :value_type, :string, null: false
+  end
+end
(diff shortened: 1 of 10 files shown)


spec/requests/api/person_escort_records_controller_show_spec.rb
```diff
@@ -19,6 +19,7 @@
     let(:person_escort_record_id) { person_escort_record.id }
 
     before do
+      person_escort_record.update_status_and_progress!
```
cwrw, line 22: since responses are added after creation, force recalculation of section progress

smoothcontract (approved):
Looks good 👍

db/migrate/20210120101859_add_null_fields_to_framework_responses.rb
```diff
@@ -0,0 +1,6 @@
+class AddNullFieldsToFrameworkResponses < ActiveRecord::Migration[6.0]
+  def change
+    change_column :framework_responses, :section, :string, null: false
```
martyn-w, line 3: would be good to check that there are no existing nulls in `framework_responses.section` and `framework_responses.value_type`, otherwise this migration will fail

martyn-w (approved):

PR #1287, opened 2021-01-22
Remove deprecated rake task

cwrw:
Since we longer have the column `person_escort_record_id`, this rake task won't work anymore.


smoothcontract (approved):
LGTM 👍

PR #1288, opened 2021-01-22
Add amended_at timestamp for subsequent updates of a completed PER

smoothcontract:
### Jira link

P4-2607

### What?

- [x] Added `amended_at` timestamp for person escort records
- [x] Include new `amended_at` attribute for serialized person escort records
- [x] Include existing `completed_at` attribute for serialized person escort records and youth risk assessments

### Why?

- This attribute is initially `nil` when the PER is first completed. If the PER is updated and completed again the current time is stored. This allows the front end to distinguish between initial completion and subsequent amendments.

- For now `amended_at` is not implemented for youth risk assessments, so the shared state machine checks whether the relevant model supports `amended_at` - if not then it's ignored.

### Have you? (optional)

- Adds new attributes to existing API endpoints so should be backwards compatible

### DIFF
--- a/app/serializers/framework_assessment_serializer.rb
+++ b/app/serializers/framework_assessment_serializer.rb
@@ -9,7 +9,9 @@ class FrameworkAssessmentSerializer
   has_many_if_included :responses, serializer: FrameworkResponseSerializer, &:framework_responses
   has_many_if_included :flags, serializer: FrameworkFlagSerializer, &:framework_flags
 
-  attributes :confirmed_at, :created_at, :nomis_sync_status
+  attributes :completed_at, :confirmed_at, :created_at, :nomis_sync_status
+
+  attribute :amended_at, if: proc { |object| object.respond_to?(:amended_at) }
 
   attribute :version do |object|
     object.framework.version
--- a/app/serializers/framework_assessment_serializer.rb
+++ b/app/serializers/framework_assessment_serializer.rb
@@ -11,8 +11,6 @@ class FrameworkAssessmentSerializer
 
   attributes :completed_at, :confirmed_at, :created_at, :nomis_sync_status
 
-  attribute :amended_at, if: proc { |object| object.respond_to?(:amended_at) }
-
   attribute :version do |object|
     object.framework.version
   end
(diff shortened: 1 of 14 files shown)


app/serializers/framework_assessment_serializer.rb
```diff
@@ -9,7 +9,9 @@ class FrameworkAssessmentSerializer
   has_many_if_included :responses, serializer: FrameworkResponseSerializer, &:framework_responses
   has_many_if_included :flags, serializer: FrameworkFlagSerializer, &:framework_flags
 
-  attributes :confirmed_at, :created_at, :nomis_sync_status
+  attributes :completed_at, :confirmed_at, :created_at, :nomis_sync_status
+
+  attribute :amended_at, if: proc { |object| object.respond_to?(:amended_at) }
```
cwrw, line 14: instead of here, can we put this in the dedicated PER and PERs serialisers to avoid the responds to? https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/serializers/person_escort_record_serializer.rb and https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/master/app/serializers/person_escort_records_serializer.rb
Also could you add the `completed_at` to the PERs serialiser as it doesn't inherit from the framework assessment serialiser

cwrw (commented):
Looks good, just the comment around moving the attribute to the dedicated serialisers, and making sure all PER serialisers have the same attributes

app/serializers/framework_assessment_serializer.rb
smoothcontract: Good shout - done 👍

spec/serializers/framework_assessment_serializer_spec.rb
```diff
@@ -5,7 +5,8 @@
 RSpec.describe FrameworkAssessmentSerializer do
   subject(:serializer) { described_class.new(assessment, includes) }
 
-  let(:move) { create(:move) }
+  let(:location) { create(:location) }
```
cwrw, line 8: why do we need define a specific location here?

cwrw (approved):
looks good 👍

spec/serializers/framework_assessment_serializer_spec.rb
smoothcontract: It's to circumvent a validation rule when creating a youth risk assessment - but actually I don't need that now as I removed the specs for the `amended_at` attribute here. Good spot, I'll clean this up.

cwrw (approved):

PR #1291, opened 2021-01-25
Add a rake task which deletes moves and allocations on staging

Nimphal:
### Jira link

P4-XXXX

### What?

I have added/removed/altered:

- [x] Added a rake task which removes all moves and allocations in the database. This task can only be run on non-production environments.
### Why?

I am doing this because:

- The frontend e2e tests create a lot of data which leads to API timeouts

### DIFF
--- a/lib/tasks/data_maintenance.rake
+++ b/lib/tasks/data_maintenance.rake
@@ -76,7 +76,7 @@ namespace :data_maintenance do
       moves.destroy_all
       allocations.destroy_all
     else
-      "You are trying to run this on a production-like environment, which is not allowed."
+      puts "You are trying to run this on a production-like environment, which is not allowed."
     end
   end
 end
--- a/lib/tasks/data_maintenance.rake
+++ b/lib/tasks/data_maintenance.rake
@@ -62,4 +62,21 @@ namespace :data_maintenance do
     GenericEvent.where(type: incident_event_types).update_all(classification: 'incident')
     GenericEvent.where(type: notification_event_types).update_all(classification: 'notification')
   end
+
+  desc 'remove moves and allocations on staging'
+  task remove_data_staging: :environment do
+    # NB: this task should never run on production or pre-production
+    is_not_production = Rails.env.development? || ENV.fetch('HOSTNAME', 'UNKNOWN') =~ /(\-(dev|staging|uat)\-)/i
+
+    if is_not_production
+      moves = Move.all
+      puts "Number of moves to be deleted: #{moves.count}"
+      allocations = Allocation.all
+      puts "Number of allocations to be deleted: #{allocations.count}"
+      moves.destroy_all
+      allocations.destroy_all
+    else
+      "You are trying to run this on a production-like environment, which is not allowed."
+    end
+  end
 end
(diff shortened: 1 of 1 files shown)


lib/tasks/data_maintenance.rake
```diff
@@ -62,4 +62,21 @@ namespace :data_maintenance do
...
+  task remove_data_staging: :environment do
+    # NB: this task should never run on production or pre-production
+    is_not_production = Rails.env.development? || ENV.fetch('HOSTNAME', 'UNKNOWN') =~ /(\-(dev|staging|uat)\-)/i
+
+    if is_not_production
+      moves = Move.all
+      puts "Number of moves to be deleted: #{moves.count}"
+      allocations = Allocation.all
+      puts "Number of allocations to be deleted: #{allocations.count}"
+      moves.destroy_all
```
cwrw, line 76: I think this won't work as we would get `ActiveRecord::InvalidForeignKey` if anything is associated to moves, for instance assessments or journeys etc.

Nimphal: Ah, would it be safer to do a delete_all then? I took the destroy all from a task further up this file which uses it moves, but perhaps have missed some context.

cwrw (commented):
I wonder if there is a bigger question here around data on staging. We've been currently using the large amount of data to flush out any performance issues and to test numbers of dashboards, as well as any other tests relating to suppliers and new features. But this in turn is breaking the FE e2e tests because of timeouts. Something that has been discussed before, but would require a bit more work is I wonder if we allow the FE e2e tests to run on another ENV where it is safe to delete all the data without disrupting the staging environment for both suppliers and any testing we might be doing. We can add test data to staging ourselves through fake data rake tasks/things we create through the interface. Ideally though we can copy some obfuscated real data here, but again that could take a lot of time to get to. Let me know your thoughts.

Nimphal:
The frontend is getting to the point where it needs to start looking at mocking some responses as it's not really possible to test all scenarios using staging since it holds state. That's no

smoothcontract:
> I wonder if there is a bigger question here around data on staging. We've been currently using the large amount of data to flush out any performance issues and to test numbers of dashboards, as well as any other tests relating to suppliers and new features. But this in turn is breaking the FE e2e tests because of timeouts. Something that has been discussed before, but would require a bit more work is I wonder if we allow the FE e2e tests to run on another ENV where it is safe to delete all the data without disrupting the staging environment for both suppliers and any testing we might be doing. We can add test data to staging ourselves through fake data rake tasks/things we create through the interface. Ideally though we can copy some obfuscated real data here, but again that could take a lot of time to get to. Let me know your thoughts.

100% agree - and the currently under utilised dev environment is pretty much ideal for this use case. I'd be much happier to see the E2E tests moving to run against that environment only - leaving staging data for the more extensive load and supplier testing, and for this PR to be amended so that it only permits running against local development or the cloud platform `dev` environment.

lib/tasks/data_maintenance.rake
cwrw: I tried running both `destroy_all` and `delete_all` locally to see which one ignores callbacks, as `delete_all` might be the one without callbacks, but they both threw the error because it had a youth risk assessment associated to it 🤔

PR #1298, opened 2021-01-28
Add rake task to backfill moves on legacy PERs

cwrw:
Before attaching a move to a PER, we previously used to attach a profile, which can be linked to multiple moves. Backfill legacy PERs with moves if possible, if there are multiple moves ignore those as they will be attached manually.

### DIFF
--- a/lib/tasks/refresh_assessments.rake
+++ b/lib/tasks/refresh_assessments.rake
@@ -9,6 +9,22 @@ namespace :assessments do
       update_section_progress(batch, YouthRiskAssessment)
     end
   end
+
+  desc 'Populate moves on assessments'
+  task populate_moves: :environment do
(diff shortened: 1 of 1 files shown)


lib/tasks/refresh_assessments.rake
```diff
@@ -9,6 +9,22 @@ namespace :assessments do
       update_section_progress(batch, YouthRiskAssessment)
     end
   end
+
+  desc 'Populate moves on assessments'
+  task populate_moves: :environment do
+    PersonEscortRecord.where(move: nil).in_batches(of: 10).each do |batch|
```
cwrw, line 15: use a small batch as larger ones were problematic on production

smoothcontract (approved):
Looks good, hopefully there aren't too many moves on production that need manual assignment 🤞

PR #1300, opened 2021-01-28
Enforce a move on a PER

cwrw:
After backfilling all legacy PERs to have moves, enforce a move relationship on a PER. Remove legacy logic which used to support a move existing and not existing.


smoothcontract (approved):
All looks good, nice cleanup 👍

cwrw:
need to study PER/move model before implementing this change, especially around detaching profiles from moves with existing PERs.

PR #1304, opened 2021-01-29
Add support for new handover_details JSONB attribute when confirming PER

smoothcontract:
### Jira link

P4-2625

### What?

- [x] Add new `handover_details` attribute to PER
- [x] Support new attribute for `PATCH /person_escort_records/:id` endpoint
- [x] Include new attribute in PER serializers (defaults to an empty hash)

### Why?

- The initial spike for this work proposed defining explicit attributes for each of the various handover attributes e.g. `handover_recipient` but already we have seen some churn in the requirements and expect this to continue. To provide future flexibility this instead supports a single `handover_details` attribute which will persist any key/value hash the front end wishes to use.

- Currently we only require handover details for PER's so this includes some refactoring to support the diverging parameter support when patching the PER.

### Have you? (optional)

- Changes an api that is used in production: new attribute is optional when updating a PER so is backwards compatible, but we should probably let suppliers know about it.

### DIFF
--- a/app/controllers/api/framework_assessments_controller.rb
+++ b/app/controllers/api/framework_assessments_controller.rb
@@ -4,13 +4,13 @@ module Api
   class FrameworkAssessmentsController < ApiController
     after_action :send_notification, only: :update
 
-    NEW_PERMITTED_PER_PARAMS = [
+    NEW_ASSESSMENT_PERMITTED_PARAMS = [
       :type,
       attributes: [:version],
       relationships: [move: {}],
     ].freeze
 
-    UPDATE_PERMITTED_PER_PARAMS = [
+    UPDATE_ASSESSMENT_PERMITTED_PARAMS = [
       :type,
       attributes: [:status],
     ].freeze
@@ -26,7 +26,7 @@ module Api
 
     def update
       FrameworkAssessments::ParamsValidator.new(update_assessment_status).validate!
-      assessment.confirm!(update_assessment_status)
+      assessment.confirm!(update_assessment_status, handover_details)
 
       render_assessment(assessment, :ok)
     end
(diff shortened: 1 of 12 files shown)


app/controllers/api/framework_assessments_controller.rb
```diff
@@ -38,17 +38,21 @@ def show
...
     def update_assessment_params
-      params.require(:data).permit(UPDATE_PERMITTED_PER_PARAMS)
+      params.require(:data).permit(UPDATE_ASSESSMENT_PERMITTED_PARAMS)
     end
 
     def update_assessment_status
       update_assessment_params.to_h.dig(:attributes, :status)
     end
 
+    def handover_details
```
cwrw, line 52: This feels like it should be in the PER controller as it's very specific to that model only. I was thinking that we could pass in the update params to the `confirm!` method instead of just the status as so:
```
assessment.confirm!(update_assessment_details)
```
meaning we just define our custom `update_assessment_params` in the PER controller to permit handover fields. For the confirm method it could look like this: 
```
def confirm!(details)
    return unless details[:status] == ASSESSMENT_CONFIRMED

    state_machine.confirm!
    self.handover_details = details[:handover_details]if details[:handover_details].present? && respond_to?(:handover_details)
   ...
```
or something similar, let me know your thoughts

cwrw (commented):
Spoke in person but for posterity: this is just a first iteration of storing handover fields on the PER, as those aren't yet fully set in stone. I think most fields should be ok within the jsonb field, however I think the handover time might need to live outside when the time comes as it is a datetime field which requires special formatting

app/controllers/api/framework_assessments_controller.rb
smoothcontract: I don't mind that, but the word `details` seems very repetitive. I'll try a variation of this suggestion to move the PER specific stuff out of the generic assessment controller 🤔

smoothcontract:
> Spoke in person but for posterity: this is just a first iteration of storing handover fields on the PER, as those aren't yet fully set in stone. I think most fields should be ok within the jsonb field, however I think the handover time might need to live outside when the time comes as it is a datetime field which requires special formatting

Yep, in the near future I might also add `handover_confirmed_at` as an independent `DateTime` attribute as a follow up PR if the requirements don't change around that.

cwrw (approved):

PR #1305, opened 2021-02-01
Add new handover_occurred_at attribute to PER

smoothcontract:
### Jira link

P4-2625

### What?

- [x] Add new `handover_occurred_at` timestamp attribute to person escort records

### Why?

- It's better if this is set explicitly rather than passed as part of the `handover_details` JSON attribute, as this needs to be incorporated into event timestamps in the near future. It's also easier to work with as a strongly typed timestamp rather than having to parse out of the other JSON details.

### Have you? (optional)

- Changes an api that is used in production; additional attribute is backwards compatible with existing API clients.


martyn-w (approved):
👍 LGTM

PR #1309, opened 2021-02-03
Add new PerHandover event

smoothcontract:
### Jira link

P4-2591

### What?

- [x] Create a `PerHandover` event when confirming PER with handover details

### Why?

- This is created instead of the PerConfirmation event when a PER is confirmed with the new `handover_details` and `handover_occurred_at` attributes populated; this will be from the updated frontend including the new PER handover screen. 
- We want to use the new PerHandover event when the handover process is used, but retain the original PerConfirmation event for the legacy PER's. The new PerHandover event includes all of the `handover_details` in the `details` section and the `handover_occurred_at` timestamp in the `occurred_at` attribute.
- The new event looks like this when serialized back to the API client:

```json
{
  "data": {
    "id": "3251297f-ec8d-4d5c-9b59-4391f5fec650",
    "type": "events",
    "attributes": {
      "event_type": "PerHandover",
      "classification": "default",
      "occurred_at": "2021-02-03T17:00:46+00:00",
      "recorded_at": "2021-02-03T17:00:46+00:00",
      "notes": "Automatically generated event",
      "created_by": "Foomanchoo",
      "details": {
        "recipient_id": "12345",
        "recipient_name": "Fulton McKay",
        "recipient_contact_number": "01-811-8055"
      }
    },
    "relationships": {
      "eventable": {
        "data": {
          "id": "fe49dd5e-d783-476e-9e67-51cdb630e948",
          "type": "person_escort_records"
        }
      },
      "supplier": {
        "data": null
      }
    }
  }
}
```

- Changes an api that is used in production; only change is a different event being created when confirming the PER so is backwards compatible. The new event will also be documented in the Wiki.

### DIFF
--- a/app/models/generic_event/per_handover.rb
+++ b/app/models/generic_event/per_handover.rb
@@ -1,0 +1,5 @@
+class GenericEvent
+  class PerHandover < GenericEvent
+    include PersonEscortRecordEventValidations
+  end
+end
--- a/app/controllers/api/person_escort_records_controller.rb
+++ b/app/controllers/api/person_escort_records_controller.rb
@@ -2,8 +2,6 @@
 
 module Api
   class PersonEscortRecordsController < FrameworkAssessmentsController
-    after_action :create_confirmation_event, only: :update # rubocop:disable LexicallyScopedActionFilter
-
     UPDATE_PER_PERMITTED_PARAMS = [
       :type,
       attributes: [:status, :handover_occurred_at, handover_details: {}],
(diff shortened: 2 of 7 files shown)


app/models/generic_event/per_handover.rb
```diff
@@ -0,0 +1,5 @@
+class GenericEvent
+  class PerHandover < GenericEvent
+    include PersonEscortRecordEventValidations
```
martyn-w, line 3: do the details attributes need to be specified too? - i.e.. 
```
{
          'status': status,
          'handover_details': handover_details,
          'handover_occurred_at': occurred_timestamp.iso8601,
}
```

smoothcontract: No, not in this case. We decided to keep the choice of `handover_details` in the front end only to allow scope for these to vary over time. So the `details` in the event only contains the `handover_details` from the PER. We don't need a status attribute in there and the `handover_occurred_at` timestamp is stored in the event `occurred_at` attribute. As we're not enforcing strong typing in the API and support any JSON structure for `handover_details` there's no need to explicitly list the details here.

martyn-w: 👍  jolly good

martyn-w:
👍  LGTM - please add the new event to https://github.com/ministryofjustice/hmpps-book-secure-move-api/wiki/Person-Escort-Record-Events

martyn-w (approved):

PR #1310, opened 2021-02-03
P4-2643 add MoveDateChanged event

martyn-w:
### Jira link

P4-2643

### What?

I have added/removed/altered:

- [x] Added MoveDateChange auto-event which is raised when a move is patched with a new date

### Why?

I am doing this because:

- so that the Hub can better track when a move's date has been changed, as it is important to CDI and MI reports

### Have you? (optional)

- [ ] TODO: Updated wiki

- minimal


app/controllers/api/v1/moves_actions.rb
```diff
@@ -139,5 +141,19 @@ def active_record_relationships
     def supported_relationships
       MoveSerializer::SUPPORTED_RELATIONSHIPS
     end
+
+    def create_auto_event(move, event_class, details = {})
```
smoothcontract, line 145: Would be nice to move this method to the moves controller, as it appears to be exactly the same in both of these modules so would be better to just define it once.

smoothcontract (commented):
Looks good, just one suggestion to DRY the `create_auto_event` method.

app/controllers/api/v1/moves_actions.rb
smoothcontract: @martyn-w After further thought, I'm tempted to move this to the api controller as there is very similar code for the person escort record controller too. I'm happy to do that work if you want to leave this as is for now.

smoothcontract (approved):

PR #1312, opened 2021-02-04
DRY up various implementations of automatic event creation

smoothcontract:
### Jira link

P4-2591

### What?

- [x] Add new `create_automatic_event!` method and use in various controllers to create associated events

### Why?

- This happens in a few controllers now, so it's good to standardise the implementation of this with a single method with sufficient flexibility to work for all current use cases. Event creation already has good existing coverage to ensure the correct attributes are populated so we can be confident this is working as expected.


martyn-w (approved):
👍 LGTM

PR #1324, opened 2021-02-09
Add webhook and email notifications for PER amendments

smoothcontract:
### Jira link

P4-2627

### What?

- [x] Lots and lots of refactoring (see below) to support new notifications when completed PER's are amended.

### Why?

- This requires a new Gov notify template as the email content is different than for moves, and updates to the webhook serializer as we want to send notifications (including a link) related directly to the PER. This presents some interesting challenges as there is a lot of shared behaviour for both notifications and we also need to be careful not to break and notification jobs that may be currently queued in sidekiq, so cannot change any method signatures or underlying behaviour.

- The approach I've taken is to add a new mailer class for the PER, but refactored the existing MoveMailer to extract common personalisation attributes to a new BaseMailer. The new PER mailer then needs to be called rather than the MoveMailer from the Notification class; that already has support for polymorphic attachment to any model, but was hardcoded to always use the MoveMailer. That has been refactored to support invoking the appropriate mailer based on the notifiable topic (either a Move or PersonEscortRecord).

- Notification jobs for all record types are ultimately delegated to PrepareMoveNotificationsJob - this too makes the assumption that the underlying topic is always a move. The new PreparePersonEscortRecordNotificationsJob descends from that class and overrides two key methods (without changing the method signature for backwards sidekiq compatibility) to resolve the correct PER and the move associated with the PER.

- The Notifier service has been updated to use the new PER job only for the case of amendments to a completed PER. For other assessment changes (PER or Youth Risk) the original PrepareAssessmentNotificationsJob is retained (which ultimately delegates to a move notification). This is slightly confusing and perhaps something that will be changed over time, but I didn't want to introduce a breaking change to the existing PER webhooks and potentially break any queued sidekiq jobs.

- Finally the FrameworkResponsesController is updated to invoke notifications for PER amendments (i.e. changes to the responses for a completed PER) - we can detect these when the PER has a populated `amended_at` attribute. Youth Risk Assessments do not have this attribute but share the same controller for updating responses.

- In a follow up PR the legacy ENV var for GOVUK_NOTIFY_TEMPLATE_ID will be removed (we have two templates now so it makes sense to rename this). The deploy repo already populates correct values for the two new ENV vars - see https://github.com/ministryofjustice/hmpps-book-secure-move-api-deploy/pull/127

### Have you? (optional)

- [x] Added environment variables to [deployment repository](https://github.com/ministryofjustice/hmpps-book-secure-move-api-deploy)

- Changes an api that is used in production; new emails and webhooks will be sent to suppliers when PER's are amended so they will need some advance warning of this. Flagging as a breaking change for this reason.

### DIFF
--- a/app/jobs/prepare_move_notifications_job.rb
+++ b/app/jobs/prepare_move_notifications_job.rb
@@ -5,7 +5,8 @@ class PrepareMoveNotificationsJob < ApplicationJob
   include QueueDeterminer
 
   def perform(topic_id:, action_name:, queue_as:, send_webhooks: true, send_emails: true, only_supplier_id: nil)
-    move = Move.find(topic_id)
+    topic = find_topic(topic_id)
+    move = associated_move(topic)
 
     # if the move has a specified supplier, use it; otherwise use the move.suppliers delegate based on from_location
     [move.supplier || move.suppliers].flatten.each do |supplier|
@@ -17,7 +18,7 @@ class PrepareMoveNotificationsJob < ApplicationJob
         # NB: always notify the webhook (if defined) on any change, even for back-dated historic moves
         if send_webhooks && subscription.callback_url.present?
           NotifyWebhookJob.perform_later(
-            notification_id: build_notification(subscription, NotificationType::WEBHOOK, move, action_name).id,
+            notification_id: build_notification(subscription, NotificationType::WEBHOOK, topic, action_name).id,
             queue_as: queue_as, # send webhook with same priority as move
           )
         end
@@ -26,7 +27,7 @@ class PrepareMoveNotificationsJob < ApplicationJob
         next unless send_emails && subscription.email_address.present? && should_email?(move)
 
         NotifyEmailJob.perform_later(
-          notification_id: build_notification(subscription, NotificationType::EMAIL, move, action_name).id,
+          notification_id: build_notification(subscription, NotificationType::EMAIL, topic, action_name).id,
           queue_as: queue_as, # send email with same priority as move
         )
       end
@@ -34,6 +35,14 @@ class PrepareMoveNotificationsJob < ApplicationJob
   end
 
 private
+
+  def find_topic(topic_id)
+    Move.find(topic_id)
+  end
+
+  def associated_move(topic)
+    topic
+  end
 
   def build_notification(subscription, type_id, topic, action_name)
     subscription.notifications.create!(
--- a/app/jobs/prepare_move_notifications_job.rb
+++ b/app/jobs/prepare_move_notifications_job.rb
@@ -1,39 +1,7 @@
 # frozen_string_literal: true
 
 # This job is responsible for preparing a set of notify jobs to run
-class PrepareMoveNotificationsJob < ApplicationJob
-  include QueueDeterminer
-
-  def perform(topic_id:, action_name:, queue_as:, send_webhooks: true, send_emails: true, only_supplier_id: nil)
-    topic = find_topic(topic_id)
-    move = associated_move(topic)
-
-    # if the move has a specified supplier, use it; otherwise use the move.suppliers delegate based on from_location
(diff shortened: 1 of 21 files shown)


app/models/notification.rb
```diff
@@ -39,4 +39,12 @@ def kept?
   def for_feed
     attributes.slice(*FEED_ATTRIBUTES)
   end
+
+  def mailer
+    if topic.is_a?(PersonEscortRecord)
```
martyn-w, line 44: do we also need to consider YouthRiskAssessment? we send a `confirm_youth_risk_assessment` webhook and `confirm_youth_risk_assessment` email

smoothcontract: It's only possible to create a notification with a topic of PER or Move - for everything else the topic ends up being delegated to the relevant move.

martyn-w: 👍

app/jobs/prepare_move_notifications_job.rb
```diff
@@ -5,7 +5,8 @@ class PrepareMoveNotificationsJob < ApplicationJob
   include QueueDeterminer
 
   def perform(topic_id:, action_name:, queue_as:, send_webhooks: true, send_emails: true, only_supplier_id: nil)
-    move = Move.find(topic_id)
+    topic = find_topic(topic_id)
```
martyn-w, line 8: Now that this class's focus is the `topic` rather than the `move`, it feels misnamed... how about a new base class `PrepareNotificationsJob` from which `PrepareMoveNotificationsJob` and `PreparePersonEscortRecordNotificationsJob` can inherit?

good work not changing the method signature, as that will break on deployment if there are existing notifications for processing 👍

app/jobs/prepare_person_escort_record_notifications_job.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+# This job is responsible for preparing a set of notify jobs to run
+class PreparePersonEscortRecordNotificationsJob < PrepareMoveNotificationsJob
+private
+
+  def find_topic(topic_id)
```
martyn-w, line 7: ah, these new methods make sense now 👍

spec/requests/api/framework_responses_controller_update_spec.rb
```diff
@@ -6,6 +6,10 @@
   describe 'PATCH /framework_responses/:framework_response_id' do
     include_context 'with supplier with spoofed access token'
 
+    subject(:patch_response) do
+      patch "/api/v1/framework_responses/#{framework_response_id}?include=#{includes}", params: framework_response_params, headers: headers, as: :json
```
martyn-w, line 10: no `v1` required in the URL - we want to deprecate those

app/serializers/notification_serializer.rb
```diff
@@ -10,11 +9,11 @@ class NotificationSerializer
 
   attribute :timestamp, &:created_at
 
-  belongs_to :topic, polymorphic: true, key: :move, if: proc { |object| move?(object) }, links: {
+  belongs_to :move, id_method_name: :topic_id, if: proc { |object| object.topic.is_a?(Move) }, links: {
```
martyn-w, line 12: much cleaner 👍

martyn-w:
There is a lot going on in this PR but it looks pretty good to me 👍 

I do think it would be a bit neater to refactor `PrepareMoveNotificationsJob` into a base class from which `PrepareMoveNotificationsJob` and `PreparePersonEscortRecordNotificationsJob ` can inherit.

spec/requests/api/framework_responses_controller_update_spec.rb
smoothcontract: Done

martyn-w (approved):
👍  excellent work in generalising this

PR #1331, opened 2021-02-10
Include `updated_at` attribute for serialized PERs and Youth risk assessments

smoothcontract:
### Jira link

P4-2659

### What?

- [x] Include `updated_at` attribute for serialized PERs and Youth risk assessments

### Why?

- This will be used in the front end to show last updated at timestamp prior to other more relevant timestamps being shown once the assessment is completed.

### Have you? (optional)

- Changes an api that is used in production; adds a new attribute to existing PER and Youth risk assessment endpoints but is backwards compatible with existing clients.


martyn-w (approved):
👍

PR #1340, opened 2021-02-15
Add new webhook (no email) for PER handover event

smoothcontract:
### Jira link

P4-2560

### What?

- [x] Send a dedicated webhook for PER handover

### Why?

- This is sent rather than PER confirmation when the update PER endpoint includes handover attributes (i.e. has been called from the updated front end handover page). This means a change in behaviour as the previous PER confirmation webhook and email will not be sent. Youth Risk assessment behaviour is unchanged.

- Suppliers need to be ready to receive the new webhook; if not then we can disable individual webhooks with an environment variable

### DIFF
--- a/app/controllers/api/person_escort_records_controller.rb
+++ b/app/controllers/api/person_escort_records_controller.rb
@@ -37,6 +37,7 @@ module Api
 
     def create_confirmation_event_and_notification!
       create_automatic_event!(eventable: assessment, event_class: GenericEvent::PerConfirmation, details: { confirmed_at: assessment.confirmed_at.iso8601 })
+      # TODO: Remove derivation of action_name 'confirm_person_escort_record' within PrepareAssessmentNotificationsJob and pass explicitly
       Notifier.prepare_notifications(topic: assessment, action_name: nil)
     end
 
--- a/app/controllers/api/person_escort_records_controller.rb
+++ b/app/controllers/api/person_escort_records_controller.rb
@@ -21,9 +21,9 @@ module Api
       assessment.confirm!(update_assessment_status, handover_details, handover_occurred_at)
 
       if handover_details.present? && handover_occurred_at.present?
-        create_handover_event!
+        create_handover_event_and_notification!
       else
-        create_confirmation_event!
+        create_confirmation_event_and_notification!
       end
     end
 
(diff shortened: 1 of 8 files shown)


app/controllers/api/person_escort_records_controller.rb
```diff
@@ -35,12 +35,14 @@ def assessment_serializer
       PersonEscortRecordSerializer
     end
 
-    def create_confirmation_event!
+    def create_confirmation_event_and_notification!
       create_automatic_event!(eventable: assessment, event_class: GenericEvent::PerConfirmation, details: { confirmed_at: assessment.confirmed_at.iso8601 })
+      Notifier.prepare_notifications(topic: assessment, action_name: nil)
```
martyn-w, line 40: could you give `action_name`  a value rather than be nil - its useful to see the reason for the notification and it is required in some of the emails

martyn-w: ah ok, I see the action_name is being set further down the chain in `PrepareAssessmentNotificationsJob`. I guess there is a reason for setting action_name there, rather than up here?

smoothcontract: Hmm... this isn't trivial to change. Previously the behaviour was shared across both assessment types and the `action_name` parameter was passed as simply `update`. I've refactored as the behaviour for each assessment has diverged but essentially for the existing notifications the same behaviour is preserved. I did find that passing a string value that was effectively ignored within the `Notifier` class was a little misleading, but as `action_name` is a required parameter I changed this to `nil`.

It would be possible to pass the `action_name` value as either `confirm_person_escort_record` or `confirm_youth_risk_assessment` rather than deriving this within `PrepareAssessmentNotificationsJob` but that would mean changing the job `perform` signature and potentially breaking any current queued jobs. So that in turn means adding a new class definition for the new method signature, deploying that and in a further PR removing the old job once we're sure there are no jobs queued.

All of which feels beyond the scope of this PR unfortunately so for now I've just documented with a `TODO` to resolve in future.

```ruby
class PrepareAssessmentNotificationsJob < ApplicationJob
  include QueueDeterminer

  def perform(topic_id:, topic_class:, **_)
    assessment = topic_class.constantize.find(topic_id)
    return unless (move = assessment.move)

    PrepareMoveNotificationsJob.perform_now(topic_id: move.id, action_name: "confirm_#{topic_class.underscore}", queue_as: move_queue_priority(move))
  end
end
```

class PrepareAssessmentNotificationsJob < ApplicationJob
  include QueueDeterminer

  def perform(topic_id:, topic_class:, **_)
    assessment = topic_class.constantize.find(topic_id)
    return unless (move = assessment.move)

    PrepareMoveNotificationsJob.perform_now(topic_id: move.id, action_name: "confirm_#{topic_class.underscore}", queue_as: move_queue_priority(move))
  end
end

martyn-w: ok, understood, thanks 👍

martyn-w (approved):
👍 LGTM

PR #1344, opened 2021-02-17
Quick and dirty hack to reinstate PER confirmation Webhook for GeoAmey

smoothcontract:
### Jira link

P4-2560

### What?

- [x] Temporarily send legacy PER confirmation webhook for PER handover

### Why?

- This is a temporary measure to support GeoAmey integration with PER Webhooks - they're not currently able to process the new PER handover webhook. This will be reverted as soon as they are able to process the new webhook.


martyn-w (approved):
👍 LGTM

PR #1346, opened 2021-02-18
Raise an error when creating/redirecting move to/from inactive location

smoothcontract:
### Jira link

P4-2636

### What?

- [x] Updated `MoveTypeValidator` to ensure that from/to location are active

### Why?

- To prevent API clients from creating or redirecting moves to locations which are no longer active in Nomis (i.e. `disabled_at` is not `nil`). Updated MoveTypeValidator to include new rules and also add an explicit code for each error for easier localisation within the front end (message shown to the user can be customised based on the error `code`).

- Leveraged existing discard gem for this - which adds handy mixin scopes and methods rather than directly working with `disabled_at` column. We should look at making further use of that in AssessmentQuestion, Ethnicity, Gender, IdentifierType, Nationality and PrisonTransferReason models, and perhaps standardise to rename the column to `discarded_at` for consistency with Document, Notification and Subscription models.

- Changes an api that is used in production: adds a new business rule which means clients will see additional error messages

### DIFF
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
 class Location < ApplicationRecord
+  include Discard::Model
+  # TODO: rename disabled_at column to discarded_at and remove this line
+  self.discard_column = :disabled_at
+
   LOCATION_TYPE_COURT = 'court'
   LOCATION_TYPE_POLICE = 'police'
   LOCATION_TYPE_PRISON = 'prison'
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -12,10 +12,6 @@ class Subscription < ApplicationRecord
   validates :email_address, format: { with: URI::MailTo::EMAIL_REGEXP, allow_nil: true }, presence: false
   validate :email_address_or_callback_url_required
 
-  def kept?
-    !discarded?
-  end
-
   def secret=(value)
     self[:encrypted_secret] = Encryptor.encrypt(value)
   end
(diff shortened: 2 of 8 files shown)


app/models/subscription.rb
```diff
@@ -12,10 +12,6 @@ class Subscription < ApplicationRecord
   validates :email_address, format: { with: URI::MailTo::EMAIL_REGEXP, allow_nil: true }, presence: false
   validate :email_address_or_callback_url_required
 
-  def kept?
-    !discarded?
-  end
-
```
smoothcontract, line 18: This is superfluous and is provided by the Discard gem: https://github.com/jhawthorn/discard/blob/master/lib/discard/model.rb#L113

martyn-w: good spot 👍

app/models/location.rb
```diff
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
 class Location < ApplicationRecord
+  include Discard::Model
+  # TODO: rename disabled_at column to discarded_at and remove this line
```
martyn-w, line 5: not sure we should rename the column - suppliers and the FE are expecting an attribute named "disabled_at" and they would need to update their systems if we renamed it to "discarded_at"

martyn-w (approved):
👍 LGTM - but see note about renaming `disabled_at`

app/models/location.rb
smoothcontract: The column could be renamed without changing the serializer, we do that a lot, it just would be nice to be more idiomatic with the discard gem. I'm not sure why but there are a few models that use the gem and a lot more that have "roll your own" soft delete functionality using a `disabled_at` column. I think it would be nice in future to standardise all of that to use the gem.

martyn-w: 👍

PR #1349, opened 2021-02-22
Add a new webhook when an assessment is completed

smoothcontract:
### Jira link

P4-2670

### What?

- [x] Add new webhook notifications (emails not supported) for `complete_person_escort_record` and `complete_youth_risk_assessment`
- [x] Remove temporary workarounds with notifications introduced for GeoAmey
- [x] Amend existing `confirm_youth_risk_assessment` webhook (also removing email support) to refer to the associated youth risk assessment rather than the move
- [x] Remove support for deprecated `confirm_person_escort_record` notification as this is replaced by existing `handover_person_escort_record`

### Why?

- This now makes the various assessment notifications consistent in both their implementation and behaviour, resolving existing tech debt, temporary workarounds and confusing code. All assessment related notifications now include a relationship to the relevant assessment and unused notifications are now remove.
- A future PR will also remove references to `PrepareAssessmentNotificationsJob` once we are sure there are no jobs queued in sidekiq.

### Have you? (optional)

- [ ] TODO: Update wiki

### DIFF
--- a/app/controllers/api/framework_responses_controller.rb
+++ b/app/controllers/api/framework_responses_controller.rb
@@ -108,7 +108,7 @@ module Api
 
     def send_notification
       if assessment.respond_to?(:amended_at) && assessment.amended_at.present?
-        Notifier.prepare_notifications(topic: assessment, action_name: 'amend_person_escort_record')
+        Notifier.prepare_notifications(topic: assessment, action_name: "amend_#{assessment.class.name.underscore}")
       elsif assessment.completed_at.present?
         Notifier.prepare_notifications(topic: assessment, action_name: "complete_#{assessment.class.name.underscore}")
       end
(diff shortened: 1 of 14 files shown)


app/controllers/api/framework_responses_controller.rb
```diff
@@ -109,6 +109,8 @@ def render_bulk_update_error(exception)
     def send_notification
       if assessment.respond_to?(:amended_at) && assessment.amended_at.present?
         Notifier.prepare_notifications(topic: assessment, action_name: 'amend_person_escort_record')
```
martyn-w, line 111: might we want to amend completed youth assessments in the future? if so, perhaps this line should be:
```
Notifier.prepare_notifications(topic: assessment, action_name: "amend_#{assessment.class.name.underscore}")
```
to avoid misnaming youth assessments as PERs ?

martyn-w (approved):
👍  LGTM

**DO NOT MERGE until Geoamey have confirmed support for the new webhooks**

app/controllers/api/framework_responses_controller.rb
smoothcontract: Good idea, it's more future proof that way 👍

PR #1352, opened 2021-02-25
refactor: Update circle config to use new Slack orb

Nimphal:
### What?

I have added/removed/altered:
Update the CircleCI config to use the new Slack orb.

### Why?

I am doing this because:

- moving to the MoJ Slack workspace requires the new major version of the orb


martyn-w (approved):
👍  LGTM

PR #1355, opened 2021-03-01
Add support for importing prison addresses from Nomis

smoothcontract:
### Jira link

P4-2683

### What?

- [x] Update `reference_data:create_locations` rake task to import prison addresses from Nomis
- [x] Extend `LocationSerializer` to include new address attributes
- [x] Add new `NomisClient::LocationDetails` class to fetch prison addresses from Nomis 

### Why?

- This allows address details for prisons to be exposed to API clients. A future PR will geocode these addresses to provide latitude and longitude coordinates to API clients, which will support the display of locations on a map. We want to do this geocoding in the backend so that it only has to be done once (locations are reference data) and are available to multiple clients.

### Have you? (optional)

- Changes an api that is used in production; extends existing location JSON with additional attributes. This affects all endpoints that include location details.

### DIFF
--- a/app/services/locations/importer.rb
+++ b/app/services/locations/importer.rb
@@ -2,21 +2,33 @@
 
 module Locations
   class Importer
-    attr_accessor :items, :added_locations, :updated_locations, :disabled_locations
+    attr_accessor :locations, :location_details, :added_locations, :updated_locations, :disabled_locations
 
-    def initialize(items)
-      @items = items
+    def initialize(locations, location_details)
+      @locations = locations
+      @location_details = location_details
       @added_locations = []
       @updated_locations = []
       @disabled_locations = []
     end
 
     def call
+      update_locations
+      disable_unused_locations
+    end
+
+  private
+
+    def update_locations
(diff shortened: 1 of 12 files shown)


app/services/locations/importer.rb
```diff
@@ -26,8 +38,10 @@ def call
         end
         location.save
       end
+    end
 
-      active_agency_ids = items.map { |item| item[:nomis_agency_id] }
+    def disable_unused_locations
+      active_agency_ids = locations.map { |item| item[:nomis_agency_id] }
```
martyn-w, line 44: may be simplify to
active_agency_ids =  locations.map(&:nomis_agency_id)

martyn-w (approved):
👍  good stuff

app/services/locations/importer.rb
smoothcontract: I don't think that is the same thing - it equates to calling `item.nomis_agency_id` rather than `item[:nomis_agency_id]`. And given `item` is just a hash we need to call the `[]` method passing the symbol argument.

martyn-w: oh yes, you are quite right, my bad 👍

PR #1359, opened 2021-03-04
refactor: Fix errors in slack message template

Nimphal:
### What?

Fix emoji attribute and make sure every text element has the correct type attribute.


smoothcontract (approved):
👍

PR #1361, opened 2021-03-04
refactor: Fix slack fields definition to match accepted format

Nimphal:
### What?

The fields attribute was defined incorrectly leading to an error when trying to send slack notification. Additionally, if a text field is of type markdown, emojis automatically work.


martyn-w (approved):
👍

PR #1362, opened 2021-03-04
refactor: Replace sentry-raven with new Ruby and Rails SDKs. Update config and tests

Nimphal:
### Jira

[P4-2638](https://dsdmoj.atlassian.net/browse/P4-2638)

### What?

I have added/removed/altered:

- [x] Added sentry-ruby, sentry-rails and sentry-sidekiq
- [x] Removed sentry-raven
- [x] Refactored all references to Raven and old methods
- [x] Updated Sentry config with new way to mask filtered params
- [x] Updated Sentry config with a traces variable to enable performance monitoring

### Why?

I am doing this because:

- sentry-raven is now in maintenance mode https://docs.sentry.io/platforms/ruby/migration/
- the new SDKs enable us to do performance monitoring in Sentry, giving us more debugging power


config/initializers/sentry.rb
```diff
@@ -1,6 +1,19 @@
 return unless ENV['SENTRY_DSN'].present?
 
-Raven.configure do |config|
+Sentry.init do |config|
   config.dsn = ENV['SENTRY_DSN']
-  config.sanitize_fields = Rails.application.config.filter_parameters.map(&:to_s)
+
+
+  filter = ActiveSupport::ParameterFilter.new(Rails.application.config.filter_parameters.map(&:to_s))
```
Nimphal, line 7: This is supposed to replace the `sanitize_fields` method, which has been removed. I am unsure whether the `to_s` is still right here, advice welcome!

Nimphal: Have checked with the ParameterFilter documentation and it seems the to_s won't hurt. It could potentially be removed as I think the filter will auto-cast it, but it's not a problem to have it here.

app/jobs/notify_webhook_job.rb
```diff
@@ -33,7 +33,10 @@ def perform(notification_id:, **_)
         delivery_attempts: notification.delivery_attempts.succ,
         delivery_attempted_at: Time.zone.now,
       )
-      Raven.capture_exception(e)
+      Sentry.with_scope do |scope|
+        scope.set_tags(supplier: subscription.supplier.name)
+        Sentry.capture_exception(e)
+      end
```
Nimphal, line 39: Include the supplier name in the notification exception.

app/jobs/notify_email_job.rb
```diff
@@ -28,7 +28,10 @@ def perform(notification_id:, **_)
         delivery_attempts: notification.delivery_attempts.succ,
         delivery_attempted_at: Time.zone.now,
       )
-      Raven.capture_exception(e)
+      Sentry.with_scope do |scope|
+        scope.set_tags(supplier: subscription.supplier.name)
+        Sentry.capture_exception(e)
+      end
```
Nimphal, line 34: Include the supplier name in the notification exception.

.circleci/config.yml
```diff
@@ -211,6 +211,8 @@ references:
         echo ${TESTS}
         bundle exec rspec ${TESTS}
         #COVERAGE=1 bundle exec rspec
+      environment:
```
martyn-w, line 214: rather than add this variable to CircleCI, how about adding it to the existing file `.env.test` - that way it can be used for local tests as well as CircleCI. Also, where did the value `//` come from? maybe a slightly more conventional would be clearer, e.g.:
```
SENTRY_DSN='http://something@example.test/'
```

.env.example
```diff
@@ -4,6 +4,7 @@ NOMIS_SITE_FOR_API='https://nomis-api.example.com'
 NOMIS_AUTH_SCHEME='basic_auth'
 NOMIS_SITE_FOR_AUTH='https://nomis-api.example.com/auth'
 NOMIS_PRISON_API_PATH_PREFIX='/api'
+SENTRY_DSN="//"
```
martyn-w, line 7: I'd suggest a more conventional test value, e.g. `http://something@example.test/`

martyn-w (approved):
👍  see a comment about adding the SENTRY_DSN variable to `.env.test` rather than the CircleCI (so it can be used locally), otherwise looks good 👍

config/initializers/sentry.rb
```diff
@@ -1,6 +1,19 @@
...
+  config.before_send = lambda do |event, hint|
+    event.request.data = filter.filter(event.request.data)
+    event
+  end
+
+  # This is the default value for this option, putting here for visiblity
+  # This will remove the request body from the information sent to sentry
+  config.send_default_pii = false
```
martyn-w, line 15: Just thinking about this, maybe we could have the PII on for dev/staging/uat but off on pre-prod / production?
```
config.send_default_pii = Rails.env.development? || ENV.fetch('HOSTNAME', 'UNKNOWN') =~ /(\-(dev|staging|uat)\-)/i
```

Nimphal: Yes, that's what @smoothcontract was suggesting in the ticket. I was going to do that as a separate PR after this one got merged, just to keep the decision encapsulated in one small change for easy retrieval later if needed :)

martyn-w: jolly good 👍

PR #1363, opened 2021-03-04
Add geocoder gem and mixin to support geocoding location postcodes

smoothcontract:
### Jira link

P4-2686

### What?

- [x] Add geocoder gem with configuration for postcodes.io API lookup (free and anonymous based on open data)
- [x] Add `Geocodeable` concern to include geocoding support for `Location` model
- [x] Expose new `latitude` and `longitude` location attributes in JSONAPI

### Why?

- Geocoding is automatically performed on updating a location (this only happens when running the `create_locations` rake task) and only when the location postcode is present (currently only populated for Nomis prisons). Geocoding is via the postcodes.io API which is free and does not require an API key so no external configuration is required.
Also added latitude and longitude attributes to location serializer so these are available to API clients.

### Have you? (optional)

- Changes an api that is used in production; adds new latitude and longitude attributes to all included locations


martyn-w (approved):
👍  nice clean implementation 👍

PR #1370, opened 2021-03-09
Add new rake task to import postcodes.csv file and perform location geocoding

smoothcontract:
### Jira link

P4-2692

### What?

- [x] Add new rake task `reference_data:import_postcodes`
- [x] Add new `Locations::PostcodeImporter` service to populate postcodes for non-prison locations

### Why?

- Prison address details including postcode can be obtained from Nomis (and are already imported via `reference_data:create_locations` task). For other location types we need to use an external source, so these are held in a simple CSV file with `nomis_agency_id` and `postcode` columns. Any prison locations within the CSV file are not imported as we should use Nomis as the source of truth for those. Similarly any disabled or unknown locations are ignored.


martyn-w (approved):
👍  LGTM

PR #1374, opened 2021-03-10
Add include support to move journeys endpoint

smoothcontract:
### Jira link

P4-2693

### What?

- [x] Add support for `include=from_location,to_location` to `GET /moves/{id}/journeys` endpoint

### Why?

- This was previously documented as supporting the same includes as a single journey, but in fact had no include support at all. Added support for a limited set of includes (just the from and to location) as it's less useful to return the suppliers for multiple journeys. Added new serializers for journeys and locations to avoid N+1 queries.

### Have you? (optional)

### DIFF
--- a/app/controllers/api/journeys_controller.rb
+++ b/app/controllers/api/journeys_controller.rb
@@ -22,7 +22,7 @@ module Api
     ].freeze
 
     def index
-      paginate journeys, serializer: JourneySerializer
+      paginate journeys, serializer: JourneysSerializer, include: included_relationships
     end
 
     def show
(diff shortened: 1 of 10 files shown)


app/controllers/api/journeys_controller.rb
```diff
@@ -47,7 +47,12 @@ def render_journey(journey, status)
     end
 
     def supported_relationships
-      JourneySerializer::SUPPORTED_RELATIONSHIPS
+      # for performance reasons, we support fewer include relationships on the index action
+      if action_name == 'index'
+        JourneysSerializer::SUPPORTED_RELATIONSHIPS
```
martyn-w, line 52: 👍  good stuff

martyn-w (approved):
👍  LGTM

PR #1415, opened 2021-04-14
refactor: Fetch person nomis data on move creation

Thource:
### Jira link

P4-2717

### What?

I have added/removed/altered:

- [ ] Altered move creation to pull new data from nomis for existing people

### Why?

I am doing this because:

- This will mean that when a move is created for a person, we will have up to date images, alerts and health data for them.

### Have you? (optional)


app/services/people/retrieve_image.rb
```diff
@@ -2,8 +2,8 @@
 
 module People
   class RetrieveImage
-    def self.call(person)
-      return true if person.image.attached?
+    def self.call(person, force_update = false)
```
martyn-w, line 5: very small niggle - could `force_update` be a named parameter, i.e.
```
def self.call(person, force_update: false)
```

spec/rails_helper.rb
```diff
@@ -79,6 +79,13 @@
 
   config.include FactoryBot::Syntax::Methods
   config.include RSpec::JsonExpectations::Matchers
+
+  config.before do
```
martyn-w, line 83: these stubs will apply to _all_ tests - which is probably not a good idea. Could you stub in just the individual tests as required?

Thource: I can, it'll be over 100 files though, as a lot of the spec files have `let(:move) { create(:move...` in

Nimphal: I think there 56 tests or so that broke when Adam first did this, which is quite the number to be stubbing individually. Is there a middle ground way?

martyn-w: Ahhh... I see what you mean. Let me have a think about this...

app/models/move.rb
```diff
@@ -116,6 +116,15 @@ class Move < VersionedModel
   before_validation :set_reference
   before_validation :set_move_type
 
+  after_create do
```
martyn-w, line 119: @Thource  thinking more about this, maybe the model is the wrong place for this code. How about instead moving this to `create_and_render` in `app/controllers/api/v2/moves_actions.rb` AND to `app/controllers/api/v1/moves_actions.rb` Those action files are where do other Nomis actions, e.g. for allocations. 

Keeping the Nomis link outside of the model will simplify the test changes, as only creates via the api will trigger a call to Nomis.

Thource: That sounds good to me, I'll get on it :)

app/models/person.rb
```diff
@@ -82,6 +82,13 @@ def category
     @category ||= Categories::FindByNomisBookingId.new(latest_nomis_booking_id).call
   end
 
+  def update_nomis_data
+    People::RetrieveImage.call(self, force_update: true)
+    return if @prison_number.blank?
+
+    Profiles::ImportAlertsAndPersonalCareNeeds.new(@profile, @prison_number).call
```
martyn-w, line 89: I think this needs a bit of rejigging:
1. you probably want to refresh `person.latest_nomis_booking_id` _before_ retrieving the person's image (however see point 3 below)
2. a `Person` can have multiple `Profiles`. The variable `@profile` here isn't defined and so will always be nil
3. more generally, there is a complicated dance the frontend does with the backend: searching for a person by prison_number will _create_ new person records from Nomis data. And creating a blank profile for a person with a prison_number will fetch the latest Nomis alerts/data, prior to creating the move. So possibly this step is redundant because it would have have already been called for someone creating the move through BaSM front end. See this walkthrough of the typical interactions between FE and BE: https://github.com/ministryofjustice/hmpps-book-secure-move-api/wiki/API-Walkthroughs-7.1-Prison-to-Court-move

martyn-w: in short, it might not be necessary to call `Profiles::ImportAlertsAndPersonalCareNeeds` - the only refresh required might be of the person's image. But this depends on the use-case: whether the intention is to improve data quality on moves created via the front end, or to improve data quality on moves created elsewhere (e.g. directly by suppliers)

Nimphal (approved):
Looks good! Would be good to get @martyn-w 's opinion on it as well :)

spec/models/person_spec.rb
```diff
@@ -165,4 +165,19 @@
       expect { create(:person, :pre1900) }.to raise_exception(ActiveRecord::RecordInvalid, 'Validation failed: Birth date must be after 1900-01-01.')
     end
   end
+
+  describe '#update_nomis_data' do
+    let(:import_alerts_and_personal_care_needs) { instance_double('Profiles::ImportAlertsAndPersonalCareNeeds', call: true) }
+
+    before do
+      allow(Profiles::ImportAlertsAndPersonalCareNeeds).to receive(:new).and_return(import_alerts_and_personal_care_needs)
```
martyn-w, line 173: this statement `allow(Profiles::ImportAlertsAndPersonalCareNeeds)...`  isn't required anymore

Thource: Doh

spec/requests/api/moves_controller_role_access_spec.rb
```diff
@@ -81,6 +81,13 @@
     end
 
     context 'when successful' do
+      before do
```
martyn-w, line 84: This block isn't being tested / isn't relevant to the test? is a test missing or is this in the wrong place?

spec/services/people/retrieve_image_spec.rb
```diff
@@ -3,13 +3,44 @@
...
+          expect(person).not_to have_received(:attach_image)
+        end
+      end
+
+      context 'when force_update: true' do
+        it 'updates the image' do
+          described_class.call(person, force_update: true)
 
-    expect(retrieve_image).to eq(false)
+          expect(person).to have_received(:attach_image)
```
martyn-w, line 41: good test 👍

martyn-w (approved):
There seems to be an unnecessary block of code in `spec/requests/api/moves_controller_role_access_spec.rb` - otherwise looks good

spec/requests/api/moves_controller_role_access_spec.rb
Thource: I think this was a leftover from when I was updating the nomis data on move#new

PR #1419, opened 2021-04-21
P4 2030 move statemachine

martyn-w:
### Jira link

P4-<2030>

### What?

I have added/removed/altered:

- [x] Added a move state machine
- [x] Relaxed some move state transitions to support legacy behaviours
- [x] Updated relevant tests

### Why?

I am doing this because:

- to better enforce data quality on the move model

### Have you? (optional)

- Potentially risky - should be trialed on Staging / UAT for a few days prior to deployment to production


app/services/moves/updater.rb
```diff
@@ -44,20 +64,22 @@ def document_ids
...
+        attributes[:profile] = Profile.find_by(id: profile_attributes.dig(:data, :id)) if profile_attributes.present?
 
-      profile = attributes[:profile] || move.profile
-      profile.documents = Document.where(id: document_ids) unless document_attributes.nil?
+        profile = attributes[:profile] || move.profile
+        profile.documents = Document.where(id: document_ids) unless document_attributes.nil?
 
-      attributes
+        attributes
+      end
```
Nimphal, line 82: Just to refresh my Ruby memory, we're doing this because we're manipulating the attributes in the `call` method?

app/controllers/api/v2/moves_actions.rb
```diff
@@ -92,6 +103,15 @@ def common_move_attributes
...
 
+    def validate_move_status
+      status = move_params.fetch(:attributes, {})[:status]
+      if status.present?
+        validator = Moves::StatusValidator.new(status: status, cancellation_reason: move_params.fetch(:attributes, {})[:cancellation_reason], rejection_reason: move_params.fetch(:attributes, {})[:rejection_reason])
+        raise ActiveModel::ValidationError, validator unless validator.valid?
+      end
+      status
+    end
+
```
Nimphal, line 114: This logic looks like it's repeated in two other places with the difference being which object we take status from. Is there a case for abstracting it to keep it consistent? Or is it the same now, but that's by chance and we'd expect them to diverge in the future?

app/services/moves/params_validator.rb
```diff
@@ -34,6 +36,7 @@ def initialize(filter_params, sort_params)
       @date_of_birth_from = filter_params[:date_of_birth_from]
       @date_of_birth_to = filter_params[:date_of_birth_to]
       @move_type = filter_params[:move_type]
+      @status = filter_params[:status]
```
Nimphal, line 39: Is the node application going to care about the status soon, is that the plan? It will currently ignore this attribute because it doesn't know about it.

app/controllers/api/v2/moves_actions.rb
Nimphal: Actually, ignore me! I can see the validation logic itself is abstracted!

Nimphal (approved):

app/services/moves/params_validator.rb
martyn-w: interesting question - when I was looking at the validators I noticed that `status` was missing from the list of things to validate and I assumed that was a mistake - but perhaps not if the node app isn't using status. But no harm adding it here in case it is used in future.

app/services/moves/updater.rb
martyn-w: I'm memoizing the attributes here for efficiency - to avoid running the same code multiple times unnecessarily. The `begin` and `end` block store the result in the `@attributes` variable

app/controllers/api/v2/moves_actions.rb
martyn-w: 😅

app/services/moves/params_validator.rb
martyn-w: Hmm, famous last words; adding this validation has broken something; I'll remove it in another PR

PR #1424, opened 2021-04-23
P4 2148 per counts for dashboards deploydev

martyn-w:
### Jira link

P4-2148

### What?

I have added/removed/altered:

- [x] Added new PER counts to metrics system

### Why?

I am doing this because:

- so we can keep track of PER usage

### Have you? (optional)


Nimphal (approved):

app/models/metrics/person_escort_records/counts_by_move_time_bin.rb
```diff
@@ -0,0 +1,33 @@
...
+      def initialize(supplier: nil)
+        setup_metric(
+          supplier: supplier,
+          label: 'PER counts by move time bin',
+          file: 'counts_by_move_time_bin',
+          interval: 5.minutes,
+          columns: {
+            name: COUNT,
+            field: :itself,
+            values: [COUNT],
```
Nimphal, line 17: Just checking my understanding here - in the other metrics calculators we append to an array whereas here we replace the array. Is this because this is a timebound query which doesn't need to hold historical data? I assume the appending is to preserve historical data.

martyn-w: Hi Nev, its because this particular metric only has one column (Count), e.g.:

Screenshot 2021-04-27 at 14 06 23


...whereas other metrics can have multiple columns, e.g.

Screenshot 2021-04-27 at 14 07 18

PR #1448, opened 2021-05-13
P4-2831: Add created_at to locations + filter

Thource:
### Jira link

P4-2831

### What?

I have added/removed/altered:

- [x] Added created_at to the locations filters, it takes a date in `yyyy-MM-dd` format
- [x] Added created_at to the locations serialization, it returns the datetime the location was created in ISO-8601 format

### Why?

I am doing this because:

- It will allow JPC to automatically recognise new locations

### Have you? (optional)

### DIFF
--- a/app/services/locations/finder.rb
+++ b/app/services/locations/finder.rb
@@ -72,7 +72,7 @@ module Locations
     def apply_created_at_filters(scope)
       return scope if filter_params[:created_at].blank?
 
-      date = Date.strptime(filter_params[:created_at])
+      date = Date.strptime(filter_params[:created_at], '%Y-%m-%d')
       scope = scope.where(created_at: date.midnight..date.end_of_day)
       scope
     end
--- a/app/services/locations/finder.rb
+++ b/app/services/locations/finder.rb
@@ -36,6 +36,7 @@ module Locations
       scope = apply_location_filters(scope)
       scope = apply_region_filters(scope)
       scope = apply_young_offender_institution_filters(scope)
+      scope = apply_created_at_filters(scope)
       scope
     end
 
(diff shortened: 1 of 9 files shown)


swagger/v1/location.yaml
```diff
@@ -97,6 +97,13 @@ Location:
           type: boolean
           example: false
           description: Indicates that the location is a Youth Offenders Institute (note that some locations may have mixed populations so have a differing location type).
+        created_at:
+          oneOf:
+          - type: string
+            format: date-time
+          - type: 'null'
+          example: '2017-07-21T17:32:28Z'
+          description: The date-time at which the location was created, in ISO-8601 format
```
martin-ballhatchet-moj, line 106: Might be better to say "The date-time at which the location was imported into Book a Secure Move, in ISO-8601 format" - currently locations are created in NOMIS but not imported into our service until we run the sync

app/services/locations/finder.rb
```diff
@@ -67,5 +68,13 @@ def apply_young_offender_institution_filters(scope)
       scope = scope.where(young_offender_institution: false) if filter_params[:young_offender_institution].to_s == 'false'
       scope
     end
+
+    def apply_created_at_filters(scope)
+      return scope if filter_params[:created_at].blank?
+
+      date = Date.strptime(filter_params[:created_at])
```
martyn-w, line 75: elsewhere we have been explicit about the date format, e.g. `Date.strptime(value, '%Y-%m-%d')` - I think it would be good to do so here to avoid any future misunderstanding

martyn-w (approved):
👍  LGTM

PR #1472, opened 2021-06-08
Update the tracing setup to only be used in prod and ignore health

Nimphal:
Update the tracing sampler to check the environment and the name of the transaction before sending it to sentry. We don't want to record tracing on anything that's not production and we also don't want to record ping or health transactions.

### DIFF
--- a/config/initializers/sentry.rb
+++ b/config/initializers/sentry.rb
@@ -1,5 +1,7 @@
 return unless ENV['SENTRY_DSN'].present?
 
+EXCLUDE_PATHS = %w[/ping /health].freeze
+
 Sentry.init do |config|
   config.dsn = ENV['SENTRY_DSN']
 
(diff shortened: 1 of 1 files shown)


config/initializers/sentry.rb
```diff
@@ -17,8 +19,16 @@
   # Don't log RetryJobError exceptions in Sentry as they will have already been logged as part of the failed job
   config.excluded_exceptions += ['RetryJobError']
 
-  # Half of all requests will be used in performance sampling.
-  # Currently the MoJ plan does not allow this. Turn on when
-  # plan has been updated in July 2021
```
Thource, line 22: Should we keep this comment?

Nimphal: Nah, it was relevant before, but it's now permanently on, so we can get rid of this :)

Thource (approved):

PR #1476, opened 2021-06-10
[P4-2892] Store CSRA from NOMIS on Profile create

Thource:
### Jira link

P4-2892

### What?

I have altered:

- [x] Altered ProfileController#Create to pull the CSRA from NOMIS and save it on the profile

### Why?

I am doing this because:

- It will make it accessible by the frontend.


martyn-w (approved):
👍  looks good - and good job tidying up the category code

PR #1482, opened 2021-06-23
P4-2902 auto confirm PER on move start

martyn-w:
### Jira link

P4-2902

### What?

I have added/removed/altered:

- [x] Auto-confirms completed PERs when a move starts

### Why?

I am doing this because:

- because a large number of PERs are not being confirmed, which means they cannot be used for pre-fill later

### Have you? (optional)

### DIFF
--- a/spec/models/generic_event/move_start_spec.rb
+++ b/spec/models/generic_event/move_start_spec.rb
@@ -1,18 +1,140 @@
 require 'rails_helper'
 
 RSpec.describe GenericEvent::MoveStart do
-  subject(:generic_event) { build(:event_move_start) }
+  subject(:move_start_event) { build(:event_move_start, eventable: create(:move, move_status)) }
 
-  it_behaves_like 'a move event'
-
-  describe '#trigger' do
-    it 'does not persist changes to the eventable' do
-      generic_event.trigger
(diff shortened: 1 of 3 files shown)


spec/models/generic_event/move_start_spec.rb
```diff
@@ -1,18 +1,140 @@
...
+        it_behaves_like 'a GenericEvent::PerConfirmation event is created'
+        it_behaves_like 'a per confirmation webhook is sent'
+      end
+
+      context 'when the PER status is confirmed' do
+        let(:person_escort_record_status) { 'confirmed' }
+
+        it_behaves_like 'the move status changes to in_transit'
+        it_behaves_like 'the PER status does not change'
+        it_behaves_like 'a GenericEvent::PerConfirmation event is not created'
```
Nimphal, line 117: Just so I know I understand it - we don't expect an event here because the confirmation has happened in the past (and so the event would've been created) and the change to the move should not re-issue that event?

Nimphal (approved):

spec/models/generic_event/move_start_spec.rb
martyn-w: yep, that's correct - it would have been handed over and instead would have a PerHandover event

PR #1483, opened 2021-06-24
[P4-2923] Convert feed rake tasks to sidekiq jobs

Thource:
### Jira link

P4-2923

### What?

I have added/removed/altered:

- [x] Added TimeSince for more simple duration tracking
- [x] Changed feed generation from rake tasks to sidekiq workers

### Why?

I am doing this because:

- It gives increased visibility, a safe way of retrying and also makes use of our current alerting for failed jobs allowing us to tackle issues before they hit CJVP.

### Have you? (optional)


config/sidekiq.yml
```diff
@@ -13,3 +13,8 @@
   - ["notifications_low", 2]
   - ["notifications", 1] # TODO: this queue can be deleted in a future PR once high/medium/low are in place
   - ["default", 1]
+
+:schedule:
```
Nimphal, line 17: ah, I thought this was a paid feature of sidekiq that we're not paying for. Perhaps I've misunderstood, I thought what we could do was use our existing cron job which called `feeds:all` and convert the individual feeds to cronjobs which were invoked by the existing cronjob.

Thource: This uses a separate gem https://github.com/moove-it/sidekiq-scheduler :)

martyn-w: Thinking this through, I'm not sure if it is safe to schedule directly in the Sidekiq pods? in most environments there are 2 sidekiq pods. So presumably the scheduling will run on both pods and hence the job will be scheduled twice? It might be safer to get a rake task triggered by a cronjob to add the tasks to the sidekiq queue, and then whichever worker is free can pick it up. And we would have no worries if we want to increase the number of sidekiq pods in the future.

martyn-w: it would be good to test on dev : e.g. schedule a job once and see if it runs twice?

app/models/time_since.rb
```diff
@@ -0,0 +1,13 @@
+class TimeSince
```
martyn-w, line 1: good idea to formalise this calculation  👍 👍   - but maybe this should be a service rather than a model ?

config/sidekiq.yml
```diff
@@ -13,3 +13,8 @@
...
+  all_worker:
+    cron: '0 * * * * *'
```
martyn-w, line 19: we definitely do not want to run all the feed jobs every minute 🙀

lib/tasks/feed.rake
```diff
@@ -0,0 +1,27 @@
+require_relative '../../../app/workers/feeds/all_worker'
```
martyn-w, line 1: this require statement seems to be breaking CircleCI:
```
rake aborted!
LoadError: cannot load such file -- /home/circleci/app/workers/feeds/all_worker
....
```

check the path is correct - it might be helpful to ssh into CircleCi and run a rails console there, where you can issue some `require` statements and debug the issue

config/sidekiq.yml
Thource: Haha, good catch, changed this to once per minute to test it actually scheduled them, forgot to change it back 🤦

lib/tasks/feed.rake
```diff
@@ -0,0 +1,27 @@
...
+    puts "#{feed_name} is not a valid feed name. Valid feed names are: all, #{Feeds::AllWorker.feed_names.join(', ')}"
+    next
+  end
+
+  Time.zone = 'London'
+  date = ENV['REPORT_ON_DATE'] ? Date.parse(ENV['REPORT_ON_DATE']) : Time.zone.yesterday
+
+  if feed_name == 'all'
+    Feeds::AllWorker.feed_names.each { |fn| Feeds::FeedWorker.new.perform(fn, date) }
+  else
```
martyn-w, line 22: to ensure the job runs as a Sidekiq job, use `JobName.perform_later(a,b,c...)` rather than `JobName.new.perform(a,b,c...)` (i.e. perform_later class method, not perform instance method)

spec/models/time_since_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'rails_helper'
+
+RSpec.describe TimeSince, type: :model do
+  let(:time) { Time.zone.local(2021) }
```
martyn-w, line 4: good tests 👍 

you may want to consider using Timecop to freeze time, rather than stubbing the Time class with allow().to receive, e.g.:

```
RSpec.describe TimeSince do
  around do |example|
    Timecop.freeze('2021-01-01 00:00:00')
    example.run
    Timecop.return
  end

 it 'should test something' do
   Time.local.now # this returns 01/01/2021 at 00:00:00
  ....
```

check the other rspecs for Timecop

lib/tasks/feed.rake
Thource: I'm a little undecided on this, the idea behind it was that it'd run the jobs while you wait, rather than having to check manually in sidekiq after to see if they ran successfully. I'm happy to change it to schedule the jobs instead though

docker-compose.yml
```diff
@@ -30,8 +30,8 @@ services:
       - DATABASE_URL=postgresql://postgres:postgres@db/hmpps-book-secure-move-api
       - SECRET_KEY_BASE=h463472fhaa91f4a277002e9652f86f330e636358e2090d86ab4a4f06845ege1
       - SERVE_API_DOCS=true
-      - AWS_ACCESS_KEY_ID=fakeid
```
martyn-w, line 33: these keys (`AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`) I think are used for document uploads and document retrieval when booking moves (youth moves). They point to different S3 buckets with different credentials to the bucket for storing feeds... so unfortunately I think you'll need to support access to both buckets simultaneously

martyn-w:
I think this is looking pretty good as a concept 👍  
There is a bit more work to do to ensure there is no clash of S3 env vars between those used for reporting feeds vs those used for other purposes (document storage / image storage / metrics storage etc)

docker-compose.yml
Thource: I thought that'd be the case too, but I couldn't find `AWS_ACCESS_KEY_ID` anywhere else in this project, any ideas?

martyn-w: its probably set in the deploy repo... I'll see if I can find it

Thource: Also these didn't match up, the AWS_ACCESS_KEY_ID was the reporting access key, but we have S3_BUCKET_NAME defined as the documents bucket in here, a bit confusing :D

martyn-w: Take a look at `app/lib/cloud_data/reports_feed.rb` vs `app/lib/cloud_data/metrics_feed.rb`
The reports feed assumes the default AWS env vars; whereas the metrics feed explicitly names them:

```
 @client = client || Aws::S3::Client.new(
        region: ENV['S3_METRICS_REGION'],
        access_key_id: ENV['S3_METRICS_ACCESS_KEY_ID'],
        secret_access_key: ENV['S3_METRICS_SECRET_ACCESS_KEY'],
      )
```

to avoid confusion and naming clashing lets use the env var names you have in your [deploy PR](https://github.com/ministryofjustice/hmpps-book-secure-move-api-deploy/pull/150/files#diff-2029cf907f6ef6b89c9bcaf88b53791aa6899564eff891b20a94c6920e1ba724R116), i.e.:
* `S3_REPORTING_ACCESS_KEY_ID`
* `S3_REPORTING_SECRET_ACCESS_KEY`
* `S3_REPORTING_BUCKET_NAME`

- so could you update `app/lib/cloud_data/reports_feed.rb` to use these new names ^ ?

Thource: Sounds like a good idea to me, I'm on it

martyn-w (approved):
👍  good work

This will need careful testing on dev/staging before deploying to production

PR #1494, opened 2021-07-08
Improve additional data for Prison API Sentry errors

thomasleese:
### Jira link

[P4-2982](https://dsdmoj.atlassian.net/browse/P4-2982)

### What?

I have added/removed/altered:

- Added more information to Prison API errors in Sentry
- Refactored tests by including a shared example for testing a message is captured in Sentry

**Probably easier to review commit by commit because of the refactor.**

### Why?

This should allow us to filter Sentry errors on the status of the response from the Prison API meaning ultimately we can focus specifically on the ones which return a `500` error.

### DIFF
--- a/app/lib/nomis_client/base.rb
+++ b/app/lib/nomis_client/base.rb
@@ -76,10 +76,8 @@ module NomisClient
                                extra: {
                                  route: path,
                                  body_params: params,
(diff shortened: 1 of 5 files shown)


Nimphal (approved):
Nice!

PR #1506, opened 2021-07-21
Add GPS data reports

Thource:
### Jira link

P4-2944

### What?

I have added/removed/altered:

- [x] Added GPS data reports

### Why?

I am doing this because:

- This will allow us to check if our suppliers are sending us accurate gps data for their moves.

### Have you? (optional)


app/services/gps_report.rb
```diff
@@ -0,0 +1,131 @@
+# frozen_string_literal: true
+
+class InvalidSupplierError < StandardError; end
+
+class GPSReport
+  include ActionView::Helpers::TextHelper
+
+  def initialize(date_range, supplier_name)
+    @supplier = Supplier.find_by(key: supplier_name)
+    raise InvalidSupplierError, "Invalid supplier: `#{supplier_name}`. Valid values are: `#{Supplier.all.map(&:key).join('`, `')}`." if @supplier.nil?
```
thomasleese, line 10:
```suggestion
    raise InvalidSupplierError, "Invalid supplier: `#{supplier_name}`. Valid values are: `#{Supplier.pluck(:key).join('`, `')}`." if @supplier.nil?
```

I think this avoids fetching all the fields from the database when we only need the `:key`.

app/services/gps_report.rb
```diff
@@ -0,0 +1,131 @@
...
+  end
+
+private
+
+  def failures
+    return @failures unless @failures.nil?
+
+    @failures = moves.filter_map { |move|
+      { move: move, reason: 'no_journeys' } if move.journeys.empty?
+    }.concat(journey_failures)
```
thomasleese, line 42:
```suggestion
    @failures ||= begin
      moves.filter_map { |move|
        { move: move, reason: 'no_journeys' } if move.journeys.empty?
      }.concat(journey_failures)
    end
```

I think you can do this to avoid needing to check if `@failures` is `nil`, and same for `@moves`.

app/services/gps_report.rb
```diff
@@ -0,0 +1,131 @@
...
+
+  def journeys
+    @journeys ||= moves.flat_map(&:journeys)
+  end
+
+  def moves
+    return @moves unless @moves.nil?
+
+    @moves = []
+    Move.where(status: 'completed', date: @from_date..@to_date, supplier: @supplier).order(:id).find_each do |move|
```
thomasleese, line 87:
```suggestion
    Move.includes(:journeys).where(status: 'completed', date: @from_date..@to_date, supplier: @supplier).order(:id).find_each do |move|
```

I notice that we do `move.journeys` in a few places, so we might want them to be pre-fetched from the database in a single query.

app/services/gps_report.rb
```diff
@@ -0,0 +1,131 @@
...
+      @moves << move
+    end
+
+    @moves
+  end
```
thomasleese, line 92:
```suggestion
  def moves
    @moves ||= Move.where(status: 'completed', date: @from_date..@to_date, supplier: @supplier).order(:id).to_a
  end
```

I think this is the same behaviour.

app/services/gps_report.rb
```diff
@@ -0,0 +1,131 @@
...
+    time_since = TimeSince.new
+    res = athena_query("select journey_id, tracking_timestamp from #{@supplier.key} where journey_id in ('#{journeys.map(&:id).join("', '")}') order by journey_id, tracking_timestamp;")
+
+    @gps_data_map = {}
+    res.result_set.rows[1..].each do |data_row|
+      journey_id, timestamp = *data_row.data.map(&:var_char_value)
+      @gps_data_map[journey_id] = [] if @gps_data_map[journey_id].nil?
+
+      @gps_data_map[journey_id] << Time.zone.parse(timestamp)
+    end
```
thomasleese, line 72:
```suggestion
    @gps_data_map = res.result_set.rows[1..].each_with_object do |data_row, data_map|
      journey_id, timestamp = *data_row.data.map(&:var_char_value)
      data_map[journey_id] = [] if data_map[journey_id].nil?

      data_map[journey_id] << Time.zone.parse(timestamp)
    end
```

I think this is slightly nicer as it avoids needing to assign the variable to an empty hash first.

app/services/gps_report.rb
```diff
@@ -0,0 +1,131 @@
...
+      begin
+        return athena_client.get_query_results(query_execution_id: query_id)
+      rescue Aws::Athena::Errors::InvalidRequestException => e
+        unless e.message.starts_with?('Query has not yet finished')
+          Rails.logger.error athena_client.get_query_execution(query_execution_id: query_id)
+
+          raise e
+        end
+      end
+    end
```
thomasleese, line 122: I think we can use [`get_query_executation`](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/Athena/Client.html#get_query_execution-instance_method) to check the state of the query, which might be better than this loop (it protects us if the error message changes).

I'm imagining something like this:

```suggestion
    while %w[QUEUED RUNNING].include?(athena_client.get_query_execution(query_execution_id: query_id).status.state)
      sleep 1.second
    end
      
    return athena_client.get_query_results(query_execution_id: query_id)
```

app/workers/gps_report_worker.rb
```diff
@@ -0,0 +1,100 @@
...
+    @date_range = Date.yesterday - 6.days..Date.yesterday
+    threads = suppliers.map { |supplier| Thread.new { GPSReport.new(@date_range, supplier).generate } }
+    post_results(threads.each(&:join).map(&:value))
+  end
+
+private
+
+  def suppliers
+    %w[geoamey serco]
+  end
```
thomasleese, line 19:
```suggestion
  SUPPLIERS = %w[geoamey serco]
```

I wonder if it's clearer to make this an explicit constant?

app/workers/gps_report_worker.rb
```diff
@@ -0,0 +1,100 @@
...
+
+    reasons = failures.map { |r| r[:reason] }.uniq
+    failure_rows = []
+    reasons.each_with_index do |reason, reason_i|
+      failures.filter { |f| f[:reason] == reason }.each_with_index do |failure, failure_i|
+        failure_rows[failure_i] = [] if failure_rows[failure_i].nil?
+
+        failure_rows[failure_i][reason_i] = failure[:move].id
+      end
+    end
```
thomasleese, line 51:
```suggestion
    failure_rows = reasons.with_index.each_with_object([]) do |(reason, reason_i), failure_rows|
      failures.filter { |f| f[:reason] == reason }.each_with_index do |failure, failure_i|
        failure_rows[failure_i] = [] if failure_rows[failure_i].nil?

        failure_rows[failure_i][reason_i] = failure[:move].id
      end
    end
```

I think using `each_with_object` here is slightly neater!

thomasleese (commented):
Looks good, I've left a few comments but I think they're mostly personal preference really.

I wonder if it's worth adding some context to the commit messages too, just to explain why the change was made, especially for the earlier ones.

app/services/gps_report.rb
Thource: You're right, this one doesn't even need the block :D

app/services/gps_report.rb
Thource: I believe we might have issues if we try to get too many moves from the db at once, which is why I'm using find_each here

app/services/gps_report.rb
Thource: Wow, nice!

app/services/gps_report.rb
thomasleese: Ah interesting, I imagined that since we're storing all the moves in the array anyway, the memory improvements of using `find_each` would be negated, but you're right that we would no longer be selecting in batches.

`find_in_batches.to_a.flatten` seems like a good improvement on my suggestion! 👍🏻

thomasleese (approved):
Code changes look good!

I'd be tempted to do a `rebase` to remove the merge commit, and perhaps squash together some of the commits as well just to make things slightly neater.

thomasleese (approved):

PR #1507, opened 2021-07-22
Fix name of GPS data report rake task

Thource:
### Jira link

P4-2944

### What?

I have added/removed/altered:

- [x] Changed GPS data report rake task name

### Why?

I am doing this because:

- Copy/paste error


thomasleese (approved):
🙊

PR #1513, opened 2021-07-29
Allow filtering moves by profile or person

thomasleese:
### Jira link

[P4-3019](https://dsdmoj.atlassian.net/browse/P4-3019)

### What

I've added two new filters to the move endpoint, `profile_id` and `person_id` that allows a user to filter the moves based on those values. I've also refactored some of the code in this area.

### Why

We want to build a page which displays the moves for a specific profile/person, so we need this filter in the API to support that. We don't necessarily need both filters, but I figured there was no harm in having more options.

### DIFF
--- a/app/controllers/api/moves_controller.rb
+++ b/app/controllers/api/moves_controller.rb
@@ -13,10 +13,7 @@ module Api
     end
 
     def csv
-      csv_moves = Moves::Finder.new(filter_params: filter_params,
-                                    ability: current_ability,
-                                    order_params: params[:sort] || {},
-                                    active_record_relationships: CSV_INCLUDES).call
+      csv_moves = find_moves(active_record_relationships: CSV_INCLUDES)
       send_file(Moves::Exporter.new(csv_moves).call, type: 'text/csv', disposition: :inline)
     end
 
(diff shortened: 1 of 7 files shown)


app/controllers/api/moves_controller.rb
```diff
@@ -38,19 +35,43 @@ def filtered
...
+      to_location_id
+      location_id
+      supplier_id
+      move_type
+      cancellation_reason
+      rejection_reason
+      has_relationship_to_allocation
+      ready_for_transit
+      profile_id
+      person_id
```
teneightfive, line 74: I know it's technically a separate piece of work, but how straightforward would it be to support `reference` as a filter too?

spec/swagger/swagger_doc_v2.yaml
```diff
@@ -465,6 +465,26 @@
           schema:
             type: boolean
             example: true
+        - name: filter[profile_id]
```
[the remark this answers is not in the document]
thomasleese: Yes, the code which does the filtering is applied to both endpoints the `GET` and the `POST` one. I didn't realise the frontend was only using the `POST` version though!

app/controllers/api/moves_controller.rb
thomasleese: Since it's just a plain field on the move it would be very easy to do. I don't mind adding it as another commit in this PR.

thomasleese: Added in d98bd8e.

Thource (approved):

spec/swagger/swagger_doc_v2.yaml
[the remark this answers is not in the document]
teneightfive: Yeah, we actually removed the use of the `GET /moves` in favour of that because of a legacy problem with where the locations are retrieved and stored 😞

app/controllers/api/moves_controller.rb
teneightfive: This will unlock a really nice option for us to provide a simple reference search to start with on the frontend 👍 

And also just allow us to debug queries using the API endpoints if all we have is a reference. Before that we had to search directly on the DB...

PR #1533, opened 2021-08-05
Add Athena query_id to sentry extras

Thource:
### Jira link

P4-2944

### What?

I have added/removed/altered:

- [x] Added failed athena query ids to sentry logging

### Why?

I am doing this because:

- It will help us understand why the gps data report has failed


thomasleese (approved):
Happy to try it and see if we get duplicate events or not.

PR #1566, opened 2021-08-23
Consider completed PERs as 'ready for transit'

thomasleese:
This updates the filter for moves such that a completed per is considered ready for transit, rather than just a confirmed one. This change means that moves with a completed, but not confirmed, PER will show in the "ready for transit" section rather than the "incomplete" section.

The old behaviour makes sense when PERs used to have to be confirmed, but now that PERs are automatically confirmed when a move starts, it doesn't make sense to keep the move in the incomplete state until that point.

[Jira Ticket](https://dsdmoj.atlassian.net/browse/P4-2920)


Thource (approved):

teneightfive:
🙌

PR #1597, opened 2021-10-04
P4 3117/validate event occurred at order

Thource:
### Jira link

P4-3117

### What?

I have added/removed/altered:

- [x] Added order validation to GenericEvent
- [x] Added order validation to Move/Journey Start/Complete

### Why?

I am doing this because:

- It will ensure that events occur in the right order, for example: JourneyComplete can not occur before JourneyStart

### DIFF
--- a/app/models/generic_event.rb
+++ b/app/models/generic_event.rb
@@ -198,6 +198,22 @@ class GenericEvent < ApplicationRecord
     end
   end
 
+  def self.validate_occurs_before(before_type)
+    validate do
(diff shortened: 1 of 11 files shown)


thomasleese (commented):
The validation looks good! The only thing I would say is that I don't think we want the validation to _go live_ yet in the sense that it shouldn't actually block events from being created, we just want to put a warning in Sentry so we can start to get a better picture of when this happens (perhaps including metadata such as the supplier).

PR #1630, opened 2021-11-02
Record failure reason in import results

thomasleese:
This adds the ability to record a reason if a record fails to import, so we can better investigate why it might have failed.

### DIFF
--- a/app/lib/imports/missing_move_ending_events.rb
+++ b/app/lib/imports/missing_move_ending_events.rb
@@ -30,14 +30,14 @@ private
 
   def results
     @results ||= records.each_with_object(Imports::Results.new) do |record, results|
-      move = Move.find_by(id: record[:move_id], status: EXPECTED_MOVE_STATUSES)
-      if move.nil?
-        results.record_failure(record)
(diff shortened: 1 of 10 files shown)


app/lib/imports/missing_move_ending_events.rb
```diff
@@ -32,12 +32,12 @@ def results
...
       move = Move.find_by(id: record[:move_id], status: EXPECTED_MOVE_STATUSES)
       if move.nil?
-        results.record_failure(record)
+        results.record_failure(record, reason: 'Could not find move.')
         next
       end
 
       unless ALLOWED_EVENT_TYPES.include?(record[:event_type])
-        results.record_failure(record)
+        results.record_failure(record, reason: 'Event type not allowed.')
```
matt-crowson-moj, line 40: Minor - might be me but couldn't see a test case for this.

matt-crowson-moj (approved):
LGTM

app/lib/imports/missing_move_ending_events.rb
thomasleese: Good point, I've amended the commit to include that now.

PR #1632, opened 2021-11-03
Use events to transition move status

thomasleese:
This modifies the import process for moves without an ending state to rely on the existing event which should already be there. This change is so we can support cancelled or rejected moves that do more than just change the status of the move, they also assign a `cancellation_reason` field.

### DIFF
--- a/app/lib/imports/moves_without_ending_state.rb
+++ b/app/lib/imports/moves_without_ending_state.rb
@@ -30,8 +30,54 @@ private
         next
       end
 
-      move.status = record[:new_status].downcase
+      new_status = record[:new_status]
+
+      event = find_event(move, new_status)
+      if event.nil?
+        results.record_failure(record, reason: 'Could not find associated event.')
+        next
+      end
+
+      move.state_machine.restore!(find_pre_trigger_status(new_status)) # to ensure event state transition is valid
+      event.eventable = move  # to ensure we are modifying the same object
+
+      event.trigger(dry_run: true)
+
+      if move.status != find_expected_status(new_status)
+        results.record_failure(record, reason: 'Event did not trigger state change.')
+        next
+      end
+
       results.save(move, record)
+    end
+  end
+
+  def find_event(move, new_status)
+    case new_status
+    when 'cancelled'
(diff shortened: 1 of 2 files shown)


app/lib/imports/moves_without_ending_state.rb
```diff
@@ -30,17 +30,63 @@ def results
...
+      GenericEvent::MoveCancel.find_by(eventable: move)
+    when 'rejected'
+      GenericEvent::MoveReject.find_by(eventable: move)
+    when 'completed'
+      GenericEvent::MoveComplete.find_by(eventable: move)
+    end
+  end
+
+  def find_pre_trigger_status(new_status)
+    case new_status
```
Thource, line 67: Minor: could use `return :requested if new_status == 'rejected'

:in_transit`

app/lib/imports/moves_without_ending_state.rb
```diff
@@ -30,17 +30,63 @@ def results
...
+    when 'rejected'
+      :requested
+    else
+      :in_transit
+    end
```
Thource, line 72: Minor: 
```suggestion
    return :requested if new_status == 'rejected'

    :in_transit
```

app/lib/imports/moves_without_ending_state.rb
```diff
@@ -30,17 +30,63 @@ def results
...
+  end
+
+  def find_expected_status(new_status)
+    case new_status
+    when 'completed'
+      'completed'
+    else
+      'cancelled'
+    end
```
Thource, line 81: Minor: 
```suggestion
    return 'completed' if new_status == 'completed'

    'cancelled'
```

app/lib/imports/moves_without_ending_state.rb
thomasleese: Yeah, I wondered about this, and thought it looked more ugly 😆 But I know `case`/`when` is not used that often in Ruby.

Thource: Haha, I think it's usually done by the amount of conditions, if it's more than 2 use a case/when instead. I don't mind either way :)

Thource (approved):

PR #1634, opened 2021-11-04
Strip whitespace from spreadsheet cells

thomasleese:
It looks like in some cases the spreadsheets we receive from the suppliers contain extra unexpected whitespace at the end of some cells which needs to be stripped away.

[Jira Ticket](https://dsdmoj.atlassian.net/browse/P4-3289)

### DIFF
--- a/app/lib/imports/journeys_missing_vehicle.rb
+++ b/app/lib/imports/journeys_missing_vehicle.rb
@@ -9,7 +9,7 @@ class Imports::JourneysMissingVehicle
 
   def initialize(csv_path:, columns:)
     @csv_path = csv_path
-    @columns = columns
+    @column_mapper = Imports::ColumnMapper.new(columns)
   end
 
   private_class_method :new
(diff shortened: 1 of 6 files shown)


app/lib/imports/journeys_missing_vehicle.rb
```diff
@@ -38,9 +38,9 @@ def results
   def records
     @records ||= CSV.table(csv_path).map do |record|
       {
-        journey_id: record[columns.fetch(:journey_id)],
-        move_id: record[columns.fetch(:move_id)],
-        vehicle_registration: record[columns.fetch(:vehicle_registration)],
+        journey_id: record[columns.fetch(:journey_id)].strip,
```
matt-crowson-moj, line 41: Minor gripe - test coverage?

matt-crowson-moj (approved):
LGTM

app/lib/imports/journeys_missing_vehicle.rb
thomasleese: Fair point, I've reworked this PR to introduce a class which handles the mapping of columns and encapsulates the `strip` and `downcase` logic so it could be tested on its own.

Thource (approved):

PR #1646, opened 2021-11-11
Add move and journey state machine diagrams

thomasleese:
This adds Graphviz diagrams of the state machines for moves and journeys, which should hopefully help to understand the system.

Recently I was helping one of the suppliers with move events and I created these diagrams to help them understand the order they should transition the moves or journeys (via events).


matt-crowson-moj (approved):
LGTM

PR #1655, opened 2021-11-29
Add task for importing moves without a to location

thomasleese:
This adds a task for importing and fixing moves which are missing a `to_location` value.

[Jira Ticket](https://dsdmoj.atlassian.net/browse/P4-3364)


matt-crowson-moj (approved):
LGTM

PR #1657, opened 2021-11-29
Ensure moves and journeys locations are different

thomasleese:
This adds validation on moves and journeys to ensure that the `from_location` and `to_location` fields are different. It shouldn't be valid for a move to have the same source/destination location.

We may need to hold off on deploying this until both of our suppliers are aware and comfortable of the validation.

[Jira Ticket](https://dsdmoj.atlassian.net/browse/P4-3363)

### DIFF
--- a/app/validators/different_to_from_location_validator.rb
+++ b/app/validators/different_to_from_location_validator.rb
@@ -1,0 +1,9 @@
+class DifferentToFromLocationValidator < ActiveModel::Validator
+  def validate(record)
+    return if record.to_location_id.blank?
+    return if record.from_location_id.blank?
+    return if record.to_location_id != record.from_location_id
+
+    record.errors.add(:to_location_id, 'should be different to the from location')
+  end
+end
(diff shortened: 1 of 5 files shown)


matt-crowson-moj (approved):
LGTM

app/validators/different_to_from_location_validator.rb
```diff
@@ -0,0 +1,10 @@
+class DifferentToFromLocationValidator < ActiveModel::Validator
+  def validate(record)
+    return if record.to_location_id.blank?
+    return if record.from_location_id.blank?
+
+    if record.to_location_id == record.from_location_id
```
Thource, line 6: Minor, but to keep with the guard theme I'd `return if record.to_location_id != record.from_location_id`

Thource (approved):

app/validators/different_to_from_location_validator.rb
thomasleese: Good point, I'll amend the commit.

Thource (approved):

PR #1663, opened 2021-12-08
Remove unused events table

thomasleese:
This follows on from https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/1171 to remove the underlying table for the now unused `Event` model. It looks like this was simply missed when the model was deleted, as it's now been entirely replaced by the `GenericEvent` model and is no longer used at all.


matt-crowson-moj (approved):
LGTM

PR #1664, opened 2021-12-15
Add BULMIL to Serco locations

thomasleese:
This is a new location we want to support in BaSM so we need to assign a supplier location to it so moves can be booked against it.

[Jira Ticket](https://dsdmoj.atlassian.net/browse/P4-3388)


matt-crowson-moj (approved):
LGTM

Thource (approved):

PR #1666, opened 2021-12-16
Update category and region for HMP Morton Hall

thomasleese:
It's recently re-opened and the category and region need to be updated to reflect the latest information.

[Jira Ticket](https://dsdmoj.atlassian.net/browse/P4-3387)


matt-crowson-moj (approved):
LGTM

PR #1678, opened 2022-01-04
Add cancellation reason for database corrections

thomasleese:
We need to manually fix the data on moves which were left uncancelled, however if the supplier has lost information such as the cancellation reason we need to pick one. Rather than using `other`, we can be explicit and say that we don't know the correct cancellation reason and that the data is not 100% accurate.

[Jira Ticket](https://dsdmoj.atlassian.net/browse/P4-3400)


thomasleese:
I've decided to go with `database_correction` as the internal name for this cancellation reason, but that doesn't mean we can't call it something completely different for our users on the frontend.

matt-crowson-moj (approved):
LGTM

PR #1687, opened 2022-01-13
Expand scope of supplier locations endpoint

thomasleese:
Currently, this endpoint only returns locations for a supplier if there's a move booked from that location for the associated supplier. This means when new locations are added and assigned to a supplier, it won't be included in this endpoint until a move has been booked from it.

We'd like to expand the scope of this so we have this existing behaviour, plus we return any locations where moves may (in the future) be booked from, allowing the suppliers to be aware of all possible locations. In particular, this helps our supplier users (for example contract delivery managers) see all the locations in the system.

[Jira Ticket](https://dsdmoj.atlassian.net/browse/P4-3389)

### DIFF
--- a/app/controllers/api/suppliers_controller.rb
+++ b/app/controllers/api/suppliers_controller.rb
@@ -11,9 +11,12 @@ module Api
     def locations_from_supplier_moves
       supplier = Supplier.find(params[:supplier_id])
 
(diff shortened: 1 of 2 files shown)


matt-crowson-moj (approved):
LGTM

PR #1688, opened 2022-01-13
Allow importing journeys to be rejected

thomasleese:
Currently we can only import journeys which are then to be completed or cancelled. This change allows us to import journeys which are proposed and to be rejected.

### DIFF
--- a/app/lib/imports/missing_journey_ending_events.rb
+++ b/app/lib/imports/missing_journey_ending_events.rb
@@ -25,16 +25,17 @@ private
 
   attr_reader :csv_path, :column_mapper
 
-  ALLOWED_JOURNEY_STATES = %w[Completed Cancelled].freeze
+  ALLOWED_EXISTING_JOURNEY_STATES = %w[proposed in_progress].freeze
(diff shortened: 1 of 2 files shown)


app/lib/imports/missing_journey_ending_events.rb
```diff
@@ -25,16 +25,17 @@ def call
 
   attr_reader :csv_path, :column_mapper
 
-  ALLOWED_JOURNEY_STATES = %w[Completed Cancelled].freeze
+  ALLOWED_EXISTING_JOURNEY_STATES = %w[proposed in_progress].freeze
```
matt-crowson-moj, line 28: Is there test coverage for this?

thomasleese: Although not explicit, it's covered here: https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/1688/files#diff-51febc3e46a5dfffe774de9c450420b67347e50e4cbc062a6c4800ec34a98eb4R8

PR #1712, opened 2022-01-25
Add delay to import move start events

thomasleese:
Serco are saying that they are experiencing performance issues with this backfill as notifications are sent to them at a high rate. I'm hoping that by introducing a small delay after each import should reduce those performance issues.

This will mean the imports take longer, but at the moment we're having to run the import in small batches with delays between each batch which is a lot of manual work whereas this should allow us to import large batches where the script itself introduces the delay.

The number I've chosen for the delay is arbitrary, and might need to be tweaked.

### DIFF
--- a/app/lib/imports/missing_move_start_events.rb
+++ b/app/lib/imports/missing_move_start_events.rb
@@ -46,6 +46,8 @@ private
       @current_move = nil
 
       results.save(move, record)
+
+      sleep 3
     end
   end
 


app/lib/imports/missing_move_start_events.rb
```diff
@@ -46,6 +46,8 @@ def results
       @current_move = nil
 
       results.save(move, record)
+
+      sleep 3
```
Thource, line 50: :/

Thource (approved):

app/lib/imports/missing_move_start_events.rb
thomasleese: Haha yes, it's not great... I'm hoping to revert this once the import has been done.

PR #1772, opened 2022-03-03
Add lockout events

thomasleese:
These will be used in lockouts and overnight lodges to record any changes to the PER which would be appropriate for the second leg of any move. Specifically, while in police custody, these might be recorded by police personnel.

[Jira Ticket](https://dsdmoj.atlassian.net/browse/P4-3335)

### DIFF
--- a/spec/factories/generic_event.rb
+++ b/spec/factories/generic_event.rb
@@ -22,6 +22,17 @@ FactoryBot.define do
     end
   end
 
+  factory :per_incident, parent: :generic_event do
+    eventable { association(:person_escort_record) }
+    classification { 'incident' }
+    details do
+      {
+        location_id: create(:location).id,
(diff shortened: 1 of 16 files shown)


spec/factories/generic_event.rb
```diff
@@ -22,6 +22,17 @@
     end
   end
 
+  factory :per_incident, parent: :generic_event do
```
matt-crowson-moj, line 25: Not sure this really matters but looking at the file this is getting very large, admittedly is a Factory.

matt-crowson-moj (approved):

spec/factories/generic_event.rb
thomasleese: Yeah, I reckon it is worth splitting this file out for sure.

PR #1782, opened 2022-03-11
Add support for events from notifications

thomasleese:
This makes it possible to send out notifications for the creation of events. Specifically, we're adding this functionality to support lockouts and overnight lodgings, where we want police users to add events and to let the suppliers know.

This won't go live as I'll disable these notifications until the suppliers are ready to handle them using: 
https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/0f69b45111b8b5d831f5cfe1439280161c28f885/app/jobs/prepare_base_notifications_job.rb#L50

[Jira Ticket](https://dsdmoj.atlassian.net/browse/P4-3427)

### DIFF
--- a/app/controllers/api/generic_events_controller.rb
+++ b/app/controllers/api/generic_events_controller.rb
@@ -17,6 +17,10 @@ module Api
       event = event_type.constantize.create!(event_attributes)
       run_event_logs
 
+      unless doorkeeper_application_owner.is_a?(Supplier)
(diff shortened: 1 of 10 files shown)


georgeschena (approved):

georgeschena:
I really like the readability of this structure `topic.is_a?(PersonEscortRecord)`, really made it easy to read and understand!

PR #1795, opened 2022-03-25
Add is_lockout field to move object and trigger change when lockout occurs

georgeschena:
### Jira link

[P4-3340](https://dsdmoj.atlassian.net/browse/P4-3340)

### What?

Added the `is_lockout` field to the move model, as well as trigger an update to the `is_lockout` field whenever a lockout occurs

### Why?

I am doing this because: This allows the frontend to receive this specific field without the need for us to pull in more data than we need into the frontend

### DIFF
--- a/app/models/generic_event/move_lockout.rb
+++ b/app/models/generic_event/move_lockout.rb
@@ -10,6 +10,10 @@ class GenericEvent
     include LocationValidations
     include LocationFeed
 
+    def trigger(*)
+      eventable.is_lockout = true
+    end
+
     enum reason: {
       unachievable_ptr_request: 'unachievable_ptr_request', # (PECS - police only)
       no_space: 'no_space', # (PECS)
--- a/spec/acceptance/generic_events/runner_spec.rb
+++ b/spec/acceptance/generic_events/runner_spec.rb
@@ -310,14 +310,15 @@ RSpec.describe GenericEvents::Runner do
     end
 
     context 'when event_name=lockout' do
-      # NB: lockout events have should have no effect on a move, they are purely for auditing
       before { create(:event_move_lockout, eventable: move) }
+
+      it 'does update the moves is_lockout value' do
+        expect { runner.call }.to change(move, :is_lockout).to(true)
+      end
 
       it 'does not update the move status' do
         expect { runner.call }.not_to change(move, :status).from('requested')
       end
-
-      it_behaves_like 'it does not call the Notifier'
     end
 
     context 'when the move record fails to save' do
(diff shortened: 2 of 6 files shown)


matt-crowson-moj (approved):

app/models/generic_event/move_lockout.rb
```diff
@@ -10,6 +10,11 @@ class MoveLockout < GenericEvent
     include LocationValidations
     include LocationFeed
 
+    def trigger(dry_run: false)
+      eventable.is_lockout = true
+      Allocations::RemoveFromNomis.call(eventable) unless dry_run
```
thomasleese, line 15: We shouldn't remove the move from an allocation in the case of a lockout, I think this may have been accidentally copied from the move cancel event!

spec/acceptance/generic_events/runner_spec.rb
```diff
@@ -310,14 +310,11 @@
     end
 
     context 'when event_name=lockout' do
-      # NB: lockout events have should have no effect on a move, they are purely for auditing
       before { create(:event_move_lockout, eventable: move) }
 
-      it 'does not update the move status' do
```
thomasleese, line 316: I think it's worth keeping this test as we still wouldn't want the status of a move to change after this event.

thomasleese (commented):
I think it's worth adding a Rake task or something similar which can retrospectively set this field on moves which has the lockout event before this PR.

app/models/generic_event/move_lockout.rb
georgeschena: Ohh yes! good spot!

georgeschena:
> I think it's worth adding a Rake task or something similar which can retrospectively set this field on moves which has the lockout event before this PR.

As in we need to run a rake task on the db's to set `is_lockout` to true if there is a lockout? As I guess currently everything will be false? @thomasleese

thomasleese:
> As in we need to run a rake task on the db's to set `is_lockout` to true if there is a lockout? As I guess currently everything will be false? @thomasleese

Yeah, I think it's worth checking the suppliers haven't sent any `MoveLockout` events, if they haven't then we don't need to do this.

Thource:
This seems to have no way of being unset, is this intentional? If lockouts only last one day (do they?), would this be more useful as `lockout_date`? That way we could only display the `Prison Lockout` tag in frontend if the date is todays date.

thomasleese:
> This seems to have no way of being unset, is this intentional? If lockouts only last one day (do they?), would this be more useful as `lockout_date`? That way we could only display the `Prison Lockout` tag in frontend if the date is todays date.

Once a move is a lockout, I don't believe we can unset this, my understand is that it's considered a lockout (i.e. we show the lockout label on the move listings page) for the entirety of the move.

PR #1796, opened 2022-03-28
Add a rake task to retrospectively set is_lockout to true

georgeschena:
This PR allows us with a rake task to retrospectively set `is_lockout` to true if a move has had any generic events with the type of `GenericEvent::MoveLockout`.


lib/tasks/lockout_moves.rake
```diff
@@ -0,0 +1,17 @@
...
+    Move.all.each do |move|
+      move_events = move.generic_events
+
+      move_events.each do |event|
+        if event.type == 'GenericEvent::MoveLockout'
+          move.is_lockout = true
+          break
+        end
+      end
+    end
```
thomasleese, line 15:
```suggestion
    move_ids = GenericEvent::MoveLockout.pluck(:eventable_id)
    Move.where(id: move_ids).update_all(is_lockout: true)
```

I think this will be a quicker way of running the query - I'm concerned that going through every move and then every event will be quite slow.

lib/tasks/lockout_moves.rake
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+namespace :lockout_moves do
+  desc 'lockouts a move which containts a MoveLockut in its generic events'
+  task lockout_moves: :environment do
```
thomasleese, line 5: I'd suggest putting this in the existing [`data_maintenance.rb`](https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/main/lib/tasks/data_maintenance.rake) group of Rake tasks.

Thource (approved):

PR #1804, opened 2022-03-30
Expose the is_lockout field to the move serializer

georges1996:
Expose the is_lockout field to the moves serializer, so we can check if a specific move is locked out. This is so we can show this on the frontend ticket: [P4-3342](https://dsdmoj.atlassian.net/browse/P4-3342)

### DIFF
--- a/app/serializers/v2/move_serializer.rb
+++ b/app/serializers/v2/move_serializer.rb
@@ -21,7 +21,8 @@ module V2
                :rejection_reason,
                :status,
                :time_due,
-               :updated_at
+               :updated_at,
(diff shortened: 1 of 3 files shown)


thomasleese (commented):
I think we should update the tests to include this field: https://github.com/ministryofjustice/hmpps-book-secure-move-api/blob/b4c9ef06f881847ff6eae45505f59cd4a580e45e/spec/serializers/v2/move_serializer_spec.rb#L18-L34

thomasleese (approved):
Look good, I'd suggest the commits can be squashed together before merging.

PR #1821, opened 2022-04-13
Add MoveLodgingEnd to the move important events

georgeschena:


app/models/move.rb
```diff
@@ -321,7 +321,10 @@ def all_events_for_timeline
   end
 
   def important_events
-    incident_events + (profile&.person_escort_record&.important_events || [])
+    eventable_types = %w[Move PersonEscortRecord]
+    classifications = %w[medical incident]
+
+    GenericEvent.where(eventable: eventable_types).where(classification: classifications).or(GenericEvent.where(type: ['GenericEvent::PerPropertyChange', 'GenericEvent::MoveLodgingEnd']))
```
thomasleese, line 327: Do you think we can remove the `important_events` method on the `PersonEscortRecord` model now? Just to clean things up and make it clear that it's not expected to be used.

thomasleese: I'd probably pull out `['GenericEvent::PerPropertyChange', 'GenericEvent::MoveLodgingEnd']` in to a variable as well.

PR #1835, opened 2022-04-27
Allow fault_classification to be nil if the supplier_id is nil

georgeschena:
We now allow the `fault_classification` to be nil if the `supplier_id` is not provided, this has been done as in the frontend we allow a user to submit events via a form. 

Previous to this we would have had to provide a `fault_classification` attribute with some data, which would have had to be hard coded.

### DIFF
--- a/spec/models/generic_event/incident_spec.rb
+++ b/spec/models/generic_event/incident_spec.rb
@@ -19,5 +19,13 @@ RSpec.describe GenericEvent::Incident, type: :model do
 
       it { is_expected.to eq(:incident) }
     end
+
+    describe 'when a supplier_id is nil' do
+      context 'when we allow fault_classification to be nil' do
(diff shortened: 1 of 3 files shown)


spec/models/generic_event/incident_spec.rb
```diff
@@ -19,5 +19,13 @@
...
       it { is_expected.to eq(:incident) }
     end
+
+    describe 'when a supplier_id is nil' do
+      context 'when we allow fault_classification to be nil' do
+        let(:event) { create(:event_person_move_used_force, supplier_id: nil, fault_classification: nil) }
+
+        it { is_expected.to eq(:incident) }
+      end
+    end
```
thomasleese, line 29: I don't think we need this test as it's covered by `an_event_about_an_incident.rb`.

thomasleese (approved):

PR #1849, opened 2022-05-23
Add responded_by to PerConfirmation event details

Thource:
This is the API side of this work, there is still frontend work to do, but I'm releasing this so that I can complete the frontend work more easily.

### Jira link

P4-3575

### What?

I have added/removed/altered:

- [ ] Added `responded_by` field to PerConfirmation events.

### Why?

I am doing this because:

- It will allow the frontend to display the authors of PERs.

### DIFF
--- a/app/controllers/api_controller.rb
+++ b/app/controllers/api_controller.rb
@@ -57,6 +57,8 @@ class ApiController < ApplicationController
 private
 
   def set_sentry_transaction_id
+    return unless Sentry.initialized?
+
     Sentry.set_tags(transaction_id: request.headers['X-Transaction-Id'])
   end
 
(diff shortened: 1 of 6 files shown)


lib/tasks/populate_per_confirmation_with_responded_by.rake
```diff
@@ -0,0 +1,18 @@
+GITHUB_FRAMEWORK_URI = 'https://github.com/ministryofjustice/hmpps-book-secure-move-frameworks.git'.freeze
```
matt-crowson-moj, line 1: For my benefit what are lines 1-3 doing?

Thource: Copy pasta :D I'll remove them

app/controllers/api_controller.rb
```diff
@@ -57,6 +57,8 @@ def info_for_paper_trail
 private
 
   def set_sentry_transaction_id
+    return unless Sentry.initialized?
```
matt-crowson-moj, line 60: Is this to resolve an issue with the initialisation of Sentry?

Thource: Yeah, this is unrelated to the PER author stuff, it just means that we can run the API locally without setting the sentry env vars now

PR #1882, opened 2022-06-29
Add PerUpdated event

Thource:
### Jira link

P4-3577

### What?

I have added/removed/altered:

- [x] Added the PerUpdated event, it is automatically added when a PER is updated after being completed.

### Why?

I am doing this because:

- It's helpful to see which users updated which sections of the PER.

### DIFF
--- a/app/state_machines/framework_assessment_state_machine.rb
+++ b/app/state_machines/framework_assessment_state_machine.rb
@@ -25,6 +25,13 @@ class FrameworkAssessmentStateMachine < FiniteMachine::Definition
             details: { completed_at: Time.zone.now },
           )
         end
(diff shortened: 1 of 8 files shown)


app/state_machines/framework_assessment_state_machine.rb
```diff
@@ -25,6 +25,13 @@ class FrameworkAssessmentStateMachine < FiniteMachine::Definition
             details: { completed_at: Time.zone.now },
           )
         end
+      elsif target.is_a?(PersonEscortRecord)
```
matt-crowson-moj, line 28: Is there test coverage for this?

matt-crowson-moj (approved):

PR #1923, opened 2022-08-18
P4-3888 intitial cut at including per change history available via diagnostics

matt-crowson-moj:


app/services/diagnostics/move_inspector.rb
```diff
@@ -338,6 +360,16 @@ def generate
       @output
     end
 
+    def diff(old_paper_trail, new_paper_trail)
+      old_values = old_paper_trail.attributes
```
matt-crowson-moj, line 364: As I understand it reify is not needed here as this is a Paper Trail object.

Thource: `reify` is still needed here, as it stands this will get the attributes of the Version object

docker-compose.yml
```diff
@@ -22,20 +22,20 @@ services:
     volumes:
       - './.data/localstack:/tmp/localstack'
       - '/var/run/docker.sock:/var/run/docker.sock'
-  web:
```
Thource, line 25: accident?

matt-crowson-moj: Doh! indeed.

Thource (approved):

PR #1929, opened 2022-08-19
P4-3888 changing to look at the response value and not the attributes too much noise with the attributes.

matt-crowson-moj:


app/services/diagnostics/move_inspector.rb
```diff
@@ -261,13 +261,15 @@ def generate
...
 
             per.framework_responses.group_by(&:section).each do |section, responses|
-              @output << "section:\t#{section}\n"
+              @output << "START OF SECTION:\t#{section}\n"
               responses.each do |response|
-                @output << "question:\t#{response.framework_question.key}\n"
+                @output << "QUESTION:\t#{response.framework_question.key}\n"
                 response.versions.each do |version|
-                  diff(version.reify&.attributes || {}, version.next&.reify&.attributes || response.attributes)
+                  diff(version.reify&.value || {}, version.next&.reify&.value || response.value, version.whodunnit)
```
Thource, line 268: just `version.reify&.value` is fine here because we're not always comparing hashes anymore

Thource (approved):

PR #2004, opened 2023-01-17
[P4-3956] Add new event: MoveOvernightLodge

Thource:
### Jira link

P4-3956

### What?

I have added/removed/altered:

- [x] Added a new event: MoveOvernightLodge

### Why?

I am doing this because:

- This event is being added as part of the overnight lodging work, it's purpose is to make users aware that a move has an overnight lodging planned.


jimbali (approved):
lgtm

PR #2008, opened 2023-01-24
feat: Add recall_date field to moves [P4-3970]

jimbali:
Add recall date field to moves.

### Jira link

[P4-P4-3970](https://dsdmoj.atlassian.net/browse/P4-3970)

### What?

I have added/removed/altered:

- Added recall_date field to Move model

### Why?

I am doing this because:

- This field is being moved to the booking form instead of the PER. Only police are able to provide this information, and it's easier to restrict access to sections of the booking form than the PER.

### Have you? (optional)

- The code expects the recall_date column to exist in the moves table so the migrations should be run prior to deployment
- Changes an API that is used in production - suppliers need to be kept informed

### DIFF
--- a/app/controllers/api/moves_controller.rb
+++ b/app/controllers/api/moves_controller.rb
@@ -73,6 +73,7 @@ module Api
       profile_id
       person_id
       reference
+      recall_date
     ].freeze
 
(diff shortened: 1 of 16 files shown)


Thource (approved):
LGTM :)

PR #2015, opened 2023-02-22
chore: Always send create_move notification

jimbali:
We don't send notifications if the move has not yet been approved. This means that sometimes there is no create_move notification. We have decided to always send this notification at the point when we start sending notifications to the subscriber, so at that point we create one and send it if it doesn't yet exist.

### Jira link

[P4-4118](https://dsdmoj.atlassian.net/browse/P4-4118)

### What?

I have added/removed/altered:

- create_move notifications are now always sent out before any others

### Why?

I am doing this because:

- Serco asked for it

### Have you? (optional)

- We need to liaise with the suppliers to make sure this is what they want and that it won't break anything

[P4-4118]: https://dsdmoj.atlassian.net/browse/P4-4118?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/jobs/prepare_base_notifications_job.rb
+++ b/app/jobs/prepare_base_notifications_job.rb
@@ -9,13 +9,17 @@ class PrepareBaseNotificationsJob < ApplicationJob
 
     subscriptions(move, only_supplier_id: only_supplier_id).find_each do |subscription|
       if send_webhooks && subscription.callback_url.present? && should_webhook?(subscription, move, action_name)
-        notification = build_notification(subscription, NotificationType::WEBHOOK, topic, action_name)
-        NotifyWebhookJob.perform_later(notification_id: notification.id, queue_as: queue_as)
+        notifications = build_notifications(subscription, NotificationType::WEBHOOK, topic, action_name)
+        notifications.each do |notification|
+          NotifyWebhookJob.perform_later(notification_id: notification.id, queue_as: queue_as)
+        end
       end
 
       if send_emails && subscription.email_address.present? && should_email?(move)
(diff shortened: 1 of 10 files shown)


app/jobs/prepare_base_notifications_job.rb
```diff
@@ -38,6 +42,17 @@ def subscriptions(move, only_supplier_id: nil)
     Subscription.kept.enabled.where(supplier: suppliers)
   end
 
+  def build_notifications(subscription, type_id, topic, action_name)
+    notifications = []
+
+    if %w[update update_status].include?(action_name) &&
```
Thource, line 48: I think we need to check that `topic` is move, right? otherwise we could create duplicate create notifications for other types

Thource (changes requested):

app/jobs/prepare_base_notifications_job.rb
jimbali: Ah yes, fair point. Or should I move the logic into the PrepareMoveNotificationsJob?

Thource: Yeah moving the logic into the PrepareMoveNotificationsJob makes more sense I think, good call

jimbali: ok I've moved some of the logic to the PrepareMoveNotificationsJob. It turns out that the other jobs that have `update` actions all also override the `prepare` method, so there wasn't any broken functionality to test for, but I've beefed up their tests a bit anyway to be safe.

Thource (approved):

PR #2024, opened 2023-03-09
P4-4100 Add logging to create and validate move methods

danbenton-mojdt:
### Jira link

P4-4100

### What?

I have added:

- Log statements to the move status validation method

### Why?

I am doing this because:

- Make it easier to debug 
-
-

### Have you? (optional)

### DIFF
--- a/app/controllers/api/v2/moves_actions.rb
+++ b/app/controllers/api/v2/moves_actions.rb
@@ -15,6 +15,7 @@ module Api::V2
     end
 
     def create_and_render
+      Rails.logger.info('V2 Move creation started')
       validate_move_status
       move = Move.new(move_attributes)
       move.supplier = doorkeeper_application_owner || SupplierChooser.new(move).call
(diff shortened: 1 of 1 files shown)


app/controllers/api/v2/moves_actions.rb
```diff
@@ -113,9 +114,12 @@ def common_move_attributes
 
     def validate_move_status
       status = move_params.fetch(:attributes, {})[:status]
+      Rails.logger.debug("Validating current status of move - [#{status}]")
       if status.present?
         validator = Moves::StatusValidator.new(status: status, cancellation_reason: move_params.fetch(:attributes, {})[:cancellation_reason], rejection_reason: move_params.fetch(:attributes, {})[:rejection_reason])
-        raise ActiveModel::ValidationError, validator unless validator.valid?
+        valid = validator.valid?
+        Rails.logger.debug("Move status - [#{valid}]")
```
jimbali, line 121: Might be clearer to say "Move validation status" so we don't confuse it with `move.status`?

danbenton-mojdt: It is validating the move.status. Reworded to clarify

jimbali: ahh ok makes sense. Thanks

jimbali (approved):

PR #2025, opened 2023-03-09
feat: Update person data from NOMIS asynchronously

jimbali:
When a move is created we update the person's image from NOMIS. Sometimes this times out or fails, meaning that a 502 response is returned without JSON data containing the newly created move's ID.

To avoid this, we now queue up a job to update the person's image asynchronously instead, and after that is done we send an update_move notification.

### Jira link

[P4-4122](https://dsdmoj.atlassian.net/browse/P4-4122)

### What?

I have added/removed/altered:

- When a move is created we now update the person's image asynchronously

### Why?

I am doing this because:

- Suppliers are sometimes encountering a 502 response code and not receiving the ID of a move which was created successfully, meaning that they have no reference to the move and are therefore unable to track it or make any updates to it

- Changes the behaviour of the API when creating a move, so we should run this by suppliers before deploying to production

[P4-4122]: https://dsdmoj.atlassian.net/browse/P4-4122?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/api/v2/moves_actions.rb
+++ b/app/controllers/api/v2/moves_actions.rb
@@ -25,12 +25,12 @@ module Api::V2
 
       Rails.logger.info("Move saved with reference [#{move.reference}] - status [#{move.status}]")
 
-      move.person.update_nomis_data if move.person.present?
-
       Notifier.prepare_notifications(topic: move, action_name: 'create')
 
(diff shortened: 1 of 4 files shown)


Thource (approved):
Nice

PR #2031, opened 2023-03-23
P4-4102 Add access log table

danbenton-mojdt:
### Jira link

P4-4102

### What?

I have added/removed/altered:

- [ ] Added a table to log all access to the API (including read-only access) 

### Why?

I am doing this because:

- To make access to the application and data therein more auditable


jimbali (approved):
Lgtm. Might benefit from an index of some kind but we can add one later.

PR #2048, opened 2023-04-06
Update allocation date

jimbali:
### Jira link

[P4-4139](https://dsdmoj.atlassian.net/browse/P4-4139)

### What?

I have added/removed/altered:

- [x] Allocation date can now be updated

### Why?

I am doing this because:

- It was requested by PMU

### Have you? (optional)

- Changes an api that is used in production (non-breaking change)

[P4-4139]: https://dsdmoj.atlassian.net/browse/P4-4139?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/allocations/updater.rb
+++ b/app/services/allocations/updater.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+module Allocations
+  class Updater
+    attr_accessor :allocation_params, :allocation_id, :allocation
+
+    def initialize(allocation_params:, allocation_id:)
+      self.allocation_params = allocation_params
+      self.allocation_id = allocation_id
+    end
+
+    def call
+      self.allocation = Allocation.find(allocation_id)
+      existing_date = allocation.date
+
+      allocation.assign_attributes(allocation_params[:attributes])
+      allocation.validate!
+
+      update_move_dates if allocation.date != existing_date
+
+      allocation.save!
+    end
+
+  private
+
+    def update_move_dates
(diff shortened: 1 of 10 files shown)


danbenton-mojdt (approved):

app/services/allocations/updater.rb
```diff
@@ -0,0 +1,47 @@
...
+          move.date = allocation.date
+          move.save!
+
+          create_automatic_event!(
+            eventable: move,
+            event_class: GenericEvent::MoveDateChanged,
+            details: { date: move.date.iso8601 },
+          )
+
+          Notifier.prepare_notifications(topic: move, action_name: 'update_status')
```
Thource, line 42: this should be 'update', as the status doesn't change

Thource: since this is called before `allocation.save!`, if the save fails, the notifications will still be sent, maybe not a problem but something to keep in mind

jimbali: Good spot on both counts. With the `update_status` I thought I had copied what the move controller does but I had misread it. And the test condition I was using to catch the allocation failing to save was also causing the move to fail to save, so I've improved the tests to catch the specific scenario you outlined and moved the `allocation.save!` inside the transaction so that it won't send the notifications (as they won't be created in the DB).

PR #2223, opened 2024-03-28
Fix issue with update lodgings accepting location

Thource:


app/controllers/api/lodgings_controller.rb
```diff
@@ -202,7 +202,11 @@ def update_lodging_params
     end
 
     def update_lodging_attributes
-      @update_lodging_attributes ||= update_lodging_params.to_h[:attributes].tap do |attribs|
+      attributes = update_lodging_params.to_h[:attributes] || {}
+      location_id = update_lodging_params.dig(:relationships, :location, :data, :id)
+      attributes[:location_id] = location_id if location_id.present?
+
+      @update_lodging_attributes ||= attributes.tap do |attribs|
```
jimbali, line 209: could do with a test for this...?

danbenton-mojdt (approved):

PR #2227, opened 2024-04-12
MAP-949 Preprod refresh scripts and cronjob

danbenton-mojdt:
### Jira link

MAP-949

### What?

I have added bash/sql scripts to run partial updates of preprod from production

### Why?

I am doing this because this currently runs on my machine and that not good

### Deployment risks

- Includes destructive/migratory actions which may affect critical offender data


Thource:
looks like script 2 is missing? presumably to copy the prod data?

danbenton-mojdt:
> looks like script 2 is missing? presumably to copy the prod data?

The sql in the bash methods does the job of what might have been script 2 (copying the prod data to the temp tables). I named the others 3 onwards in case I wanted to replace the individual calls with a prewritten sql script in the same way as the others.

Thource (approved):

PR #2232, opened 2024-04-19
[MAP-967] Add the preprod refresh and reports scripts to the Helm chart

jimbali:
### Jira link

MAP-967

### What?

I have added/removed/altered:

- Add the preprod refresh and reports scripts to the Helm chart

### Why?

I am doing this because:

- So that they get deployed automatically


danbenton-mojdt (approved):
Reports need to be disabled in values.yaml and overridden in prod

PR #2238, opened 2024-04-26
MAP-531: Modified duplicated per month query to be in line with updated incomp…

jonesj2193:
…lete per query. Fixed missing line in incomplete per query.

### Jira link

MAP-531

### What?

I have added/removed/altered:

- [ ] Modified the duplicated per month query to be in line with updated incomplete per query
- [ ] Added in missing line to the incomplete per query

### Why?

I am doing this because:

- For consistency and improved performance of the duplicates query
- Missing order by and correct ordering of variables defined causing misallocation of values in incomplete per report.


tobyprivett:
No longer needed

PR #2252, opened 2024-05-20
MAP-967 Use updated configmap refs

danbenton-mojdt:


jimbali (approved):
LGTM. Tested using `helm diff` and it looks right.

PR #2265, opened 2024-05-28
Remove non-helm reports directory

danbenton-mojdt:
Remove the reports sql files that are now defined in the Helm directory, to save confusion


helm_deploy/hmpps-book-secure-move-api/templates/reports/README.md
jimbali, line 1: I've already put this file in the helm chart but in the root of the chart folder called `REPORTS.md`, so I'd either delete that or copy this to there instead.

jimbali (approved):

PR #2266, opened 2024-05-28
feat: Send notify_move notification to cross-deck supplier

jimbali:
### Jira link

[MAP-322]

### What?

I have added/removed/altered:

- Where the receiving location is operated by another supplier, we now send them a `notify_move` notification at the same time as we send the departing supplier a `create_move` notification
- Any subsequent `update_move` notifications will go to both suppliers
- A `MoveRedirect` which results in the move becoming cross-deck will also result in a `notify_move` notification being sent to the receiving supplier
- A `MoveRedirect` which results in the move ceasing to be cross-deck will result in a `disregard_move` notification being sent to any supplier that previously received a `notify_move` notification

### Why?

I am doing this because:

- So that the receiving supplier can add it to their systems to be ready for when they take over the move

[MAP-322]: https://dsdmoj.atlassian.net/browse/MAP-322?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/jobs/prepare_base_notifications_job.rb
+++ b/app/jobs/prepare_base_notifications_job.rb
@@ -9,13 +9,11 @@ class PrepareBaseNotificationsJob < ApplicationJob
 
     subscriptions(move, only_supplier_id:).find_each do |subscription|
       if send_webhooks && subscription.callback_url.present? && should_webhook?(subscription, move, action_name)
-        notification = build_notification(subscription, NotificationType::WEBHOOK, topic, action_name)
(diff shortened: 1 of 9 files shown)


danbenton-mojdt (changes requested):
We only need to notify the other supplier where the receiving location is a court

danbenton-mojdt (approved):

PR #2269, opened 2024-05-30
MAP-1040 Add extra column heading to incomplete PER report

danbenton-mojdt:
to match updated query output, splitting `supplier from court` column into separate `to court` and `from court` columns

### Jira link

MAP-1040


tobyprivett:
Closing this as the task is completed: https://dsdmoj.atlassian.net/browse/MAP-1040

PR #2281, opened 2024-07-10
MAP-1380 Made changes to include a 'cancelled_at' column to the download showing the time when the move had been cancelled

jonesj2193:
### Jira link

MAP-1380

### What?

I have added/removed/altered:

- [ ] Added a "cancelled at" column to the csv download.

### Why?

I am doing this because:

- Users want to be able to see when a move had been cancelled.


jimbali (approved):
lgtm!

PR #2308, opened 2024-08-22
Script release commands

danbenton-mojdt:


release.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+git stash
```
jimbali, line 2: I'm not sure I'd want it to do git stash...
Maybe we just assume you're on main branch?

danbenton-mojdt: Big assumption - could check that we are and diff with origin/main, non-zero exiting if there is any difference? 
Stash ensures we can check it out without there being any conflicting unstaged change

jimbali: Most important for me is that I'd like to be warned if a script is going to stash my code

jimbali (approved):
lgtm!

release.sh
danbenton-mojdt: Fair! It no longer makes changes to your working tree at all and will just exit if it thinks you're trying to release something that's not the latest commit in main

PR #2311, opened 2024-08-28
chore: Upgrade Rails to 7.1 - first pass

jimbali:
### Jira link

MAP-1477

### What?

Main actions so far:
- Update Rails to 7.1.4 in Gemfile
- Run `bin/rails app:update` to automatically update various config files
- Add explicit attributes for enums without a corresponding DB column
- Rename 422 unprocessable_entity to unprocessable_content
- Update some Swagger methods to the new OpenAPI equivalents
- Make sure that AWS library is properly included
- Use Database Cleaner instead of RSpec's own transactional fixtures as it seems to work more reliably on CircleCI
- Fix deprecation warnings where possible
- Finish the previous Rails 7.0 upgrade by enabling all 7.0 defaults

### Deployment risks

Various defaults have changed, so we need to be careful that nothing's broken before releasing to prod.


Mjwillis (approved):
LGTM

PR #2316, opened 2024-08-30
Rails 7.1 pt2

jimbali:
### Jira link

MAP-1477

### What?

I have added/removed/altered:

- Enable most Rails 7.1 defaults
- Get rid of the unnecessary guff that gets printed to the console during tests

### Why?

I am doing this because:

- To continue with the Rails 7.1 upgrade and get ready for 7.2
- To make it easier to spot deprecation errors logged during test runs

- Some of these changes needed to only be deployed after all pods were running 7.1 (which they now are)


Mjwillis (approved):
LGTM

PR #2327, opened 2024-09-02
Group Dependabot updates

jimbali:
Hopefully this will make Dependabot raise a single PR instead of 10 random ones


Mjwillis (approved):
lgtm

PR #2408, opened 2024-11-15
MAP-1781 Cleanup Access Logs more than 3 months old

tobyprivett:
### Jira link

https://dsdmoj.atlassian.net/browse/MAP-1781

### What?

Adds a daily cronjob to delete stale Access Logs. 
DELETE can be a very expensive operation, so we do this in batches of 1000.

### Why?

200k Access Logs are created every day. 
We use them for debugging production issues, so it should be okay to remove anything older than three months.

### DIFF
--- a/lib/tasks/access_logs.rake
+++ b/lib/tasks/access_logs.rake
@@ -2,13 +2,13 @@ namespace :access_logs do
   desc 'Cleanup access_logs older than 3 months'
   task cleanup: [:environment] do
     cutoff_date = 3.months.ago
-    access_logs_count = AccessLog.where("timestamp < ?", cutoff_date).count
+    access_logs_count = AccessLog.where('timestamp < ?', cutoff_date).count
     number_of_iterations = (access_logs_count.to_f / 1000).ceil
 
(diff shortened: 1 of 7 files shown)


jimbali:
I think three months is too short tbh, I've had to debug issues with older moves before

lib/tasks/access_logs.rake
```diff
@@ -0,0 +1,14 @@
...
+    access_logs_count = AccessLog.where('timestamp < ?', cutoff_date).count
+    number_of_iterations = (access_logs_count.to_f / 1000).ceil
+
+    puts "Cleaning up #{access_logs_count} access_logs in #{number_of_iterations} iterations"
+
+    1.upto(number_of_iterations).each do
+      AccessLog.where('timestamp < ?', cutoff_date).limit(1000).delete_all
+    end
+  end
+end
```
jimbali, line 14: I would suggest possibly adding a test for this task but it's up to you

jimbali (dismissed):

danbenton-mojdt:
> I think three months is too short tbh, I've had to debug issues with older moves before

Agree, minimum 12 months imo

jimbali (approved):

PR #2409, opened 2024-11-15
MAP-1672 Upgrade to Rails 8

tobyprivett:
### Jira link

[MAP-1672](https://dsdmoj.atlassian.net/browse/MAP-1672)

### What?

Upgrade to Rails 8.0.0
Remove Flipper

### Why?

Rails 8 has been released
Flipper was used once in 2020, and the gem is outdated 

[MAP-1672]: https://dsdmoj.atlassian.net/browse/MAP-1672?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Gemfile
```diff
@@ -5,7 +5,7 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
 ruby File.read('.ruby-version')
 
-gem 'rails', '~> 7.1.4'
+gem 'rails', '~> 8.0'
```
jimbali, line 8: This technically permits 8.1

config/application.rb
```diff
@@ -21,7 +21,7 @@
 module BookASecureMoveApi
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
-    config.load_defaults 7.1
+    config.load_defaults 8.0
```
jimbali, line 24: The Rails upgrade guide does not recommend changing this value on the first release, as I understand it. We're supposed to try it out with the values loaded from `config/initializers/new_framework_defaults_8_0.rb`, and when we know we don't need to roll back we can change `config.load_defaults`.

config/initializers/new_framework_defaults_8_0.rb
```diff
@@ -0,0 +1,30 @@
+# Be sure to restart your server when you modify this file.
```
jimbali, line 1: See comment on `config/application.rb`, is this file not redundant if we are already changing `config.load_defaults` there?

db/migrate/20241118095830_remove_flipper_tables.rb
jimbali, line 1: Is it safe to delete tables that are still referenced in the code? Usually I would release the application with the code that references the tables removed, and then do a subsequent release with a migration to remove the tables. This ensures backwards compatibility with the currently running code, as the migration is run before deployment. But I guess if we're not using Flipper then it probably doesn't matter.

tobyprivett: This is the only row in the Flipper Features table: 
```i
rb(main):001> Flipper::Adapters::ActiveRecord::Feature.all
=>
[#<Flipper::Adapters::ActiveRecord::Feature:0x00007fe6b75c5c60
  id: 1,
  key: "prefill",
  created_at: Tue, 27 Oct 2020 17:18:28.632256000 GMT +00:00,
  updated_at: Tue, 27 Oct 2020 17:18:28.632256000 GMT +00:00>]
```

PR #2413, opened 2024-11-19
MAP-1812 Resolve duplicates on the Weekly Cancellation report

tobyprivett:
### Jira link

https://dsdmoj.atlassian.net/browse/MAP-1812

### What?

Resolve duplicates on the Weekly Cancellation report.

Update the SQL to report on a journey's to/from location, rather than the move's to/from location.

### Why?

Recipients of this report noticed duplicate rows. Digging in, we discovered that the report is correctly writing a row for each move's journey, but is displaying the move's to/from location. To the end user, these rows look like duplicates, because the to/from location is repeated for each journey.

We have also learned that suppliers may be cancelling a journey that has the wrong location and creating a new one when, perhaps, they should be updating the journey with the correct location.


Mjwillis (approved):
LGTM

PR #2465, opened 2025-01-07
Avoid migrating audit-trail data to pre-prod

tobyprivett:
### Jira link

https://dsdmoj.atlassian.net/browse/MAP-1975

### What?

Currently, there are 165 million records in the versions table and the migrate script is not working.

### Why?

Rather than debugging the script, it makes more sense to disable it as there is no use case for Papertail version history on pre-prod.


jnioche-jd (approved):

BarryGee-MOJ (approved):
lgtm

PR #2494, opened 2025-02-11
MAP-2128 Move rejection emails

tobyprivett:
### Jira link

[MAP-2128](https://dsdmoj.atlassian.net/browse/MAP-2128)

### What?
Added an `after_create` callback to the `GenericEvent::MoveReject` that looks up the email address of the person who proposed the move and, if found, sends them an email.

The logic:
- When a move is REJECTED a MoveReject event is created
- When that event is created, we attempt to send a move-reject email ([template on Notify](https://www.notifications.service.gov.uk/services/c267a60e-e4e3-45b0-8aaf-f64546a1839a/templates/e728365d-5b0c-41ee-8c76-aca57f59c843)) to the Booker
- A Proposer is either the person who Proposed (MoveProposed event) the move
- To get the Proposer’s email address, we look them up in the Manage Users API
- We only attempt to send an email if we can find an email address

### Why?
When prison users propose prison transfers moves these go to PMU to approve or reject. If the move is rejected this will get stated on the move in BaSM but the person who booked the move is not notified directly. 

### Have you?  

Added MANAGE_USERS_API_BASE_URL: https://manage-users-api-[env].hmpps.service.justice.gov.uk/ to
- [x] Staging
- [x] UAT
- [x] Preprod
- [x] Production

### Deployment risks
- Adds a call to a new API (Manage Users), so MANAGE_USERS_API_BASE_URL must be set before deployment.

[MAP-2128]: https://dsdmoj.atlassian.net/browse/MAP-2128?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/models/generic_event/move_reject.rb
```diff
@@ -15,5 +17,27 @@ def for_feed
...
+      return unless email
+
+      MoveRejectMailer.notify(email, move, self).deliver_now!
+    end
+
+    def move_proposed_by_email
+      username =
+        move.generic_events
+          .where(type: 'GenericEvent::MoveProposed')
+          .first&.created_by
```
Thource, line 38: created_by can be an email address, I think it would be a good idea to check if this value is an email and if so, just use the email?

tobyprivett: @Thource Added a check for an email address

Thource (approved):

PR #2510, opened 2025-03-10
Update Rubocop

tobyprivett:
Updating rubocop-govuk manually becuase it introducted a new rule:

`Rails/Delegate: Use delegate to define delegations.`

Which required a test fix


michalnawrockidj (approved):
LGTM

PR #2538, opened 2025-04-03
Introduce a Client Audit Log that captures IP and client

tobyprivett:
### Jira link

MAP-<TODO>

### What?

Add a new table `client_audit_logs`
Add a before handler to the API controller to `set_client_audit_log`

Ensure that:
- Only unique records are created
- Records older than the months are deleted

### Why?

We need a way to store client IP addresses so we can run an audit and check for clients accessing from unauthorised IP address.

### Have you? (optional)

-  Updated helm charts with new cron job: `client-audit-log-cleanup`

- Includes database migration


tobyprivett:
Closing this as we can get the data needed from App Insights and Ingress logs

PR #2600, opened 2025-06-18
MAP-2530 Avoid invalid calls to the Prisoner API

tobyprivett:
### Jira link

https://dsdmoj.atlassian.net/browse/MAP-2530

### What?

Prevents API calls when prison_number is nil/blank.

### Why?

This was causing 404 "No static resource prisoner" errors in BaSM logs.


dafhoustonmoj (approved):
LGTM

PR #2624, opened 2025-07-15
MAP-2534 Fix ImportPeople to use Alerts API per-person calls

tobyprivett:
### Jira link

https://dsdmoj.atlassian.net/browse/MAP-2534

### What?

Changed `ImportPeople` service to make individual API calls to AlertsApiClient 
for each person instead of batching all prison numbers in a single request.

### Why?

The Alerts API expects per-person calls rather than batch operations. 
The previous batch approach was incompatible with the API's design.

There is no evidence that anyone is using this particular endpoint (create move v1)
as the code wouldn't have worked in the current state, but as part of the overall
refactor to move away from NOMIS, we should ensure everything is working.

### DIFF
--- a/app/services/moves/import_people.rb
+++ b/app/services/moves/import_people.rb
@@ -16,21 +16,21 @@ module Moves
 
     def import_people
       people = NomisClient::People.get(prison_numbers).index_by { |p| p.fetch(:prison_number) }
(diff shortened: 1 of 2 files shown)


Thource (approved):
Looks good, I think V1 is unused and should probably be deprecated

PR #2629, opened 2025-07-22
Handle empty response body in UserEmail.get

tobyprivett:
### Jira link

[MAP-<TODO>](https://dsdmoj.atlassian.net/browse/MAP-2606)

### What?

Add defensive code to the UserEmail API client:
- Return nil when API response body is blank to avoid JSON::ParserError
- Update shared auth context to handle empty JSON responses

### Why?
 
The [Manage Users API ](https://manage-users-api.hmpps.service.justice.gov.uk/swagger-ui/index.html#/user-controller/getUserEmail) can return an empty string for the body


BarryGee-MOJ (approved):
LGTM

PR #2632, opened 2025-07-23
Refactor test setup

tobyprivett:
### Jira link

N/A

### What?

In this test, we were assigning the response body to an empty string and then checking for an empty string before attempting to parse it as JSON.

### Why?

I amThe test was testing itself, so removing the unnecessary code


GurnankCheema (approved):
lgtm

PR #2643, opened 2025-08-06
Fix api_docs CI failure with Ruby 3.4.5

tobyprivett:
### Jira link

N/A

### What?

Fixes the CI job: `api_docs` by adding `RUBYOPT: "--parser=parse.y"`

### Why?

Ruby 3.4.5 switched from the classic parser to Prism as the default parser, but `rswag` v2.1.6 (used to generate Swagger docs) doesn't like it. So we force the job to use the classic parser.


dafhoustonmoj (approved):
LGTM

PR #2730, opened 2026-05-14
Doorkeeper 5.9.0

danbenton-mojdt:


jimbali (approved):
Approving this to test it in staging but, as discussed, it may break things.

